PR #1, opened 2016-04-12
Add `no-process-exit` rule

jfmengels:
Adds `no-process-exit` rule. Same as the ESLint core rule, but with the exception of files that start with a hasbang `#!`.

@sindresorhus asked me if I could implement this rule, and I thought it was fit to make a PR here.

### DIFF
--- a/rules/no-process-exit.js
+++ b/rules/no-process-exit.js
@@ -1,0 +1,20 @@
+'use strict';
+
+module.exports = function (context) {
+	var startsWithHashBang = context.getSourceCode().lines[0].indexOf('#!') === 0;
+	if (startsWithHashBang) {
+		return {};
+	}
+
+	return {
+		CallExpression: function (node) {
+			var callee = node.callee;
+
+			if (callee.type === 'MemberExpression' &&
+					callee.object.name === 'process' &&
+					callee.property.name === 'exit') {
+				context.report(node, 'Only use process.exit() in CLI apps. Throw an error instead.');
+			}
+		}
+	};
+};
--- a/test/no-process-exit.js
+++ b/test/no-process-exit.js
@@ -1,0 +1,41 @@
+import test from 'ava';
+import {RuleTester} from 'eslint';
+import rule from '../rules/no-process-exit';
+
+const ruleTester = new RuleTester({
+	env: {
+		es6: true
+	}
+});
+
+const errors = [{
+	ruleId: 'no-process-exit',
+	message: 'Only use process.exit() in CLI apps. Throw an error instead.',
+	type: 'CallExpression'
+}];
+
+test(() => {
+	ruleTester.run('no-process-exit', rule, {
+		valid: [
+			'Process.exit()',
+			'const exit = process.exit;',
+			'x(process.exit)',
+			'#!/usr/bin/env node\nprocess.exit(0);',
+			''
+		],
+		invalid: [
+			{
+				code: 'process.exit();',
+				errors
+			},
+			{
+				code: 'process.exit(1);',
+				errors
+			},
+			{
+				code: 'x(process.exit(1));',
+				errors
+			}
+		]
+	});
+});
--- a/docs/rules/no-process-exit.md
+++ b/docs/rules/no-process-exit.md
@@ -1,0 +1,18 @@
+# Disallow `process.exit()`
+
+This rule is an extension to [ESLint's `no-process-exit`](http://eslint.org/docs/rules/no-process-exit) rule, that allows `process.exit()` to be called in files that start with a hashbang `#!/usr/bin/env node`.
+
+
+## Fail
+
+```js
+process.exit(0);
+```
+
+
+## Pass
+
+```js
+#!/usr/bin/env node
+process.exit(0);
+```
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 
 module.exports = {
 	rules: {
+		'no-process-exit': require('./rules/no-process-exit'),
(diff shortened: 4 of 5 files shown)


jamestalmage:
Has anyone filed an issue w/ ESLint requesting the hashbang exception? Seems like a pretty reasonable improvement to core. 

:+1: for adopting this rule for XO (no matter where it lands).

jfmengels:
Yeah, @sindresorhus did https://github.com/eslint/eslint/issues/2666.

Oh yeah, forgot to mention: Pretty sure I need to do something copyright-wise (I pretty much copy-pasted from https://github.com/eslint/eslint/blob/master/lib/rules/no-process-exit.js and the test file), and wanted to ask whether you guys knew what I should do about it. Keep the original @author tags?

rules/no-process-exit.js
```diff
@@ -0,0 +1,19 @@
...
+		CallExpression: function (node) {
+			if (startsWithHashBang) {
+				return;
+			}
+
+			var callee = node.callee;
+			if (callee.type === "MemberExpression" &&
+					callee.object.name === "process" &&
+					callee.property.name === "exit") {
+				context.report(node, "Don't use process.exit(); throw an error instead.");
```
sindresorhus: Use single-quotes

sindresorhus:
I don't really see how this is a copyright issue. It's so little code anyone would have written it exactly like that.

test/no-process-exit.js
```diff
@@ -0,0 +1,47 @@
...
+	env: {
+		es6: true
+	}
+});
+
+test(() => {
+	ruleTester.run('no-process-exit', rule, {
+		valid: [
+			'Process.exit()',
+			'var exit = process.exit;',
```
sindresorhus: `var` => `const`

I know it doesn't matter, but for niceness.

test/no-process-exit.js
```diff
@@ -0,0 +1,47 @@
...
+			'f(process.exit)',
```
sindresorhus: `f` => `x`

test/no-process-exit.js
```diff
@@ -0,0 +1,47 @@
...
+			''
+		],
+		invalid: [
+			{
+				code: 'process.exit(0);',
+				errors: [{
+					ruleId: 'no-process-exit',
+					message: 'Don\'t use process.exit(); throw an error instead.',
+					type: 'CallExpression'
+				}]
```
sindresorhus: The error is the same so just reuse it like we do in other rules.

test/no-process-exit.js
```diff
@@ -0,0 +1,47 @@
...
+		valid: [
+			'Process.exit()',
+			'var exit = process.exit;',
+			'f(process.exit)',
+			'#!/usr/bin/env node\nprocess.exit(0);',
+			''
+		],
+		invalid: [
+			{
+				code: 'process.exit(0);',
```
sindresorhus: `code: 'process.exit();',`

rules/no-process-exit.js
sindresorhus: > Only use `process.exit()` in CLI apps. Throw and error instead.

rules/no-process-exit.js
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports = function (context) {
+	var startsWithHashBang = context.getSourceCode().lines[0].indexOf('#!') === 0;
+	return {
+		CallExpression: function (node) {
+			if (startsWithHashBang) {
+				return;
+			}
+
+			var callee = node.callee;
```
sindresorhus: add an empty line below here.

rules/no-process-exit.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+module.exports = function (context) {
+	var startsWithHashBang = context.getSourceCode().lines[0].indexOf('#!') === 0;
```
sindresorhus: add an empty line below here.

rules/no-process-exit.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
```
sindresorhus: single-quotes

rules/no-process-exit.js
sindresorhus: Can't you just do the early return here?

jamestalmage:
Re: `Copyright`

I believe the correct way of doing it would be to add the following line to the LICENSE file, underneath the main copyright:

```
no-process-exit rule: Copyright (c) 2013 Nicholas C. Zakas. All rights reserved.
```

If you truly copy-pasted code, you probably should do it that way (no real harm in doing so).

sindresorhus:
@jamestalmage This all applies to the final code and not what's here in this PR. When this PR is done there will be nothing left from the copy-paste.

rules/no-process-exit.js
jfmengels: I thought about that, and lazily did not try it out, but it seems to work :+1:

jfmengels:
Updated. Haven't re-added the author tag.

>  It's so little code anyone would have written it exactly like that.

True, I would have written it in very close to the same code. I mostly copy-pasted it so we'd get the same messages as the core rule (on purpose I mean).

As you guys wish. I'd say that it'd be proper and cost-less to add it (though it is indeed a very simple rule).

sindresorhus:
Woo! Glad to finally be able to use this rule :)

leo-smile

*(Can you ever have enough Leo? I don't think so 😝)*

jfmengels:
Glad you're happy with it :)

(and I don't think it's possible to have enough of Leo)

PR #14, opened 2016-04-18
Add `catch-error-name` rule

### LINKED ISSUE #10: Enforce error name in `catch`
I want to enforce the use of `err` as the error name in try/catch and promises:

```js
try {} catch (err) {}
```

```js
foo.catch(err => {});
```

-

Should be easy to make it fixable.

jamestalmage:
Fixes #10.

Disregards if there are nested CatchClauses.


rules/caught-error-name.js
```diff
@@ -0,0 +1,25 @@
...
+		'CatchClause:exit': function (node) {
+			if (!stack.pop()) {
+				context.report({
+					node: node.param,
+					message: 'The catch clause parameter should be named ' + name
+				});
+			}
+		}
+	};
+};
```
sindresorhus: Can you provide a options schema to validate the option: https://github.com/sindresorhus/eslint-plugin-xo/blob/1c287ab3e0a0c9a57a4194eac6b17d8bbeb7fda6/rules/filename-case.js#L49-L60

rules/caught-error-name.js
```diff
@@ -0,0 +1,25 @@
...
+	var stack = [];
+
+	return {
+		'CatchClause': function (node) {
+			if (stack.length === 1) {
+				stack[0] = true;
+			}
+			stack.push(stack.length > 0 || node.param.name === name);
+		},
+		'CatchClause:exit': function (node) {
```
sindresorhus: Why does it need to wait until exiting the node?

rules/caught-error-name.js
```diff
@@ -0,0 +1,25 @@
...
+			if (!stack.pop()) {
+				context.report({
+					node: node.param,
+					message: 'The catch clause parameter should be named ' + name
```
sindresorhus: End it with a `.`

rules/caught-error-name.js
sindresorhus: To ignore nested catch clauses, right?

rules/caught-error-name.js
sindresorhus: Include autofix?

Should be as easy as:

```js
fix: function (fixer) {
	return fixer.replaceText(node.param, name);
}
```

sindresorhus:
Can it support Promise catch too? `foo.catch(err => {});`

sindresorhus:
`catch-error-name`?

rules/caught-error-name.js
sindresorhus: http://eslint.org/docs/developer-guide/working-with-rules#applying-fixes

sindresorhus:
Needs some docs, but I'd be happy to fix that if you want.

rules/caught-error-name.js
jamestalmage: You would also need to rename all other references to it, which isn't so easy with ESLints current fixer implementation.

rules/caught-error-name.js
jamestalmage: Yes

rules/caught-error-name.js
sindresorhus: Oh, that's a good point. I guess we can skip it for now then.

jfmengels:
LGTM, except for the points already mentionned by @sindresorhus :)

jamestalmage:
> Can it support Promise catch too? `foo.catch(err => {});`

Without type information? Not with 100% reliability. But I think it's pretty safe to assume `.catch(FunctionExpression)` means it's a Promise.

jamestalmage:
Updated.
Now handles promises. 
Added documentation.

jfmengels:
LGTM :+1:

sindresorhus:
LGTM2. Yay :)

68747470733a2f2f6d65646961342e67697068792e636f6d2f6d656469612f75344163337942656b623272362f67697068792e676966

PR #20, opened 2016-04-30
Add `PascalCase` option to the `filename-case` rule

gokaygurcan:
Hello,

I need to use **PascalCase** for one of my projects and I thought it'd be nice to have it in this repo. I know that there're tons of people who uses PascalCase for their class names or model names, etc. So yea, I hope it'll help. 

```bash
$ npm test

> eslint-plugin-xo@0.3.1 test /Users/gokaygurcan/Projects/GitHub/eslint-plugin-xo
> xo && nyc ava


   4 passed
----------------------|----------|----------|----------|----------|----------------|
File                  |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
----------------------|----------|----------|----------|----------|----------------|
 rules/               |    95.38 |     87.5 |      100 |    95.38 |                |
  catch-error-name.js |    90.32 |       80 |      100 |    90.32 |        8,14,18 |
  filename-case.js    |      100 |      100 |      100 |      100 |                |
  no-process-exit.js  |      100 |      100 |      100 |      100 |                |
  throw-new-error.js  |      100 |      100 |      100 |      100 |                |
----------------------|----------|----------|----------|----------|----------------|
All files             |    95.38 |     87.5 |      100 |    95.38 |                |
----------------------|----------|----------|----------|----------|----------------|
```

### DIFF
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -21,7 +21,7 @@ Enforces all linted files to have their names in a certain case style. Default i
 - `foo_bar.test.js`
 - `foo_bar.test_utils.js`
 
-### `PascalCase`
+### `pascalCase`
 
 - `FooBar.js`
 - `FooBar.Test.js`
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -21,6 +21,12 @@ Enforces all linted files to have their names in a certain case style. Default i
 - `foo_bar.test.js`
 - `foo_bar.test_utils.js`
 
+### `PascalCase`
+
+- `FooBar.js`
+- `FooBar.Test.js`
+- `FooBar.TestUtils.js`
+
 
 ## Options
 
--- a/rules/filename-case.js
+++ b/rules/filename-case.js
@@ -3,7 +3,11 @@ var path = require('path');
 var camelCase = require('lodash.camelcase');
 var kebabCase = require('lodash.kebabcase');
 var snakeCase = require('lodash.snakecase');
-var pascalCase = require('pascal-case');
+var upperfirst = require('lodash.upperfirst');
+
+var pascalCase = function (str) {
+	return upperfirst(camelCase(str));
+};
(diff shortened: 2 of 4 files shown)


rules/filename-case.js
```diff
@@ -3,6 +3,7 @@ var path = require('path');
 var camelCase = require('lodash.camelcase');
 var kebabCase = require('lodash.kebabcase');
 var snakeCase = require('lodash.snakecase');
+var pascalCase = require('pascal-case');
```
jamestalmage: This brings in a whole bunch of duplicate dependencies. Instead let's stick with lodash:

```js
var capitalize = require('lodash.capitalize');

var pascalCase = function(string) {
  return capitalize(camelCase(string));
};

**Comment by jamestalmage** [comment]
@gokaygurcan - No need to copy paste the test output. We have a CI server that runs the tests for every pull request.

File: docs/rules/filename-case.md [hunk 2]
```diff
@@ -21,6 +21,12 @@ Enforces all linted files to have their names in a certain case style. Default i
 - `foo_bar.test.js`
 - `foo_bar.test_utils.js`
 
+### `PascalCase`
```
jfmengels: This should start with a lowercase letter, like the name of the option

rules/filename-case.js
gokaygurcan: Hmm, it was the one that I normally use but you're right. I'll update the deps and stick with Lodash only.

docs/rules/filename-case.md
gokaygurcan: Arghh! Fixing asap.

jfmengels:
LGTM :)

sindresorhus:
Thanks @gokaygurcan :)

gokaygurcan:
Thank you all, for your kindness and suggestions. Now, I can use XO without any problem ;)

PR #21, opened 2016-05-01
Handle linting from text in `filename-case`. Fixes #19

### LINKED ISSUE #19: `filename-case` rule should gracefully handle linting from text
If the ESLint (and the one proxied in XO) `fn.lintText()` method is used without passing the filename option, `context.getFilename()` returns `<text>`. The rule should gracefully handle that by returning early.

jfmengels:
Handle linting from text in `filename-case`. Fixes #19


sindresorhus:
Thanks for the quick turnaround :)

jfmengels:
NP ;)

PR #28, opened 2016-05-16
Make exception for numbers in `filename-case`. Fixes #25

### LINKED ISSUE #25: `filename-case` rule false positive
```
spec/iss47-spec.js:9:1
  ✖   9:1   Filename is not in kebab case. Rename it to iss-47-spec.js  xo/filename-case
```

Should we maybe make an exception here? I don't see why we would enforce the number to be independent.

jfmengels:
Make exception for numbers in `filename-case`. Fixes #25.

This should allow (depending on the case) both `iss47` and `iss-47`. Let me know what you guys think.


rules/filename-case.js
```diff
@@ -9,6 +9,29 @@ var pascalCase = function (str) {
 	return upperfirst(camelCase(str));
 };
 
+var numberRegex = /(\d+)/;
+var PLACEHOLDER = 'œœœœœœœœœœ';
```
jfmengels: This is a bit of a hack, but could not find any (open to suggestions) placeholders that would remain after being snakeCased or anything.

sindresorhus: @jfmengels Maybe some non-printable unicode characters?

jfmengels:
I tried using the backspace character as a placeholder, but it doesn't work. I use `\uFFFF` now, which does the trick. Open to suggestions from more unicode-aware people.

sindresorhus:
Perfect. Thanks for fixing @jfmengels :)

sindresorhus:
leo-yay

jfmengels:
No problem :)

Will try to play with the ~~lodash~~ underscore prefixed files tonight

PR #31, opened 2016-05-21
Allow leading underscores in filename. Fixes #30

### LINKED ISSUE #30: `filename-case` rule should allow underscore prefixed names
```
  test/_mock-module-system.js:1:1
  ✖  1:1   Filename is not in kebab case. Rename it to mock-module-system.js  xo/filename-case
```

jfmengels:
Allow leading underscores in filename. Fixes #30

### DIFF
--- a/rules/filename-case.js
+++ b/rules/filename-case.js
@@ -58,6 +58,15 @@ function fixFilename(chosenCase, filename) {
 		.join('.');
 }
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
That was quick. Good stuff. Thanks @jfmengels :)

leo-happy

jfmengels:
My pleasure

PR #34, opened 2016-06-04
Improve test coverage for `catch-error-name`

jfmengels:
Improve test coverage for `catch-error-name` (100% code coverage for the project :smile:) and make sure `catch-error-name` reports multiple errors


sindresorhus:
❤️

PR #35, opened 2016-06-08
Get nice feedback with RuleTester using AVA

jfmengels:
Get nice feedback with RuleTester using AVA.
You know the drill :)


sindresorhus:
👌😎

leo-cat

PR #44, opened 2016-07-27
Allow `no-process-exit` to be used in process event handlers (fixes #32)

### LINKED ISSUE #32: Exceptions to `no-process-exit` rule
Sometimes it does make sense to use `process.exit()`:

```js
process.on('uncaughtException', err => {
	console.error('Exception:', err);
	process.exit(1);
});
```

```js
process.on('SIGINT', () => {
	console.log('Got SIGINT');
	process.exit(1);
});
```

I think we should make an exception when it's inside a `process` event listener.

jfmengels:
Allow `no-process-exit` to be used in process event handlers (fixes #32)

I think that we should probably change the error message to mention it can be used in `process.on(...)`, but not sure. I don't want to invite people to use it everywhere, and the message would be pretty long. What do you think?

### DIFF
--- a/docs/rules/no-process-exit.md
+++ b/docs/rules/no-process-exit.md
@@ -1,6 +1,6 @@
 # Disallow `process.exit()`
 
-This rule is an extension to ESLint's [`no-process-exit` rule](http://eslint.org/docs/rules/no-process-exit), that allows `process.exit()` to be called in files that start with a [hashbang](https://en.wikipedia.org/wiki/Shebang_(Unix)) → `#!/usr/bin/env node`.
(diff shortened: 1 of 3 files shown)


sindresorhus:
Awesome!

> I think that we should probably change the error message to mention it can be used in process.on(...), but not sure. I don't want to invite people to use it everywhere, and the message would be pretty long. What do you think?

Nah. What you have here is good. It's better documented in the docs than the error message.

jfmengels:
:smile:

sindresorhus:
leo-yay

PR #45, opened 2016-07-27
Add `explicit-length-check` rule (fixes #12)

### LINKED ISSUE #12: Explicit length check
I want to enforce doing `if (arr.length > 0) {` over just `if (arr.length) {`. Same with `!arr.length`. I think we can safely assume `.length` is a number. I generally favor explicitness. Should be autofixable.

jfmengels:
Add `explicit-length-check` rule (fixes #12)

### DIFF
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -1,12 +1,12 @@
-# Enforce explicitly checking the length of an array
+# Enforce explicitly comparing the `length` property of a value
 
-Enforce explicitly checking the length of an array in an `if` condition, rather than checking the truthiness of the length.
+Enforce explicitly checking the length of a value array in an `if` condition, rather than checking the truthiness of the length.
 
 ## Fail
 
 ```js
+if (string.length) {}
 if (array.length) {}
-
 if (!array.length) {}
 ```
 
@@ -14,8 +14,8 @@ if (!array.length) {}
 ## Pass
 
 ```js
+if (string.length > 0) {}
 if (array.length > 0) {}
 if (array.length !== 0) {}
-
 if (array.length === 0) {}
 ```
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -1,0 +1,21 @@
+# Enforce explicitly checking the length of an array
+
+Enforce explicitly checking the length of an array in an `if` condition, rather than checking the truthiness of the length.
+
+## Fail
+
+```js
+if (array.length) {}
+
+if (!array.length) {}
+```
+
+
+## Pass
+
+```js
+if (array.length > 0) {}
+if (array.length !== 0) {}
+
+if (array.length === 0) {}
+```
--- a/test/explicit-length-check.js
+++ b/test/explicit-length-check.js
@@ -1,0 +1,56 @@
+import test from 'ava';
+import avaRuleTester from 'eslint-ava-rule-tester';
+import rule from '../rules/explicit-length-check';
+
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true
+	}
+});
+
+const errors = [{
+	ruleId: 'explicit-length-check',
+	message: '`length` property should be compared to a value'
+}];
+
+ruleTester.run('explicit-length-check', rule, {
+	valid: [
+		'array.foo',
+		'array.length',
+		'array.length === 0',
+		'array.length !== 0',
+		'array.length > 0',
+		'if (array.foo) {}',
+		'if (length) {}',
+		'if (array.length > 0) {}',
+		'if (array.length === 0) {}',
+		'if (array.length !== 0) {}',
+		'if (array.length !== 0 && array[0] === 1) {}'
+	],
+	invalid: [
+		{
+			code: 'if (array.length) {}',
+			errors
+		},
+		{
+			code: 'if (!array.length) {}',
+			errors
+		},
+		{
+			code: 'if (array.foo.length) {}',
+			errors
+		},
+		{
+			code: 'if (!!array.length) {}',
+			errors
+		},
+		{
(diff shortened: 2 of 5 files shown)


docs/rules/explicit-length-check.md
```diff
@@ -0,0 +1,21 @@
+# Enforce explicitly checking the length of an array
```
sindresorhus: This applies to any value with a `.length` property though. Like `string`, `TypedArray`, `Buffer`.

docs/rules/explicit-length-check.md
```diff
@@ -0,0 +1,21 @@
...
+if (!array.length) {}
+```
+
+
+## Pass
+
+```js
+if (array.length > 0) {}
+if (array.length !== 0) {}
+
```
sindresorhus: Why empty line here?

docs/rules/explicit-length-check.md
```diff
@@ -0,0 +1,21 @@
+# Enforce explicitly checking the length of an array
+
+Enforce explicitly checking the length of an array in an `if` condition, rather than checking the truthiness of the length.
+
+## Fail
+
+```js
+if (array.length) {}
+
```
sindresorhus: https://github.com/sindresorhus/eslint-plugin-xo/pull/45/files#r72902531

docs/rules/explicit-length-check.md
jfmengels: True.
Enforce explicitly comparing the `length` property of a value?

docs/rules/explicit-length-check.md
jfmengels: I was trying to map the previous incorrect code to the right correct code, by delimiting them by an empty line. Do you prefer the line to be removed?

docs/rules/explicit-length-check.md
sindresorhus: 👍 

And show some examples of other values, like `string`.

docs/rules/explicit-length-check.md
sindresorhus: Oh ok, that was not clear at all. I would just remove the empty lines.

test/explicit-length-check.js
```diff
@@ -0,0 +1,65 @@
...
+		'if ([].length > 0) {}',
+		'if ("".length > 0) {}',
+		'if (array.length === 0) {}',
+		'if (array.length !== 0) {}',
+		'if (array.length !== 0 && array[0] === 1) {}'
+	],
+	invalid: [
+		{
+			code: 'if ([].length) {}',
+			errors
```
sindresorhus: Unrelated to this PR, but I've noticed we always duplicate this a lot. Would it be possible to pass a default error to the `.run()` function? Or add ability for that? That way we would only need to pass the `errors` when it's different.

sindresorhus:
Groovy! Thanks for doing this rule. It's going to help me a lot :)

leo-clapping

test/explicit-length-check.js
jfmengels: There isn't anything like that in the core RuleTester, so we could add that in `eslint-ava-rule-tester`, but then it'd be more than just a wrapper.

We could add that there, or add our own wrapper only in this plugin. Anyway we'd do it, I'd push against doing that, as it would make it harder for newcomers to override the default error, as they would have few examples of it in the codebase.

If we'd go for it, then I'd suggest to also do as is done for the valid cases: allow for test cases to be simple strings. Something like this

```js
const defaultError = {
  // ...
};
ruleTesterWrapper.run('rule-name', rule, {
  valid: [
    'array.foo',
    'array.length',
  ],
  invalid: [
    'if ([].length) {}',
    {
      code: 'if ([].length) {} if ([].length) {}',
      errors: [defaultError, defaultError]
    }
  ]
}, {
  errors: [defaultError]
});
```

jfmengels:
Haha no problem, hope you'll have to do less nitpicking from now on :smile:

test/explicit-length-check.js
sindresorhus: > There isn't anything like that in the core RuleTester, so we could add that in eslint-ava-rule-tester, but then it'd be more than just a wrapper.

Maybe try adding it to ESLint RuleTester first? If they're not interested, we'll just add it to the AVA RuleTester.

> Anyway we'd do it, I'd push against doing that, as it would make it harder for newcomers to override the default error, as they would have few examples of it in the codebase.

That seems like a documentation problem though.

> If we'd go for it, then I'd suggest to also do as is done for the valid cases: allow for test cases to be simple strings. Something like this

I like that.

jfmengels: > That seems like a documentation problem though.

Sure, but in this case, I'd rather stick to what the core RuleTester allows you to do. So that beginners and people familiar with the core RuleTester do not get confused.

This would be a lot more interesting if it allows setting any default options (environment options for instance), and then also the simple string thing I mentioned.

> Maybe try adding it to ESLint RuleTester first

Sounds good. I'm still a bit torn on this, do you feel like creating an issue?

jfmengels: In other rules, we went around this problem using this kind of construct. Might be simpler to do it this way.
https://github.com/sindresorhus/eslint-plugin-xo/blob/master/test/catch-error-name.js#L11-L17

sindresorhus: Then you need to pass in lots of `null` and `true` arguments. I would personally still prefer being able to set defaults.

https://github.com/jfmengels/eslint-ava-rule-tester/issues/1

sindresorhus: > Sounds good. I'm still a bit torn on this, do you feel like creating an issue?

Feel free, but I'm personally done creating issues on ESLint. My views are just too incompatible with theirs and I don't have the energy or time for endless discussions that usually ends up with them closing it as won't fix.

PR #52, opened 2016-10-02
Fixes `xo` typo in `catch-error-name` documentation.

manovotny:
Should be `unicorn` instead of `xo`.


jfmengels:
THanks again @manovotny! :smile:

PR #53, opened 2016-10-02
Fixes `xo` typo in `filename-case` documentation.

manovotny:
Should be `unicorn` instead of `xo`.


jfmengels:
THanks @manovotny! :smile:

PR #57, opened 2016-10-17
add number-literal-case rule - fixes #55

### LINKED ISSUE #55: `number-literal-case` rule
[JS has different number literals.](http://www.2ality.com/2015/04/numbers-math-es6.html) Casing is of course not enforced, because why would you want consistency or readable numbers?...

This rule should enforce lowercase literal identifier and uppercase value.

No option for this as none of the failing cases makes sense.

Should be auto-fixable.

### Fail

```js
0XFF
0xff
0Xff

0B11

0O10
```

### Pass

```js
0xFF

0b11

0o10
```

SamVerschueren:
This rule implements #55. Not sure about the description though.

### DIFF
--- a/docs/rules/number-literal-case.md
+++ b/docs/rules/number-literal-case.md
@@ -1,0 +1,27 @@
+# Enforce a lowercase literal identifier and an uppercase value
+
+Enforces a convention in defining number literals where the literal identifier is written in lowercase and the value in uppercase.
+
+
+## Fail
+
+```js
+const foo = 0XFF;
+const foo = 0xff;
+const foo = 0Xff;
+
+const foo = 0B11;
+
+const foo = 0O10;
+```
+
+
+## Pass
+
+```js
+const foo = 0xFF;
+
+const foo = 0b11;
+
+const foo = 0o10;
+```
--- a/docs/rules/number-literal-case.md
+++ b/docs/rules/number-literal-case.md
@@ -1,4 +1,4 @@
-# Enforce a lowercase identifier and an uppercase value for number literals
+# Enforce lowercase identifier and uppercase value for number literals
 
 Enforces a convention of defining number literals where the literal identifier is written in lowercase and the value in uppercase. Differentiating the casing of the identifier and value clearly separates them and makes your code more readable.
 
--- a/docs/rules/number-literal-case.md
+++ b/docs/rules/number-literal-case.md
@@ -1,6 +1,6 @@
-# Enforce a lowercase literal identifier and an uppercase value
+# Enforce a lowercase identifier and an uppercase value for number literals
 
-Enforces a convention in defining number literals where the literal identifier is written in lowercase and the value in uppercase.
+Enforces a convention of defining number literals where the literal identifier is written in lowercase and the value in uppercase. Differentiating the casing of the identifier and value clearly separates them and makes your code more readable.
 
 
 ## Fail
@@ -9,9 +9,13 @@ Enforces a convention in defining number literals where the literal identifier i
 const foo = 0XFF;
 const foo = 0xff;
 const foo = 0Xff;
+```
 
+```js
 const foo = 0B11;
+```
 
+```js
 const foo = 0O10;
 ```
 
@@ -20,8 +24,12 @@ const foo = 0O10;
 
 ```js
 const foo = 0xFF;
+```
 
+```js
 const foo = 0b11;
+```
 
+```js
 const foo = 0o10;
 ```
--- a/index.js
+++ b/index.js
@@ -18,7 +18,8 @@ module.exports = {
 				'unicorn/filename-case': ['error', {case: 'kebabCase'}],
 				'unicorn/no-abusive-eslint-disable': 'error',
 				'unicorn/no-process-exit': 'error',
-				'unicorn/throw-new-error': 'error'
+				'unicorn/throw-new-error': 'error',
+				'unicorn/number-literal-case': 'error'
 			}
 		}
 	}
--- a/rules/number-literal-case.js
+++ b/rules/number-literal-case.js
@@ -1,0 +1,41 @@
+'use strict';
+const fix = value => {
+	if (!/^0[a-zA-Z]/.test(value)) {
+		return value;
+	}
+
+	const indicator = value[1].toLowerCase();
+	const val = value.substring(2).toUpperCase();
+
+	return `0${indicator}${val}`;
+};
+
+const create = context => {
+	return {
+		Literal: node => {
+			const value = node.raw;
+			const fixedValue = fix(value);
+
+			if (value !== fixedValue) {
+				context.report({
+					node,
+					message: 'Invalid number literal casing',
+					fix: fixer => fixer.replaceText(node, fixedValue)
+				});
+			}
+		}
+	};
+};
+
+module.exports = {
+	meta: {
+		meta: {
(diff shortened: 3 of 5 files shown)


SamVerschueren:
Also not sure about the implementation when I read @sindresorhus his comment

> You just need to tell ESLint that you want to listen to any ExpressionStatement. You then ensure node.expression.type === 'Literal', and you can check its .raw property

That didn't work. Maybe ESLint changed something? Or Sindre didn't drink his morning coffee before writing that :).

sindresorhus:
> That didn't work. Maybe ESLint changed something? Or Sindre didn't drink his morning coffee before writing that :).

☕️🙉 You're right.

docs/rules/number-literal-case.md
```diff
@@ -0,0 +1,27 @@
...
+
+```js
+const foo = 0XFF;
+const foo = 0xff;
+const foo = 0Xff;
+
+const foo = 0B11;
+
+const foo = 0O10;
+```
```
sindresorhus: I think the separation between the types might be clearer as different code-blocks than newlines.

docs/rules/number-literal-case.md
```diff
@@ -0,0 +1,27 @@
+# Enforce a lowercase literal identifier and an uppercase value
+
+Enforces a convention in defining number literals where the literal identifier is written in lowercase and the value in uppercase.
```
sindresorhus: convention in => convention of

docs/rules/number-literal-case.md
```diff
@@ -0,0 +1,27 @@
...
+
```
sindresorhus: We should include a reasoning here. Maybe:

> Differentiating the casing of the identifier and value clearly separates them and makes your code more readable.

Improvement suggestions welcome.

index.js
```diff
@@ -18,7 +18,8 @@ module.exports = {
 				'unicorn/filename-case': ['error', {case: 'kebabCase'}],
 				'unicorn/no-abusive-eslint-disable': 'error',
 				'unicorn/no-process-exit': 'error',
-				'unicorn/throw-new-error': 'error'
+				'unicorn/throw-new-error': 'error',
+				'unicorn/number-literal-case': 'error'
```
sindresorhus: @jfmengels Should we do the same as the AVA plugin here?

docs/rules/number-literal-case.md
```diff
@@ -0,0 +1,27 @@
+# Enforce a lowercase literal identifier and an uppercase value
```
sindresorhus: > Enforce a lowercase identifier and uppercase value for number literals

?

rules/number-literal-case.js
```diff
@@ -0,0 +1,41 @@
...
+module.exports = {
+	meta: {
+		meta: {
+			docs: {
+				description: 'Enforce a lowercase literal identifier and an uppercase value',
+				recommended: 'error'
+			}
+		},
+		fixable: 'code'
+	},
```
sindresorhus: We don't really use the `meta` property. It's not used by ESLint, but the ESLint team uses it for their docs.

@jfmengels ?

rules/number-literal-case.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+const fix = value => {
+	if (!/^0[a-zA-Z]/.test(value)) {
+		return value;
+	}
+
+	const indicator = value[1].toLowerCase();
+	const val = value.substring(2).toUpperCase();
```
sindresorhus: I generally prefer `.slice()` when possible as it's more common. @jfmengels Maybe we should have a rule about that?

sindresorhus:
Congrats on your first ESLint rule, Sam!

1ab7e29c-117e-11e6-9cbb-f0828bacea8e

docs/rules/number-literal-case.md
sindresorhus: Drop the `a` too:

> Enforce lowercase identifier and uppercase value for number literals

rules/number-literal-case.js
SamVerschueren: Oh, good to know. I thought that it was used internally or something. I'll drop it.

SamVerschueren:
> Congrats on your first ESLint rule, Sam!

Thanks! I enjoyed it to be honest :). Glad I did it.

rules/number-literal-case.js
jfmengels: I rarely use either of those. I personally have to experiment with or find the docs for `slice` every time I use it, while substring is more self-explanatory.
But I have a weak opinion on this. Feel free to write a proposal

index.js
jfmengels: Not sure what you mean?

jfmengels:
I can't find the comment about the meta, but I started [using it](https://github.com/jfmengels/eslint-plugin-fp/blob/master/rules/no-class.js#L16-L21) in my other plugins, to:
- [Generate my recommended config](https://github.com/jfmengels/eslint-plugin-fp/blob/master/index.js#L12-L15)
- [Generate my README](https://github.com/jfmengels/eslint-plugin-fp/blob/master/rule-description.js#L27-L31) with all the rules

There is talk in [`eslint-find-rules`](https://github.com/sarbbottam/eslint-find-rules/issues/172) to use the `deprecated` field in core rules and plugins, and there's talk of using that field in plugin like eslint-plugin-react and [`import`](https://github.com/benmosher/eslint-plugin-import/pull/625).

Short to say: I like the meta object, and I think I'll use it more and more.

test/number-literal-case.js
```diff
@@ -0,0 +1,75 @@
...
+	ruleId: 'number-literal-case',
+	message: 'Invalid number literal casing'
+};
+
+ruleTester.run('number-literal-case', rule, {
+	valid: [
+		'const foo = 0xFF',
+		'const foo = 0b11',
+		'const foo = 0o10',
+		`const foo = '0Xff'`
```
jfmengels: This case does not respect the title of the rule `Enforce lowercase identifier and uppercase value for number literals`.

Sure, the casing is not the same, but I thought we were enforcing lowercase then uppercase.

SamVerschueren: @jfmengels It's a string that I test here. When I created the rule, I was on my bike thinkering about this "should I explicitely test if the literal is of type number". This is not needed as the `raw` value of a string is `"'0Xff'"`. With this test, I just wanted to make sure this is not enforced on string values. Does it make sense somehow :p?

sindresorhus:
@jfmengels If we're gonna use the `meta` object, I think that's a separate discussion and we should do it to all the rules at once and wait until we actually use a thing that needs it, like your recommended workflow generator. So I think we should leave it out in this PR. Probably better to do the discussion in eslint-plugin-ava first, as we have more eyes there.

test/number-literal-case.js
jfmengels: Ah, my bad! Didn't look closely enough. That's a good test to have, as you often get surprising results when you're writing tests like that. :+1:

jfmengels:
> If we're gonna use the meta object, I think that's a separate discussion and we should do it to all the rules at once

Absolutely. Will create an issue about it in the AVA plugin repo.

And thanks @SamVerschueren! :D

jfmengels:
Opened an issue at https://github.com/avajs/eslint-plugin-ava/issues/151

PR #58, opened 2016-10-17
Add `no-array-instanceof` rule (fixes #11)

### LINKED ISSUE #11: Prefer `Array.isArray()` over `arr instanceof Array`
The latter doesn't work across realms (frames/tabs in the browser, `vm` in Node.js). I still see people using the latter, so would be nice to prevent them.

-

Should be possible to make this autofixable.

-

pvskarthikeya:
Hey @sindresorhus. Please check the below pull request. I hope everything is fine. Thank you.

### DIFF
--- a/rules/no-array-instanceof.js
+++ b/rules/no-array-instanceof.js
@@ -3,9 +3,8 @@
 const create = context => ({
 	BinaryExpression: node => {
 		if (node.operator === 'instanceof' && node.right.type === 'Identifier' && node.right.name === 'Array') {
-			// Get the source code and exteact the array.
+			// Get the source code and extract the left part.
 			const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
-			console.log(arraySourceCode);
 			context.report({
 				node,
 				message: 'Use `Array.isArray()` instead of `instanceof Array`',
--- a/rules/no-array-instanceof.js
+++ b/rules/no-array-instanceof.js
@@ -1,0 +1,31 @@
+'use strict';
+
+const create = context => ({
+	BinaryExpression: node => {
+		if (node.operator === 'instanceof' && node.right.type === 'Identifier' && node.right.name === 'Array') {
+			if (node.left.type === 'Identifier') {
+				context.report({
+					node,
+					message: 'Use `Array.isArray()` instead of `instanceof Array`',
+					fix: fixer => fixer.replaceText(node, 'Array.isArray(' + node.left.name + ')')
+				});
+			}
+			if (node.left.type === 'ArrayExpression') {
+				// Get the source code and exteact the array
+				const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
+				context.report({
+					node,
+					message: 'Use `Array.isArray()` instead of `instanceof Array`',
+					fix: fixer => fixer.replaceText(node, 'Array.isArray(' + arraySourceCode + ')')
+				});
+			}
+		}
+	}
+});
+
+module.exports = {
+	create,
+	meta: {
+		fixable: 'code'
+	}
+};
--- a/rules/no-array-instanceof.js
+++ b/rules/no-array-instanceof.js
@@ -3,22 +3,14 @@
 const create = context => ({
 	BinaryExpression: node => {
 		if (node.operator === 'instanceof' && node.right.type === 'Identifier' && node.right.name === 'Array') {
-			if (node.left.type === 'Identifier') {
-				context.report({
-					node,
-					message: 'Use `Array.isArray()` instead of `instanceof Array`',
-					fix: fixer => fixer.replaceText(node, `Array.isArray(${node.left.name})`)
-				});
-			}
-			if (node.left.type === 'ArrayExpression') {
-				// Get the source code and exteact the array
-				const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
-				context.report({
-					node,
-					message: 'Use `Array.isArray()` instead of `instanceof Array`',
-					fix: fixer => fixer.replaceText(node, `Array.isArray(${arraySourceCode})`)
-				});
-			}
+			// Get the source code and exteact the array.
+			const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
+			console.log(arraySourceCode);
+			context.report({
+				node,
+				message: 'Use `Array.isArray()` instead of `instanceof Array`',
+				fix: fixer => fixer.replaceText(node, `Array.isArray(${arraySourceCode})`)
+			});
 		}
 	}
 });
--- a/rules/no-array-instanceof.js
+++ b/rules/no-array-instanceof.js
@@ -7,7 +7,7 @@ const create = context => ({
 				context.report({
 					node,
 					message: 'Use `Array.isArray()` instead of `instanceof Array`',
-					fix: fixer => fixer.replaceText(node, 'Array.isArray(' + node.left.name + ')')
+					fix: fixer => fixer.replaceText(node, `Array.isArray(${node.left.name})`)
 				});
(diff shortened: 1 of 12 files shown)


rules/no-array-instanceof.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const create = context => ({
+	BinaryExpression: node => {
+		if (node.operator === 'instanceof' && node.right.type === 'Identifier' && node.right.name === 'Array') {
+			if (node.left.type === 'Identifier') {
+				context.report({
+					node,
+					message: 'Use `Array.isArray()` instead of `instanceof Array`',
+					fix: fixer => fixer.replaceText(node, 'Array.isArray(' + node.left.name + ')')
```
kevva:
```js
`Array.isArray(${node.left.name})`
```

rules/no-array-instanceof.js
```diff
@@ -0,0 +1,31 @@
...
+				});
+			}
+			if (node.left.type === 'ArrayExpression') {
+				// Get the source code and exteact the array
+				const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
+				context.report({
+					node,
+					message: 'Use `Array.isArray()` instead of `instanceof Array`',
+					fix: fixer => fixer.replaceText(node, 'Array.isArray(' + arraySourceCode + ')')
```
kevva:
```js
`Array.isArray(${arraySourceCode})`
```

test/no-array-instanceof.js
```diff
@@ -0,0 +1,69 @@
...
+ruleTester.run('avoid-array-instanceof', rule, {
+	valid: [
+		'Array.isArray(arr)',
+		'Array.isArray(arr) === true',
+		'Array.isArray([])',
+		'Array.isArray([]) === true',
+		'Array.isArray([1,2,3])',
+		'Array.isArray([1,2,3]) === true',
+		'if(Array.isArray(arr)){}',
+		'if(Array.isArray([1,2,3])){}'
```
jfmengels: In the sense of testing your rule, these cases do not test the rule's boundaries, as you never test the presence of `Array.isArray`. You should not remove them, but you should add tests containing `instanceof` comparisons and the sorts. Here are a few examples

// Valid
```js
a instanceof array;
a instanceof 'array';
a instanceof ArrayA;
a instanceof Object;
```

// Invalid
```js
a.b instanceof Array;
foo() instanceof Array;
foo`bar${value}` instanceof Array;
```

rules/no-array-instanceof.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+const create = context => ({
+	BinaryExpression: node => {
+		if (node.operator === 'instanceof' && node.right.type === 'Identifier' && node.right.name === 'Array') {
+			if (node.left.type === 'Identifier') {
```
pvskarthikeya: Yup. I just realized that i dont need to explicitly check for `ArrayExpression` or an identifier. All i need is if i have `instanceof Array`. Thank you.

rules/no-array-instanceof.js
```diff
@@ -3,22 +3,14 @@
...
-				const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
-				context.report({
-					node,
-					message: 'Use `Array.isArray()` instead of `instanceof Array`',
-					fix: fixer => fixer.replaceText(node, `Array.isArray(${arraySourceCode})`)
-				});
-			}
+			// Get the source code and exteact the array.
+			const arraySourceCode = context.getSourceCode().text.slice(node.left.range[0], node.left.range[1]);
+			console.log(arraySourceCode);
```
jfmengels: forgot to remove this one ;)

test/no-array-instanceof.js
```diff
@@ -16,13 +18,11 @@ const errors = [{
...
-		'Array.isArray([]) === true',
-		'Array.isArray([1,2,3])',
-		'Array.isArray([1,2,3]) === true',
-		'if(Array.isArray(arr)){}',
-		'if(Array.isArray([1,2,3])){}'
+		'arr instanceof Object',
+		'arr instanceof array',
+		'a instanceof \'array\'',
+		'a instanceof ArrayA',
+		`a.x[${value}] instanceof Object`
```
jfmengels: Don't bother with creating a variable. In this case you're just writing `"a.x[2]` anyway.
Replace this by:
```js
a.x[2] instanceof Object;
```

test/no-array-instanceof.js
```diff
@@ -31,38 +31,23 @@ ruleTester.run('avoid-array-instanceof', rule, {
...
-			code: 'if(arr instanceof Array){}',
-			output: 'if(Array.isArray(arr)){}',
+			code: 'obj.arr instanceof Array',
+			output: 'Array.isArray(obj.arr)',
 			errors
 		},
 		{
-			code: 'if([1,2,3] instanceof Array){}',
-			output: 'if(Array.isArray([1,2,3])){}',
+			code: `foo.bar${value} instanceof Array`,
```
jfmengels: Same as before. Do you mean to have a test with a template literal? Because as it is right now, you are making a test case that will evaluate to `foo.bar2 instanceof Array`

test/no-array-instanceof.js
```diff
@@ -31,38 +31,23 @@ ruleTester.run('avoid-array-instanceof', rule, {
...
-			output: 'Array.isArray([]) === true',
-			errors
-		},
-		{
-			code: '[1,2,3] instanceof Array',
-			output: 'Array.isArray([1,2,3])',
-			errors
-		},
-		{
-			code: '[1,2,3] instanceof Array === true',
```
jfmengels: You should keep this case and the others that use a comparison.

test/no-array-instanceof.js
jfmengels: You should have a test where the right side of the operator is not an identifier (ok, you already have strings, but another test can't hurt). Maybe `arr instanceof foo()`

pvskarthikeya: Okay. @jfmengels  Thank you.

test/no-array-instanceof.js
pvskarthikeya: @jfmengels I wanted to see how the rule would handle this case. I tried the same template literal in AST explorer and it doesn't evaluate to BinaryExpression at all. All it does is parse it as TemplateLiteral. So, I am currently confused how the tests are passing. :(

[AST Explorer Sample](https://astexplorer.net/#/hyq1piy29D)

jfmengels: Did you mean to do [this](https://astexplorer.net/#/hyq1piy29D/1)?

jfmengels: > What I wanted to see was if the instanceof operator was part of a string, what would happen? (Sorry, i am just curious, I have no experience with AST's ).

And that is a very good question to ask yourself :)
If you're not well versed with AST, then you are bound to get surprises (and even if you are :sweat_smile:)

I asked myself the same question, but could not find a test case that would make much sense (`arr "instanceof" Array`?) . Feel free to add a `"arr instanceof Array"` though!

sindresorhus:
> this rule and the numbers literals case have been added as errors in the recommended config. Are you planning a major release for these next or should they be turned off?

Yes, planning to do a new major.

sindresorhus:
@pvskarthikeya Looks very good now. As @jfmengels mentions, this needs docs and to be added to the readme. See existing rules docs on how it's done: https://github.com/sindresorhus/eslint-plugin-unicorn/tree/master/docs/rules And example of what needs to be added to the readme: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/718578aadc696dbc506bce32cd8e5bec3fb12c3a#diff-0730bb7c2e8f9ea2438b52e419dd86c9L39

sindresorhus:
Yay! Very nice work @pvskarthikeya 🌟

Keep the pull requests coming :)

PR #61, opened 2016-10-20
Add `no-new-buffer` rule - fixes #59

### LINKED ISSUE #59: Enforce `Buffer.from` instead of `new Buffer`
Since `new Buffer()` is being deprecated since Node.js >= 6, it might be nice if we could enforce this some day in the future. Not sure if it makes sense to already create this rule because ESLint can't target specific Node.js versions, right?

SamVerschueren:
:tada: Feedback more then welcome!

I was struggling with using `fixer.replaceText` and thought I had to parse the arguments myself untill I found out about `fixer.replaceTextRange` :).

### DIFF
--- a/docs/rules/no-new-buffer.md
+++ b/docs/rules/no-new-buffer.md
@@ -1,0 +1,27 @@
+# Enforce the use of `Buffer.from()` and `Buffer.alloc()` instead of the deprecated `new Buffer()`
+
+Enforces the use of [Buffer.from](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array) and [Buffer.alloc()](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_alloc_size_fill_encoding) instead of [new Buffer()](https://nodejs.org/api/buffer.html#buffer_new_buffer_array), which has been deprecated since Node.js 4.
+
+
+## Fail
+
+```js
+const buf = new Buffer('7468697320697320612074c3a97374', 'hex');
+const buf = new Buffer([0x62, 0x75, 0x66, 0x66, 0x65, 0x72]);
+```
+
+```js
+const buf = new Buffer(10);
+```
+
+
+## Pass
+
+```js
+const buf = Buffer.from('7468697320697320612074c3a97374', 'hex');
+const buf = Buffer.from([0x62, 0x75, 0x66, 0x66, 0x65, 0x72])
+```
+
+```js
+const buf = Buffer.alloc(10);
+```
--- a/readme.md
+++ b/readme.md
@@ -40,7 +40,8 @@ Configure it in `package.json`.
 			"unicorn/no-process-exit": "error",
 			"unicorn/throw-new-error": "error",
 			"unicorn/number-literal-case": "error",
-			"unicorn/no-array-instanceof": "error"
+			"unicorn/no-array-instanceof": "error",
+			"unicorn/no-new-buffer": "error"
 		}
 	}
 }
@@ -57,6 +58,7 @@ Configure it in `package.json`.
 - [throw-new-error](docs/rules/throw-new-error.md) - Require `new` when throwing an error. *(fixable)*
 - [number-literal-case](docs/rules/number-literal-case.md) - Enforce lowercase identifier and uppercase value for number literals. *(fixable)*
 - [no-array-instanceof](docs/rules/no-array-instanceof.md) - Require `Array.isArray()` instead of `instanceof Array`. *(fixable)*
+- [no-new-buffer](docs/rules/no-new-buffer.md) - Enforce the use of `Buffer.from()` and `Buffer.alloc()` instead of the deprecated `new Buffer()`. *(fixable)*
 
 
 ## Recommended config
--- a/rules/no-new-buffer.js
+++ b/rules/no-new-buffer.js
@@ -1,0 +1,29 @@
+'use strict';
+const inferMethod = args => (args.length > 0 && typeof args[0].value === 'number') ? 'alloc' : 'from';
+
+const create = context => {
+	return {
+		NewExpression: node => {
+			if (node.callee.name === 'Buffer') {
+				const method = inferMethod(node.arguments);
+				const range = [
+					node.start,
+					node.callee.end
+				];
+
+				context.report({
+					node,
+					message: `\`new Buffer()\` is deprecated, use \`Buffer.${method}()\` instead.`,
+					fix: fixer => fixer.replaceTextRange(range, `Buffer.${method}`)
+				});
+			}
+		}
+	};
+};
+
+module.exports = {
+	create,
+	meta: {
+		fixable: 'code'
+	}
+};
--- a/index.js
+++ b/index.js
@@ -20,7 +20,8 @@ module.exports = {
 				'unicorn/no-process-exit': 'error',
 				'unicorn/throw-new-error': 'error',
 				'unicorn/number-literal-case': 'error',
-				'unicorn/no-array-instanceof': 'error'
+				'unicorn/no-array-instanceof': 'error',
+				'unicorn/no-new-buffer': 'error'
 			}
 		}
 	}
--- a/test/no-new-buffer.js
+++ b/test/no-new-buffer.js
@@ -1,0 +1,80 @@
+import test from 'ava';
+import avaRuleTester from 'eslint-ava-rule-tester';
+import rule from '../rules/no-new-buffer';
+
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true
+	},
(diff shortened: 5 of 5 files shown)


docs/rules/no-new-buffer.md
```diff
@@ -0,0 +1,27 @@
+# Enforce the use of `Buffer.from()` and `Buffer.alloc()` instead of the deprecated `new Buffer()`
+
+Enforces the use of [Buffer.from](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_array) and [Buffer.alloc()](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_alloc_size_fill_encoding) instead of [new Buffer()](https://nodejs.org/api/buffer.html#buffer_new_buffer_array) which has been deprecated.
```
sindresorhus: Should there be a comma before `which has been deprecated.`?

docs/rules/no-new-buffer.md
```diff
@@ -0,0 +1,27 @@
...
+
```
sindresorhus: Maybe we should be explicit about those two methods being available in Node.js 4 and up.

readme.md
```diff
@@ -55,7 +55,7 @@ Configure it in `package.json`.
 - [no-process-exit](docs/rules/no-process-exit.md) - Disallow `process.exit()`.
 - [throw-new-error](docs/rules/throw-new-error.md) - Require `new` when throwing an error. *(fixable)*
 - [number-literal-case](docs/rules/number-literal-case.md) - Enforce lowercase identifier and uppercase value for number literals. *(fixable)*
-
```
sindresorhus: Don't remove the newline.

readme.md
```diff
@@ -55,7 +55,7 @@ Configure it in `package.json`.
...
+- [no-new-buffer](docs/rules/no-new-buffer.md) - Enforce the use of `Buffer.from()` and `Buffer.alloc()` instead of the deprecated `new Buffer()`
```
sindresorhus: Needs to be marked as fixable.

rules/no-new-buffer.js
```diff
@@ -0,0 +1,35 @@
...
+			if (node.callee.name === 'Buffer') {
+				const method = inferMethod(node.arguments);
+				const range = [
+					node.start,
+					node.callee.end
+				];
+
+				context.report({
+					node,
+					message: `\`new Buffer()\` is being deprecated, use \`Buffer.${method}()\` instead`,
```
sindresorhus: Drop the `is being` part. It *is* deprecated.

rules/no-new-buffer.js
```diff
@@ -0,0 +1,35 @@
...
+			}
+		}
+	};
+};
+
+module.exports = {
+	meta: {
+		fixable: 'code'
+	},
+	create
```
sindresorhus: Can you put `create` at the top to be consistent with the other rules?

rules/no-new-buffer.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+const inferMethod = args => {
+	if (args.length > 0 && typeof args[0].value === 'number') {
+		return 'alloc';
+	}
+
+	return 'from';
+};
```
sindresorhus: This could be a one-liner.

docs/rules/no-new-buffer.md
SamVerschueren: Not sure, which one do you think sounds better :D?

docs/rules/no-new-buffer.md
SamVerschueren: Good point btw, people should be well informed.

docs/rules/no-new-buffer.md
SamVerschueren: With the new sentence `which has been deprecated since Node.js 4`, I believe a comma makes sense.

sindresorhus:
@SamVerschueren Can you fix the merge conflict?

kevva:
Nice work with this one!

docs/rules/no-new-buffer.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Fail
+
+```js
+const buf = new Buffer('7468697320697320612074c3a97374', 'hex');
+const buf = new Buffer([0x62, 0x75, 0x66, 0x66, 0x65, 0x72]);
+```
+
+```js
```
jfmengels: Why the two distinct code blocks?

SamVerschueren:
The [first invalid](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/61/files#diff-bdd2d7fa0e7a33141bf1a62d561e2fc8R33) test is a test with `new Buffer()`

And the two last invalid tests use an identifier. Or maybe you mean something else?

docs/rules/no-new-buffer.md
SamVerschueren: Good question, I wanted to make a difference between `new Buffer` being converted to `Buffer.from` and the `new Buffer` being converted to `Buffer.alloc` as you can see in the `Pass` section below. To keep those blocks consistent, I splitted them up. I could merge them though, whatever you guys prefer.

jfmengels:
Ah, my bad! LGTM then.

> But the fix should probably not fix if it can't determine the right method to use, otherwise you could inject bugs in the codebase.

Still wondering about this. But I leave that to someone who knows it better.

SamVerschueren:
> But the fix should probably not fix if it can't know for sure, otherwise you could inject bugs in the codebase.

You mean `new Buffer()` to `Buffer.from()` which is actually invalid? I agree, this is an edge-case and I went for converting it. `new Buffer()` is invalid as well though. If someone else has a strong opinion on this, feel free to chime in :).

jfmengels:
> You mean new Buffer() to Buffer.from() which is actually invalid?

I mean this

```js
let value = 10;
new Buffer(value) // should be fixed to Buffer.alloc
value = 'foo';
new Buffer(value) // should be fixed to Buffer.from
```
With most expressions or identifiers, you can't know for sure which fix to apply. Maybe not apply a fix when you can't determine that, and advise to use either `alloc` or `from`.

SamVerschueren:
Darn, good catch. Even haven't though about that. Maybe not fixing that might be better indeed... I will look at that.

sindresorhus:
> Maybe not apply a fix when you can't determine that, and advise to use either alloc or from.

I think it's safe to always assume `Buffer.from()` as even if we guess incorrectly, `const foo = 10; Buffer.from(foo);` would throw, as `Buffer.from()` doesn't accept a number. In addition, I've never seen anyone use a number there, so very much and edge-case.

> Welcome to the world of I wish JS had types so that I could know the type of X statically :D

Flow could have easily inferred it in this case without any type info. In most cases actually, it's ESLint being naive, not lack of types that is the show stopper.

jfmengels:
> Flow could have easily inferred it in this case without any type info. In most cases actually, it's ESLint being naive, not lack of types that is the show stopper.

It probably could have in most cases yes. But you're right, ESLint could infer stuff on its own.

sindresorhus:
So, are you guys good with merging this as is?

SamVerschueren:
I'm ok. If someone complains, I could always try to track down the value of the reference by traversing the tree. But that might be a little bit overkill for now.

jfmengels:
I think it's dangerous to apply the fix whenever, but if you think it's a sufficient edge case, and that people will notice it soon enough, let's merge this. But I have a bad gut feeling about doing so TBH.

SamVerschueren:
Should I look into traversing the tree and try to infer the type ourselves? Could be easily extracted in a separate library. If the type could not be inferred (easily), we do not apply the fix, if it could be inferred, we apply it. Not sure if possible though, but could be very useful for other rules if we had a basic type inference system.

sindresorhus:
> I think it's dangerous to apply the fix whenever

You're right, but I don't think this is practically going to affect anyone. Only very few libraries allocates the buffer size, and those will get an error about incorrect usage from `Buffer.from()` anyways, so what is really the harm other than a slightly incorrect transformation? It won't introduce any hidden bugs. We have to weigh complexity against risk, and I don't think the risk outweighs the complexity here.

jfmengels:
If you feel that way, let's do it.

jfmengels:
Thanks again @SamVerschueren :D

sindresorhus:
I still think adding better type inference is a good idea. Especially if we can make something reusable as many rules would be better with some light type inference. Type inference could also increase the chances here of guessing the correct type, and making @jfmengels a little bit more at ease.

sindresorhus:
Thanks @SamVerschueren. Great as always :)

SamVerschueren:
Nice, thanks for all the feedback guys! I will add it to my todo list :).

PR #62, opened 2016-10-20
Add `no-hex-escape` rule - fixes #51

### LINKED ISSUE #51: Prefer unicode escapes over hexadecimal escapes
`\x1B` → `\u001B`

For consistency.

[Reading material.](https://mathiasbynens.be/notes/javascript-escapes#unicode)

-

Rule name suggestion?

-

Should be auto-fixable.

matijs:
ToDo:

- [x] docs
- [x] process feedback :]
- [x] apply the rule for TemplateLiterals

### DIFF
--- a/docs/rules/no-hex-escape.md
+++ b/docs/rules/no-hex-escape.md
@@ -1,15 +1,11 @@
-# Enforce the use of unicode encoded escapes instead of hexadecimal encoded escapes
+# Enforce the use of unicode escapes instead of hexadecimal escapes
 
-Enforces a convention of using unicode encoded escapes instead of hexadecimal encoded escapes to enfoce code consistency. Works on both Literals and TemplateLiterals.
+Enforces a convention of using unicode escapes instead of hexadecimal escapes for consistency and clarity.
 
 ## Fail
 
 ```js
 const foo = '\x1B';
-const foo = '\x1b';
-```
-
-```js
 const foo = `\x1B${bar}`;
 ```
 
@@ -17,9 +13,5 @@ const foo = `\x1B${bar}`;
 
 ```js
 const foo = '\u001B';
-const foo = '\u001b';
-```
-
-```js
 const foo = `\u001B${bar}`;
 ```
--- a/docs/rules/no-hex-escape.md
+++ b/docs/rules/no-hex-escape.md
@@ -1,0 +1,25 @@
+# Enforce the use of unicode encoded escapes instead of hexadecimal encoded escapes
+
+Enforces a convention of using unicode encoded escapes instead of hexadecimal encoded escapes to enfoce code consistency. Works on both Literals and TemplateLiterals.
+
+## Fail
+
+```js
+const foo = '\x1B';
+const foo = '\x1b';
+```
+
+```js
+const foo = `\x1B${bar}`;
+```
+
+## Pass
+
+```js
+const foo = '\u001B';
+const foo = '\u001b';
+```
+
+```js
+const foo = `\u001B${bar}`;
+```
--- a/rules/no-hex-escape.js
+++ b/rules/no-hex-escape.js
@@ -12,7 +12,7 @@ const create = context => {
 			if (value !== fixedValue) {
 				context.report({
 					node,
-					message: 'Use unicode escapes over hexadecimal escapes',
+					message: 'Use unicode escape codes instead of hexadecimal escapes.',
 					fix: fixer => fixer.replaceText(node, fixedValue)
 				});
 			}
--- a/rules/no-hex-escape.js
+++ b/rules/no-hex-escape.js
@@ -27,8 +27,8 @@ const create = context => {
 };
 
 module.exports = {
+	create,
 	meta: {
 		fixable: 'code'
-	},
-	create
+	}
 };
--- a/rules/no-hex-escape.js
+++ b/rules/no-hex-escape.js
@@ -1,13 +1,7 @@
 'use strict';
 const hexRE = /\\x[a-fA-F0-9]{2}/;
 
-const fix = value => {
-	if (!hexRE.test(value)) {
-		return value;
-	}
-
-	return value.replace(/\\x/g, '\\u00');
-};
+const fix = value => hexRE.test(value) ? value.replace(/\\x/g, '\\u00') : value;
 
 const create = context => {
 	return {
--- a/rules/no-hex-escape.js
+++ b/rules/no-hex-escape.js
@@ -1,14 +1,12 @@
 'use strict';
-const hexRE = /\\x[a-fA-F0-9]{2}/;
-
 function checkEscape(context, node, value) {
-	const fixedValue = hexRE.test(value) ? value.replace(/\\x/g, '\\u00') : value;
+	const fixedValue = typeof value === 'string' ? value.replace(/\\x/g, '\\u00') : value;
 
 	if (value !== fixedValue) {
 		context.report({
 			node,
-			message: 'Use unicode escape codes instead of hexadecimal escapes.',
-			fix: fixer => fixer.replaceText(node, fixedValue)
+			message: 'Use unicode escapes instead of hexadecimal escapes.',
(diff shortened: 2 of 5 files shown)


rules/no-hex-escape.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const fix = value => {
+	if (!/\\x[a-fA-F0-9]{2}/.test(value)) {
```
jfmengels: Assign `/\\x[a-fA-F0-9]{2}/` to a top-level variable. That way, the regex won't get recreated at every iteration (and every time the linter encounters a Literal).

test/no-hex-escape.js
```diff
@@ -0,0 +1,62 @@
...
+		{
+			code: `const foo = 'foo\\xb1'`,
+			errors: [error],
+			output: `const foo = 'foo\\u00b1'`
+		},
+		{
+			code: `const foo = 'foo\\x12foo\\x34'`,
+			errors: [error],
+			output: `const foo = 'foo\\u0012foo\\u0034'`
+		}
```
jfmengels: Can you add tests with numeral values, because they are considered as `Literal` too. At least one like this
```
const foo = 3;
```

test/no-hex-escape.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+ruleTester.run('no-hex-escape', rule, {
+	valid: [
+		`const foo = 'foo'`,
+		`const foo = '\\u00b1'`,
+		`const foo = '\\u00b1\\u00b1'`,
+		`const foo = 'foo\\u00b1'`,
+		`const foo = 'foo\\u00b1foo'`,
+		`const foo = '\\u00b1foo'`
```
jfmengels: Can you add tests with numeral values, because they are considered as `Literal` too. At least one like this
```
const foo = 3;
```

rules/no-hex-escape.js
```diff
@@ -0,0 +1,33 @@
...
+			}
+		}
+	};
+};
+
+module.exports = {
+	meta: {
+		fixable: 'code'
+	},
+	create
```
SamVerschueren: `create` should be placed before `meta` to keep it in line with other rules.

rules/no-hex-escape.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const fix = value => {
```
SamVerschueren: Not sure if @sindresorhus prefers to remove this empty line. If he prefers that, maybe we should write a rule for that :D.

rules/no-hex-escape.js
```diff
@@ -0,0 +1,33 @@
...
+const create = context => {
+	return {
+		Literal: node => {
+			const value = node.raw;
+			const fixedValue = fix(value);
+
+			if (value !== fixedValue) {
+				context.report({
+					node,
+					message: 'Use unicode escapes over hexadecimal escapes',
```
sindresorhus: `over` => `instead of`

sindresorhus: And end it in a `.`

rules/no-hex-escape.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+
+const fix = value => {
+	if (!/\\x[a-fA-F0-9]{2}/.test(value)) {
+		return value;
+	}
+
+	return value.replace(/\\x/g, '\\u00');
+};
```
sindresorhus: This function could be a one-liner with a ternary.

rules/no-hex-escape.js
matijs: Not entirely sure what you mean :> could you clarify a bit?

SamVerschueren: ```message: `Use unicode escape codes instead of hexadecimal escapes.` ```

matijs: Ah :) Misread it as do something something with a fat arrow :') Fixed now.

docs/rules/no-hex-escape.md
```diff
@@ -0,0 +1,25 @@
+# Enforce the use of unicode encoded escapes instead of hexadecimal encoded escapes
```
sindresorhus: > Enforce unicode escapes instead of hexadecimal escapes

Same with the readme description.

docs/rules/no-hex-escape.md
```diff
@@ -0,0 +1,25 @@
...
+
+Enforces a convention of using unicode encoded escapes instead of hexadecimal encoded escapes to enfoce code consistency. Works on both Literals and TemplateLiterals.
```
sindresorhus: `to enforce code consistency` => `for consistency and clarity`

sindresorhus: Drop the `Works on both Literals and TemplateLiterals.` part. It's already illustrated in the example.

docs/rules/no-hex-escape.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Fail
+
+```js
+const foo = '\x1B';
+const foo = '\x1b';
```
sindresorhus: Drop this. We only need one example. Casing doesn't matter here.

docs/rules/no-hex-escape.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+```js
```
sindresorhus: Merge this code block with the above.

docs/rules/no-hex-escape.md
```diff
@@ -0,0 +1,25 @@
...
+const foo = `\x1B${bar}`;
+```
+
+## Pass
+
+```js
+const foo = '\u001B';
+const foo = '\u001b';
```
sindresorhus: Drop this.

docs/rules/no-hex-escape.md
```diff
@@ -0,0 +1,25 @@
...
+```
+
+```js
```
sindresorhus: Merge codeblock with above.

rules/no-hex-escape.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+const hexRE = /\\x[a-fA-F0-9]{2}/;
+
+function checkEscape(context, node, value) {
+	const fixedValue = hexRE.test(value) ? value.replace(/\\x/g, '\\u00') : value;
+
+	if (value !== fixedValue) {
+		context.report({
+			node,
+			message: 'Use unicode escape codes instead of hexadecimal escapes.',
```
sindresorhus: > Use unicode escapes instead of hexadecimal escapes.

test/no-hex-escape.js
```diff
@@ -0,0 +1,75 @@
...
+			errors: [error],
+			output: `const foo = 'foo\\u0012foo\\u0034'`
+		},
+		{
+			code: `const foo = '42\\x1242\\x34'`,
+			errors: [error],
+			output: `const foo = '42\\u001242\\u0034'`
+		},
+		{
+			code: 'const foo = `\\xb1`',
```
sindresorhus: Add some more template literal tests, especially with multiple elements. Also add more passing tests like that.

See: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/63#discussion_r84426600

rules/no-hex-escape.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+const hexRE = /\\x[a-fA-F0-9]{2}/;
+
+function checkEscape(context, node, value) {
+	const fixedValue = hexRE.test(value) ? value.replace(/\\x/g, '\\u00') : value;
```
sindresorhus: Why test first and then replace? Why not just replace directly?

jfmengels: `value` might not be a string, it could be a number.

jfmengels:
LGTM, thanks again @matijs :)

sindresorhus:
Excellent work @matijs. Hope you had fun :)

superhighfive

PR #63, opened 2016-10-20
Add `escape-case` rule - fixes #54

### LINKED ISSUE #54: `escape-case` rule
In JS you can write escapes with either uppercase or lowercase value. Would be useful to have a rule to enforce the value to be uppercase. Don't want an option for this as lowercase is a bad practice.

For example `\xa9` and `\xA9` are equivalent, but the latter is clearer. `\ud834\udf06` and `\uD834\uDF06` are too.

The following should be supported (the values are just there as an example):
- `\xA9`
- `\uD834`
- `\u{1D306}`
- `\cA`
## 

Suggestions for a better rule name?

Alternative: `uppercase-escape`
## 

Would be useful if it works in a regex too, though not essential for the first version: https://github.com/sindresorhus/filename-reserved-regex/blob/af9c88cd73ba944ec27c64173aa9017ec80e7d2a/index.js#L4
## 

Should also be possible to auto-fix it, right?

Bebersohl:
Hello :smile:

Right now the rule only applies to string literals and not declarations. I also had trouble with the auto fix feature and could use some help on that.


docs/rules/escape-case.md
```diff
@@ -0,0 +1,22 @@
...
+
+Enforces a convention of defining escape sequences with uppercase characters rather than lowercase ones.
+
+## Fail
+
+```js
+var foo = "\xa9"
+var foo = "\ud834"
+var foo = "\u{1d306}"
+var foo = "\ca"
```
sindresorhus: Use `const` instead of `var` and semicolons and single-quotes.

readme.md
```diff
@@ -55,7 +56,7 @@ Configure it in `package.json`.
 - [no-process-exit](docs/rules/no-process-exit.md) - Disallow `process.exit()`.
 - [throw-new-error](docs/rules/throw-new-error.md) - Require `new` when throwing an error. *(fixable)*
 - [number-literal-case](docs/rules/number-literal-case.md) - Enforce lowercase identifier and uppercase value for number literals. *(fixable)*
-
```
sindresorhus: Don't remove the newline.

rules/escape-case.js
```diff
@@ -0,0 +1,23 @@
+const create = context => {
```
sindresorhus: Missing `'use strict';` statement at the top. See the other rule files.

test/escape-case.js
```diff
@@ -0,0 +1,41 @@
...
+		{
+			code: 'var foo = "\\ud834"',
+			errors
+		},
+		{
+			code: 'var foo = "\\u{1d306}"',
+			errors
+		},
+		{
+			code: 'var foo = "\\ca"',
```
sindresorhus: Use const instead of var and semicolons and single-quotes in all of these.

rules/escape-case.js
```diff
@@ -0,0 +1,23 @@
...
+			if (typeof node.value !== 'string') {
+				return;
+			}
+
+			const match = node.raw.match(/\\(x[a-f0-9]{2})|(u[a-f0-9]{4})|(u\{([0-9a-f]{1,})\})|(c[a-z])/);
+
+			if (match) {
+				context.report({
+					node,
+					message: 'Use uppercase characters for escape sequences'
```
sindresorhus: Should mention it's the value part of the escape sequence.

docs/rules/escape-case.md
```diff
@@ -0,0 +1,22 @@
+# Enforce escape sequences to use uppercase values
+
+Enforces a convention of defining escape sequences with uppercase characters rather than lowercase ones.
+
```
sindresorhus: Should include reasoning for why this rule exists. Basically that it's more readable to have the value uppercase so it's distinct from the escape identifier.

docs/rules/escape-case.md
```diff
@@ -0,0 +1,22 @@
+# Enforce escape sequences to use uppercase values
```
sindresorhus: Maybe use `Require` instead of `Enforce`

sindresorhus:
Nice start @Bebersohl :)

> I also had trouble with the auto fix feature and could use some help on that.

I would recommend looking at how it was solved for the `number-literal-case`. Basically, always fix it and for the validation part check if the original value matches the fixed value: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d576ea528d6ea341f7336ce65174ec14f11af7a7/rules/number-literal-case.js#L19

sindresorhus:
Should also work on `TemplateLiteral`, see: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/62#pullrequestreview-5061720

rules/escape-case.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const create = context => {
+	return {
+		Literal(node) {
+			if (typeof node.value !== 'string') {
+				return;
+			}
+
+			const match = node.raw.match(/\\(x[a-f0-9]{2}|u[a-f0-9]{4}|u\{([0-9a-f]{1,})\}|c[a-z])/);
```
sindresorhus: This should probably work on `node.value` instead.

rules/escape-case.js
```diff
@@ -0,0 +1,55 @@
...
+
+			if (match) {
+				context.report({
+					node,
+					message: 'Use uppercase characters for the value of the escape sequence',
+					fix: fixer => fixer.replaceText(node, fix(node.raw))
```
sindresorhus: Usually safer to use `fixer.replaceRange()` whenever possible. (Using `node.value` instead of `node.raw`).

rules/escape-case.js
```diff
@@ -0,0 +1,55 @@
...
+				if (typeof element.value.raw !== 'string') {
+					return;
+				}
+
+				const match = element.value.raw.match(/\\(x[a-f0-9]{2}|u[a-f0-9]{4}|u\{([0-9a-f]{1,})\}|c[a-z])/);
+
+				if (match) {
+					context.report({
+						node,
+						message: 'Use uppercase characters for the value of the escape sequence',
```
sindresorhus: The message should also end in a `.`

rules/escape-case.js
```diff
@@ -0,0 +1,55 @@
...
+			if (match) {
+				context.report({
+					node,
+					message: 'Use uppercase characters for the value of the escape sequence',
+					fix: fixer => fixer.replaceText(node, fix(node.raw))
+				});
+			}
+		},
+
+		TemplateLiteral(node) {
```
jfmengels: I agree. That would make the fixer much simpler as it would target only the static part.

test/escape-case.js
```diff
@@ -0,0 +1,61 @@
...
+		{
+			code: 'const foo = \'\\ca\';',
+			errors
+		},
+		{
+			code: 'const foo = `\\xa9`;',
+			errors
+		},
+		{
+			code: 'const foo = `\\ud834`;',
```
jfmengels: Could you add cases with template literal syntax that contain other things than the element to escape? I'm not sure the fix is well applied (and as @sindresorhus said, the tests should test the fix output).
A few examples:
```js
const foo = `\\ud834foo`;
const foo = `foo\\ud834`;
const foo = `foo \\ud834`;
const foo = `${'\\ud834 foo'} \\ud834`;
// and a few strings containing the same thing
const foo = '\\ud834foo';
const foo = 'foo\\ud834';
const foo = 'foo \\ud834';
```

rules/escape-case.js
```diff
@@ -0,0 +1,55 @@
...
+					return;
+				}
+
+				const match = element.value.raw.match(/\\(x[a-f0-9]{2}|u[a-f0-9]{4}|u\{([0-9a-f]{1,})\}|c[a-z])/);
+
+				if (match) {
+					context.report({
+						node,
+						message: 'Use uppercase characters for the value of the escape sequence',
+						fix: fixer => fixer.replaceText(node, '`' + fix(element.value.raw) + '`')
```
jfmengels: This will probably only work when the whole template string contains only the static part, but not if it contains multiple quasi and/or dynamic expressions. I think the tests I suggested adding in another comment should demonstrate that.

test/escape-case.js
```diff
@@ -0,0 +1,61 @@
...
+			errors
+		},
+		{
+			code: 'const foo = `\\u{1d306}`;',
```
jfmengels: Why is this not valid? :confused: 
(It might be, I'm not familiar enough with the syntax to know for sure)

jfmengels:
This looks pretty good @Bebersohl, thanks a lot! :D

test/escape-case.js
Bebersohl: I believe it should be `u{1D306}` instead of `u{1d306}`

sindresorhus:

rules/escape-case.js
```diff
@@ -0,0 +1,56 @@
...
+const create = context => {
+	return {
+		Literal(node) {
+			if (typeof node.value !== 'string') {
+				return;
+			}
+
+			const match = node.raw.match(escapeWithLowercase);
+
+			if (match) {
```
jfmengels:
```js
if (typeof node.value === 'string' && escapeWithLowercase.test(node.raw)) {
  // ...
}
```

rules/escape-case.js
```diff
@@ -0,0 +1,56 @@
...
+		},
+
+		TemplateElement(node) {
+			if (typeof node.value.raw !== 'string') {
+				return;
+			}
+
+			const match = node.value.raw.match(escapeWithLowercase);
+
+			if (match) {
```
jfmengels: Same as previous comment

test/escape-case.js
```diff
@@ -0,0 +1,111 @@
...
+});
+
+const errors = [{
+	ruleId: 'escape-case',
+	message: 'Use uppercase characters for the value of the escape sequence.'
+}];
+
+ruleTester.run('escape-case', rule, {
+	valid: [
+		'const foo = \'\\xA9\';',
```
jfmengels: Instead of escaping the inner `'`, use `"` like `'const foo = "\\xA9";'`, which makes test cases easier to read.

Bebersohl: @jfmengels Thank you for your feedback so far.

I agree with you that double quotes are easier to read but in a previous comment sindresorhus, if I understood him correctly, requested single quotes--so I'm not sure which one to go with :P.

Any thoughts?

sindresorhus: I'm ok with `"` inside the string to reduce escaping.

Bebersohl:
Are there any other changes I should make to this pull request?

sindresorhus:
@Bebersohl No. This is super. Sorry about the delay. We've just been busy in real life.

sindresorhus:
👍✨🦄🌈😎

superhighfive

PR #64, opened 2016-10-21
Add `prefer-starts-ends-with` rule - fixes #13

### LINKED ISSUE #13: `prefer-starts-ends-with` rule
[`.startsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) instead of:
- `foo.indexOf('bar') === 0`
- `foo[0] === 'b'`
- `foo.charAt(0) === 'b'` (and `foo.charAt()`)
- `foo.slice(0, 2) === 'bar'` (and .substring & .substr)
- `/^bar/.test(foo)` (and `.match()`)

[`.endsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith) instead of:
- `.lastIndexOf`
- `foo[foo.length - 1] === 'r'`
- `foo.charAt(foo.length - 1) === 'b'`
- `foo.slice(0, -3) === 'bar'` (and .substring & .substr)
- `/bar$/.test(foo)` (and `.match()`)

The regex case is especially useful as I've seen a lot of regexes used in this scenario.
## 

anglee:
The changes here add the `prefer-starts-ends-with` for ONLY the regex cases.

I am not so sure how people use `.match` for the purpose of `.startsWith` and `.endsWith` and am concerned that the way I had currently might cause false negative because `str.match(/^foo/)` could be a valid use. Should we test if the code only checks whether the returned result is empty? Please advise. 

Also, this rule doesn't report error for cases like `/^\\/` (false negative), I am not sure what's the best thing to do there. Please shed some light. Thanks!

### DIFF
--- a/docs/rules/prefer-starts-ends-with.md
+++ b/docs/rules/prefer-starts-ends-with.md
@@ -1,6 +1,6 @@
-# Prefer String#startsWidth & String#endsWidth over more complex alternatives
+# Prefer `String#startsWidth` & `String#endsWidth` over more complex alternatives
 
-To know whether a string starts or ends, you have several methods, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduces simpler alternatives named `.startsWith` and `.endsWith`. This rule enforces the use of those whenever possible.
+To know whether a string starts or ends, you have several methods, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduces simpler alternatives named [`String#startsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) and [`String#endsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith). This rule enforces the use of those whenever possible.
 
 
 ## Fail
--- a/docs/rules/prefer-starts-ends-with.md
+++ b/docs/rules/prefer-starts-ends-with.md
@@ -1,0 +1,18 @@
+# Prefer `.startsWith` and `.endsWith` 
+
+Determine whether a string begins/ends with characters with `.startWith` and `.endsWith` over using regex
+
+## Fail
+
+```js
+/^bar/.test(foo)
+/bar$/.test(foo)
+```
+
+
+## Pass
+
+```js
+foo.startsWith('bar')
+foo.endsWith('bar')
+```
--- a/docs/rules/prefer-starts-ends-with.md
+++ b/docs/rules/prefer-starts-ends-with.md
@@ -1,6 +1,7 @@
 # Prefer `.startsWith` and `.endsWith` 
 
-Determine whether a string begins/ends with characters with `.startWith` and `.endsWith` over using regex.
+To know whether a string starts or ends, you have several methods, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduces simpler alternatives named `.startsWith` and `.endsWith`. This rule enforces the use of those whenever possible.
+
 
 ## Fail
 
--- a/docs/rules/prefer-starts-ends-with.md
+++ b/docs/rules/prefer-starts-ends-with.md
@@ -1,18 +1,18 @@
 # Prefer `.startsWith` and `.endsWith` 
 
-Determine whether a string begins/ends with characters with `.startWith` and `.endsWith` over using regex
+Determine whether a string begins/ends with characters with `.startWith` and `.endsWith` over using regex.
 
 ## Fail
 
 ```js
-/^bar/.test(foo)
-/bar$/.test(foo)
+/^bar/.test(foo);
+/bar$/.test(foo);
 ```
 
 
 ## Pass
 
 ```js
-foo.startsWith('bar')
-foo.endsWith('bar')
+foo.startsWith('bar');
+foo.endsWith('bar');
 ```
--- a/docs/rules/prefer-starts-ends-with.md
+++ b/docs/rules/prefer-starts-ends-with.md
@@ -1,4 +1,4 @@
-# Prefer `.startsWith` and `.endsWith` 
+# Prefer String#startsWidth & String#endsWidth over more complex alternatives
 
 To know whether a string starts or ends, you have several methods, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduces simpler alternatives named `.startsWith` and `.endsWith`. This rule enforces the use of those whenever possible.
 
--- a/docs/rules/prefer-starts-ends-with.md
+++ b/docs/rules/prefer-starts-ends-with.md
@@ -1,6 +1,6 @@
 # Prefer `String#startsWidth` & `String#endsWidth` over more complex alternatives
 
-To know whether a string starts or ends, you have several methods, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduces simpler alternatives named [`String#startsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) and [`String#endsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith). This rule enforces the use of those whenever possible.
+There are several ways of checking whether a string starts or ends with a certain string, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduced simpler alternatives named [`String#startsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) and [`String#endsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith). This rule enforces the use of those whenever possible.
 
 
 ## Fail
--- a/readme.md
+++ b/readme.md
@@ -40,7 +40,8 @@ Configure it in `package.json`.
 			"unicorn/no-process-exit": "error",
 			"unicorn/throw-new-error": "error",
 			"unicorn/number-literal-case": "error",
-			"unicorn/no-array-instanceof": "error"
+			"unicorn/no-array-instanceof": "error",
+			"unicorn/prefer-starts-ends-with": "error"
 		}
 	}
 }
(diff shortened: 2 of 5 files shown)


docs/rules/prefer-starts-ends-with.md
```diff
@@ -0,0 +1,18 @@
+# Prefer `.startsWith` and `.endsWith`
```
sindresorhus: > Prefer `String#startsWidth` & `String#endsWidth` over a regex

@jfmengels @SamVerschueren ⬆️ Any better suggestions?

Use this in the readme description too.

docs/rules/prefer-starts-ends-with.md
```diff
@@ -0,0 +1,18 @@
...
+/^bar/.test(foo)
+/bar$/.test(foo)
+```
+
+
+## Pass
+
+```js
+foo.startsWith('bar')
+foo.endsWith('bar')
```
sindresorhus: Semicolons

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+const doesNotContain = (string, chars) => chars.every(char => string.indexOf(char) === -1);
```
sindresorhus: You can use `String#includes` here instead of indexOf. It's supported in Node.js 4.

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
...
+			} else if (callee.property.name === 'match' && args && args[0] && args[0].regex) {
+				pattern = args[0].regex.pattern;
+			} else {
+				return;
+			}
+
+			if (pattern.startsWith('^') && isSimpleString(pattern.substr(1))) {
+				context.report({
+					node,
+					message: 'Prefer `.startsWith` over regex with `^`'
```
sindresorhus: Message should end in a `.`

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
...
+				});
+			} else if (pattern.endsWith('$') && isSimpleString(pattern.substr(0, pattern.length - 1))) {
```
sindresorhus: Use `.slice` instead of `.substr`

docs/rules/prefer-starts-ends-with.md
```diff
@@ -0,0 +1,18 @@
+# Prefer `.startsWith` and `.endsWith` 
+
+Determine whether a string begins/ends with characters with `.startWith` and `.endsWith` over using regex
```
sindresorhus: Should end in a `.`

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+const doesNotContain = (string, chars) => chars.every(char => !string.includes(char));
+
+const isSimpleString = string => doesNotContain(
+	string,
+	['^', '$', '+', '[', '{', '(', '\\', '.', '?', '*']
```
anglee: Thanks for the kind feedback, @jfmengels!

Just to make sure we are on the same page, here is what my code does currently: 

1. `/^foo/,` is correctly reported as _NOT valid_
2. `/\^foo/,` is correctly reported as _valid_
3. `/^\^foo/,` is reported as _valid_ (which is not ideal and as I had pointed it out in the PR description is what I am hoping to get help for)

Can you please confirm what we are looking to improve here is the 3rd case?

btw, just want to explain my thought process and to get feedback. My idea was to catch the use cases of `^ + simple string` and `simple string + $` and to deem `^ + not-so-simple string` as a legit use of regex (as `string#startsWith` doesn't cover such cases). My current way of determining whether a string is simple is obviously very naive. However, to handle cases like the 3rd case above, we have to be more aggressively categorizing a string as 'simple'. While we want to identify as many cases of misusing regex that can be replaced with `.startsWith` or `.endsWith` (by correctly categorizing more strings as 'simple') as possible, we probably want to be careful because if we wrongly categorizing a string as 'simple' the consequence is marking a legit use of regex as error(false positive), which could be worse than not reporting error on an improper use of regex(false negative).

readme.md
```diff
@@ -57,6 +58,7 @@ Configure it in `package.json`.
 - [throw-new-error](docs/rules/throw-new-error.md) - Require `new` when throwing an error. *(fixable)*
 - [number-literal-case](docs/rules/number-literal-case.md) - Enforce lowercase identifier and uppercase value for number literals. *(fixable)*
 - [no-array-instanceof](docs/rules/no-array-instanceof.md) - Require `Array.isArray()` instead of `instanceof Array`. *(fixable)*
+- [prefer-starts-ends-with](docs/rules/prefer-starts-ends-with.md) - Prefer `foo.startsWith('bar)` over `/^bar/.test(foo)`.
```
sindresorhus: Description should be identical to the docs heading.

docs/rules/prefer-starts-ends-with.md
```diff
@@ -0,0 +1,19 @@
+# Prefer String#startsWidth & String#endsWidth over more complex alternatives
```
sindresorhus: Quote the `String#startsWidth` and `String#endsWidth` in backticks.

docs/rules/prefer-starts-ends-with.md
```diff
@@ -0,0 +1,19 @@
...
+
+To know whether a string starts or ends, you have several methods, such as `string.indexOf('foo') === 0` or using regexes with `/^foo/` or `/foo$/`. ES2015 introduces simpler alternatives named `.startsWith` and `.endsWith`. This rule enforces the use of those whenever possible.
```
sindresorhus: .startsWith => String#startsWith
.endsWith => String#endsWith

and linkify them to their MDN page.
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
...
+);
+
+const create = context => {
+	return {
+		CallExpression(node) {
+			const callee = node.callee;
+			const args = node.arguments;
+
```
sindresorhus: It needs to take in account that it doesn't work if the user uses the `i` flag. For example: `/^bar/i.test(foo);`

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
...
+			} else if (callee.property.name === 'match' && args && args[0] && args[0].regex) {
+				pattern = args[0].regex.pattern;
+			} else {
+				return;
+			}
+
+			if (pattern.startsWith('^') && isSimpleString(pattern.slice(1))) {
+				context.report({
+					node,
+					message: 'Prefer `.startsWith` over regex with `^`.'
```
sindresorhus: .startsWith => String#startsWith

regex => a regex

rules/prefer-starts-ends-with.js
```diff
@@ -0,0 +1,43 @@
...
+				});
+			} else if (pattern.endsWith('$') && isSimpleString(pattern.slice(0, -1))) {
+				context.report({
+					node,
+					message: 'Prefer `.endsWith` over regex with `$`.'
```
sindresorhus: .endsWith => String#endsWith

regex => a regex

rules/prefer-starts-ends-with.js
anglee: Good catch. I added the code to handle that. Thanks!

rules/prefer-starts-ends-with.js
sindresorhus: @jfmengels ⬆️

jfmengels: Sorry, totally forgot to answer.

Yes, we're on the same page.

I really don't know how to work with this.
I think we could ask help from @firasdi, who made the (awesome) https://regex101.com/. @firasdi, would you know how to tell whether a string is a normal string or has symbols (with the issue of escaping/unescaping) that makes it a regex that does more than just `.contain()`? If you know a library that does this, that'd be even better :sweat_smile: 

In case this gets too hard, we could just merge the PR without the regex part and handle that later.

sindresorhus: @jfmengels Your mention isn't linked, so you probably typo'd the username.

jfmengels: Yup, typo'd it. I meant @firasdib.

firasdib: @jfmengels If I understood you properly, something like this
```js
/^(?:\\.|[\w ])+$/gu
```
Should do the trick to match a string with only "words" in it or escaped symbols.

See: https://regex101.com/r/3WKMiS/1/tests

anglee: Thanks, @firasdib but I am not sure if that's exactly what we can use here. Please allow me to clarify. We want to catch the case that can be replaced with `string#startsWith` and `string#endsWith`, and the way I am approaching it is to take the test string, check if the first character is '$' and then check if the rest of the string is a _simple_ (and use similar logic for checking `endsWith`). It boils down to how to tell if that rest part of the test string is a _simple_ string. By _simple_ string I mean it doesn't use regex special characters (because otherwise it is a legit use of regex and can't be replaced with `string#startsWith`). Here is some examples of what I'd consider to be _simple_ strings:
```
foo
\^foo
!!!
\$
我
```

and some examples of NOT _simple_ strings
```
fo.o
foo+
[f,g]oo
\wfoo
a\b
$
```
I am not sure how we can modify the regex to tell these groups apart. I updated the test cases on https://regex101.com/r/DPdFPg/4/tests

anglee: @firasdib yeah, thank you. I was under the same impression so I asked for help to see if I missed something.

@sindresorhus, @jfmengels I am not sure if there is much I can add here. Please let me know if there is anything else I can do. Thanks.

jfmengels: Unless someone happens to find a handy regex parser, I think we should drop support for regexes until the time we find one or someone happens to find a different solution.

By the way @firasdib, I would love to see the one used in regex101 as a library :D

firasdib: @jfmengels I might do that at some point. I'm actually writing a blog post about it as we speak.

sindresorhus:
@anglee Would you like to handle `indexOf` support in this pull request or should we open an issue about doing it later?

rules/prefer-starts-ends-with.js
jfmengels: yes, just drop reporting if it contains any regex special character, that sounds good.
And you're welcome, it's been plenty useful to me :)

sindresorhus:
ping @anglee

anglee:
I am sorry about the delay, @sindresorhus.

As @jfmengels pointed out in an [earlier comment](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/13#issuecomment-211338957), for `indexOf` to work, we will have to infer that `foo` in`foo.indexOf` is a String. I am not sure I know how to do that immediately, so I think it makes sense to handle `indexOf` support in a separate issue.

sindresorhus:
Thank you @anglee! Glad we could eventually land this :)

jfmengels:
Thanks a lot @anglee! :D

anglee:
Thank you both for your guidance. I really enjoyed it!

PR #67, opened 2016-10-24
Add `prefer-string-slice` rule - fixes #65

matijs:
Prefer `String.slice` instead of `String.substr` or `String.substring`.
- [x] Documentation
- [ ] Process feedback
- [ ] Add fixer

PR #68, opened 2016-10-27
Make `no-hex-escape` more concise

matijs:
Was tinkering with #67 and realised I could clean up a few things in `no-hex-escape`.

### DIFF
--- a/rules/no-hex-escape.js
+++ b/rules/no-hex-escape.js
@@ -1,25 +1,21 @@
 'use strict';
 
-function checkEscape(context, node, value) {
-	const fixedValue = typeof value === 'string' ? value.replace(/\\x/g, '\\u00') : value;
-
-	if (value !== fixedValue) {
-		context.report({
-			node,
-			message: 'Use unicode escapes instead of hexadecimal escapes.',
-			fix: fixer => fixer.replaceTextRange([node.start, node.end], fixedValue)
-		});
-	}
-}
-
 const create = context => {
-	return {
-		Literal: node => {
-			checkEscape(context, node, node.raw);
-		},
-		TemplateElement: node => {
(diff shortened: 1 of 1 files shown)


rules/no-hex-escape.js
```diff
@@ -1,25 +1,21 @@
...
-const create = context => {
 	return {
-		Literal: node => {
-			checkEscape(context, node, node.raw);
-		},
-		TemplateElement: node => {
-			checkEscape(context, node, node.value.raw);
-		}
+		Literal: node => checkEscape(node, node.raw),
+		TemplateElement: node => checkEscape(node, node.value.raw)
```
sindresorhus: I prefer only doing inline arrow functions for when you care about the return value. This would be more readable like it was before.

sindresorhus: @jfmengels ⬆️ Wish we could write a rule for that...

sindresorhus:
👎 I find it more readable to have functions top-level when possible instead of nested.

rules/no-hex-escape.js
matijs: Fair enough.

matijs:
@sindresorhus even though `context` can be used instead of having to pass it to a top-level function?

rules/no-hex-escape.js
jfmengels: You could enforce that with a type checker like Flow/TS, though a bit troublesome.

I'm starting to think about having a plugin to write rules, not sure what I'd put in there yet. Having Flow types for the whole AST would be so much more useful though, so thinking about it (but not doing much ATM :sweat_smile:).

sindresorhus:
@matijs Yes, since it's only one extra argument. Had it been more it would have made sense.

PR #70, opened 2016-10-30
Add `prefer-includes` rule - fixes #8

### LINKED ISSUE #8: `prefer-includes` rule
Prefer `.includes()` over `.indexOf()` when checking for existence. Since ESLint has no type analysis we'll have to assume all properties named `.indexOf()` have a `.includes()` counterpart. This is luckily true for all builtins: `String#includes`, `Array#includes`, `TypedArray#includes`, `Buffer#includes`.

See https://github.com/eslint/eslint/issues/4209 for more.

`.includes()` cases:
- `'foobar'.indexOf('foo') !== -1`
- `'foobar'.indexOf('foo') != -1`
- `'foobar'.indexOf('foo') > -1`
- `'foobar'.indexOf('foo') >= 0`
## 

jeremyjs:
Implements #8 except for the regex part which is tricky due to needing to distinguish a simple string from a regex with special characters.

### DIFF
--- a/docs/rules/prefer-includes.md
+++ b/docs/rules/prefer-includes.md
@@ -24,11 +24,11 @@ const str = 'foobar';
 !str.includes < 0
 ```
 
-```js
+<!-- ```js
 const str = 'foobar';
 
 /foo/.test(str);
-```
+``` -->
 
 
 ## Pass
--- a/docs/rules/prefer-includes.md
+++ b/docs/rules/prefer-includes.md
@@ -1,0 +1,54 @@
+# Prefer .includes() over .indexOf() when checking for existence.
+
+Since ESLint has no type analysis we'll have to assume all properties named .indexOf() have a .includes() counterpart.
+
+This is luckily true for all builtins: String#includes, Array#includes, TypedArray#includes, Buffer#includes.
+
+
+## Fail
+
+```js
+const str = 'foobar';
+
+str.indexOf('foo') !== -1
+str.indexOf('foo') != -1
+str.indexOf('foo') > -1
+str.indexOf('foo') >= 0
+```
+
+```js
+const str = 'foobar';
+
+!str.includes === -1
+!str.includes == -1
+!str.includes < 0
+```
+
+```js
+const str = 'foobar';
+
+/foo/.test(str);
+```
+
+
+## Pass
+
+```js
+const str = 'foobar';
+
+str.includes('foo')
+```
+
+TODO:
+
+Would also be useful to catch cases where .includes() would be better than a regex:
+
+```js
+/\r\n/.test(foo);
+```
+
+Could be:
+
+```js
+foo.includes('\r\n');
+```
--- a/docs/rules/prefer-includes.md
+++ b/docs/rules/prefer-includes.md
@@ -24,12 +24,6 @@ const str = 'foobar';
 !str.includes < 0
 ```
 
-<!-- ```js
-const str = 'foobar';
-
-/foo/.test(str);
-``` -->
-
 
 ## Pass
 
--- a/rules/prefer-includes.js
+++ b/rules/prefer-includes.js
@@ -1,0 +1,104 @@
+'use strict';
+
+const isIndexOfCallExpression = node => {
+	if (node.type !== 'CallExpression') {
+		return false;
+	}
+
+	const property = node.callee.property;
+
+	return property.name === 'indexOf';
+};
+
+const isUnaryNotExpression = node => (
+	node.type === 'UnaryExpression' && node.operator === '!'
+);
+
+const isNegativeOne = (operator, value) => operator === '-' && value === 1;
+
+const report = (context, node) => {
+	context.report({
+		node,
+		message: 'Use `.includes()`, not .indexOf(), when checking for existence.'
+		// fix: fixer => fixer.replaceText(node, `Array.isArray(${arraySourceCode})`)
+	});
+};
+
+const create = context => ({
+	BinaryExpression: node => {
+		const left = node.left;
+		const right = node.right;
+
+		if (isIndexOfCallExpression(left)) {
+			if (right.type === 'UnaryExpression') {
+				const argument = right.argument;
+
+				if (argument.type !== 'Literal') {
+					return false;
+				}
+
+				const value = argument.value;
+
+				if (node.operator === '!==' && isNegativeOne(right.operator, value)) {
+					report(context, node);
+				}
+				if (node.operator === '!=' && isNegativeOne(right.operator, value)) {
+					report(context, node);
+				}
+				if (node.operator === '>' && isNegativeOne(right.operator, value)) {
+					report(context, node);
+				}
+			}
+
+			if (right.type !== 'Literal') {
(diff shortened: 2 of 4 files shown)


sindresorhus:
> except for the regex part which is tricky due to needing to distinguish a simple string from a regex with special characters.

Indeed. See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/64#discussion_r84574596 for relevant discussion. Add your thoughts there if any.

docs/rules/prefer-includes.md
```diff
@@ -0,0 +1,48 @@
+# Prefer .includes() over .indexOf() when checking for existence.
```
sindresorhus: Wrap the methods in backticks and drop the `.` at the end.

docs/rules/prefer-includes.md
```diff
@@ -0,0 +1,48 @@
...
+
+Since ESLint has no type analysis we'll have to assume all properties named .indexOf() have a .includes() counterpart.
```
sindresorhus: I would form it a bit different:

> The rule assume all properties named `.indexOf()` have a `.includes()` counterpart, as ESLint doesn't do type analysis. This is luckily true for all built-ins: `String#includes`, `Array#includes`, `TypedArray#includes`, `Buffer#includes`.

docs/rules/prefer-includes.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Fail
+
+```js
+const str = 'foobar';
+
+str.indexOf('foo') !== -1
+str.indexOf('foo') != -1
+str.indexOf('foo') > -1
+str.indexOf('foo') >= 0
```
sindresorhus: Semicolons

rules/prefer-includes.js
```diff
@@ -0,0 +1,116 @@
...
+const getSourceCode = (context, node) => (
+	context.getSourceCode().text.slice(node.range[0], node.range[1])
+);
+
+const report = (context, node, target, pattern) => {
+	const targetSource = getSourceCode(context, target);
+	const patternSource = getSourceCode(context, pattern);
+	context.report({
+		node,
+		message: 'Use `.includes()`, not .indexOf(), when checking for existence.',
```
sindresorhus: > Use `.includes()` rather than `.indexOf()` when checking for existence.

rules/prefer-includes.js
```diff
@@ -0,0 +1,116 @@
...
+};
+
+const isUnaryNotExpression = node => (
+	node.type === 'UnaryExpression' && node.operator === '!'
+);
+
+const isNegativeOne = (operator, value) => operator === '-' && value === 1;
+
+const getSourceCode = (context, node) => (
+	context.getSourceCode().text.slice(node.range[0], node.range[1])
```
sindresorhus: Would be better to use `context.getSourceCode().getText(node)`

sindresorhus:
Very good work @jeremyjs! :)

Have you worked with AST's before?

----

// @jfmengels Mind doing a review too?

rules/prefer-includes.js
```diff
@@ -0,0 +1,116 @@
...
+
+			if (right.type !== 'Literal') {
+				return false;
+			}
+
+			if (node.operator === '>=' && right.value === 0) {
+				report(context, node, target, pattern);
+			}
+
+			return false;
```
jfmengels:
```js
if (right.type === 'Literal' && node.operator === '>=' && right.value === 0) {
  report(context, node, target, pattern);
}
return;
```

test/prefer-includes.js
```diff
@@ -0,0 +1,69 @@
...
+			output: 'str.includes(\'foo\')',
+			errors
+		},
+		{
+			code: '\'foobar\'.indexOf(\'foo\') >= 0',
+			output: '\'foobar\'.includes(\'foo\')',
+			errors
+		},
+		{
+			code: '!str.indexOf(\'foo\') === -1',
```
jfmengels: Could you add another example with two `!!`: `!str.indexOf(\'foo\') === -1`. I want to make sure the result is correct.

test/prefer-includes.js
```diff
@@ -0,0 +1,69 @@
...
+			errors
+		},
+		{
+			code: 'str.indexOf(\'foo\') > -1',
+			output: 'str.includes(\'foo\')',
+			errors
+		},
+		{
+			code: '\'foobar\'.indexOf(\'foo\') >= 0',
+			output: '\'foobar\'.includes(\'foo\')',
```
jfmengels: Instead of escaping `\'` everywhere, you can use backticks around the whole string.
```js
output: `'foobar'.includes('foo')`
```

test/prefer-includes.js
```diff
@@ -0,0 +1,69 @@
...
+ruleTester.run('prefer-includes', rule, {
+	valid: [
+		'str.indexOf(\'foo\') !== -n',
+		'str.indexOf(\'foo\') !== 1',
+		'!str.indexOf(\'foo\') === 1',
+		'!str.indexOf(\'foo\') === -n',
+		'str.includes(\'foo\')',
+		'\'foobar\'.includes(\'foo\')',
+		'[1,2,3].includes(4)',
+		'"str.indexOf(\'foo\') !== -1"'
```
jfmengels: Could you add, just to be sure that nothing odds is happening, the following tests:
```js
`a + b` // you target binary expressions. If your checks have gaps, code like this could crash the plugin or create false positives
`indexOf !== -1`
// And whatever else you feel like adding :D
```

jfmengels:
Hey @jeremyjs! Are you still up to continue working on this?

sindresorhus:
ping @jeremyjs

jeremyjs:
Merry Christmas! 🎅 Sorry for the neglect. 😅  I'll see if I can iron out these kinks in my spare time before the new year. 😄

rules/prefer-includes.js
```diff
@@ -0,0 +1,116 @@
+'use strict';
+
+const isIndexOfCallExpression = node => {
+	if (node.type !== 'CallExpression') {
+		return false;
+	}
+
+	const property = node.callee.property;
+
+	return property.name === 'indexOf';
```
jeremyjs: I'm not sure I totally understand. Could you provide an example in context where `indexOf(foo)` wouldn't be a `MemberExpression`?

jfmengels:
```js
function indexOf(input) {
  // ...
}

indexOf(foo);
```

But in this implementation, you assume that any CallExpression's callee is a `MemberExpression`, which will cause a crash before you even check whether the property is `indexOf`. Add `indexOf(foo)` and `a()` in your valid tests and the problems should be made clear right awaya IIRC.

jfmengels:
Merccy Christmas (or rather, happy new year now). No problem for the delay, I'm not the quick to review these days anyway ^^'

sindresorhus:
Hey @jeremyjs. Still interested in finishing this? :)

sindresorhus:
Closing for lack of activity. Happy to reopen if you ever have time to finish this :)

PR #73, opened 2016-11-04
Add `type-error` rule - fixes #27

### LINKED ISSUE #27: Prevent using `Error` when doing type-checking
I often forget to use a `TypeError` when doing type-checking in a module.

Example:

``` js
if (Array.isArray(foo)) {
    throw new Error('Expected an array');
}
```

Would be nice to error the conditional only does type-checking and prevent the use of `Error`. Custom errors should be allowed.
## 

florianb:
Finally my draft for the `typeerror`-rule which fixes #27. Since this is my **1st** eslint-rule i'd really appreciate any feedback! Thanks in advance! ☕️ 

# Reviewers Wishlist

🎅  **Ho Ho Ho - Merry Christmas**

- [x] Add top level throw-statement
- [x] Add valid `if (!Number.isNaN(foo) && foo === 10) { throw new Error(); }` case
- [x] Remove `isFrozen` and `isSealed` from the type checking identifiers set
- [x] Remove `isTypeCheckingProvider()`
- [x] Add valid `isNaN`-expression throwing an Error
- [x] Add additional `throw new Error()`-conditions to the tests
- [x] Add one test per `tcIdentifiers`-item
- [x] Add tests for `isFor()`/`_.isFoo()` to check isX-calls
- [x] throw new Error(args)
- [x] Add test for `throw Error()`
- [x] Add test for `throw new Error`
- [x] Add test for `new Error()` (not thrown)
- [x] Add test for a condition with multiple statements and preceeding throw
- [x] Add test for `throw new Error.foo()`
- [x] Add test for `throw new Foo.Error()`
- [x] Add test for `throw new CustomError()`
- [x] Add test with a condition with non-type checking parts
- [x] Add test with a condition call but without arguments of isArray()
- [x] Add test for a thrown Error with arguments
- [x] Slap uncovered lines out of nyc 👋 
- [x] Proove and fix `isNaN`-treatment
- [x] Fix broken tests 💣 
- [x] Check if the `typeof` and `instanceof` operator may be combined or should be treated differently in unary/binary expressions.

> :warning: I considered the expressions from below and it is very likely that the list is not complete. So any hint on additional typechecks are welcome.

## ECMA262 ES17

- [`instanceof`](https://tc39.github.io/ecma262/#sec-instanceofoperator)
- [`typeof`](https://tc39.github.io/ecma262/#sec-typeof-operator)
-  [`Object.getPrototypeOf()`](https://tc39.github.io/ecma262/#sec-object.getprototypeof)
- [`Object.prototype.__proto__`](https://tc39.github.io/ecma262/#sec-object.prototype.__proto__)
- [`Object.prototype.isPrototypeOf()`](https://tc39.github.io/ecma262/#sec-object.prototype.isprototypeof)
- [`Object.prototype.toString()`](https://tc39.github.io/ecma262/#sec-object.prototype.tostring)
- [`Number.isInteger()`](https://tc39.github.io/ecma262/#sec-number.isinteger)
- [`Number.isSafeInteger()`](https://tc39.github.io/ecma262/#sec-number.issafeinteger)
- [`Number.isFinite()`](https://tc39.github.io/ecma262/#sec-number.isfinite)
- [`Number.isNaN()`](https://tc39.github.io/ecma262/#sec-number.isnan)
- [`Array.isArray()`](https://tc39.github.io/ecma262/#sec-array.isarray)
- [`ArrayBuffer.isView()`](https://tc39.github.io/ecma262/#sec-arraybuffer.isview)
- [`isFinite()`](https://tc39.github.io/ecma262/#sec-isfinite-number)
- [`isNaN()`](https://tc39.github.io/ecma262/#sec-isnan-number)
- [`Object.is()`](https://tc39.github.io/ecma262/#sec-object.is)

### Rejected Expressions

- [Property Descriptors](https://tc39.github.io/ecma262/#sec-property-descriptor-specification-type) like `isAccessorDescriptor`
- `isExtinsible`, `isFrozen`, `isSealed`, `isConcatSpreadable`
- [`IsArray()`](https://tc39.github.io/ecma262/#sec-isarray)
- [`IsConstructor()`](https://tc39.github.io/ecma262/#sec-isconstructor)
- [`IsInteger()`](https://tc39.github.io/ecma262/#sec-isinteger)
- [`IsPropertyKey()`](https://tc39.github.io/ecma262/#sec-ispropertykey)
- [`IsRegExp()`](https://tc39.github.io/ecma262/#sec-isregexp)
- [`IsCallable()`](https://tc39.github.io/ecma262/#sec-iscallable)
- [`Object.isFrozen()`](https://tc39.github.io/ecma262/#sec-object.isfrozen)
- [`Object.isSealed()`](https://tc39.github.io/ecma262/#sec-object.issealed)

## Utility-Libraries

| Library | npms.io-score | Downloads per month |
|--------|:---------------:|----------------------:|
| [lodash](https://github.com/lodash/lodash) | 99 | 25m |
| [underscore](https://github.com/jashkenas/underscore) | 85 | 11m |
| [jquery](https://github.com/jquery/jquery) | 94 | 2.2m |
| [kind-of](https://github.com/jonschlinkert/kind-of) | 83 | 9.9m |
| [check-types](https://github.com/philbooth/check-types.js) | 76 | 189k |

### Type checking Expressions

- `_.isArguments()`
- `_.isArray()`
- `_.isArrayBuffer()`
- `_.isArrayLike()`
- `_.isArrayLikeObject()`
- `_.isBoolean()`
- `_.isBuffer()`
- `_.isDate()`
- `_.isElement()`
- `_.isError()`
- `_.isFinite()`
- `_.isFunction()`
- `_.isInteger()`
- `_.isLength()`
- `_.isMap()`
- `_.isNaN()`
- `_.isNative()`
- `_.isNil()`
- `_.isNull()`
- `_.isNumber()`
- `_.isObject()`
- `_.isObjectLike()`
- `_.isPlainObject()`
- `_.isRegExp()`
- `_.isSafeInteger()`
- `_.isSet()`
- `_.isString()`
- `_.isSymbol()`
- `_.isTypedArray()`
- `_.isUndefined()`
- `_.isWeakMap()`
- `_.isWeakSet()`
- `jQuery.isArray()`
- `jQuery.isEmptyObject()`
- `jQuery.isFunction()`
- `jQuery.isWindow()`
- `jQuery.isXMLDoc()`

### Rejected Expressions

- `_.isEmpty()`
- `_.isMatch()`
- `kindOf()`
- `check.array()`
- `check.emptyArray()`
- `check.arrayLike()`
- `is[A-Z]\w+`: there are several libraries providing this scheme of typecheck (f.e. [is-number](https://github.com/jonschlinkert/is-number) or [is-stream](https://github.com/sindresorhus/is-stream))


index.js
```diff
@@ -22,7 +22,8 @@ module.exports = {
 				'unicorn/number-literal-case': 'error',
 				'unicorn/no-array-instanceof': 'error',
 				'unicorn/no-new-buffer': 'error',
-				'unicorn/no-hex-escape': 'error'
+				'unicorn/no-hex-escape': 'error',
+				'unicorn/typeerror': 'error'
```
sindresorhus: `typeerror` => `type-error`

Applies to all places where that spelling is used.

test/typeerror.js
```diff
@@ -0,0 +1,71 @@
...
+		`if (typeof foo === 'object' || (Blob.isBlob(foo) || 'String' === typeof foo)) {
+			throw new TypeError();
+		}`,
+		`if (Blob.isBlob(foo) && Array.isArray(foo)) {
+			throw new TypeError();
+		}`
+	],
+	invalid: [
+		{
+			code:	`if (Array.isArray(foo)) {throw new Error();}`,
```
sindresorhus: The code should be multi-line, like in the `valid` cases.

test/typeerror.js
```diff
@@ -0,0 +1,71 @@
...
+			errors: [typeError]
+		},
+		{
+			code:	`if (foo instanceof bar) {throw new Error();}`,
```
sindresorhus: Too much whitespace after `code:`

rules/typeerror.js
```diff
@@ -0,0 +1,72 @@
...
+
+const isTypecheckingProvider = node => node.type === 'Identifier' && node.name === 'check';
+
+const isTypecheckingOperator = operator => operator === 'typeof' || operator === 'instanceof';
+
+const isTypecheckingIdentifier = node => {
+	return node.type === 'Identifier' &&
+	(node.name.match(/[Ii]s[A-Z0-9][\w_]*/) !== null ||
+	node.name === 'kindOf');
+};
```
sindresorhus: Can be an inline arrow function. Same with line 13.

```js
const isTypecheckingIdentifier = node =>
	node.type === 'Identifier' &&
	(node.name.match(/[Ii]s[A-Z0-9][\w_]*/) !== null ||
	node.name === 'kindOf');
```

test/typeerror.js
```diff
@@ -0,0 +1,71 @@
...
+		},
+		{
+			code:	`if (typeof foo == 'Foo' || 'Foo' === typeof foo) {throw new Error();}`,
+			errors: [typeError]
+		},
+		{
+			code:	`if (Array.isArray(foo) || (Blob.isBlob(foo) || isBlip(foo))) {
+				throw new Error();
+			}`,
+			errors: [typeError]
```
sindresorhus: You also need to test the fixer output on all of these. See: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/8c39be47bab25b282aaa0015e3b7d64bbf721167/test/no-array-instanceof.js#L30

rules/typeerror.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+const isTypecheckingProvider = node => node.type === 'Identifier' && node.name === 'check';
+
+const isTypecheckingOperator = operator => operator === 'typeof' || operator === 'instanceof';
+
+const isTypecheckingIdentifier = node => {
+	return node.type === 'Identifier' &&
+	(node.name.match(/[Ii]s[A-Z0-9][\w_]*/) !== null ||
```
sindresorhus: I think this is too loose. It will match anything that starts with `is`. Not everything `is` is typechecking. For example `isCookieMonster()`. Instead, I think we should hardcode every method we want to support. I also don't think we should target everything. Just the absolute most popular typechecks.

sindresorhus:
> IsArray()
IsConstructor()
IsInteger()
IsPropertyKey()
IsRegExp()
IsCallable()

These are internal JS engine methods and not exposed to userland.

---

Lodash, underscore, and jquery looks good to support, but we shouldn't bother with `kind-of` or `check-types`.

sindresorhus:
Good start! 🤘🦄

florianb:
thank you for your feedback - i will refine the stuff as soon as possible! 😀

sindresorhus:
The list is missing these:

Number.isFinite()
Number.isNaN()
Object.is()
Object.isFrozen()
Object.isSealed()

florianb:
@sindresorhus, i fixed all the requested changes, only one thing isn't implemented properly, yet. The prove of `__proto__`-, as well as `prototype.toString()`-comparisons.

I guess both can be dropped for the sake of ES17+ support but i just want to be sure.

If there are no further changes i will write the documentation.. then it is 🏁 👏

rules/type-error.js
```diff
@@ -1,20 +1,62 @@
...
+			case 'isSet':
+			case 'isString':
+			case 'isSymbol':
+			case 'isTypedArray':
+			case 'isUndefined':
+			case 'isView':
+			case 'isWeakMap':
+			case 'isWeakSet':
+			case 'isWindow':
+			case 'isXMLDoc':
```
sindresorhus: This should be a `Set` and use `set.has()`.

florianb:
Done.

florianb:
🏁  🛎  @sindresorhus

rules/type-error.js
florianb: This is fixed in a later commit. I am really confused by GitHubs 🍝  Review UI. Probably you were waiting for me adding a comment here?! Here we go..

sindresorhus:
Sorry about the delay. I was AFK most of december.

Generally looks good. I noticed that it doesn't handle the following. Could you add a test for it and handle cases like it?

```js
// should fail, does not
if (!Array.isArray(foo)) {
	throw new Error();
}
```

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+const isTypechecking = node => node.type === 'IfStatement' && isTypecheckingExpression(node.test);
+
+const create = context => {
+	return {
+		ThrowStatement: node => {
+			if (throwsErrorObject(node) &&
+				isTypechecking(node.parent.parent)) {
+				context.report({
+					node,
+					message: '`new Error()` is too unspecific for a typecheck, use `new TypeError()` instead.',
```
sindresorhus: typecheck => type check

sindresorhus:
Note that I pushed some docs improvements: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/73/commits/fba367f6128ac54aff8c269e48da5b3999f67da2 (So you don't force push over it)

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+		case 'CallExpression':
+			return isTypecheckingExpression(node.callee);
+		case 'UnaryExpression':
+			return isTypecheckingOperator(node.operator);
+		case 'BinaryExpression':
+			return isTypecheckingOperator(node.operator) ||
+				(isTypecheckingExpression(node.left) || isTypecheckingExpression(node.right));
+		case 'LogicalExpression':
+			return isTypecheckingExpression(node.left) && isTypecheckingExpression(node.right);
+		default:
```
SamVerschueren: I would `return false` immediately after `default`:

```js
default:
    return false;
```

This looks weird imo.

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+	return false;
+};
+
+const isTypechecking = node => node.type === 'IfStatement' && isTypecheckingExpression(node.test);
+
+const create = context => {
+	return {
+		ThrowStatement: node => {
+			if (throwsErrorObject(node) &&
+				isTypechecking(node.parent.parent)) {
```
SamVerschueren: Put line 89 and 90 on one line?

test/type-error.js
```diff
@@ -0,0 +1,105 @@
...
+		}`,
+		`if (Array.isArray(foo) || (Blob.isBlob(foo) || Blip.isBlip(foo))) {
+			throw new TypeError();
+		}`,
+		`if (typeof foo === 'object' || (Object.isFrozen(foo) || 'String' === typeof foo)) {
+			throw new TypeError();
+		}`,
+		`if (Object.isSealed(foo) && _.isArguments(foo)) {
+			throw new TypeError();
+		}`
```
SamVerschueren: According to the docs, this should be valid as well

```js
if (!Number.isNaN(foo) && foo === 10) {
    throw new Error();
}
```

Because all the expressions should be a typecheck before this rule fails. Maybe it's a good idea to add it something like this to the valid tests.

rules/type-error.js
florianb: Seems like a relict from the past. 🕸 

- [x] fixed, thanks!

test/type-error.js
florianb: True - i will add this case! Thanks! 😄 

- [ ] fixed

rules/type-error.js
florianb: It looks weird to me too, i was swinging between this solution and having an explicit return at the end of the function.

- [x] fixed

rules/type-error.js
florianb: - [x] fixed, thanks! 😄

rules/type-error.js
jfmengels: Could you add a test with a top-level throw statement? I'm worried there may not always be a parent.parent for a throw statement

test/type-error.js
```diff
@@ -0,0 +1,105 @@
...
+});
+
+const errors = [{
+	ruleId: 'type-error',
+	message: '`new Error()` is too unspecific for a typecheck, use `new TypeError()` instead.'
+}];
+
+ruleTester.run('type-error', rule, {
+	valid: [
+		`if (Array.isArray(foo) || ArrayBuffer.isView(foo)) {
```
jfmengels: Testing-wise, these valid tests are not very interesting because you don't test when `new Error()` is allowed, which is what we want to know. "Does this case forbid new Error() when I do X?"
Keep the actual tests, but I'd like to see new Error() thrown in more diverse conditions. Inside a condition, outside one, inside a loop, at the top-level, thrown with a custom error or, with non-type checking conditions, ...

test/type-error.js
```diff
@@ -0,0 +1,105 @@
...
+			errors
+		},
+		{
+			code: `if (Number.isFinite(foo) && Number.isSafeInteger(foo) && Number.isInteger(foo)) {
+				throw new Error();
+			}`,
+			output: `if (Number.isFinite(foo) && Number.isSafeInteger(foo) && Number.isInteger(foo)) {
+				throw new TypeError();
+			}`,
+			errors
```
jfmengels: Could you add a test with new Error having arguments? To make sure they are not removed when fixed to TypeError.

test/type-error.js
jfmengels: Also, tests with isFoo / _.isFoo to make sure that not every function call or isX call is considered to be a type check

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+const isTypecheckingProvider = node => node.type === 'Identifier' && false;
+const isTypecheckingOperator = operator => operator === 'typeof' || operator === 'instanceof';
+const isTypecheckingIdentifier = node => node.type === 'Identifier' && tcIdentifiers.has(node.name);
+
+const throwsErrorObject = node =>
+	node.argument.type === 'NewExpression' &&
+	node.argument.callee.type === 'Identifier' &&
+	node.argument.callee.name === 'Error';
+
+const isTypecheckingMemberExpression = node => {
```
jfmengels: I think this function considers `isNaN.foo()` and `isNaN` (without being called) as type checking expressions. Add those as valid tests

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+			return isTypecheckingIdentifier(node);
+		case 'MemberExpression':
+			return isTypecheckingMemberExpression(node);
+		case 'CallExpression':
+			return isTypecheckingExpression(node.callee);
+		case 'UnaryExpression':
+			return isTypecheckingOperator(node.operator);
+		case 'BinaryExpression':
+			return isTypecheckingOperator(node.operator) ||
+				(isTypecheckingExpression(node.left) || isTypecheckingExpression(node.right));
```
jfmengels: No need for the parents here

jfmengels: Parens (sorry, no editing on mobile).
And should this condition not be && ? Both left and right should be type checking expressions, or did I misunderstand?

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+	'isUndefined',
+	'isView',
+	'isWeakMap',
+	'isWeakSet',
+	'isWindow',
+	'isXMLDoc'
+]);
+
+const isTypecheckingProvider = node => node.type === 'Identifier' && false;
+const isTypecheckingOperator = operator => operator === 'typeof' || operator === 'instanceof';
```
jfmengels: These are not really the same kind of operator. I would remove the function and compare with "instanceof" in the Binary expression case, and compare with typeof in the unary case. Bte, should `!` not be allowed somehow somewhere. Would love some tests pushing that one to the edge.

test/type-error.js
jfmengels: More test ideas:
- throw new Error(args)
- throw Error()
- throw new Error
- new Error() (not thrown)
- a condition with multiple statements and then a throw
- Throw new Error.foo()
- throw new Foo.Error()
- throw new CustomError()
- with a condition with non-type checking parts

To be clear, not all tests I am requesting will be valid.

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+	'isArrayLike',
+	'isArrayLikeObject',
+	'isBoolean',
+	'isBuffer',
+	'isDate',
+	'isElement',
+	'isEmptyObject',
+	'isError',
+	'isFinite',
+	'isFrozen',
```
florianb: Thanks @jfmengels - i had to reread why i chose adding these (isSealed, too).Trying to change sealed/frozen Objects in a permitted way throws a `TypeError`, this is why i chose to treat isFrozen/isSealed-checks as a type check.

But this probably needs more discussion?

P.s.: A good [SO-answer](http://stackoverflow.com/questions/21402108/difference-between-freeze-and-seal-in-javascript) about sealed vs frozen.

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+	'isNil',
+	'isNull',
+	'isNumber',
+	'isObject',
+	'isObjectLike',
+	'isPlainObject',
+	'isPrototypeOf',
+	'isRegExp',
+	'isSafeInteger',
+	'isSealed',
```
florianb: See previous comment.

rules/type-error.js
```diff
@@ -0,0 +1,106 @@
...
+	'isTypedArray',
+	'isUndefined',
+	'isView',
+	'isWeakMap',
+	'isWeakSet',
+	'isWindow',
+	'isXMLDoc'
+]);
+
+const isTypecheckingProvider = node => node.type === 'Identifier' && false;
```
florianb: I kept this statement as a placeholder for future improvements. But you're right i should just get rid of it.

- [x] Remove `isTypeCheckingProvider()`

rules/type-error.js
SamVerschueren: I agree with @jfmengels. The reason we should throw a `TypeError` is because there is a mismatch in type. `Object.isFrozen(obj)` does not tell us anything about the type and is not meant to check a type. So imo they should be removed.

rules/type-error.js
florianb: 🤔  the binary negation-operator should supported by the subsequent `BinaryExpression`-condition.

rules/type-error.js
florianb: isNaN without being called isn't handled correctly, thanks! 😃 

- [x] Add valid `isNaN`-expression throwing an Error
- [ ] Proove and fix `isNaN`-treatment

rules/type-error.js
florianb: You're right with the parentheses, thanks! 😄 

This BinaryExpression is mentioned to catch f.e. `typeof foo === 'object'` where `typeof foo` is a type checking unary expression and `'object'` is a literal. The whole expression is therefor treated as type checking. I guess this is right. But i will dig into your edge cases to prove it.

jfmengels: My mistake, I got it confused with LogicalExpression. Bu yeah, lots of tests can't hurt :)

rules/type-error.js
florianb: Thanks @SamVerschueren - i really have no opinion on that. But just for completeness i could argue that testing for immutability is aiming for an constant type and throwing a `TypeError` as being the best matching error "[indicat[ing] an unsuccessful operation when none of the other NativeError objects are an appropriate indication of the failure cause](https://tc39.github.io/ecma262/#sec-native-error-types-used-in-this-standard-typeerror)".

However, since i don't bother, and you both voted for removing them, i am going to remove them. 🔨

test/type-error.js
florianb: 😅 💦 💦 💦

- [ ] Add 1 bazillion tests. 🗻

jfmengels: Sorry for the addionnal work :sweat_smile: 
But this will most definitely pay off :wink:

test/type-error.js
florianb: - [ ] Add one more test. 🦄

rules/type-error.js
florianb: Okay - i proved the `TypeCheckingOperator` function and for me it seems warrantable since its purpose is "just" to identify if this is a type checking operation. However, since this is adding a little performance penalty i removed it as suggested.. Thanks for the hint.. 🐎

florianb:
🛎  @sindresorhus, @SamVerschueren, @jfmengels: I am done - still open for feedback, though. 😺

test/type-error.js
```diff
@@ -0,0 +1,563 @@
...
+			code: `if (_.isElement(foo)) {
+				throw new Error;
+			}`,
+			output: `if (_.isElement(foo)) {
+				throw new TypeError;
+			}`,
+			errors
+		},
+		{
+			code: `if (wrapper._.isElement(foo)) {
```
jfmengels: What is this `wrapper`? Is it a special known variable ?

Do we accept anything as the object? Should we consider `foo.isElement(foo)` as a type check? I'm not saying we should/shouldn't, as that would make detecting Lodash and friends very hard.

test/type-error.js
```diff
@@ -0,0 +1,563 @@
...
+			errors
+		},
+		{
+			code: `if (wrapper.f.g.n.isFinite(foo) && wrapper.g.n.isSafeInteger(foo) && wrapper.n.isInteger(foo)) {
+				throw new Error();
+			}`,
+			output: `if (wrapper.f.g.n.isFinite(foo) && wrapper.g.n.isSafeInteger(foo) && wrapper.n.isInteger(foo)) {
+				throw new TypeError();
+			}`,
+			errors
```
jfmengels: By the way, since you're writing pretty much the same boilerplate everywhere, you could make a test helper.

```js
function errorTestCase(code) {
  return {
    code,
    output: code.replace('new Error(', 'new TypeError('),
    errors
  };
}

errorTestCase(`if (wrapper.f.g.n.isFinite(foo) && wrapper.g.n.isSafeInteger(foo) && wrapper.n.isInteger(foo)) {
  throw new Error();
}`)
```

rules/type-error.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
+
+const tcIdentifiers = new Set([
```
jfmengels: Some of these functions are methods on objects (like _.isFunction) , and some of them are independent functions (like isNaN).
I think there would be less false positives if we separated this list into two based on this difference. Some of theses items would appear in both lists (like for `isNaN` and `Number.isNaN` IIRC)

test/type-error.js
```diff
@@ -0,0 +1,563 @@
...
+			if (Array.isArray(foo) || ArrayBuffer.isView(foo)) {
+				return new Error('This is fo FooBar', foo);
+			}
+			return foo;
+		}`,
+		`if (Array.isArray(foo) || ArrayBuffer.isView(foo)) {
+			lastError = new Error('This is fo FooBar', foo);
+		}`,
+		`if (!isFinite(foo)) {
+			throw new TypeError();
```
jfmengels: Can you add this test case too but throwing `new Error()`? That should be reported

test/type-error.js
florianb: The `wrapper` tests for property-chains ending in a type checking identifier. I added that one when realizing that 1) i can't be sure the hosting objects are named as we would expect and 2) type checking functions might be attached to a property chain.

But writing this reminds me of checking for calling-chains, too. 🤔 

I did think about checking for Object-names too, but i can't think of any reliable and performant solution being able to handle all cases of attachments.

test/type-error.js
florianb: Absolutely! I just spend 5 minutes regexing the tests. The next 5 minutes will go into the test-helper..

florianb: Spent - thanks!

test/type-error.js
florianb: - [x] Add new invalid for unary expressions

rules/type-error.js
florianb: As far as i know only `isNaN` and `isFinite` are affected. I added a check for global type checking identifiers, too. 👍

florianb:
🛎  I added some nice new tests.

@jfmengels suggested renaming the rule into something more descriptive. F.e. favor-type-error?

I have no real opinion about it, but the rule should be short, for the case you have to add an exception the eslint-comments blow easily the line-lengths.

sindresorhus:
> suggested renaming the rule into something more descriptive. F.e. favor-type-error?

Maybe `prefer-type-error` to stay consistent with our other rules?

florianb:
Right - `prefer-type-error` then?

jfmengels:
`prefer-type-error` it is

florianb:
Aye:

- [ ] Change rule name to `prefer-type-error`

sindresorhus:
@jfmengels @SamVerschueren Can you approve the PR if it looks good to you?

readme.md
```diff
@@ -67,6 +68,7 @@ Configure it in `package.json`.
 - [no-hex-escape](docs/rules/no-hex-escape.md) - Enforce the use of unicode escapes instead of hexadecimal escapes. *(fixable)*
 - [custom-error-definition](docs/rules/custom-error-definition.md) - Enforces the only valid way of `Error` subclassing. *(fixable)*
 - [prefer-starts-ends-with](docs/rules/prefer-starts-ends-with.md) - Prefer `String#startsWidth` & `String#endsWidth` over more complex alternatives.
+- [prefer-type-error](docs/rules/prefer-type-error.md) - Enforce throwing `TypeError` in type checking conditions.
```
SamVerschueren: Add `*(fixable)*` at the end of the line

rules/prefer-type-error.js
```diff
@@ -0,0 +1,127 @@
...
+	'isBoolean',
+	'isBuffer',
+	'isDate',
+	'isElement',
+	'isEmptyObject',
+	'isError',
+	'isFinite',
+	'isFunction',
+	'isInteger',
+	'isLength',
```
SamVerschueren: What is this check? Should it throw a `TypeError`? On mobile now but couldn't find what it does. So just double checking.

rules/prefer-type-error.js
```diff
@@ -0,0 +1,127 @@
...
+	'isArrayBuffer',
+	'isArrayLike',
+	'isArrayLikeObject',
+	'isBoolean',
+	'isBuffer',
+	'isDate',
+	'isElement',
+	'isEmptyObject',
+	'isError',
+	'isFinite',
```
SamVerschueren: Checking if something is finite doesn't imply that the type is wrong imo. What do you guys think?

rules/prefer-type-error.js
```diff
@@ -0,0 +1,127 @@
...
+	'isArguments',
+	'isArray',
+	'isArrayBuffer',
+	'isArrayLike',
+	'isArrayLikeObject',
+	'isBoolean',
+	'isBuffer',
+	'isDate',
+	'isElement',
+	'isEmptyObject',
```
SamVerschueren: Same comment as isFinite.

rules/prefer-type-error.js
```diff
@@ -0,0 +1,127 @@
...
+	'isView',
+	'isWeakMap',
+	'isWeakSet',
+	'isWindow',
+	'isXMLDoc'
+]);
+
+const tcGlobalIdentifiers = new Set([
+	'isNaN',
+	'isFinite'
```
SamVerschueren: Same as isFinite above.

rules/prefer-type-error.js
sindresorhus: I assume it's https://lodash.com/docs/4.17.2#isLength

SamVerschueren: Aha, thanks! Alright, this check is a valid one.

rules/prefer-type-error.js
florianb: Hmm.. [ES treats +/-Infinity as special Number type](https://tc39.github.io/ecma262/#sec-ecmascript-language-types-number-type) (like NaN). So from a technical perspective i'd say it is a `TypeError` like NaN.

From a users perspective this might not be intuitive in a semantical way, especially if you compare this meaning to the infinity of math. Where in my opinion infinity is mostly used to create a line where the number's values aren't of interest anymore. Whereas ES creates the line where numbers can't be treated anymore. So i personally take that as a sign of an invalid result of a math operation.

SamVerschueren: It just looks odd if you are forced to write this

```js
if (!Number.isFinite(x)) {
    throw new TypeError('The number is not finite'));
}
```

They might be interpreted like a special type, users don't see it like that. So I agree, it's a type error to JavaScript (technically), but not to developers. It's counter-intuitive.

rules/prefer-type-error.js
florianb: This function is borrowed from [jQuery](https://api.jquery.com/jQuery.isEmptyObject/) and checks

> "properties on the object itself and properties inherited from prototypes (in that it doesn't use hasOwnProperty). The argument should always be a plain JavaScript Object as other types of object (DOM elements, primitive strings/numbers, host objects) may not give consistent results across browsers."

I added this one because the check also checks for the presence of a PlainObject.

But you're right since this is including a non-typecheck i drop that identifier.

- [ ] Remove `isEmptyObject` from `tcIdentifiers`

readme.md
florianb: Right - thanks! 👍 

- [ ] Add `*(fixable)*` to the readme entry.

rules/prefer-type-error.js
florianb: Right - i remove it.

- [ ] remove `isFinite` from any identifiers lists

SamVerschueren: It's just my opinion though. Not sure with @sindresorhus and @jfmengels think?

sindresorhus: I dunno. I've been using `TypeError` for this: https://github.com/sindresorhus/crypto-random-string/blob/21960c996c3f77cf5f7bf6765e837e7c8a07e200/index.js#L5-L7

I guess my thinking is that, even though it's not a different type, it's not the correct form of the type.

jfmengels: I don't mind keeping it this way. `number` in JS is very large, and contains multiple number sub-types: positive numbers, negative numbers, floating numbers, infinites and not numbers (NaN. Actually, maybe this is the definition of an imaginary number :thinking: :open_mouth:).

I could very well imagine a function saying: "I'm expecting a finite number. You gave me an infinite number, which is not the correct type". I'm :+1: with keeping this.

SamVerschueren: Allright, let's keep it this way then. We can always reconsider if someone has issues with it later on.

SamVerschueren:
Great work @florianb! I Will approve when the latest changes are pushed.

sindresorhus:
@florianb Can you push the latest changes?

florianb:
You're welcome @jfmengels, @SamVerschueren & @sindresorhus!

Thank you very much for your advice - i'm taking a lot away - so i guess i will keep contributing. 😁

SamVerschueren:
Only the `isEmptyObject` test and done we're done :tada:!

florianb:
Thanks for your patience.. 😗  🎶

sindresorhus:
Yay! Excited to see this land. Thanks for working on it @florianb. Hope it was a good experience for you too.

5fa79994-b5ee-11e3-9a4d-b19557d7480b

florianb:
Yay - happy it landed! 🛩  I would really like to contribute further, so please feel free to assign me a new task (even in a different project like ava), i don't see any issue here which i'd think is adequate and is **not** picked by other contributors.

ready-dog-opt

sindresorhus:
@florianb That's great to hear! You could start looking through the AVA ["good for beginner"](https://github.com/avajs/ava/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3A%22good%20for%20beginner%22%20-label%3Aassigned) issues. Just comment on one you want or ask follow-up questions. Happy to chat about it in https://gitter.im/avajs/ava too :)

If you would like to create more ESLint rules, we're also looking for help in the [AVA ESlint plugin](https://github.com/avajs/eslint-plugin-ava).<br>[This issue](https://github.com/avajs/eslint-plugin-ava/issues/139), for example, shouldn't be too hard.

florianb:
That's great, thanks @sindresorhus. I am offline for 2 weeks but i will join the Ava-/eslint-issues after that. :smile:
Sindre Sorhus <notifications@github.com> schrieb am 29.01.2017 9:35 nachm.:

@florianb<https://github.com/florianb> That's great to hear! You could start looking through the AVA "good for beginner"<https://github.com/avajs/ava/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3A%22good%20for%20beginner%22%20-label%3Aassigned> issues. Just comment on one you want or ask follow-up questions. Happy to chat about it in https://gitter.im/avajs/ava too :)

If you would like to create more ESLint rules, we're also looking for help in the AVA ESlint plugin<https://github.com/avajs/eslint-plugin-ava>.
This issue<https://github.com/avajs/eslint-plugin-ava/issues/139>, for example, shouldn't be too hard.

-

PR #75, opened 2016-11-24
Add `custom-error-definition` rule - fixes #74

### LINKED ISSUE #74: Rule proposal: Enforce correct Error subclassing
There are many things people do wrong when subclassing an error. We only care about the class syntax in this rule.

## Useless `this.message` assignment:

```js
class CustomError extends Error {
	constructor(message) {
		super(message);
		this.message = message;
		this.name = this.constructor.name;
	}
}
```

Could just be:

```js
class CustomError extends Error {
	constructor(message) {
		super(message);
		this.name = this.constructor.name;
	}
}
```

## Forgetting to set `this.name`

```js
class CustomError extends Error {
	constructor(message) {
		super(message);
	}
}
```

Without it just displays `Error` when thrown. Should be:

```js
class CustomError extends Error {
	constructor(message) {
		super(message);
		this.name = this.constructor.name;
	}
}
```

## Forgetting to pass `message` to `super()`

```js
class CustomError extends Error {
	constructor(message) {
		super();
		this.message = message;
		this.name = this.constructor.name;
	}
}
```

Should be:

```js
class CustomError extends Error {
	constructor(message) {
		super(message);
		this.name = this.constructor.name;
	}
}
```

## Use `this.constructor.name`

```js
class CustomError extends Error {
	constructor() {
		super();
		this.name = this.constructor.name;
	}
}
```

instead of manually writing the name:

```js
class CustomError extends Error {
	constructor() {
		super();
		this.name = 'CustomError';
	}
}
```

Less chance for a typo/inconsistency between them.

---

Should be auto-fixable.

---

Thoughts? Any other things we should include? Name suggestion?

SamVerschueren:
Implemented #74. Not sure about the name though. Feedback on the error messages are welcome as well. And feedback for the implementation off course :).

TODO's
- [x] Pick the correct name
- [x] Write fixers

### DIFF
--- a/docs/rules/custom-error-definition.md
+++ b/docs/rules/custom-error-definition.md
@@ -75,7 +75,7 @@ class foo extends Error {
 }
 ```
 
-The class name is invalid, it should be capitalized and end with `Error`. In this case it should be `FooError`.
+The class name is invalid. It should be capitalized and end with `Error`. In this case it should be `FooError`.
 
 
 ## Pass
--- a/docs/rules/custom-error-definition.md
+++ b/docs/rules/custom-error-definition.md
@@ -66,6 +66,18 @@ class CustomError extends Error {
 The `name` property should be set to the class name.
 
 
+```js
+class foo extends Error {
+	constructor(message) {
+		super(message);
+		this.name = 'foo';
+	}
+}
+```
+
+The class name is invalid, it should be capitalized and end with `Error`. In this case it should be `FooError`.
+
+
 ## Pass
 
 ```js
--- a/docs/rules/custom-error.md
+++ b/docs/rules/custom-error.md
@@ -1,0 +1,96 @@
+# Enforce correct Error subclassing
+
+Enforces the only valid way of `Error` subclassing. It works with any super class that ends in `Error`.
+
+
+## Fail
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super(message);
+		this.message = message;
+		this.name = 'CustomError';
+	}
+}
+```
+
+The `this.message` assignment is useless as it's already set via the `super()` call.
+
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super();
+		this.message = message;
+		this.name = 'CustomError';
+	}
+}
+```
+
+Pass the error message to `super()` instead of setting `this.message`.
+
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super(message);
+	}
+}
+```
+
+No `name` property set. The name property is needed so the error shows up as `[CustomError: foo]` and not `[Error: foo]`.
+
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super(message);
+		this.name = this.constructor.name;
+	}
+}
+```
+
+Use a string literal to set the `name` property as it will not change after minifying.
+
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super(message);
+		this.name = 'MyError';
+	}
+}
+```
+
+The `name` property should be set to the class name.
+
+
+## Pass
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super(message);
+		this.name = 'CustomError';
+	}
+}
+```
+
+```js
+class CustomError extends Error {
+	constructor() {
+		super('My custom error');
+		this.name = 'CustomError';
+	}
+}
+```
+
+```js
+class CustomError extends TypeError {
+	constructor() {
+		super();
+		this.name = 'CustomError';
+	}
+}
+```
--- a/readme.md
+++ b/readme.md
@@ -43,7 +43,8 @@ Configure it in `package.json`.
 			"unicorn/escape-case": "error",
 			"unicorn/no-array-instanceof": "error",
 			"unicorn/no-new-buffer": "error",
-			"unicorn/no-hex-escape": "error"
+			"unicorn/no-hex-escape": "error",
+			"unicorn/custom-error-definition": "error"
 		}
 	}
 }
--- a/readme.md
+++ b/readme.md
@@ -63,6 +63,7 @@ Configure it in `package.json`.
 - [no-array-instanceof](docs/rules/no-array-instanceof.md) - Require `Array.isArray()` instead of `instanceof Array`. *(fixable)*
 - [no-new-buffer](docs/rules/no-new-buffer.md) - Enforce the use of `Buffer.from()` and `Buffer.alloc()` instead of the deprecated `new Buffer()`. *(fixable)*
 - [no-hex-escape](docs/rules/no-hex-escape.md) - Enforce the use of unicode escapes instead of hexadecimal escapes. *(fixable)*
+- [custom-error-definition](docs/rules/custom-error-definition.md) - Enforces the only valid way of `Error` subclassing. *(fixable)*
 
 
 ## Recommended config
--- a/rules/custom-error.js
+++ b/rules/custom-error.js
@@ -1,0 +1,140 @@
+'use strict';
+const upperfirst = require('lodash.upperfirst');
+
+const nameRegexp = /^(?:[A-Z][a-z0-9]*)*Error$/;
+
+const getClassName = name => {
+	name = upperfirst(name);
+
+	const suffix = name.slice(-5);
+
+	if (suffix === 'error') {
+		return name.slice(0, name.length - 5) + 'Error';
+	}
+
+	return suffix === 'Error' ? name : `${name}Error`;
+};
+
+const getConstructorMethod = className => `
+	constructor() {
+		super();
+		this.name = '${className}';
+	}
+`;
+
+const isSuperExpression = node => node.type === 'ExpressionStatement' && node.expression.type === 'CallExpression' && node.expression.callee.type === 'Super';
+
+const isAssignmentExpression = (node, name) => {
+	if (node.type !== 'ExpressionStatement' || node.expression.type !== 'AssignmentExpression') {
+		return false;
+	}
+
+	const lhs = node.expression.left;
+
+	if (!lhs.object || lhs.object.type !== 'ThisExpression') {
+		return false;
+	}
+
+	return lhs.property.name === name;
+};
+
+const create = context => {
+	return {
+		ClassDeclaration: node => {
+			if (!node.superClass || !nameRegexp.test(node.superClass.name)) {
+				return;
+			}
+
+			const name = node.id.name;
+			const className = getClassName(name);
+
+			if (name !== className) {
+				context.report({
+					node: node.id,
+					message: `Invalid class name, use \`${className}\`.`,
+					fix: fixer => fixer.replaceText(node.id, className)
+				});
+			}
+
+			const body = node.body.body;
+
+			const constructor = body.find(x => x.kind === 'constructor');
+
+			if (!constructor) {
+				context.report({
+					node,
+					message: 'Add a constructor to your error.',
+					fix: fixer => fixer.insertTextAfterRange([
+						node.body.start,
+						node.body.start + 1
+					], getConstructorMethod(className))
+				});
+				return;
+			}
+
+			const constructorBodyNode = constructor.value.body;
+			const constructorBody = constructorBodyNode.body;
+
+			const superExpression = constructorBody.find(isSuperExpression);
+			const messageExpressionIndex = constructorBody.findIndex(x => isAssignmentExpression(x, 'message'));
+
+			if (!superExpression) {
+				context.report({
+					node: constructorBodyNode,
+					message: 'Missing call to `super()` in constructor.'
+				});
+			} else if (messageExpressionIndex !== -1 && superExpression.expression.arguments.length === 0) {
+				const rhs = constructorBody[messageExpressionIndex].expression.right;
+
+				context.report({
+					node: superExpression,
+					message: 'Pass the error message to `super()`.',
+					fix: fixer => fixer.insertTextAfterRange([
+						superExpression.start,
+						superExpression.start + 6
+					], rhs.raw || rhs.name)
+				});
+			}
+
+			if (messageExpressionIndex !== -1) {
+				const expression = constructorBody[messageExpressionIndex];
+
+				context.report({
+					node: expression,
+					message: 'Pass the error message to `super()` instead of setting `this.message`.',
+					fix: fixer => fixer.removeRange([
+						messageExpressionIndex === 0 ? constructorBodyNode.start : constructorBody[messageExpressionIndex - 1].end,
+						expression.end
+					])
+				});
+			}
+
+			const nameExpression = constructorBody.find(x => isAssignmentExpression(x, 'name'));
+			const nameMessage = `The \`name\` property should be set to \`${className}\`.`;
+
+			if (!nameExpression) {
+				context.report({
+					node: constructorBodyNode,
+					message: nameMessage,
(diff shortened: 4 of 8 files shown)


rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
...
+				const superExpression = constructorBody.find(findSuperExpression);
+
+				if (!superExpression) {
+					context.report({
+						node: constructor.value.body,
+						message: 'No call to `super()`'
+					});
+				} else if (superExpression.expression.arguments.length === 0) {
+					context.report({
+						node, // superExpression.expression -> Tests fail
```
SamVerschueren: I wanted to set `superExpression.expression` here but if I do that, tests are failing. I have no clue why.

docs/rules/custom-error.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+```js
+class CustomError extends Error {
+	constructor(message) {
+		super(message);
+		this.name = this.constructor.name;
+	}
+}
+```
```
sindresorhus: I think each of the examples should say why they're incorrect as it's probably not clear to the user.

rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	const lhs = node.expression.left;
+
+	if (!lhs.object || lhs.object.type !== 'ThisExpression') {
+		return false;
+	}
+
+	return lhs.property.name === name;
+};
```
sindresorhus: @jfmengels Maybe these could be useful for https://github.com/jfmengels/eslint-ast-utils ?

rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
...
+
+const create = context => {
+	return {
+		ClassDeclaration: node => {
+			if (node.superClass && node.superClass.name === 'Error') {
```
sindresorhus: To reduce indentation, I would inverse this statement and return early.

test/custom-error.js
```diff
@@ -0,0 +1,144 @@
...
+					this.name = 'FooError';
+				}
+			}
+		`
+	],
+	invalid: [
+		{
+			code: 'class FooError extends Error { }',
+			errors: [
+				{ruleId: 'custom-error', message: 'Add a constructor to your error'}
```
sindresorhus: Put all the messages in variables or an object, so they can be reused in all the invalid rules here.

rules/custom-error.js
sindresorhus: Weird, but I don't think this should be reported like this regardless. It should be allowed not to pass a message to `super()`. We should report only when `constructor()` has at least one argument and `super()` has none.

test/custom-error.js
```diff
@@ -0,0 +1,144 @@
...
+			code: `
+				class FooError extends Error {
+					constructor() {
+						super();
+						this.message = 'My custom message';
+					}
+				}
+			`,
+			errors: [
+				{ruleId: 'custom-error', message: 'No `message` passed to `super()`'},
```
sindresorhus: This shouldn't report here.

rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
...
+						message: `The \`name\` property should be set to \`${name}\``
+					});
+				}
+
+				const messageExpression = constructorBody.find(x => findAssignmentExpression(x, 'message'));
+
+				if (messageExpression) {
+					context.report({
+						node: messageExpression,
+						message: 'Do not set the `message` property'
```
sindresorhus: > Pass the error message to `super()` instead of setting `this.message`

SamVerschueren:
Thanks for all the feedback. I also have to add support for other native error types right? Like `TypeError`, `RangeError`, ...

sindresorhus:
@SamVerschueren Yes, and would be nice to support subclassing of custom errors too, so maybe we can just check if the super class matches `/^(?:[A-Z][a-z0-9]*)*Error$/`?

SamVerschueren:
And also enforce the `Error` suffix naming convention.

sindresorhus:
> And also enforce the Error suffix naming convention.

Yeah, we should say in the docs that it works on any super class that ends in `Error`.

SamVerschueren:
Also, is there an easy way of fixing this error

```js
	constructor() { }
}
```

to something like this

```js
	constructor(message) {
		super(message);
		this.name = 'FooError';
	}
}
```

I can easily fix the right hand side of `this.name` and remove the `this.message` assignment. But adding new nodes with the correct indentation and syntax (some might not use `;` for instance).

Or should we just focus on the easy ones and improve in the future when things like that are easier to do?

sindresorhus:
I would just replace the constructor body node with a string of your second example. I assume the other rules would handle the correct indentation and semicolon, if not, the user can just fix it themselves. I think ESLint intends to implement multipass eventually, where this would be solved for us. @jfmengels might have a better idea than me though.

SamVerschueren:
That's what I had in mind as well but the user might have set some other custom properties as well which (I believe) we don't want to remove. Could extra those first of course.

sindresorhus:
@SamVerschueren I was just talking about your example where `constructor` is empty.

SamVerschueren:
Oh yes, sorry missed that. Will try to cover as much as possible without making it to complex. It doesn't take that long to fix it by hand either.

rules/custom-error.js
SamVerschueren: Maybe we should drop this check entirely, also the one you suggested. What if, not sure why though, you accept an argument in the constructor but aren't setting a message.

Just a stupid example which doesn't make much sense but for the sake of showing you.

```js
class HTTPError extends Error {
    constructor(statusCode) {
        super();
        this.statusCode = statusCode;
    }
}
```

docs/rules/custom-error.md
```diff
@@ -15,24 +15,33 @@ class CustomError extends Error {
...
 	constructor(message) {
-		super();
-		this.message = message;
-		this.name = = 'CustomError';
+		super(message);
 	}
 }
 ```
 
+No `name` property set.
```
sindresorhus: Maybe also add:

> The name property is needed so the error shows up as `[CustomError: foo]` and not `[Error: foo]`

docs/rules/custom-error.md
```diff
@@ -42,6 +51,20 @@ class CustomError extends Error {
 }
 ```
 
+Use a string literal to set the `name` property as it will not change after uglifying.
```
sindresorhus: uglifying => minifying

rules/custom-error.js
jfmengels: And a `.` at the end, to be consistent with other rules

rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
+'use strict';
+const findSuperExpression = node => node.type === 'ExpressionStatement' && node.expression.type === 'CallExpression' && node.expression.callee.type === 'Super';
+
+const findAssignmentExpression = (node, name) => {
```
jfmengels: isAssignmentExpression. Same for findSuperExpression.

rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
...
+					context.report({
+						node,
+						message: 'Add a constructor to your error'
+					});
+					return;
+				}
+
+				const constructorBody = constructor.value.body.body;
+
+				const superExpression = constructorBody.find(findSuperExpression);
```
SamVerschueren: Lol, yes :p

rules/custom-error.js
```diff
@@ -0,0 +1,78 @@
+'use strict';
+const findSuperExpression = node => node.type === 'ExpressionStatement' && node.expression.type === 'CallExpression' && node.expression.callee.type === 'Super';
```
SamVerschueren: I hope/think you are confusing the callee type with something else? ASTExplorer shows me `Super`. But might be missing something.

screen shot 2016-11-27 at 13 14 09

SamVerschueren:
I think I processed all the feedback.

Some of the fixed outputs might have weird indentation though but it works :).

test/custom-error.js
```diff
@@ -0,0 +1,331 @@
...
+const invalidClassNameError = {ruleId: 'custom-error', message: 'Invalid class name, use `FooError`.'};
+const constructorError = {ruleId: 'custom-error', message: 'Add a constructor to your error.'};
+const noSuperCallError = {ruleId: 'custom-error', message: 'Missing call to `super()` in constructor.'};
+const invalidNameError = {ruleId: 'custom-error', message: 'The `name` property should be set to `FooError`.'};
+const passMessageToSuperError = {ruleId: 'custom-error', message: 'Pass the error message to `super()`.'};
+const invalidMessageAssignmentError = {ruleId: 'custom-error', message: 'Pass the error message to `super()` instead of setting `this.message`.'};
+
+ruleTester.run('custom-error', rule, {
+	valid: [
+		'class Foo { }',
```
jfmengels: Could you add:
- a test case with a class that extends something that is not an Error
- a test case with a class without name (to make sure it doesn't crash at any point), both when extending Error and not extending it (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Classes#Class_expressions)
- a test case where the superclass is not an Identifier but something else: `class A extends Foo() {}` (just ignore it)
- a test case where the superclass is not an Identifier but a MemberExpression: `class A extends Foo.Bar {}`

For the last one, I think we could check if the property's name ends with Error (like `Http.ProtocolError`), but I'm fine ignoring it for know and add it if it comes up at some point. Anyway, this looks like it would crash the plugin at the moment.

rules/custom-error.js
```diff
@@ -0,0 +1,140 @@
...
+			}
+
+			const name = node.id.name;
+			const className = getClassName(name);
+
+			if (name !== className) {
+				context.report({
+					node: node.id,
+					message: `Invalid class name, use \`${className}\`.`,
+					fix: fixer => fixer.replaceText(node.id, className)
```
jfmengels: Are we sure we want this to be autofixed ?
If you are writing the new Error class and the linter fixes it right away, I agree that it's useful. But if I apply this rule to an existing code base and the error is used elsewhere, this would probably cause some crashes. IMO this should be fixed either manually or with a codemod (and even that would be hard).

SamVerschueren: I guess you are right. Let's see what @sindresorhus thinks about it and then I will remove it.

test/custom-error.js
SamVerschueren: Thanks for the feedback, will add these.

rules/custom-error.js
sindresorhus: I agree with @jfmengels

test/custom-error.js
sindresorhus: > I think we could check if the property's name ends with Error (like Http.ProtocolError)

👍

sindresorhus:
Alternative name: `error-subclass`. Undecided which I prefer.

jfmengels:
I'm thinking along `custom-error-definition` / `custom-error-declaration`.

SamVerschueren:
`custom-error-definition` sounds good to me.

rules/custom-error.js
SamVerschueren: This rule isn't fixed automatically, but I assume we should still set `this.name` automatically to the correct class name?

```js
class fooError extends Error {
    constructor() {
        super();
        this.name = 'foo';
    }
}
```

would become


```js
class fooError extends Error {
    constructor() {
        super();
        this.name = 'FooError';
    }
}
```

test/custom-error.js
SamVerschueren: > a test case with a class without name (to make sure it doesn't crash at any point), both when extending Error and not extending it 

What should we pick as `name` property? The name of the identifier?

SamVerschueren: Added the extra tests.

rules/custom-error.js
sindresorhus: @SamVerschueren No, for the same reason, the name might be checked somewhere and cause problems if changed. Reporting is enough.

rules/custom-error.js
sindresorhus: Done: https://github.com/jfmengels/eslint-ast-utils/issues/9

docs/rules/custom-error.md
```diff
@@ -0,0 +1,96 @@
...
+	constructor(message) {
+		super(message);
+		this.name = 'MyError';
+	}
+}
+```
+
+The `name` property should be set to the class name.
+
+
```
sindresorhus: Can you add an example of invalid class name?

Per the `Invalid class name, use ${className}` violation.

sindresorhus:
> custom-error-definition

👍

sindresorhus:
Can you also add it to the readme?

rules/custom-error.js
SamVerschueren:
```js
class foo extends Error {
    constructor() {
        super();
    }
}
```

Would the error in this case be

1. The `name` property should be set to `FooError`.
2. The `name` property should be set to `foo`.

Meaning, should we match with the current class name, or should we match with what the class name should be. In the second case, if they resolve the class name manually, it will eventually show the first error message though. So the second one might be more correct.

rules/custom-error.js
```diff
@@ -0,0 +1,140 @@
...
+	name = upperfirst(name);
+
+	const suffix = name.slice(-5);
+
+	if (suffix === 'error') {
+		return name.slice(0, name.length - 5) + 'Error';
+	}
+
+	return suffix === 'Error' ? name : `${name}Error`;
+};
```
SamVerschueren: Apparently yes, nice!

SamVerschueren:
Added the tweaks. Only thing we have to discuss is https://github.com/sindresorhus/eslint-plugin-unicorn/pull/75#discussion_r89957322.

Thanks for all the feedback! Appreciate the time you guys put into it!

readme.md
```diff
@@ -63,6 +63,7 @@ Configure it in `package.json`.
 - [no-array-instanceof](docs/rules/no-array-instanceof.md) - Require `Array.isArray()` instead of `instanceof Array`. *(fixable)*
 - [no-new-buffer](docs/rules/no-new-buffer.md) - Enforce the use of `Buffer.from()` and `Buffer.alloc()` instead of the deprecated `new Buffer()`. *(fixable)*
 - [no-hex-escape](docs/rules/no-hex-escape.md) - Enforce the use of unicode escapes instead of hexadecimal escapes. *(fixable)*
+- [custom-error-definition](docs/rules/custom-error-definition.md) - Enforces the only valid way of `Error` subclassing. *(fixable)*
```
sindresorhus: It also needs to be added to the rule config above. (In `"rules": {`)

docs/rules/custom-error-definition.md
```diff
@@ -66,6 +66,18 @@ class CustomError extends Error {
...
+```js
+class foo extends Error {
+	constructor(message) {
+		super(message);
+		this.name = 'foo';
+	}
+}
+```
+
+The class name is invalid, it should be capitalized and end with `Error`. In this case it should be `FooError`.
```
sindresorhus:
```
The class name is invalid. It should be capitalized and end with `Error`. In this case it should be `FooError`.
```

rules/custom-error.js
sindresorhus: 2

SamVerschueren:
I think I covered all the feedback.

jfmengels:
Thanks a lot @SamVerschueren :D

SamVerschueren:
Thanks for the support! :)

sindresorhus:
🎉✨🦄🌸🤘

PR #86, opened 2017-02-06
Bug fix for number only escapes in `escape-case` rule

### LINKED ISSUE #83: `escape-case` rule false positive with number only Unicode escapes
For example `'\u2500'`.

Bebersohl:
Fixes #83


sindresorhus:
Thank you :)

PR #88, opened 2017-03-07
Add support for custom errors in the `throw-new-error` rule

### LINKED ISSUE #4: Add option to the `throw-new-error` rule for working with custom errors
From: https://github.com/sindresorhus/eslint-plugin-xo/commit/53b7133c910695557d9c4a10079fac791b32c4dc#commitcomment-17077491

> An option that accepts an array of string/regex of error names? Could default to [/Error$/].
## 

liubinyi:
added customer error by using /^(?:[A-Z][a-z0-9]*)*Error$/

fixes #4

### DIFF
--- a/rules/throw-new-error.js
+++ b/rules/throw-new-error.js
@@ -9,14 +9,14 @@ const errorTypes = [
 	'TypeError',
 	'URIError'
 ];
-const customerError = /^(?:[A-Z][a-z0-9]*)*Error$/;
+const customeError = /^(?:[A-Z][a-z0-9]*)*Error$/;
 
 const create = context => ({
 	ThrowStatement: node => {
 		const arg = node.argument;
 		const error = arg.callee;
 
-		if (arg.type === 'CallExpression' && (errorTypes.indexOf(error.name) !== -1 || customerError.test(error.name))) {
+		if (arg.type === 'CallExpression' && customeError.test(error.name)) {
 			context.report({
 				node,
 				message: 'Use `new` when throwing an error.',
--- a/rules/throw-new-error.js
+++ b/rules/throw-new-error.js
@@ -1,14 +1,5 @@
 'use strict';
 
-const errorTypes = [
-	'Error',
-	'EvalError',
-	'RangeError',
-	'ReferenceError',
-	'SyntaxError',
-	'TypeError',
-	'URIError'
-];
 const customeError = /^(?:[A-Z][a-z0-9]*)*Error$/;
 
 const create = context => ({
--- a/rules/throw-new-error.js
+++ b/rules/throw-new-error.js
@@ -1,13 +1,13 @@
 'use strict';
 
-const customeError = /^(?:[A-Z][a-z0-9]*)*Error$/;
+const customError = /^(?:[A-Z][a-z0-9]*)*Error$/;
 
 const create = context => ({
 	ThrowStatement: node => {
 		const arg = node.argument;
 		const error = arg.callee;
 
-		if (arg.type === 'CallExpression' && customeError.test(error.name)) {
+		if (arg.type === 'CallExpression' && customError.test(error.name)) {
 			context.report({
 				node,
 				message: 'Use `new` when throwing an error.',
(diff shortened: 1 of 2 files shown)


rules/throw-new-error.js
```diff
@@ -9,13 +9,14 @@ const errorTypes = [
 	'TypeError',
 	'URIError'
 ];
+const customerError = /^(?:[A-Z][a-z0-9]*)*Error$/;
```
jfmengels: It's actually a "custom" error, not a "customer" error.

rules/throw-new-error.js
```diff
@@ -9,13 +9,14 @@ const errorTypes = [
...
 
 const create = context => ({
 	ThrowStatement: node => {
 		const arg = node.argument;
 		const error = arg.callee;
 
-		if (arg.type === 'CallExpression' && errorTypes.indexOf(error.name) !== -1) {
+		if (arg.type === 'CallExpression' && (errorTypes.indexOf(error.name) !== -1 || customerError.test(error.name))) {
```
jfmengels: If the regex is well made, we don't need the `errorTypes.indexOf(error.name) !== -1` check anymore. If you remove it and all tests pass, it should be fine to remove.

test/throw-new-error.js
```diff
@@ -14,7 +14,8 @@ ruleTester.run('new-error', rule, {
 	valid: [
 		'throw new Error()',
 		'new Error()',
-		'throw new TypeError()'
+		'throw new TypeError()',
+		'throw new CustomerError()'
```
jfmengels: everywhere you use "customer", replace by "custom"

rules/throw-new-error.js
jfmengels: Is there a use for the `?:`? I'm not sure what it does.

test/throw-new-error.js
```diff
@@ -28,6 +29,11 @@ ruleTester.run('new-error', rule, {
 			errors
 		},
 		{
+			code: `throw CustomerError('foo')`,
+			output: `throw new CustomerError('foo')`,
+			errors
+		},
```
jfmengels: Can you add a few tests that make sure the regex is well built? I'm thinking things like the following, but add more if you think of any

Valid
```
throw fooBarBazError('foo') // does not start with an upper case character
```

Invalid
```
throw FooBarBazError('foo') // With multiple upper case characters
throw ABCError('foo') // With multiple upper case characters in a row
throw Abc3Error('foo') // With numbers
```

rules/throw-new-error.js
liubinyi: will update them:
i believe ?: is look ahead expression.

test/throw-new-error.js
liubinyi: sure.

rules/throw-new-error.js
liubinyi: ah yeah, i will update it.

liubinyi:
@jfmengels Thanks for reviewing and providing useful feedback! I updated the PR base on you comments. i will update it again once other questions are cleared out.

rules/throw-new-error.js
```diff
@@ -1,21 +1,13 @@
...
-const errorTypes = [
-	'Error',
-	'EvalError',
-	'RangeError',
-	'ReferenceError',
-	'SyntaxError',
-	'TypeError',
-	'URIError'
-];
+const customeError = /^(?:[A-Z][a-z0-9]*)*Error$/;
```
sindresorhus: `customeError` => `customError`

liubinyi: update coming soon!

liubinyi: @sindresorhus thanks for reviewing, updated.

sindresorhus:
Looks good. Thank you! :)

rules/throw-new-error.js
kj35878: > 将更新它们。 I believe ？： 表示该组已匹配，但未被捕获以进行反向引用
我是俊爷，你300多RMB礼物再来一份，不准再坑了，联系方式，手机+8619372278999，我有很多问题问你，看在6年朋友的份上，马上联系我，我得推特是kuangjun，头像是一条龙。脸书联系不上你。

PR #96, opened 2017-06-21
Add non-zero length style option to `explicit-length-check` rule

### LINKED ISSUE #56: Enforce length check style in the `explicit-length-check` rule
`explicit-length-check` checks that `length` properties get compared to a value instead of checking the truthiness of the property.

In https://github.com/avajs/eslint-plugin-ava/pull/149#pullrequestreview-3799305, we remembered that there are some opinions on how that value then gets compared :sweat_smile: 

For instance:

``` js
a.length === 0
// vs
a.length < 1 // not sure I've ever seen this one TBH

a.length !== 0
// vs
a.length > 0
// vs
a.length >= 1

// + some I might have forgotten
```

It could be nice to be able to specify how you want your comparisons to be made, and make that consistent across your project.

I'm not sure what the options would look like. At the top of my head, I'm thinking:

``` json
{
  "unicorn/explicit-length-check": ["error", {
    "empty": ["eq", 0] | ["lt", 1], // one or the other
    "not-empty": ["ne", 0] | ["gt", 0] | ["gte", 1]
  }]
}
```

Should one be omitted (`empty` for instance), then you'd simply get no errors reports on how that one gets checked (both `a.length < 1` `a.length === 0` would be fine).

This sounds pretty complicated at the moment (especially if we want to support `==` too), so I'm not too fond of it at the moment, especially it would not make sense to have `empty: ["equal", 1200]` as the config. Maybe we could simplify it to

``` json
{
  "unicorn/explicit-length-check": ["error", {
    "empty": "eq" | "lt",
    "not-empty": "ne" | "gt" | "gte"
  }]
}
```

and then we have predefined values associated to each of those operators. We could also have longer names for the operators, but you'd then get `greater-than-or-equal` (which is okay, but pretty long).

I'm kind of proposing this system so that we could have the same system for a different rule checking the `indexOf` uses. It might be nice to think of a similar "API" for it.

Oh, and obviously, this should not report any check for specific indices: `a.length === 2` and `a.length == 2` should always be fine, whatever the settings may be.

paulobmarcos:
Fixes #56

In short it aims to add functionality to the rule `explicit-length-check` by providing some options to determine the preferred way of checking for length properties.

The `explicit-length-check` rule now enforces length comparison with `!== 0` when checking for an empty array. It auto-fixes when the check is made with '< 1'. 

#### Fail
```js
if (string.length < 1) {}
```
#### Pass
```js
if (array.length !== 0) {}
```

For non zero length comparisons we can define our preferred way of checking it by providing an option:
```js
{
  'unicorn/explicit-length-check': ['error', {
    'non-zero': 'not-equal'
  }]
}
```
where `non-zero` can be one of the following:
- `not-equal`
	- this option makes sure that non-zero is checked with: `array.length !== 0`
- `greater-than`
	- this option makes sure that non-zero is checked with: `array.length > 0`
- `greater-than-or-equal`
	- this option makes sure that non-zero is checked with: `array.length >= 1`

*If no options are provided it will work as before, allowing any of this possibilities.* 

What I did was to check for `BinaryExpressions` and look for the `empty` case and, if defined in the options, the `not-empty` case.

It will check if any of the cases, `empty` or `not-empty`, are contemplated in the `BinaryExpression`  by comparing the operator and the literal value. 
If any of the cases have a match and the operator+value are of other option types, than it is rejected and a fix is provided with the preferred way.

### DIFF
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -39,17 +39,17 @@ if (array.length !== 0) {}
 ## Non-Zero Comparisons
 
 You can define your preferred way of checking non-zero length by providing a `non-zero` option:
-```json
+```js
 {
-  "unicorn/explicit-length-check": ["error", {
-    "non-zero": "not-equal"
+  'unicorn/explicit-length-check': ['error', {
+    'non-zero': 'not-equal'
   }]
 }
 ```
-where `"non-zero"` can be configured with one of the following:
-- `"not-equal"`
+where `non-zero` can be configured with one of the following:
+- `not-equal`
 	- this option makes sure that non-zero is checked with: `array.length !== 0`
-- `"greater-than"`
+- `greater-than`
 	- this option makes sure that non-zero is checked with: `array.length > 0`
-- `"greater-than-or-equal"`
+- `greater-than-or-equal`
 	- this option makes sure that non-zero is checked with: `array.length >= 1`
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -2,7 +2,7 @@
 
 Enforce explicitly checking the length of a value array in an `if` condition, rather than checking the truthiness of the length.
 
-## Fail
+### Fail
 
 ```js
 if (string.length) {}
@@ -11,7 +11,7 @@ if (!array.length) {}
 ```
 
 
-## Pass
+### Pass
 
 ```js
 if (string.length > 0) {}
@@ -20,28 +20,37 @@ if (array.length !== 0) {}
 if (array.length === 0) {}
 ```
 
-## Options
+## Empty Comparisons
 
-You can set options for `empty` and `not-empty` comparisons like this:
+Enforce length comparison with `!== 0` when checking for an empty array.
 
-```json
-"unicorn/explicit-length-check": ["error", {
-  "empty": "emptyOption",
-  "not-empty": "notEmptyOption"
-}]
+### Fail
+
+```js
+if (string.length < 1) {}
 ```
 
-where:
-- `"emptyOption"` can be one of the following:
-	- `"eq"` (equal)
-		- this option makes sure that empty is checked with: `a.length === 0 `
-	- `"lt"` (less-than)
-		- this option makes sure that empty is checked with: `a.length < 1`
+### Pass
 
-- `"notEmptyOption"` can be one of the following:
-	- `"ne"` (not-equal)
-		- this option makes sure that empty is checked with: `a.length !== 0`
-	- `"gt"` (greater-than)
-		- this option makes sure that empty is checked with: `a.length > 0`
-	- `"gte"` (greater-than or equal-to)
-		- this option makes sure that empty is checked with: `a.length >= 1`
+```js
+if (array.length !== 0) {}
+```
+
+## Non Zero Comparisons
+
+You can define your preferred way of checking non zero length by providing an option:
+```js
+{
+  "unicorn/explicit-length-check": ["error", {
+    "not-empty": "notEmptyOption"
+  }]
+}
+```
+where `"notEmptyOption"` can be one of the following:
+- `"ne"` (not-equal)
+	- this option makes sure that non zero is checked with: `array.length !== 0`
+- `"gt"` (greater-than)
+	- this option makes sure that non zero is checked with: `array.length > 0`
+- `"gte"` (greater-than or equal-to)
+	- this option makes sure that non zero is checked with: `array.length >= 1`
+
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -19,3 +19,29 @@ if (array.length > 0) {}
 if (array.length !== 0) {}
 if (array.length === 0) {}
 ```
+
+## Options
+
+You can set options for `empty` and `not-empty` comparisons like this:
+
+```json
+"unicorn/explicit-length-check": ["error", {
+  "empty": "emptyOption",
+  "not-empty": "notEmptyOption"
+}]
+```
+
+where:
+- `"emptyOption"` can be one of the following:
+	- `"eq"` (equal)
+		- this option makes sure that empty is checked with: `a.length === 0 `
+	- `"lt"` (less-than)
+		- this option makes sure that empty is checked with: `a.length < 1`
+
+- `"notEmptyOption"` can be one of the following:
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for working on this :)

I think we should just drop the `emptyOption` option. `a.length < 1` makes no sense` and I've never seen it in the wild.

I'm ok with the `notEmptyOption` option.

sindresorhus:
> Also added a new check. When the literal comes before the length property it also rejects it.
1 > array.length should be array.length < 1. What do you think?

That should already be handled by the ESLInt [`yoda`](http://eslint.org/docs/rules/yoda) rule.

rules/explicit-length-check.js
```diff
@@ -1,5 +1,93 @@
...
+			node,
+			message
+		});
+	}
+}
+
+function checkEmptyType(type, operator, value) {
+	switch (type) {
+		case 'eq':
+			if (types.lt.includes(operator) && value === 1) {
```
sindresorhus: You can't use `.includes()`. We still support Node.js 4.

sindresorhus:
Would you also be able to add a fixer? So it can be auto-fixed.

rules/explicit-length-check.js
```diff
@@ -1,5 +1,93 @@
...
+	}
+}
+
+function checkNotEmptyType(type, operator, value) {
+	switch (type) {
+		case 'gt':
+			if ((types.gte.includes(operator) && value === 1) ||
+				(types.ne.includes(operator) && value === 0)
+			) {
+				return 'not empty `length` should be compared with `> 0`.';
```
sindresorhus:
```js
return 'Non-zero `.length` should be compared with `> 0`.';
```

rules/explicit-length-check.js
```diff
@@ -1,5 +1,93 @@
...
+			node,
+			message
+		});
+	}
+}
+
+function checkEmptyType(type, operator, value) {
+	switch (type) {
+		case 'eq':
+			if (types.lt.indexOf(operator) >= 0 && value === 1) {
```
sindresorhus: `>= 0` → `!== -1`

paulobmarcos:
Hey @sindresorhus! Sorry, didn't have the opportunity to look into this right away.
I removed the `emptyOption` and it now check for `!== 0` as the default. 
Also added an auto-fixer for both empty and non zero length comparisons. 
And you are right, the ESLint `yoda` rule already handles one of the cases 👍 

Hope it looks good 🎉

rules/explicit-length-check.js
paulobmarcos: Good one! It reads better that way. ✌️

rules/explicit-length-check.js
paulobmarcos: Yes I noticed that and pushed a fix for it right after 😁

docs/rules/explicit-length-check.md
```diff
@@ -11,11 +11,46 @@ if (!array.length) {}
...
+if (array.length !== 0) {}
+```
+
+## Non Zero Comparisons
+
+You can define your preferred way of checking non zero length by providing an option:
+```js
+{
+  "unicorn/explicit-length-check": ["error", {
+    "not-empty": "notEmptyOption"
```
sindresorhus: I think `not-empty` should be `non-zero`.

And use single-quotes.

docs/rules/explicit-length-check.md
```diff
@@ -11,11 +11,46 @@ if (!array.length) {}
...
+if (string.length < 1) {}
+```
+
+### Pass
+
+```js
+if (array.length !== 0) {}
+```
+
+## Non Zero Comparisons
```
sindresorhus: Non-Zero

docs/rules/explicit-length-check.md
```diff
@@ -11,11 +11,46 @@ if (!array.length) {}
...
+  }]
+}
+```
+where `"notEmptyOption"` can be one of the following:
+- `"ne"` (not-equal)
+	- this option makes sure that non zero is checked with: `array.length !== 0`
+- `"gt"` (greater-than)
+	- this option makes sure that non zero is checked with: `array.length > 0`
+- `"gte"` (greater-than or equal-to)
+	- this option makes sure that non zero is checked with: `array.length >= 1`
```
sindresorhus: I would prefer to use verbose option values for readability. So:

- `not-equal`
- `greater-than`
- `greater-than-or-equal`

rules/explicit-length-check.js
```diff
@@ -1,5 +1,95 @@
...
+	ne: ['!==', '!=']
+};
+
+const getFixDetails = (node, operator, value) => {
+	return {
+		node,
+		operator,
+		value
+	};
+};
```
sindresorhus: This feels kinda moot. Just use an object in the actual calls instead.

rules/explicit-length-check.js
```diff
@@ -33,5 +125,7 @@ const create = context => {
 
 module.exports = {
 	create,
-	meta: {}
```
sindresorhus: The option should be defined using a `schema` so it can be validated by ESLint

See: http://eslint.org/docs/developer-guide/working-with-rules#options-schemas

rules/explicit-length-check.js
```diff
@@ -1,5 +1,95 @@
...
+			);
+		})
+	});
+}
+
+function checkEmptyType(context, node) {
+	if (node.operator === '<' && node.right.value === 1) {
+		reportError(context,
+			node,
+			'Empty `.length` should be compared with `=== 0`.',
```
sindresorhus: Empty => Zero

sindresorhus:
Sorry about the slow response. Summer ;)

Looks pretty good. Just some last minor nitpick.

docs/rules/explicit-length-check.md
paulobmarcos: `non-zero` seems good 👍 

One thing, since this configuration is done in a `.json` file and since the JSON standard is double-quotes, shouldn't the example also be in double-quotes?

sindresorhus: ESLint config can be in both JS and JSON.

paulobmarcos: Ok, I will change it to single-quotes!

paulobmarcos:
Hey @sindresorhus! No worries!

Just checked your review and pushed some changes. ✌️

sindresorhus:
Great, thanks :)

PR #97, opened 2017-07-05
Add `no-fn-reference-in-iterator` rule - fixes #91

### LINKED ISSUE #91: Prevent passing a function directly to iterator methods
You have a dependency:

```js
module.exports = x => x + 1;
```

You use it like this (which many do):

```js
const unicorn = require('unicorn');

[1, 2, 3].map(unicorn);
//=> [2, 3, 4]
```

The `unicorn` module now does a minor version that adds another argument:

```js
module.exports = (x, y) => x + (y ? y : 1);
```

Your code will now return something different and probably break for users:

```js
const unicorn = require('unicorn');

[1, 2, 3].map(unicorn);
//=> [2, 3, 5]
```

Instead, you should do:

```js
const unicorn = require('unicorn');

[1, 2, 3].map(x => unicorn(x));
//=> [2, 3, 4]
```

---

This would have prevented: https://github.com/avajs/ava/pull/1343

---

It should apply to:

- Array#map
- Array#forEach
- Array#every
- Array#filter
- Array#find
- Array#some
- Map#forEach
- Set#forEach

Anything I missed?

SamVerschueren:
Trying to fix #91 which prevents passing a function directly to an iterator method.

I think that we should also support [`reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce) and [`reduceRight`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/ReduceRight). Not sure about [`sort`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort). These are somewhat different though because they accept multiple arguments so should be handled separately.

Just let me know if they should be supported as well.

### DIFF
--- a/docs/rules/no-function-iterator.md
+++ b/docs/rules/no-function-iterator.md
@@ -47,6 +47,13 @@ const fn = x => x === 2;
 [1, 2, 3].some(fn);
 ```
 
+```js
+const fn = x => x === 2;
+
+[1, 2, 3].map(m({foo: 'bar'}));
+```
+
+
 ## Pass
 
 ```js
@@ -89,4 +96,10 @@ const fn = x => x === 1;
 const fn = x => x === 2;
 
 [1, 2, 3].some(x => fn(x));
+```
+
+```js
+const fn = x => x === 2;
+
+[1, 2, 3].map(x => m({foo: 'bar'})(x));
 ```
--- a/docs/rules/no-function-iterator.md
+++ b/docs/rules/no-function-iterator.md
@@ -1,6 +1,34 @@
 # Prevents passing a function directly to iterator methods
 
-Prevents passing a function directly to an iterator method to make it more clear with the function accepts.
+Suppose you have a `unicorn` module:
+
+```js
+module.exports = x => x + 1;
+```
+
+You can then use it like this:
+
+```js
+const unicorn = require('unicorn');
+
+[1, 2, 3].map(unicorn);
+//=> [2, 3, 4]
+```
+
+The `unicorn` module now does a minor version that adds another argument:
+
+```js
+module.exports = (x, y) => x + (y ? y : 1);
+```
+
+Your code will now return something different and probably break for users because it is now passing the index of the item as second argument.
+
+```js
+const unicorn = require('unicorn');
+
+[1, 2, 3].map(unicorn);
+//=> [2, 3, 5]
+```
 
 
 ## Fail
--- a/docs/rules/no-function-iterator.md
+++ b/docs/rules/no-function-iterator.md
@@ -1,0 +1,92 @@
+# Prevents passing a function directly to iterator methods
+
+Prevents passing a function directly to an iterator method to make it more clear with the function accepts.
+
+
+## Fail
+
+```js
+const fn = x => x + 1;
+
+[1, 2, 3].map(fn);
+```
+
+```js
+const fn = x => console.log(x);
+
+[1, 2, 3].forEach(fn);
+```
+
+```js
+const fn = x => x < 10;
+
+[1, 2, 3].every(fn);
+```
+
+```js
+const fn = x => x % 2;
+
+[1, 2, 3].filter(fn);
+```
+
+```js
+const fn = x => x === 1;
+
+[1, 2, 3].find(fn);
+```
+
+```js
+const fn = x => x === 1;
+
+[1, 2, 3].findIndex(fn);
+```
+
+```js
+const fn = x => x === 2;
+
+[1, 2, 3].some(fn);
+```
+
+## Pass
+
+```js
+const fn = x => x + 1;
+
+[1, 2, 3].map(x => fn(x));
+```
+
+```js
+const fn = x => console.log(x);
+
+[1, 2, 3].forEach(x => fn(x));
+```
+
+```js
+const fn = x => x < 10;
+
+[1, 2, 3].every(x => fn(x));
+```
+
+```js
+const fn = x => x % 2;
+
+[1, 2, 3].filter(x => fn(x));
+```
+
+```js
+const fn = x => x === 1;
+
+[1, 2, 3].find(x => fn(x));
+```
+
+```js
+const fn = x => x === 1;
+
+[1, 2, 3].findIndex(x => fn(x));
+```
+
+```js
+const fn = x => x === 2;
+
+[1, 2, 3].some(x => fn(x));
+```
--- a/rules/no-function-iterator.js
+++ b/rules/no-function-iterator.js
@@ -1,0 +1,34 @@
+'use strict';
+const iteratorMethods = [
+	'map',
+	'forEach',
+	'every',
+	'filter',
+	'find',
+	'findIndex',
+	'some'
+];
+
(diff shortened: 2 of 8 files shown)


kevva:
I definitely think `sort` should be allowed. I often pass in other modules directly in the callback. Not sure about `reduce` though.

SamVerschueren:
Don't think sort should be in the list either. It always has 2 arguments.

kevva:
Also, should we handle closures? Might be overdoing it...

```js
[1, 2, 3].map(fn({unicorn: true}));
```

SamVerschueren:
Good idea. Think we should yeah!

docs/rules/no-function-iterator.md
```diff
@@ -0,0 +1,92 @@
+# Prevents passing a function directly to iterator methods
+
+Prevents passing a function directly to an iterator method to make it more clear with the function accepts.
+
```
sindresorhus: The intro should include why it's a useful rule. See: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/91#issue-219641506 I don't think most realize the issue with doing this.

rules/no-function-iterator.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+const iteratorMethods = [
```
sindresorhus: Should be a `Set`.

SamVerschueren:
@sindresorhus any feedback on `reduce` and `reduceRight`?

docs/rules/no-function-iterator.md
```diff
@@ -0,0 +1,105 @@
+# Prevents passing a function directly to iterator methods
+
+Prevents passing a function directly to an iterator method to make it more clear with the function accepts.
```
sindresorhus: Don't repeat the title in the text here.

rules/no-function-iterator.js
```diff
@@ -0,0 +1,36 @@
...
+	'findIndex',
+	'some'
+];
+
+const isIteratorMethod = node => node.callee.property && iteratorMethods.indexOf(node.callee.property.name) !== -1;
+const hasFunctionArgument = node => node.arguments.length === 1 && (node.arguments[0].type === 'Identifier' || node.arguments[0].type === 'CallExpression');
+
+const create = context => ({
+	CallExpression: node => {
+		const sourcecode = context.getSourceCode();
```
sindresorhus: Getting the source code is expensive. We should only do it when we actually need it.

sindresorhus:
Wonder if we could use AST selectors for this: http://eslint.org/docs/developer-guide/selectors

sindresorhus:
> any feedback on `reduce` and `reduceRight`?

`reduce()` and `reduceRight()` doesn't make sense, as you always use all their arguments.

sindresorhus:
I think the rule name could be clearer. Not exactly sure what though. Something like `no-function-directly-in-iterator`, but that's too long.

SamVerschueren:
> 

I was able to do it like this

```js
const iteratorMethods = [
	'filter',
	'map',
	'forEach',
	'every',
	'filter',
	'find',
	'findIndex',
	'some'
];

const selector = 'CallExpression[callee.property.name=/^' + iteratorMethods.join('|') + '$/]';
```

That way I can drop the `isIteratorMethod` function. Don't really have a strong preference for one of them.

> I think the rule name could be clearer

Wasn't sure about it either. Can't really come up with something that really fits it.

- `no-function-ref-iterator`
- `no-function-ref-in-iterator`
- `no-fn-ref-in-iterator`
- `no-fn-directly-in-iterator`

docs/rules/no-function-iterator.md
```diff
@@ -0,0 +1,133 @@
+# Prevents passing a function directly to iterator methods
+
+Suppose you have a `unicorn` module:
```
SamVerschueren: I rewrote the example a bit. Feel free to provide feedback on wording/style/...

rules/no-function-iterator.js
```diff
@@ -0,0 +1,44 @@
...
+
+const isIteratorMethod = node => node.callee.property && iteratorMethods.has(node.callee.property.name);
+const hasFunctionArgument = node => node.arguments.length === 1 && (node.arguments[0].type === 'Identifier' || node.arguments[0].type === 'CallExpression');
+
+const parseArgument = (context, arg) => {
+	if (arg.type === 'Identifier') {
+		return arg.name;
+	}
+
+	const sourcecode = context.getSourceCode();
```
sindresorhus: This variable feels moot, just use it directly below.

sindresorhus:
> That way I can drop the `isIteratorMethod` function. Don't really have a strong preference for one of them.

Nah, I prefer what you already have.

sindresorhus:
Maybe we just go verbose `no-function-reference-in-iterator`, unless @kevva can come up with something better.

docs/rules/no-function-iterator.md
sindresorhus: Looks good!

sindresorhus:
Actually, let's go with `no-fn-reference-in-iterator`. `fn` is a common `function` abbreviation.

SamVerschueren:
I think I processed all the feedback

kevva:
> `reduce()` and `reduceRight()` doesn't make sense, as you always use all their arguments.

Not necessarily true since they have `index` and `array` too, although it might be an edge case.

> Maybe we just go verbose `no-function-reference-in-iterator`, unless [@kevva](https://github.com/kevva) can come up with something better.

I think `no-fn-reference-in-iterator` is good too. Had a look at what ESLint uses themselves to see if we could stay consistent with their naming, but they're using both `function` and `func` in their rule names.

SamVerschueren:
I prefer `fn` above `func` to be honest.

sindresorhus:
> Not necessarily true since they have `index` and `array` too, although it might be an edge case.

Oh, this is embarrassing. I didn't know that. Never used those arguments. I guess they make sense to include after all then.

> I prefer `fn` above `func` to be honest.

Me too.

sindresorhus:
Just a thought. Should we allow this if the function used is defined in the same scope?

Should be make an exception for built-in types? This is a very common pattern `[true, 1, 2].filter(Boolean)`.

SamVerschueren:
> Should we allow this if the function used is defined in the same scope?

I think we shouldn't. It's easy to overlook something and someone can also add a second argument to the function without really noticing this caveat.

> Should we make an exception for built-in types?

We could, but I'd suggest using a whitelist. `parseInt` is also built-in but you don't want someone to write `['1', '2', '3'].map(parseInt)`.

sindresorhus:
> I think we shouldn't. It's easy to overlook something and someone can also add a second argument to the function without really noticing this caveat.

Yeah, you're probably right.

> We could, but I'd suggest using a whitelist. `parseInt` is also built-in but you don't want someone to write `['1', '2', '3'].map(parseInt)`.

Actually, I can only think of `Boolean` as an actual use-case. Maybe we just add that first and add others later as needed.

sindresorhus:
Good stuff Sam. This one will be very useful :)

SamVerschueren:
Woop woop, thanks for the feedback :)

PR #99, opened 2017-07-09
Allow `_` as error name if not used inside catch block - fixes #87

### LINKED ISSUE #87: Allow `_` as a name in `catch-error-name` rule if it's not used
Thought of it when I saw: https://github.com/avajs/ava/pull/1265/files#diff-b72e3e25f31d6b3c3e5f40b5991dc5f6R43

Sometimes you don't need the error and you should be able to name it `_` to ignore it. This is a common pattern many use and it's used in some ESLint core rules too I think.

// @novemberborn

SamVerschueren:
This PR fixes #87 by allowing `_` if it's not used in the catch block.

### DIFF
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -46,6 +46,15 @@ try {
 }
 ```
 
+```js
+try {
+	doSomething();
+} catch (_) {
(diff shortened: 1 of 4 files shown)


SamVerschueren:
And thank you for the awesome library :)

PR #100, opened 2017-07-09
Add `new-regexp` rule - fixes #92

### LINKED ISSUE #92: Require `new RegExp()` in instead of `RegExp()`
`RegExp()` and `new RegExp()` are no different, but would be nice to enforce a consistent style. This rule should enforce `new RegExp()` as that's how it's commonly used and matches other constructors like `new Error()` and user-land constructors.  Should be fixable.

For reference, similar rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/throw-new-error.js

I think we could use AST selectors for this: http://eslint.org/docs/developer-guide/selectors

SamVerschueren:
Not sure about the name but couldn't come up with something better.

### DIFF
--- a/docs/rules/new-regexp.md
+++ b/docs/rules/new-regexp.md
@@ -1,16 +1,17 @@
-# Enforce the use of `new RegExp()` instead of `RegExp`
+# Enforce the use of `new RegExp()` instead of `RegExp()`
+
+They work the same, but `new RegExp()` should be preferred for consistency with other constructors.
 
 
 ## Fail
 
 ```js
-const regexp = RegExp('foo', 'g');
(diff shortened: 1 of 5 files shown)


docs/rules/new-regexp.md
```diff
@@ -0,0 +1,16 @@
+# Enforce the use of `new RegExp()` instead of `RegExp`
+
```
SamVerschueren: Should we add a description here? Docs are quite... empty. But to be honest, does it need more content? It's quite straight forward now.

sindresorhus:
Awesome! :)

PR #101, opened 2017-07-09
Use AST selector for `no-new-buffer` rule

SamVerschueren:
Looks cleaner with a selector.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,6 +64,6 @@
 		"xo": "*"
 	},
 	"peerDependencies": {
-		"eslint": ">=3.6"
+		"eslint": ">=4"
 	}
 }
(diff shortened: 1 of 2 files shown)


jfmengels:
You're right, it looks better with selectors. But the plugin currently has a peer dependency to [ESLint >= 3.6](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/package.json#L67), whereas AST selectors landed in [ESLint 3.18.0](http://eslint.org/blog/2017/03/eslint-v3.18.0-released). So landing this would break this rule for any projects using ESLint < 3.18.0 :/

sindresorhus:
@SamVerschueren Can you bump the peerDependency to ESLint v4? We'll have to do a major version anyways (because of new default rules).

SamVerschueren:
Done

PR #102, opened 2017-07-10
Add `import-index` rule - fixes #9

### LINKED ISSUE #9: Enforce import index style
I want to enforce `require('.');` and `from '.';`. Before Node.js 4 you had to do `'./'`. In Node.js 4 you can do: `'.'`.

Should catch the following cases:
- `'./'`
- `'./index'`
- `'./index.js'`

Should be easy to implement auto-fixing for this rule.

SamVerschueren:
Implemented #9.

### DIFF
--- a/rules/import-index.js
+++ b/rules/import-index.js
@@ -1,0 +1,27 @@
+'use strict';
+const isImportingIndex = m => m.endsWith('/') || m.endsWith('index.js') || m.endsWith('index');
+const normalize = m => m.replace(/(\/|\/index(.js)?)$/, '');
+
+const importIndex = (context, node, m) => {
+	if (isImportingIndex(m.value)) {
+		context.report({
+			node,
+			message: 'Import index files with `.`',
+			fix: fixer => fixer.replaceText(m, `'${normalize(m.value)}'`)
+		});
+	}
+};
+
+const create = context => {
+	return {
+		'CallExpression[callee.name="require"]': node => importIndex(context, node, node.arguments[0]),
+		ImportDeclaration: node => importIndex(context, node, node.source)
+	};
+};
+
+module.exports = {
+	create,
+	meta: {
+		fixable: 'code'
+	}
+};
--- a/rules/import-index.js
+++ b/rules/import-index.js
@@ -1,6 +1,7 @@
 'use strict';
-const isImportingIndex = m => m.endsWith('/') || m.endsWith('index.js') || m.endsWith('index');
-const normalize = m => m.replace(/(\/|\/index(.js)?)$/, '');
+const regexp = /^(@.*?\/.*?|[./]+?.*?)(?:\/(?:index(?:\.js)?)?)$/;
+const isImportingIndex = m => regexp.test(m);
+const normalize = m => m.replace(regexp, '$1');
 
 const importIndex = (context, node, m) => {
 	if (isImportingIndex(m.value)) {
--- a/rules/import-index.js
+++ b/rules/import-index.js
@@ -7,7 +7,7 @@ const importIndex = (context, node, m) => {
 	if (isImportingIndex(m.value)) {
 		context.report({
 			node,
-			message: 'Import index files with `.`',
+			message: 'Do not reference the index file directly',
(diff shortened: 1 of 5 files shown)


rules/import-index.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+const isImportingIndex = m => m.endsWith('/') || m.endsWith('index.js') || m.endsWith('index');
```
sindresorhus: `index.js` => `/index.js` and `index` => `/index`, otherwise something `require('./lindex.js')` will trigger it. Why not just reuse the regex below?

jfmengels: Related to this, I'd like to see a few valid tests with all kinds of imported modules, to make sure nothing breaks

'foobar', 'index', 'fooindex', 'indexbar', (add an at sign before the next 2) 'index/foo', 'foo/index.js', 'index.js', './foobar/' and more if you can think of some.

SamVerschueren:
Good ideas from both of you. Will try to have a look later today.

SamVerschueren:
I suddenly realised that we have dynamic imports as well, but it seems it can't be parsed yet.

rules/import-index.js
SamVerschueren: @jfmengels I guess you didn't mean `index` and `index.js`? Regarding the case `@foo/index.js`, does it mean the package is called `@foo/index.js` and actually you are importing `@foo/index.js/index.js`?

jfmengels: I did mean `index` and `index.js`. They are valid package names (and they apparently do exist).
You can use https://github.com/npm/validate-npm-package-name to check the validity of npm package names.

For `@foo/index.js`, it is as you said.

SamVerschueren:
Alright, this became a little bit more complex than initially thought. Thanks for coming up with those use cases @jfmengels because I totally missed those. The thing is, I don't think the error message is accurate anymore because `package/index` becomes `package`. Any ideas for something better?

test/import-index.js
```diff
@@ -0,0 +1,128 @@
...
+			errors: [error],
+			output: `const m = require('../..')`
+		},
+		{
+			code: `const m = require('./foo/index.js')`,
+			errors: [error],
+			output: `const m = require('./foo')`
+		},
+		{
+			code: `const m = require('./foobar/')`,
```
SamVerschueren: Is this use case correct? I think it is but not sure.

kevva:
What if the user has defined the `main` field in `package.json`? E.g. if `main` is `dist.js` and you do `require('.')` it'll `require` `dist.js`.

SamVerschueren:
People can always ignore the rule. I think it is a valid use case, but we can't take the package.json into account.

test/import-index.js
jfmengels: Style-wise, my preference is to remove the trailing `/`, so this is :ok_hand: for me

rules/import-index.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+const regexp = /^(@.*?\/.*?|[./]+?.*?)(?:\/(?:index(?:\.js)?)?)$/;
+const isImportingIndex = m => regexp.test(m);
+const normalize = m => m.replace(regexp, '$1');
+
+const importIndex = (context, node, m) => {
+	if (isImportingIndex(m.value)) {
+		context.report({
+			node,
+			message: 'Import index files with `.`',
```
jfmengels: `message :` -> `message:`

jfmengels:
This looks good to me, except as you said for the error message.
Something along the lines of `Do not reference the index file directly` maybe? I think this is explicit enough, especially as there is autofix for this rule.

rules/import-index.js
sindresorhus: This was actually caused by a Refined GitHub bug.

SamVerschueren:
Changed the error message to `Do not reference the index file directly` 👍

sindresorhus:
Yay! Been wanting this rule for a long time. Thanks Sam!

SamVerschueren:
Yeey! Thanks for the feedback all of you!

mischah:
@SamVerschueren I ❤️ this rule. Thank you very much.

PR #103, opened 2017-07-12
Allow number postfix in `catch-error-name` to prevent shadowed variables - fixes #93

### LINKED ISSUE #93: `catch-error-name` rule should allow for number postfix
Sometimes I need to use a different name so not to shadow an earlier error. This rule should accept a number postfix from `2` to `9`.

For example:

```js
const handleError = err => {
	try {
		options.logger(err);
	} catch (err2) {
		err = new Error('option.logger threw error:\n' + err2.stack);
	}
};
```

---

`_` should be allowed as a name too.

SamVerschueren:
Not sure if I made it too complex. Could've just allowed any number postfix any time. So if you rather want that, let me know and I'll change it :).

### DIFF
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -55,6 +55,16 @@ try {
 }
 ```
 
+```js
+const handleError = err => {
+	const err2 = new Error('foo bar');
+
+	obj.catch(err3 => {
+		// `err3` is allowed because of shadowed variables
+	});
+}
+```
+
 
 ## Options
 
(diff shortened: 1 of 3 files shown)


SamVerschueren:
> Could you add a failing test case where the name is overridden to `error`?

👍 

> I think that should be a valid case, but it may not be a bad idea to ask to rename to `err2`.

We currently don't have control over other variables in the scope. So not sure what you're asking here. Do you mean in this case

```js
const handleError = err => {
	const err9 = new Error('foo bar');

	try {
		doSomething();
	} catch (foo) {
		console.log(foo);
	}
}
```

We should ask to rename `foo` to `err2` or do you mean we should ask to rename `err9` to `err2`? In the first case, got you covered :). The second case is a lot harder because I think it would make the code much more complex.

> The original issue also mentioned the following, which has not been treated for promises I think:

Ugh, good catch. Should've done it in #99. Let me create a new PR for that and focus this one on the number postfix.

sindresorhus:
Looks good to me too :)

PR #104, opened 2017-07-15
Add support for `_` as error name in promises

SamVerschueren:
In #99 I allowed the use of `_` for `try-catch` blocks, but I forgot to do this for promises as well. This PR adds the possibility to use `_` as a name in promises when the error name is not used.

### DIFF
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -65,6 +65,13 @@ const handleError = err => {
 }
 ```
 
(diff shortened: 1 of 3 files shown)


SamVerschueren:
@jfmengels If you pass in `({message})` as argument, this rule will fail because you should use `err`. I added extra tests for this use case. If you want I can add those tests in a separate PR because it's not really related to the underscore issue.

PR #106, opened 2017-07-19
Add integration tests - fixes #84

### LINKED ISSUE #84: Integration tests
We should run all the rules using ESLint against a huge amount of code to ensure it never crashes. We could include the source of some larger projects.

I usually manually run master against AVA to as an ensurance, but I forgot this time.

This could have prevented https://github.com/sindresorhus/eslint-plugin-unicorn/issues/81 and https://github.com/sindresorhus/eslint-plugin-unicorn/issues/82.

SamVerschueren:
This is still WIP, but feedback always welcome. Sorry for the `package.json` tabs vs spaces thing. Will fix it when everything is ready.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ matrix:
   include:
     - node_js: 8
       env: INTEGRATION=true
+  allow_failures:
+    - node_js: 8
+      env: INTEGRATION=true
 before_install:
   - npm i -g npm
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ matrix:
 before_install:
   - npm i -g npm
 script:
-  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
+  - "[ \"$INTEGRATION\" = true ] && npm run integration || npm test"
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,6 @@ node_js:
   - '8'
   - '6'
   - '4'
-env: TEST=npm test
-matrix:
-  include:
-    - node_js: 8
-      env: TEST=npm run integration
 before_install:
   - npm i -g npm
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,12 @@ node_js:
   - '8'
   - '6'
   - '4'
-env: TEST=npm test
 matrix:
-  allow_failures:
+  include:
     - node_js: 8
-      env: TEST=npm run integration
+      env: INTEGRATION=true
 before_install:
   - npm i -g npm
 script:
-  - (eval "$TEST")
+  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ matrix:
 before_install:
   - npm i -g npm
 script:
-  - if [ "$INTEGRATION" = true ]; then "npm run integration" else "npm test"; fi
+  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ node_js:
   - '4'
 env: TEST=npm test
 matrix:
-  include:
+  allow_failures:
     - node_js: 8
       env: TEST=npm run integration
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,8 @@ matrix:
   include:
     - node_js: 8
       env: INTEGRATION=true
-  allow_failures:
-    - node_js: 8
-      env: INTEGRATION=true
 before_install:
   - npm i -g npm
 script:
-  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
+  - if [ "$INTEGRATION" = true ]; then "npm run integration" else "npm test"; fi
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ node_js:
   - '8'
   - '6'
   - '4'
+env: TEST=npm test
+matrix:
+  include:
+    - node_js: 8
+      env: TEST=npm run integration
 before_install:
   - npm i -g npm
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,13 @@ node_js:
   - '8'
   - '6'
   - '4'
+env: TEST=npm test
+matrix:
+  include:
+    - node_js: 8
+      env: TEST=npm run integration
 before_install:
   - npm i -g npm
+script:
+  - (eval "$TEST")
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ matrix:
 before_install:
   - npm i -g npm
 script:
-  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
+  - [ "$INTEGRATION" = true ] && npm run integration || npm test
 after_success: npm run coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,5 +10,5 @@ matrix:
 before_install:
   - npm i -g npm
 script:
-  - [ "$INTEGRATION" = true ] && npm run integration || npm test
+  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
 after_success: npm run coveralls
--- a/package.json
+++ b/package.json
@@ -1,80 +1,81 @@
 {
-	"name": "eslint-plugin-unicorn",
-	"version": "2.1.2",
-	"description": "Various awesome ESLint rules",
-	"license": "MIT",
-	"repository": "sindresorhus/eslint-plugin-unicorn",
-	"author": {
-		"name": "Sindre Sorhus",
-		"email": "sindresorhus@gmail.com",
-		"url": "sindresorhus.com"
-	},
-	"maintainers": [
-		{
-			"name": "James Talmage",
-			"email": "james@talmage.io",
-			"url": "github.com/jamestalmage"
-		},
-		{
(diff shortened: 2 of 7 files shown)


SamVerschueren:
Tests are working now correctly. I defined the integration tests so that they are allowed to fail. I think it's more a notification for ourselves when new rules are introduced?

Any other packages that you guys like to see tested?

jfmengels:
Nice! That works well!

A few things/ideas:

I understood the initial issue was about detecting crashes. Right now, the process fails because there are linting errors in the other projects, which is normal in this case because there are breaking changes, but normally that should not block the test. Do you know if ESLint fails returns a different exit code depending on whether there was a crash or whether there are linting errors? If not, I guess we could grep the output and check it manually.

That said, when not planning breaking changes, having the test fail when there are linting errors is great. Maybe we could have a flag that we add manually when we plan a major version that allows linting errors.

Ideally, we should test all rules with (almost) all possible configurations on, but that's hard because the projects we lint against do not enforce all the rules or with the alternative configurations. For the sake of testing crashes though, especially for new rules, there should probably be an additional test matrix like the following:
- For every project we test on
  - Test recommended config (**do not** allow linting errors, unless a major version is planned, but no crash)
  - Test config with all rules not in recommended or with alternative configurations (allow linting errors but no crashes). If there are multiple alternative configurations to test, we could create a matrix of configuration files to run the test with.

Alternatively, we could clone some of the projects once (`avajs/ava` --> `@SamVerschueren/ava-eslint-test-repo`), adapt them to follow the alternative configurations, and run the tests against those. Those would only need to be updated at every major version.

Anyway, this is a great start and my comments could perhaps be the basis of a new issue/PR rather than integrate this all into this PR. Great work @SamVerschueren!

SamVerschueren:
> I understood the initial issue was about detecting crashes. Right now, the process fails because there are linting errors in the other projects

A good catch! You are right, it shouldn't fail when the code does not passes the linter, it should fail when a rule forgets something and it crashes.

> Do you know if ESLint fails returns a different exit code depending on whether there was a crash or whether there are linting errors? If not, I guess we could grep the output and check it manually.

We'll have to check it manually, I couldn't find a difference.

> That said, when not planning breaking changes, having the test fail when there are linting errors is great. Maybe we could have a flag that we add manually when we plan a major version that allows linting errors.

Think that might be hard to maintain. Because after a breaking change release, we'll have to turn off the flag again. If we forget that, it will silently pass. If you have any ideas on that, feel free to elaborate :).

> Ideally, we should test all rules with (almost) all possible configurations on

You're right, good suggestion. I'm going to think about how we can do that in an easy manner. Think I already have some ideas without forking or whatsoever. But maybe let's focus on the initial use case and improve further in later PRs like you suggest.

---------------------------
So the next thing that I will do is make sure the integration tests just passes when no actual error occurred (and remove the `allow_failures` flag in `.travis.yml`.

Any additional feedback? Extra repositories that we can lint?

test/integration/test.js
```diff
@@ -0,0 +1,68 @@
...
+'use strict';
+const path = require('path');
+const Listr = require('listr');
+const tempy = require('tempy');
+const execa = require('execa');
+const del = require('del');
+
+const packages = new Map([
+	['got', 'https://github.com/sindresorhus/got'],
+	['ava', 'https://github.com/avajs/ava']
```
sindresorhus: Some more:

- https://github.com/sindresorhus/pageres
- https://github.com/sindresorhus/np
- https://github.com/sindresorhus/ora
- https://github.com/sindresorhus/p-map
- https://github.com/sindresorhus/detect-indent
- https://github.com/sindresorhus/os-locale
- https://github.com/sindresorhus/execa
- https://github.com/sindresorhus/pify
- https://github.com/sindresorhus/boxen
- https://github.com/sindresorhus/make-dir


@kevva Can you list some of your larger repos?

@SamVerschueren You should add `listr` and some of your other repos.

test/integration/test.js
```diff
@@ -0,0 +1,68 @@
...
+		}
+	], {
+		exitOnError: false
+	});
+};
+
+const list = new Listr([
+	{
+		title: 'Setup',
+		task: () => execa('npm', ['install', '../../../', 'eslint'], {cwd})
```
sindresorhus: `../../../` => `../../..`

test/integration/test.js
```diff
@@ -0,0 +1,68 @@
...
+	},
+	{
+		title: 'Integration tests',
+		task: () => {
+			const tests = new Listr({concurrent: true});
+
+			for (const [name, url] of packages.entries()) {
```
sindresorhus: `.entries()` is moot here. Just do `for (const [name, url] of packages) {`

test/integration/test.js
```diff
@@ -0,0 +1,68 @@
...
+	}
+]);
+
+list.run()
+	.catch(err => {
+		for (const error of err.errors) {
+			console.error(error.message);
+		}
+
+		process.exit(-1);
```
sindresorhus: Why `-1`?

.travis.yml
```diff
@@ -3,6 +3,15 @@ node_js:
...
+  include:
+    - node_js: 8
+      env: INTEGRATION=true
+  allow_failures:
+    - node_js: 8
+      env: INTEGRATION=true
 before_install:
   - npm i -g npm
+script:
+  - if [ "$INTEGRATION" = true ]; then npm run integration else npm test; fi
```
sindresorhus: Can you quote the commands? It's a bit hard to read without quotes or syntax highlighting.

test/integration/test.js
SamVerschueren: Should've been `1` :).

.travis.yml
SamVerschueren: What do you mean with `quote the commands`? I can't do this:

```
- if [ "$INTEGRATION" = true ]; then "npm run integration" else "npm test"; fi
```

I can do this if you like it more

```
- if [ "$INTEGRATION" = true ]; then npm run integration; fi
- if [ -z "$INTEGRATION ]; then npm test; fi
```

Not sure about the `-z`, have to verify. But you'll get the idea :).

sindresorhus: > - if [ "$INTEGRATION" = true ]; then "npm run integration" else "npm test"; fi

Oh, I assumed that would work.

How about:

```
 [ "$INTEGRATION" = true ] && npm run integration || npm test
```

test/integration/test.js
```diff
@@ -0,0 +1,92 @@
...
+	const dest = tempy.directory();
+
+	return new Listr([
+		{
+			title: 'Cloning',
+			task: () => execa('git', ['clone', packages.get(name), '--single-branch', dest])
+		},
+		{
+			title: 'Running tests',
+			task: () => execa('./node_modules/.bin/eslint', ['--config', path.join(cwd, 'index.js'), dest], {cwd})
```
sindresorhus: You could use the `localDir` option here.

test/integration/test.js
```diff
@@ -0,0 +1,92 @@
...
+				.catch(err => {
+					if (!/✖ [0-9]+ problems? \([0-9]+ errors?, [0-9]+ warnings?\)/.test(err.message)) {
```
sindresorhus: `[0-9]` => `\d`

package.json
```diff
@@ -1,70 +1,81 @@
...
+    "xo": "*"
+  },
+  "peerDependencies": {
+    "eslint": ">=4"
+  },
+  "ava": {
+    "files": [
+      "test/*.js"
+    ]
+  }
```
sindresorhus: You changed the indentation of package.json

SamVerschueren: I don't use npm@5 yet :p

.travis.yml
SamVerschueren: Doesn't seem to be valid yaml https://travis-ci.org/sindresorhus/eslint-plugin-unicorn/builds/256575546

test/integration/test.js
SamVerschueren: Fixed. If you have a better idea on how to find out if the error are ESLint warnings or something else, feel free to let me know.

test/integration/test.js
SamVerschueren: Good catch!

.travis.yml
sindresorhus: Ah, you have to quote it:

```
  - "[ \"$INTEGRATION\" = true ] && npm run integration || npm test"
```

Another reason I hate YAML. You never know when you have to quote or not.

SamVerschueren: Ugh! That seems to work, thanks!

test/integration/test.js
kevva: I don't have that many large repos anymore, but here are some more suggestions:

* https://github.com/kevva/to-ico
* https://github.com/kevva/download
* https://github.com/kevva/brightness
* https://github.com/kevva/decompress
* https://github.com/kevva/npm-conf
* https://github.com/imagemin/imagemin
* https://github.com/chalk/wrap-ansi

sindresorhus: And https://github.com/chalk/chalk https://github.com/qix-/color-convert

SamVerschueren: Added! Thanks guys!

test/integration/test.js
```diff
@@ -0,0 +1,102 @@
...
+	['bragg-router', 'http://github.com/SamVerschueren/bragg-router'],
+	['dev-time', 'http://github.com/SamVerschueren/dev-time'],
+	['decode-uri-component', 'https://github.com/SamVerschueren/decode-uri-component'],
+	['to-ico', 'https://github.com/kevva/to-ico'],
+	['download', 'https://github.com/kevva/download'],
+	['brightness', 'https://github.com/kevva/brightness'],
+	['decompress', 'https://github.com/kevva/decompress'],
+	['npm-conf', 'https://github.com/kevva/npm-conf'],
+	['imagemin', 'https://github.com/imagemin/imagemin'],
+	['ctolor-conver', 'https://github.com/qix-/color-convert']
```
sindresorhus: typo

SamVerschueren: Doh, thanks :)

SamVerschueren:
Anything else I can do for the integration tests?

sindresorhus:
🙌 It's great!

SamVerschueren:
Thanks for the feedback guys! :)

PR #109, opened 2017-07-28
Fix number-only escapes in strings with lowercase letters

### LINKED ISSUE #108: escape-case false positive if the string has lowercase characters but the escape has numbers only
For example: `'a\u0151'`

nkovacs:
Fixes #108


sindresorhus:
Thanks for fixing :)

PR #111, opened 2017-08-07
Add `new-for-builtins` rule - fixes #105

### LINKED ISSUE #105: `new-for-builtins` rule
Unsurprisingly my ESLint proposal was declined: https://github.com/eslint/eslint/issues/7634.
*(I will no longer waste time opening any more issues there.)*

The rule should enforce `new` for the following:

```
Array
ArrayBuffer
DataView
Date
Error
Float32Array
Float64Array
Function
Int16Array
Int32Array
Int8Array
Map
Promise
RegExp
Set
Symbol
Uint16Array
Uint32Array
Uint8Array
Uint8ClampedArray
WeakMap
WeakSet
```

And disallow it for:

```
Boolean
Number
Object
String
```

Some of these are covered by the ESLint [`new-cap`](http://eslint.org/docs/rules/new-cap) rule, but that's irrelevant as the rule should not be a prerequisite.

SamVerschueren:
This PR introduces the `new-for-builtins` rule discussed in #105. I'm just wondering if the name covers the rule entirely, because it also disallows `new` for 4 builtins.

Feedback welcome!


SamVerschueren:
Oh yes, I almost forgot. What should we do with the `new-regexp` rule? Just remove it? As it is covered by this one. If we are going to remove it, should I do it in this PR?

sindresorhus:
> What should we do with the new-regexp rule? Just remove it?

Yup, remove it.

docs/rules/new-for-builtins.md
```diff
@@ -0,0 +1,70 @@
...
+- `Uint32Array`
+- `Uint8ClampedArray`
+
+Disallows the use of `new` for following builtins.
+
+- `String`
+- `Number`
+- `Boolean`
+- `Object`
+
```
sindresorhus: Maybe add a note on why these are an exception. Something like:

> These should not use `new` as that would create object wrappers for the primitive values, which is not what you want. However, without `new` they can be useful for coercing a value to that type.

docs/rules/new-for-builtins.md
```diff
@@ -0,0 +1,70 @@
...
+- `WeakMap`
+- `Set`
+- `WeakSet`
+- `Promise`
+- `RegExp`
+- `Symbol`
+- `Uint8Array`
+- `Uint16Array`
+- `Uint32Array`
+- `Uint8ClampedArray`
```
sindresorhus: Just realized, we need to include `Object` here instead of the below. `Object` and `new Object` both wraps the object and doesn't coerce. And the `no-new-object` ESLint rule only catches `new Object()`, not `Object`.

docs/rules/new-for-builtins.md
```diff
@@ -0,0 +1,70 @@
+# Enforce the use of `new` for all builtins
```
sindresorhus: This needs to be the same title as the description in the readme.

readme.md
```diff
@@ -75,6 +76,7 @@ Configure it in `package.json`.
 - [no-fn-reference-in-iterator](docs/rules/no-fn-reference-in-iterator.md) - Prevents passing a function reference directly to iterator methods. *(fixable)*
 - [new-regexp](docs/rules/new-regexp.md) - Enforce the use of `new RegExp()` instead of `RegExp()`. *(fixable)*
 - [import-index](docs/rules/import-index.md) - Enforce importing index files with `.`. *(fixable)*
+- [new-for-builtins](docs/rules/new-for-builtins.md) - Enforce the use of `new` for all builtins, except `String`, `Number` `Boolean`, and `Object`. *(fixable)*
```
sindresorhus: Remove `Object` per my other comment.

rules/new-for-builtins.js
```diff
@@ -0,0 +1,69 @@
...
+
+const create = context => {
+	return {
+		CallExpression: node => {
+			const name = node.callee.name;
+
+			if (enforceNew.has(name)) {
+				context.report({
+					node,
+					message: `Use \`new ${name}()\` instead of \`${name}()\``,
```
sindresorhus: End the message with a `.`

docs/rules/new-for-builtins.md
SamVerschueren: 👍

SamVerschueren:
Thanks for the feedback @sindresorhus !

test/new-for-builtins.js
```diff
@@ -0,0 +1,198 @@
...
+	ruleId: 'new-for-builtins',
+	message: `Use \`new ${builtin}()\` instead of \`${builtin}()\`.`
+});
+
+const disallowNewError = builtin => ({
+	ruleId: 'new-for-builtins',
+	message: `Use \`${builtin}()\` instead of \`new ${builtin}()\`.`
+});
+
+ruleTester.run('new-regexp', rule, {
```
sindresorhus: `new-regexp` => `new-for-builtins`

SamVerschueren: Good catch :)

sindresorhus:
🙌 Looks great!

SamVerschueren:
Jeej! :tada:

PR #113, opened 2017-08-11
Add `regex-shorthand` rule - fixes #112

### LINKED ISSUE #112: Rule proposal: Regex shorthands
Prefer:

- `\d` instead of `[0-9]`
- `\D` instead of `[^0-9]`
- `\w` instead of `[A-Za-z0-9_]` or `[A-Za-z\d_]`
- `\w` instead of `[a-z0-9_]` or `[a-z\d_]` with the `i` flag
- `\W` instead of `[^A-Za-z0-9_]` or `[^A-Za-z\d_]`
- `\W` instead of `[^a-z0-9_]` or `[^a-z\d_]` with the `i` flag

For regex literals and the `RegExp` constructor (As long as it's a simple string).

Any order inside the brackets should work.

Might be more shorthands we can infer, but this would be a good start. I have to comment about the first case in PRs all the time.

---

@SamVerschueren @jfmengels Thoughts?

SamVerschueren:
This PR adds a new `regex-shorthand` rule proposed in #112. I created a new package for this, [`clean-regexp`](https://github.com/samverschueren/clean-regexp) because I realised that every regular expression should be matched in any order as explained here https://github.com/sindresorhus/eslint-plugin-unicorn/issues/112#issuecomment-321161108.

### DIFF
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -1,0 +1,27 @@
+# Enforce the use of regex shorthands to improve readability
+
+
+## Fail
+
+```js
+const regex = /[0-9]/;
+const regex = /[^0-9]/;
+const regex = /[a-zA-Z0-9_]/;
+const regex = /[a-z0-9_]/i;
+const regex = /[^a-zA-Z0-9_]/;
+const regex = /[^a-z0-9_]/i;
+const regex = /[0-9]\.[a-zA-Z0-9_]\-[^0-9]/i;
+```
+
+
+## Pass
+
+```js
+const regex = /\d/;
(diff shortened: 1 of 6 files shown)


docs/rules/regex-shorthand.md
```diff
@@ -0,0 +1,27 @@
+# Enforce the use of regex shorthands to improve readability
+
```
SamVerschueren: Do we need an additional description?

sindresorhus:
Looks great. Thanks Sam :)

SamVerschueren:
:tada: Thanks!

PR #115, opened 2017-09-17
Add `prefer-includes` rule

nunofreitasbotelho:
Implements #8 with the regex part still missing for now.
Used @jeremyjs's base code with the improvements suggested by @sindresorhus and @jfmengels 
 here: #70 due to inactivity from @jeremyjs.
Glad to work on it with more improvements that you find worthy.

### DIFF
--- a/rules/prefer-includes.js
+++ b/rules/prefer-includes.js
@@ -1,4 +1,5 @@
 'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
 
 const isIndexOfCallExpression = node => {
 	if (node.type !== 'CallExpression') {
@@ -105,6 +106,9 @@ const create = context => ({
 module.exports = {
 	create,
 	meta: {
+		docs: {
+			url: getDocsUrl()
+		},
 		fixable: 'code'
 	}
 };
--- a/rules/prefer-includes.js
+++ b/rules/prefer-includes.js
@@ -1,0 +1,110 @@
+'use strict';
+
+const isIndexOfCallExpression = node => {
+	if (node.type !== 'CallExpression') {
+		return false;
+	}
+
+	const property = node.callee.property;
+
+	return property.name === 'indexOf';
+};
+
+const isUnaryNotExpression = node => (
+	node.type === 'UnaryExpression' && node.operator === '!'
+);
+
+const isNegativeOne = (operator, value) => operator === '-' && value === 1;
+
+const getSourceCode = (context, node) => (
+	// Context.getSourceCode().text.slice(node.range[0], node.range[1])
+	context.getSourceCode().getText(node)
+);
+
+const report = (context, node, target, pattern) => {
+	const targetSource = getSourceCode(context, target);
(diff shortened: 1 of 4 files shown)


sindresorhus:
@nunofreitasbotelho Thanks for picking this up and your patience.

Can you add it to the readme too? https://github.com/sindresorhus/eslint-plugin-unicorn/commit/dd05f68f42b51c5941a59fb1509ef48bad658440#diff-0730bb7c2e8f9ea2438b52e419dd86c9

rules/prefer-includes.js
```diff
@@ -0,0 +1,114 @@
...
+};
+
+const isUnaryNotExpression = node => (
+	node.type === 'UnaryExpression' && node.operator === '!'
+);
+
+const isNegativeOne = (operator, value) => operator === '-' && value === 1;
+
+const getSourceCode = (context, node) => (
+	// Context.getSourceCode().text.slice(node.range[0], node.range[1])
```
sindresorhus: Remove commented out code.

rules/prefer-includes.js
```diff
@@ -0,0 +1,114 @@
...
+	context.getSourceCode().getText(node)
+);
+
+const report = (context, node, target, pattern) => {
+	const targetSource = getSourceCode(context, target);
+	const patternSource = getSourceCode(context, pattern);
+	context.report({
+		node,
+		message: 'Use .includes(), rather than .indexOf(), when checking for existence.',
```
sindresorhus: I don't think we need the commas here and `.includes()` and `.indexOf()` should be wrapped in backticks.

sindresorhus:
Ping @nunofreitasbotelho :)

rules/prefer-includes.js
```diff
@@ -0,0 +1,114 @@
...
+		fix: fixer => fixer.replaceText(node, `${targetSource}.includes(${patternSource})`)
```
futpib: If we want to support both "includes" and "does not include" cases, this fixer does not cut it. It does not look like it could possibly generate `!a.includes(b)` right now.

test/prefer-includes.js
```diff
@@ -0,0 +1,68 @@
...
+			output: `str.includes('foo')`,
+			errors
+		},
+		{
+			code: `'foobar'.indexOf('foo') >= 0`,
+			output: `'foobar'.includes('foo')`,
+			errors
+		},
+		{
+			code: `!str.indexOf('foo') === -1`,
```
futpib: This is equivalent too `(!str.indexOf('foo')) === -1`, instead you probably want tests for both of these:
* `str.indexOf('foo') !== -1` -> `str.includes('foo')`
* `!(str.indexOf('foo') === -1)` -> `str.includes('foo')`

Same goes for some of the tests below.

sindresorhus:
I'm gonna close this as it's stale. If anyone wants to see this merged, a PR finishing this the last things mentioned here would be lovely.

PR #123, opened 2017-11-16
Add prefer-spread rule - fixes #120

### LINKED ISSUE #120: Prefer the spread operator over `Array.from()`
## Fail

```js
const set = new Set([1, 2]);
Array.from(set).map(() => {});
```

## Pass

```js
const set = new Set([1, 2]);
[...set].map(() => {});
```

The only difference between them is that `Array.from()` also works on array-like objects, while spread requires something to implement the iterator protocol, but we don't care about that, as it's what you should do anyway.

Should be easy to make this auto-fixable.

There's already a built-in ESLint rule called [`prefer-spread`](https://eslint.org/docs/rules/prefer-spread), but it's only concerned about `.apply()`.

bdougherty:
With fixer. Doesn't check `TypedArray.from()`.


sindresorhus:
Thanks for your patience. This looks splendid. 🙌

PR #125, opened 2017-11-16
Update filename-case.md

andrewp-efundamental:


sindresorhus:
The option names are in camelCase for consistency with JS.

PR #126, opened 2017-11-16
Fix no-fn-reference-in-iterator to support static Promise methods

### LINKED ISSUE #124: `no-fn-reference-in-iterator` rule should not apply to `Promise` static methods
Bluebird adds [`Promise.map()`](http://bluebirdjs.com/docs/api/promise.map.html) which has a different semantic. Since Bluebird is a very popular module, we should add an exclusion for it. So any static method on `Promise` should be ignored.

SamVerschueren:
This PR fixes #124.

### DIFF
--- a/docs/rules/no-fn-reference-in-iterator.md
+++ b/docs/rules/no-fn-reference-in-iterator.md
@@ -165,3 +165,9 @@ const fn = x => x === 2;
 
 [1, 2, 3].map(x => m({foo: 'bar'})(x));
 ```
+
+```js
+const fn = x => x === 2;
+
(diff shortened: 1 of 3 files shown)


sindresorhus:
That was fast :)

SamVerschueren:
Hehe :tada: :)

eaviles:
What about [Promise.filter](http://bluebirdjs.com/docs/api/promise.filter.html)? Or the other Bluebird-specific Promise.x?

sindresorhus:
@eaviles As you can see in the code it ignores anything on the `Promise` object, including `.filter`.

PR #132, opened 2017-11-29
Fix custom error definition when exporting - fixes #127

### LINKED ISSUE #127: custom-error-definition doesn't report error if the Error class is exported with `module.exports`
This code report the error ``The `name` property should be set `FooError`.`` as expected:
```js
  constructor() {
    super('My awesome Foo Error');
    this.name = this.constructor.name;
  }
};
```

This code doesn't report any error:
```js
module.exports = class FooError extends Error {
  constructor() {
    super('My awesome Foo Error');
    this.name = this.constructor.name;
  }
};
```

With `export` it works as expected and the error is reported:
```js
export class FooError extends Error {
  constructor() {
    super('My awesome Foo Error');
    this.name = this.constructor.name;
  }
}
```

SamVerschueren:
Fixes #127. Add some extra tests as well regarding exported classes.

### DIFF
--- a/rules/custom-error-definition.js
+++ b/rules/custom-error-definition.js
@@ -42,85 +42,88 @@ const isAssignmentExpression = (node, name) => {
 	return lhs.property.name === name;
 };
 
+const customErrorDefinition = (context, node) => {
+	if (!hasValidSuperClass(node)) {
+		return;
+	}
+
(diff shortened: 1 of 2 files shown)


sindresorhus:
Purfect :)

Issue #133, opened 2017-12-11
no-process-exit throws on process.once (but not on process.on)


Izhaki:
From [the docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/48ef6e1de5d92f6538b8399f8f4bfe06a7c3ec8f/docs/rules/no-process-exit.md):

> It also allows process.exit() to be called in process.on('<event>', func) event handlers.

Any reason `process.once` is disallowed?

As in:
```javascript
process.once( 'SIGINT', () => {
    process.exit() // Throws Only use `process.exit()` in CLI apps. Throw an error instead. (unicorn/no-process-exit)
})

process.on( 'SIGINT', () => {
    process.exit() // Doesn't thorw
})
```

jamestalmage:
No good reason. If it's good in `on` it's good in `once`. 
PR Welcome.
Relevant line of code is here: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/no-process-exit.js#L17

Izhaki:
@jamestalmage Done.

PR #134, opened 2017-12-12
Fix: allow process.exit() to be called in process.once

Izhaki:
no-process-exit rule: Just like process.exit is allowed within process.on callbacks, allow it in process.once.

Resovles #133


Izhaki:
Forgot tests.

PR #135, opened 2017-12-12
Fix process.exit within process.once throwing no-process-exit - fixes #133

### LINKED ISSUE #133: no-process-exit throws on process.once (but not on process.on)

Izhaki:
Support process.exit process.once callback, just like with process.on.

### DIFF
--- a/rules/no-process-exit.js
+++ b/rules/no-process-exit.js
@@ -14,7 +14,7 @@ const create = context => {
 			const callee = node.callee;
 
 			if (callee.type === 'MemberExpression' && callee.object.name === 'process') {
-				if (callee.property.name === 'on') {
+				if (callee.property.name === 'on' || callee.property.name === 'once') {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks :)

PR #137, opened 2017-12-13
Add `caughtErrorsIgnorePattern` option to `catch-error-name` rule

### LINKED ISSUE #48: Allow multiple `name` for `catch-error-name`
Sometimes we have the following code:

``` js
const fs = require('fs');
try {
  fs.readFileSync('./foo';)
} catch (err) {
  // Nothing
}
```

And if enable rule `no-unused-vars` we have error. For ignoring this construction we can use http://eslint.org/docs/rules/no-unused-vars#caughterrorsignorepattern. But if i use `ignoreErr` i get error from `catch-error-name` rule.

evilebottnawi:
Fixes #48
I think better add option which error name we should ignore (as do eslint) instead allow multiple errors names

### DIFF
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -75,8 +75,40 @@ somePromise.catch(_ => {
 
 ## Options
 
+### Name
+
 You can set the `name` option like this:
 
 ```js
 "unicorn/catch-error-name": ["error", {"name": "err"}]
 ```
+
+### caughtErrorsIgnorePattern
+
+```js
+"unicorn/catch-error-name": ["error", {"caughtErrorsIgnorePattern": "^ignore"}]
+```
+
+The `caughtErrorsIgnorePattern` option specifies exceptions not to check for usage: catch arguments whose names match a regexp pattern. 
+For example, variables whose names begin with a string ‘ignore’.
+
+## Fail
+
+```js
+try {
+	doSomething();
+} catch (skipErr) {
+	// ...
+}
+```
+
+## Pass
+
+```js
+try {
+	doSomething();
+} catch (ignoreErr) {
+	// ...
+}
+```
+
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -86,11 +86,11 @@ You can set the `name` option like this:
 ### caughtErrorsIgnorePattern
 
 ```js
-"unicorn/catch-error-name": ["error", {"caughtErrorsIgnorePattern": "^ignore"}]
+"unicorn/catch-error-name": ["error", {"caughtErrorsIgnorePattern": "^_"}]
 ```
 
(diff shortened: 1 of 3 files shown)


evilebottnawi:
@sindresorhus friendly ping

docs/rules/catch-error-name.md
```diff
@@ -75,8 +75,40 @@ somePromise.catch(_ => {
...
 "unicorn/catch-error-name": ["error", {"name": "err"}]
 ```
+
+### caughtErrorsIgnorePattern
+
+```js
+"unicorn/catch-error-name": ["error", {"caughtErrorsIgnorePattern": "^ignore"}]
+```
+
+The `caughtErrorsIgnorePattern` option specifies exceptions not to check for usage: catch arguments whose names match a regexp pattern.
```
sindresorhus: This sentence could use some more work. It doesn't read very well.

rules/catch-error-name.js
```diff
@@ -38,6 +38,9 @@ function indexifyName(name, scope) {
 const create = context => {
 	const opts = context.options[0];
 	const name = (opts && opts.name) || 'err';
```
sindresorhus: Use Object.assign so you don't have to check whether options exist or not each time:

```js
const opts = Object.assign({}, {name: 'err'}, context,options[0]);
```

sindresorhus:
I think the `caughtErrorsIgnorePattern` option should default to `^_$`, so `catch (_) {}` is ignored by default (This should also be documented).

sindresorhus:
Thanks for your patience. I've just been busy in real life.

evilebottnawi:
@sindresorhus nothing terrible :star: , in the near future I will correct all comments

evilebottnawi:
@sindresorhus default value was got from https://eslint.org/docs/rules/no-unused-vars#caughterrorsignorepattern, we can change, but i think better is use as in eslint or we can add multiple default value `_*` and `ignore*`

docs/rules/catch-error-name.md
evilebottnawi: @sindresorhus also was got from https://eslint.org/docs/rules/no-unused-vars#caughterrorsignorepattern. It is difficult to imagine how to explain in more detail

sindresorhus:
> but i think better is use as in eslint or we can add multiple default value _* and ignore*

ESLint is generally very unopinionated, and I prefer good defaults. I meant it should be default, but the user could, of course, specify the option to get something else.

sindresorhus:
Looks like you had a bad rebase.

evilebottnawi:
@sindresorhus all done :+1:

sindresorhus:
The default should be `^_$`, not `^_`.

evilebottnawi:
@sindresorhus done :+1:

sindresorhus:
Thanks :)

PR #138, opened 2017-12-13
Fix regex literal inside RegExp constructor in the `regex-shorthand` rule

evilebottnawi:
Fix for syntax:
```js
new RegExp(/[0-9]/);
new RegExp(/[0-9]/, 'ig');
```
See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

### DIFF
--- a/rules/regex-shorthand.js
+++ b/rules/regex-shorthand.js
@@ -11,6 +11,11 @@ const create = context => {
 
 			const newPattern = cleanRegexp(oldPattern, flags);
 
+			// Handle regex literal inside RegExp constructor in the other handler
+			if (node.parent.type === 'NewExpression' && node.parent.callee.name === 'RegExp') {
+				return;
+			}
+
 			if (oldPattern !== newPattern) {
 				context.report({
 					node,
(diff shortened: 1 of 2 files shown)


rules/regex-shorthand.js
```diff
@@ -11,6 +11,10 @@ const create = context => {
 
 			const newPattern = cleanRegexp(oldPattern, flags);
 
+			if (node.parent.type === 'NewExpression' && node.parent.callee.name === 'RegExp') {
```
sindresorhus: Can you add a short comment about what this is for. Something like `// Handle regex literal inside RegExp constructor in the other handler`

rules/regex-shorthand.js
```diff
@@ -26,16 +30,26 @@ const create = context => {
 				return;
 			}
 
-			const oldPattern = args[0].value;
-			const flags = args[1] && args[1].type === 'Literal' ? args[1].value : '';
+			const hasRegExp = args[0].regex || false;
```
sindresorhus: The `false` here is moot.

evilebottnawi:
@sindresorhus done :+1:

sindresorhus:
Looks great! :)

PR #141, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can know where their documentation is without having to resort to external packages to guess.

### DIFF
--- a/rules/utils/get-docs-url.js
+++ b/rules/utils/get-docs-url.js
@@ -1,13 +1,11 @@
 'use strict';
+const {basename} = require('path');
 
 const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
 
-function getDocsUrl(ruleName, givenCommitHash) {
-	let commitHash = givenCommitHash;
-	if (!commitHash) {
-		commitHash = 'master';
-	}
-	return `${repoUrl}/blob/${commitHash}/docs/rules/${ruleName}.md`;
-}
+module.exports = function (ruleName, commitHash) {
+	ruleName = ruleName || basename(module.parent.filename).replace('.js', '');
+	commitHash = commitHash || 'master';
 
-module.exports = getDocsUrl;
+	return `${repoUrl}/blob/${commitHash}/docs/rules/${ruleName}.md`;
+};
--- a/rules/utils/get-docs-url.js
+++ b/rules/utils/get-docs-url.js
@@ -3,7 +3,7 @@ const path = require('path');
 
 const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
 
-module.exports = (ruleName) => {
+module.exports = ruleName => {
 	ruleName = ruleName || path.basename(module.parent.filename, '.js');
 	return `${repoUrl}/blob/master/docs/rules/${ruleName}.md`;
 };
--- a/rules/utils/get-docs-url.js
+++ b/rules/utils/get-docs-url.js
@@ -3,9 +3,7 @@ const path = require('path');
 
 const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
 
-module.exports = (ruleName, commitHash) => {
+module.exports = (ruleName) => {
 	ruleName = ruleName || path.basename(module.parent.filename, '.js');
-	commitHash = commitHash || 'master';
-
-	return `${repoUrl}/blob/${commitHash}/docs/rules/${ruleName}.md`;
+	return `${repoUrl}/blob/master/docs/rules/${ruleName}.md`;
 };
--- a/rules/utils/get-docs-url.js
+++ b/rules/utils/get-docs-url.js
@@ -1,10 +1,10 @@
 'use strict';
-const {basename} = require('path');
+const path = require('path');
 
 const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
 
-module.exports = function (ruleName, commitHash) {
-	ruleName = ruleName || basename(module.parent.filename).replace('.js', '');
+module.exports = (ruleName, commitHash) => {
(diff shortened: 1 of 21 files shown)


SamVerschueren:
I believe it would be a lot more maintainable if we created a utility function called `getDocsURL()` which uses `module.parent.filename` to dynamically calculate the last bit of the URL. All the other stuff is static.

Arcanemagus:
Moved to that model @SamVerschueren 😉.

rules/utils/get-docs-url.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
+
+function getDocsUrl(ruleName, givenCommitHash) {
```
SamVerschueren: Export this one directly.

rules/utils/get-docs-url.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
```
SamVerschueren: Move it to the top

rules/utils/get-docs-url.js
SamVerschueren: Rename `givenCommitMash` to `commitHash`. Below, write it like this

```js
commitHash = commitHash || 'master';
```

Arcanemagus: > Export this one directly.

The rest of this was written in CommonJS, so that's the style I was following. I'll switch it to a `module` 👍.

> Rename `givenCommitHash` to `commitHash`.

It's generally a bad idea to assign to parameters, but I can change it to that style.

SamVerschueren: I meant `module.exports = (ruleName, commit:ash) => {`

SamVerschueren: Strings are immutable. It's just the same as a default value, only one line extra. It's a bad idea when an object is passed in and you'd write `obj.commitHash = obj.commitHash || 'master'`

Arcanemagus:
Is that more along the lines of what you were thinking @SamVerschueren?

rules/utils/get-docs-url.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+const {basename} = require('path');
+
+const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
+
+module.exports = function (ruleName, commitHash) {
```
SamVerschueren: Use an arrow function

rules/utils/get-docs-url.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+const {basename} = require('path');
```
SamVerschueren: Use `const path = require('path')`

rules/utils/get-docs-url.js
```diff
@@ -0,0 +1,11 @@
...
+
+const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
+
+module.exports = function (ruleName, commitHash) {
+	ruleName = ruleName || basename(module.parent.filename).replace('.js', '');
```
SamVerschueren: Just pass `.js` as second argument to `path.basename()`. See https://nodejs.org/api/path.html#path_path_basename_path_ext

test/get-docs-url.js
```diff
@@ -0,0 +1,20 @@
+import test from 'ava';
+import getDocsUrl from '../rules/utils/get-docs-url';
+
+test('returns the URL of the a named rule\'s documentation', t => {
+	t.plan(1);
```
SamVerschueren: Remove

test/get-docs-url.js
```diff
@@ -0,0 +1,20 @@
...
+	const url = 'https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/foo.md';
+	t.is(getDocsUrl('foo'), url);
+});
+
+test('returns the URL of the a named rule\'s documentation at a commit hash', t => {
+	t.plan(1);
```
SamVerschueren: Remove

test/get-docs-url.js
```diff
@@ -0,0 +1,20 @@
...
+	const url = 'https://github.com/sindresorhus/eslint-plugin-unicorn/blob/bar/docs/rules/foo.md';
+	t.is(getDocsUrl('foo', 'bar'), url);
+});
+
+test('determines the rule name from the file', t => {
+	t.plan(1);
```
SamVerschueren: Remove

Arcanemagus:
Updated again @SamVerschueren 😉.

rules/utils/get-docs-url.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+const path = require('path');
+
+const repoUrl = 'https://github.com/sindresorhus/eslint-plugin-unicorn';
+
+module.exports = (ruleName, commitHash) => {
```
SamVerschueren: Not sure about the commit hash as we don't use it currently.

Arcanemagus: It's used in a few other plugins that I sent PRs to, so I figured I would leave it in here. I can remove it if you want though.

SamVerschueren: I think we should remove it as long as we don't need it. It adds "complexity" and should be maintained later on if perhaps the function gets more complex. But that's up to Sindre to decide.

sindresorhus:
Thanks for adding this @Arcanemagus :)

PR #145, opened 2018-01-18
Use link to tagged version for rule docs

sudo-suhas:
This uses the version present in package.json for generating the URL to rule documentation. I think this is good because the documentation will match the version the user has installed. This way, even if a rule is changed or removed in the future, the user can find the documentation with ease.


sudo-suhas:
Thanks @Arcanemagus 😁

PR #146, opened 2018-01-20
Add unsafe-regex rule

bdougherty:
For sindresorhus/xo#227

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,7 +33,8 @@ module.exports = {
 				'unicorn/new-for-builtins': 'error',
 				'unicorn/regex-shorthand': 'error',
 				'unicorn/prefer-spread': 'error',
-				'unicorn/error-message': 'error'
+				'unicorn/error-message': 'error',
+				'unicorn/no-unsafe-regex': 'error'
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


docs/rules/unsafe-regex.md
```diff
@@ -0,0 +1,27 @@
+# Detect unsafe regular expressions
```
sindresorhus: Let's use the word `disallow` instead of `detect`. In the other places too.

index.js
```diff
@@ -33,7 +33,8 @@ module.exports = {
 				'unicorn/new-for-builtins': 'error',
 				'unicorn/regex-shorthand': 'error',
 				'unicorn/prefer-spread': 'error',
-				'unicorn/error-message': 'error'
+				'unicorn/error-message': 'error',
+				'unicorn/unsafe-regex': 'error'
```
sindresorhus: `unsafe-regex` sounds like we enforce unsafe regexes. I think it should be `no-unsafe-regex`.

rules/unsafe-regex.js
```diff
@@ -0,0 +1,60 @@
...
+	};
+};
+
+module.exports = {
+	create,
+	meta: {
+		docs: {
+			url: getDocsUrl()
+		},
+		fixable: 'code'
```
sindresorhus: It's not fixable, so this should be removed.

bdougherty:
Should be good now!

sindresorhus:
Looks great! Thank you :)

PR #161, opened 2018-02-16
Do not use module.parent for resolving rule name

sudo-suhas:
_Note: This is very similar to the PR avajs/eslint-plugin-ava#188, but including the same description just to be through._

`module.parent` is cached after the first `require` and this results in nearly all rules having the URL to `assertion-arguments`. Here's a simple repro:

```js
'use strict';

const { rules } = require('eslint-plugin-unicorn');

const wrongLinks = [];

for (const ruleName of Object.keys(rules)) {
    const rule = rules[ruleName];
    const { url } = rule.meta.docs;
    if (!url.includes(ruleName)) wrongLinks.push(ruleName);
}

console.log(wrongLinks);
```

```
λ node rule-docs-url.js
[ 'custom-error-definition',
  'error-message',
  'escape-case',
  'explicit-length-check',
  'filename-case',
  'import-index',
  'new-for-builtins',
  'no-abusive-eslint-disable',
  'no-array-instanceof',
  'no-fn-reference-in-iterator',
  'no-hex-escape',
  'no-new-buffer',
  'no-process-exit',
  'no-unsafe-regex',
  'number-literal-case',
  'prefer-add-event-listener',
  'prefer-spread',
  'prefer-starts-ends-with',
  'prefer-type-error',
  'regex-shorthand',
  'throw-new-error' ]
```

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -115,7 +115,7 @@ module.exports = {
 	create,
 	meta: {
 		docs: {
-			url: getDocsUrl()
+			url: getDocsUrl(__filename)
 		},
 		schema
 	}
--- a/rules/custom-error-definition.js
+++ b/rules/custom-error-definition.js
@@ -132,7 +132,7 @@ module.exports = {
 	create,
 	meta: {
 		docs: {
-			url: getDocsUrl()
+			url: getDocsUrl(__filename)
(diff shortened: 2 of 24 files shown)


sindresorhus:
Thank again :)

PR #165, opened 2018-03-14
Do not auto-fix prefer-spread in certain cases - fixes #164

### LINKED ISSUE #164: Don't auto-fix `Array.from({ length: 12 })`
Before:

```js
Array.from({ length: 123 })
```

After autofix:

```js
[...{ length: 123 }]
```

But result is wrong:

image

SamVerschueren:
This PR fixes #164. Currently, I only check in the `isArrayLike` function if the argument is not of type `ObjectExpression`. I believe that's the best we can do for now because auto-fixing `Array.from(foo);` might be wrong as well if `foo` is an object for instance. As long as we don't have types in JS, we can't do this 100% for sure.

### DIFF
--- a/rules/prefer-spread.js
+++ b/rules/prefer-spread.js
@@ -12,6 +12,8 @@ const isArrayFrom = node => {
 	);
 };
 
+const isArrayLike = arg => arg && arg.type !== 'ObjectExpression';
+
 const parseArgument = (context, arg) => {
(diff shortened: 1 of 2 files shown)


sindresorhus:
This is what I was thinking too. Oh, so I wish we had native types in JS and could query them here.

PR #172, opened 2018-04-19
Add blacklist to `no-fn-reference-in-iterator` rule to prevent false-positives

### LINKED ISSUE #121: False-positives in `no-fn-reference-in-iterator` rule
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/121)

the `unicorn/no-fn-reference-in-iterator` matched all of these pieces of (valid) code and the fix applied broke every one. not sure the best way forward for this..

https://reactjs.org/docs/react-api.html#reactchildrenforeach
```
React.Children.forEach(children, (p) => {
})
```

https://github.com/airbnb/enzyme/blob/master/docs/api/ReactWrapper/find.md
```
const wrapper = shallow(<Foo />)
wrapper.find(Panel).first().childAt(1)
```

https://caolan.github.io/async/docs.html#filter
```
    async.filter(
      Object.keys(manifest),
      (f, callback) => {
        callback(null, f.match(/\.js$/))
      },
      (err, results) => {
        async.each(results, (f, callback) => {
```

https://lodash.com/docs/4.17.4#forEach
```
      _.forEach(api, (op, m) => {
```

https://github.com/react-bootstrap/react-bootstrap/blob/master/src/utils/ValidComponentChildren.js

```
import ValidComponentChildren from 'react-bootstrap/lib/utils/ValidComponentChildren'
ValidComponentChildren.map(children, child => (
```

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/121).

SamVerschueren:
Trying to fix #121. Currently only implemented a blacklist for `Async`, `async`, `_`, `React.children` and `Promise`. It wasn't entirely clear what was meant by 

> We should always ignore methods with more than 1 argument.

Does this mean that if we pass in more than 1 argument to `map`, it shouldn't do anything? Because this might be an issue for `reduce` and `reduceRight`.

### DIFF
--- a/rules/no-fn-reference-in-iterator.js
+++ b/rules/no-fn-reference-in-iterator.js
@@ -46,7 +46,7 @@ const toSelector = name => {
 };
 
 // Select all the call expressions except the ones present in the blacklist
-const selector = `CallExpression${calleeBlacklist.map(toSelector).join('')}`
+const selector = `CallExpression${calleeBlacklist.map(toSelector).join('')}`;
 
 const create = context => ({
 	[selector]: node => {
--- a/rules/no-fn-reference-in-iterator.js
+++ b/rules/no-fn-reference-in-iterator.js
@@ -41,7 +41,6 @@ const fix = (context, node) => {
 
 const toSelector = name => {
 	const splitted = name.split('.');
-
 	return `[callee.${'object.'.repeat(splitted.length)}name!="${splitted.shift()}"]`;
 };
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Does this mean that if we pass in more than 1 argument to map, it shouldn't do anything? Because this might be an issue for reduce and reduceRight.

Yes, but for `reduce` and `reduceRight` we can make an exception and allow up to two arguments.

sindresorhus:
Can you also turn the rule on again in https://github.com/sindresorhus/eslint-plugin-unicorn/blob/352ba4a0291f9210ca5c8e2e61c7e3ad14028e77/index.js#L44 and the readme?

PR #179, opened 2018-07-24
Add conditional expressions in explicit-length-check rule

christophehurpeau:
Add ConditionalExpression in the rule `explicit-length-check`, allowing to check this kind of code : 

```js
const foo = [].length === 0 ? null : undefined;
```

Before, only IfStatements where checked.

It does create a breaking change, do you want to add an option for it ?


sindresorhus:
> It does create a breaking change, do you want to add an option for it ?

No, I'll just do a major release. Most things regarding linting is a breaking change.

PR #185, opened 2018-08-19
Add prefer-exponentiation-operator rule

### LINKED ISSUE #173: Rule proposal: `prefer-exponentiation-operator`
Prefer `7 ** 2` over `Math.pow(7, 2)`.

http://exploringjs.com/es2016-es2017/ch_exponentiation-operator.html

---

Should come with an auto-fix.

MrHen:
Had two questions:

* Should the rule be added to the recommended configuration as `off`?
* Should new rules always be added to the bottom of the rule list in `readme.md`?

Fixes #173.

### DIFF
--- a/rules/prefer-exponentiation-operator.js
+++ b/rules/prefer-exponentiation-operator.js
@@ -1,0 +1,52 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const isMathPow = node => {
+	const {callee} = node;
+	return (
+		callee.type === 'MemberExpression' &&
+		callee.object.type === 'Identifier' &&
+		callee.object.name === 'Math' &&
+		callee.property.type === 'Identifier' &&
+		callee.property.name === 'pow'
+	);
+};
+
+const parseArgument = (context, arg) => {
+	if (arg.type === 'Identifier') {
+		return arg.name;
+	}
+
+	return context.getSourceCode().getText(arg);
+};
+
+const create = context => {
+	return {
+		CallExpression(node) {
+			if (isMathPow(node)) {
+				context.report({
+					node,
+					message: 'Prefer the exponentiation operator over `Math.pow()`.',
+					fix: fixer => {
+						const base = parseArgument(context, node.arguments[0]);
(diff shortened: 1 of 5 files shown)


test/prefer-exponentiation-operator.js
```diff
@@ -0,0 +1,51 @@
...
+
+ruleTester.run('prefer-exponentiation-operator', rule, {
+	valid: [
+		'a ** b;',
+		'2 ** 4;',
+	],
+	invalid: [
+		{
+			code: 'const x = Math.pow(2, 4);',
+			errors: [{message: 'Prefer the exponentiation operator over `Math.pow()`.', column: 11, line: 1}],
```
sindresorhus: Create a method like [this](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/2283cda9313deda0c527ce3af8b7607ad512540f/test/new-for-builtins.js#L11-L14) so you don't have to repeat the error for invalid item. I don't think there's any point in specifying column/line except for the test where there are multiple `Math.pow()` in one test.

rules/prefer-exponentiation-operator.js
```diff
@@ -0,0 +1,52 @@
...
+};
+
+const create = context => {
+	return {
+		CallExpression(node) {
+			if (isMathPow(node)) {
+				context.report({
+					node,
+					message: 'Prefer the exponentiation operator over `Math.pow()`.',
+					fix: fixer => {
```
sindresorhus: I would move the fixer method to a top-level function too, for readability.

test/prefer-exponentiation-operator.js
MrHen: Since there aren't any parameters, I just made it a constant.

rules/prefer-exponentiation-operator.js
MrHen: Since you need `fixer`, `context` and `node` to run the fix, I'm not sure moving it out really helps much. I gave it a shot; let me know what you think.

sindresorhus:
Perfect. Thanks for implementing this :)

PR #187, opened 2018-09-02
catch-error-name - Support optional catch binding

### LINKED ISSUE #186: [Bug] `catch-error-name` rule does not work with optional catch binding
The following code using the latest ES [optional catch binding](https://github.com/tc39/proposal-optional-catch-binding) (supported by latest Acorn, Espree and ESLint):

```
try {
  throw new Error('message');
} catch {
  console.log('failed');
}
```

fails with the `catch-error-name` rule:
```
Cannot read property 'name' of null
TypeError: Cannot read property 'name' of null
    at CatchClause (/node_modules/eslint-plugin-unicorn/rules/catch-error-name.js:88:19)
    at listeners.(anonymous function).forEach.listener (/node_modules/eslint/lib/util/safe-emitter.js:45:58)
    at Object.emit (/node_modules/eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (/node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:632:23)
    at nodeQueue.forEach.traversalInfo (/node_modules/eslint/lib/linter.js:860:28)
```

MrHen:
Fixes #186. The test runner choked on the new syntax and I couldn't figure out exactly what the problem was. My hunch is that either ava or the eslint test runner doesn't support optional catch bindings.

To test the fix, I used `npm link` and ran the scenario from the bug. Pushing `true` was the cleanest way to leave the rest of the code alone.

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -85,6 +85,11 @@ const create = context => {
 			}
 		},
 		CatchClause: node => {
+			if (!node || !node.param) {
+				push(true);
+				return;
+			}
+
 			if (node.param.name === '_') {
 				push(!astUtils.someContainIdentifier('_', node.body.body));
 				return;
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -85,6 +85,7 @@ const create = context => {
 			}
 		},
 		CatchClause: node => {
+			// Optional catch binding
 			if (!node || !node.param) {
 				push(true);
 				return;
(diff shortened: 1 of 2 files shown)


sindresorhus:
> My hunch is that either ava or the eslint test runner doesn't support optional catch bindings.

It's AVA. Fixed in the latest version, but I can't upgrade yet because of https://github.com/jfmengels/eslint-ava-rule-tester/issues/7

PR #191, opened 2018-09-03
Add `no-console-spaces` rule

### LINKED ISSUE #175: Add rule for no-space-in-console rule
see https://github.com/wulkano/kap/pull/450#discussion_r188170111

`console.log` (and .warn etc), all add spaces between the messages: 
```js
console.log('a', 'b', 'c') => `a b c`
```

This should be considered invalid:
```js
console.log('a ', 'b ', 'c') => `a  b  c`
```

MrHen:
A few questions:

1. I chose to use [`.trim()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) to apply the fix. This removes _all_ whitespace -- including `\t` and `\n`. In theory we could expose a setting to chose what values to remove but then we'd probably want to upgrade to something like [lodash.trim](https://lodash.com/docs/4.17.5#trim).

2. Similarly, `.trim()` also removes preceding whitespace. There could be a setting to switch to [`trimEnd`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimEnd) instead?

3. This implementation explicitly ignores [template literals](https://github.com/estree/estree/blob/master/es2015.md#template-literals). It wouldn't be a big deal to add them but the previous decision to remove `\n` makes some of the fixes awkward. Maybe we want to include template literals in the fix but ignore multiline literals?

4. I assumed this rule should be enabled in the recommended config.

5. Any other documentation examples and / or tests I should include?

6. Thoughts on the rule name? I personally prefer `no-console-spaces` or `no-console-whitespace` because I don't like prepositions in names when they can be avoided.

Fixes #175.


sindresorhus:
> The other question is if we should worry about someone shadowing console. The eslint no-console rule has an example of how to survive shadowing. (Maybe work on that as a follow up?)

Nah. That's an edge-case. Would be nice to solve eventually, but can be a follow-up.

sindresorhus:
Can you rebase from master? I just added back linting (It was previously removed because of problems with ESLint 4). And then run `npx xo --fix` and fix any lint issues.

rules/no-console-spaces.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const getConsoleMethod = (node) => {
+	const methods = [
+		'log',
+		'warn',
+		'error',
```
sindresorhus: There are many more log methods: https://developers.google.com/web/tools/chrome-devtools/console/console-reference We only need to support the ones that accept multiple arguments.

sindresorhus:
You might (or may not) find these useful:

- https://mysticatea.github.io/eslint-utils/guide/getting-started.html
- https://github.com/jfmengels/eslint-ast-utils

index.js
```diff
@@ -39,7 +39,8 @@ module.exports = {
 				'unicorn/error-message': 'error',
 				'unicorn/no-unsafe-regex': 'off',
 				'unicorn/prefer-add-event-listener': 'error',
-				'unicorn/prefer-exponentiation-operator': 'error'
+				'unicorn/prefer-exponentiation-operator': 'error',
+				'unicorn/no-console-spaces': 'error',
```
sindresorhus: You also need to add this to the readme, both in the example config and rule list.

rules/no-console-spaces.js
sindresorhus: There are many more log methods: https://developers.google.com/web/tools/chrome-devtools/console/console-reference We only need to support the ones that accept multiple arguments.

rules/no-console-spaces.js
```diff
@@ -0,0 +1,84 @@
...
+const create = (context) => {
+	return {
+		CallExpression(node) {
+			const method = getConsoleMethod(node);
+			if (!method) {
+				return;
+			}
+
+			const args = getTrimmableArguments(node);
+			args.forEach((arg) => {
```
sindresorhus: Use a `for-of` loop.

test/no-console-spaces.js
```diff
@@ -0,0 +1,127 @@
...
+		},
+		{
+			code: `
+				console.log(
+					'abc',
+					'def '
+				);
+			`,
+			errors: [
+				buildError({ method: 'log', column: 6, line: 4}),
```
sindresorhus: No need to specify column/line for one-line fixtures.

rules/no-console-spaces.js
MrHen: I added `debug` and `info`. The only other multi-parameter is `group` which isn't really the same. I can add that as well if you want.

MrHen:
@sindresorhus Updated PR.

* Will now only fail / fix exactly one space on either side of the string
  * Of note, I decided to leave `console.log(' ');` alone _mostly_ because I was lazy and didn't want to an explicit check for it. If you have strong feelings I can add a check -- but in a certain sense I'm not sure we really want to replace `console.log(' ');` with `console.log('');`.
* Added support for template literals
* Rebased and fixed lint errors
* Added missing documentation
* Added missing console methods [(see comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/191#discussion_r214762400)

Let me know if you want any other changes.

MrHen:
@sindresorhus gentle bump in case the earlier notification got lost in the wind.

docs/rules/no-console-spaces.md
```diff
@@ -0,0 +1,38 @@
+# Do not include spaces in `console.log` parameters.
```
sindresorhus: This is too generic. Should be something more like:

> Do not include leading/trailing space in `console.log` parameters.

rules/no-console-spaces.js
```diff
@@ -0,0 +1,108 @@
...
+
+	return value;
+};
+
+const fixValue = value => {
+	if (!value) {
+		return value;
+	}
+
+	// Find exactly one leading or tailing space
```
sindresorhus: typo, trailing

rules/no-console-spaces.js
```diff
@@ -0,0 +1,108 @@
...
+	return value.replace(/^ ?((?! ).*?[^ ]) ?$/, '$1');
```
sindresorhus: Seems like using `.startsWith()`/`.endsWith()` and then `.slice(1)` or `.slice(-1)` would be simpler and more readable.

sindresorhus:
Sorry about the delay. This got lost in a tab somewhere.

sindresorhus:
I don't think these should be failures:

- `console.log("abc ");`
- `console.log(" abc");`
- `console.log(" abc ");`

They are not really mistake. The mistake is not realizing **multiple** arguments are joined with a space.

Meaning, these should not be failures either:

- `console.log(" abc", "abc"));`
- `console.log("abc", "abc "));`
- `console.log(" abc", "abc "));`

MrHen:
@sindresorhus No worries. I made the changes requested but travis is failing due to the following:

```
  Error thrown in test:
  TypeError {
    message: 'process.stdout.getColorDepth is not a function',
  }
```

I'm not entirely sure how that is relevant to my changes but I can dig into it later today.

sindresorhus:
This is looking great now. Thanks for your patience. I think we ended up with something really good :)

PR #198, opened 2018-10-04
Add `prefer-query-selector` rule

### LINKED ISSUE #171: Rule proposal: `prefer-query-selector`
Prefer `querySelector` over `getElementById`, `querySelectorAll` over `getElementsByClassName` and `getElementsByTagName`. They are not faster than `querySelector` and it's better to be consistent.

## Fail

```js
document.getElementById('foo');
```

## Pass

```js
document.querySelector('#foo');
```

janowsiany:
fixes #171

### DIFF
--- a/docs/rules/prefer-query-selector.md
+++ b/docs/rules/prefer-query-selector.md
@@ -6,12 +6,20 @@ They are not faster than `querySelector` and it's better to be consistent.
 ## Fail
 
 ```js
-document.getElementById('foo');
+document.getElementById("foo");
+document.getElementsByClassName('foo bar');
+document.getElementsByTagName(`main`);
+document.getElementsByClassName(fn());
 ```
 
 
 ## Pass
 
 ```js
-document.querySelector('#foo');
+document.querySelector("#foo");
+document.querySelector('.bar');
+document.querySelector(`main #foo .bar`);
+document.querySelectorAll(".foo .bar");
+document.querySelectorAll('li a');
+document.querySelector(`li`).querySelectorAll("a");
 ```
--- a/docs/rules/prefer-query-selector.md
+++ b/docs/rules/prefer-query-selector.md
@@ -1,6 +1,6 @@
 # Prefer `querySelector` over `getElementById`, `querySelectorAll` over `getElementsByClassName` and `getElementsByTagName`
 
-They are not faster than querySelector and it's better to be consistent.
+They are not faster than `querySelector` and it's better to be consistent.
 
 
 ## Fail
--- a/docs/rules/prefer-query-selector.md
+++ b/docs/rules/prefer-query-selector.md
@@ -1,0 +1,17 @@
+# Prefer `querySelector` over `getElementById`, `querySelectorAll` over `getElementsByClassName` and `getElementsByTagName`
+
+They are not faster than querySelector and it's better to be consistent.
+
+
+## Fail
+
+```js
+document.getElementById('foo');
+```
+
+
+## Pass
+
+```js
+document.querySelector('#foo');
+```
--- a/docs/rules/prefer-query-selector.md
+++ b/docs/rules/prefer-query-selector.md
@@ -6,9 +6,9 @@ They are not faster than `querySelector` and it's better to be consistent.
 ## Fail
 
 ```js
-document.getElementById("foo");
+document.getElementById('foo');
 document.getElementsByClassName('foo bar');
-document.getElementsByTagName(`main`);
+document.getElementsByTagName('main');
 document.getElementsByClassName(fn());
 ```
 
@@ -16,10 +16,10 @@ document.getElementsByClassName(fn());
 ## Pass
 
 ```js
-document.querySelector("#foo");
+document.querySelector('#foo');
 document.querySelector('.bar');
-document.querySelector(`main #foo .bar`);
-document.querySelectorAll(".foo .bar");
+document.querySelector('main #foo .bar');
+document.querySelectorAll('.foo .bar');
 document.querySelectorAll('li a');
-document.querySelector(`li`).querySelectorAll("a");
+document.querySelector('li').querySelectorAll('a');
 ```
--- a/index.js
+++ b/index.js
@@ -35,6 +35,7 @@ module.exports = {
 				'unicorn/import-index': 'error',
 				'unicorn/new-for-builtins': 'error',
 				'unicorn/regex-shorthand': 'error',
+				'unicorn/prefer-query-selector': 'error',
 				'unicorn/prefer-spread': 'error',
 				'unicorn/error-message': 'error',
 				'unicorn/no-unsafe-regex': 'off',
--- a/index.js
+++ b/index.js
@@ -40,7 +40,8 @@ module.exports = {
 				'unicorn/error-message': 'error',
 				'unicorn/no-unsafe-regex': 'off',
 				'unicorn/prefer-add-event-listener': 'error',
-				'unicorn/prefer-exponentiation-operator': 'error'
+				'unicorn/prefer-exponentiation-operator': 'error',
+				'unicorn/no-console-spaces': 'error'
 			}
 		}
 	}
--- a/index.js
+++ b/index.js
@@ -10,16 +10,26 @@ module.exports = {
 				es6: true
 			},
 			parserOptions: {
-				ecmaVersion: 2018,
+				ecmaVersion: 2019,
 				sourceType: 'module'
 			},
 			plugins: [
 				'unicorn'
 			],
 			rules: {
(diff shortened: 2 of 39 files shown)


docs/rules/prefer-query-selector.md
```diff
@@ -0,0 +1,17 @@
...
+```js
+document.getElementById('foo');
+```
+
+
+## Pass
+
+```js
+document.querySelector('#foo');
+```
```
sindresorhus: Include some more examples in `Fail` and `Pass`.

index.js
```diff
@@ -35,6 +35,7 @@ module.exports = {
 				'unicorn/import-index': 'error',
 				'unicorn/new-for-builtins': 'error',
 				'unicorn/regex-shorthand': 'error',
+				'unicorn/prefer-query-selector': 'error',
```
sindresorhus: Should be added to the end. You also need to add this to https://github.com/sindresorhus/eslint-plugin-unicorn#usage

rules/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+	'getElementById',
+	'querySelectorAll',
+	'getElementsByClassName',
+	'getElementsByTagName'
+];
+
+const VALID_QUOTES = /([',",`])/;
+const getRange = (prop, node) => [prop.start, node.arguments[0].end];
+const getReplacement = (ctx, identifierName, node) => {
+	const argAsTxt = ctx.getSourceCode().getText(node.arguments[0]);
```
sindresorhus: It would be better to operate on the AST nodes instead of the text. Then you don't have to deal with possible quotes, escapes, etc, it gets complicated quickly. Looks at how some of the other rules are implemented, for inspiration.

rules/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+const forbiddenIdentifierNames = [
+	'getElementById',
+	'querySelectorAll',
+	'getElementsByClassName',
+	'getElementsByTagName'
+];
+
+const VALID_QUOTES = /([',",`])/;
+const getRange = (prop, node) => [prop.start, node.arguments[0].end];
+const getReplacement = (ctx, identifierName, node) => {
```
sindresorhus: `ctx` => `context`

Generally, use readable names instead of abbreviations.

test/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+	invalid: [
+		{
+			code: 'document.getElementById("foo");',
+			errors: [{message: 'Prefer `querySelector` over `getElementById`.'}],
+			output: 'document.querySelector("#foo");'
+		},
+		{
+			code: 'document.querySelectorAll("p");',
+			errors: [{message: 'Prefer `querySelector` over `querySelectorAll`.'}],
+			output: 'document.querySelector("p");'
```
sindresorhus: This is incorrect. It should not prefer `querySelector` over `querySelectorAll`. Both are valid.

test/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+		},
+		{
+			code: 'document.getElementsByClassName("foo");',
+			errors: [{message: 'Prefer `querySelector` over `getElementsByClassName`.'}],
+			output: 'document.querySelector(".foo");'
+		},
+		{
+			code: 'document.getElementsByClassName("foo bar");',
+			errors: [{message: 'Prefer `querySelector` over `getElementsByClassName`.'}],
```
sindresorhus: No need to repeat the error in every assertion. Just build them upfront, like done in https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/new-for-builtins.js

test/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+			output: 'document.querySelector(".foo.bar");'
+		},
+		{
+			code: 'document.getElementsByTagName("div");',
+			errors: [{message: 'Prefer `querySelector` over `getElementsByTagName`.'}],
```
sindresorhus: We should suggest `querySelectorAll` for `getElementsByTagName` and `getElementsByClassName`, since they can return multiple elements, and `querySelector` for `getElementById`.

test/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+		{
+			code: 'document.getElementById(\'foo\');',
+			errors: [{message: 'Prefer `querySelector` over `getElementById`.'}],
+			output: 'document.querySelector(\'#foo\');'
+		},
+		{
+			code: 'const qux = \'qux\'; document.getElementsByClassName(`foo ${qux}`);', // eslint-disable-line no-template-curly-in-string
+			errors: [{message: 'Prefer `querySelector` over `getElementsByClassName`.'}],
+			output: 'const qux = \'qux\'; document.querySelector(`.foo.${qux}`);' // eslint-disable-line no-template-curly-in-string
+		}
```
sindresorhus: There needs to be some more tests:

- When a variable is used: `getElementsByClassName(foo)`.
- When a variable is partly used: `getElementsByClassName(foo + 'bar')`

sindresorhus:
The rule needs to be added to the bottom of the list here: https://github.com/sindresorhus/eslint-plugin-unicorn#rules

sindresorhus:
@futpib @MrHen Would you be able to help review? :)

test/prefer-query-selector.js
futpib: We should allow usage like `getElementById(id)` (when the argument is a variable), as to not to encourage writing `querySelector('#' + id)`.

In other words: 
`getElementById(id)`
is better then
`querySelector('#' + CSS.escape(id))`
is better then
`querySelector('#' + id)`

Same goes for `getElementsByClassName` and `getElementsByTagName`.

sindresorhus: > We should allow usage like getElementById(id) (when the argument is not a string literal), as to not to encourage writing querySelector('#' + id).

I disagree. It's better to be consistent no matter what. It's an edge-case anyway, and `` querySelector(`#${id}`) `` is totally fine.

Why would you need to escape it with `querySelector` and not with `getElementById`?

futpib: Why escape:
```js
const id = 'shoot foot';
const div = document.createElement('div');
div.id = id;
document.append(div);
document.querySelector(`#${CSS.escape(id)}`); // -> HTMLDivElement
document.querySelector(`#${id}`); // -> null
document.getElementById(id) // -> HTMLDivElement
```

sindresorhus: I have done web dev for 10+ years and have yet to see an ID or class name with spaces. I honestly don't think we should care about such an obscure edge-case.

rules/prefer-query-selector.js
janowsiany: Yes i know that operating on text is not the best. I wanted to do this by returning an array of fixes one for changing `MemberExpression` and the other for changing arguments `Literal`. However, although the documentation says that it is possible to return an array of fixes i could not make it work.

sindresorhus: Can you push those changes somewhere? Hard to tell what's wrong otherwise.

It should look like something like this:

```js
context.report({
    node: node,
    message: "Missing semicolon",
    fix: function(fixer) {
        return [
        	fixer.insertTextAfter(node, ";"),
        	fixer.insertTextAfter(node, ";")
        ]
    }
});
```

It should be an array of only `fixing` objects, which are returned by the `fixer` methods, like `fixer.insertTextAfter`.

rules/prefer-query-selector.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const create = context => {
+	return {
+		CallExpression(node) {
+			const {callee} = node;
+			const prop = callee.property;
+
+			const identifierName = prop.name;
+			const preferedSelector = forbiddenIdentifierNames[identifierName];
```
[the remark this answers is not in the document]
sindresorhus: A `Map` would be the correct data structure for this.

rules/prefer-query-selector.js
janowsiany: Ok, @sindresorhus sorry if that takes too long but was investing this case with fixer returning an array. I believe it is because of our wrapper `eslint-ava-rule-tester` which has a dependency on old eslint package. I already made a PR and checked it locally. @jfmengels if you could review it here: https://github.com/jfmengels/eslint-ava-rule-tester/pull/11

sindresorhus:
@janowsiany Ping :)

janowsiany:
@sindresorhus sorry for such long delay, please revalidate

sindresorhus:
This looks good to me now. Very nice work, @janowsiany 👍

PR #199, opened 2018-10-05
Add `no-unreadable-array-destructuring` rule

### LINKED ISSUE #197: Rule proposal: No unreadable array destructuring
Destructuring is very useful, but it can also make some code harder to read.

For example: https://github.com/sindresorhus/refined-github/pull/1558/files#diff-7fa8bf1c42c5640497119c24113948a4R56

## Fail

```js
const [,, foo] = parts;
const [,,, foo] = parts;
const [,,,, foo] = parts;
// ...and so on...
```

## Pass

```js
const [, foo] = parts;
const [foo] = parts;
const foo = parts[3];
```

janowsiany:
At the moment the rule disallows more than one ignored value in series. However I can tune this, I can allow to ignore at most one value at the beginning if you find current behaviour too mild. 
Fixes #197

### DIFF
--- a/docs/rules/no-unreadable-array-destructuring.md
+++ b/docs/rules/no-unreadable-array-destructuring.md
@@ -1,6 +1,7 @@
 # Disallow unreadable array destructuring
 
-Destructuring is very useful, but it can also make some code harder to read. This limits the abusive ignoring consecutive values when destructuring from array.
+Destructuring is very useful, but it can also make some code harder to read. This rule prevents ignoring consecutive values when destructuring from an array.
+
 
 ## Fail
 
@@ -11,6 +12,7 @@ const [,,,, foo] = parts;
 const [,,...rest] = parts;
 ```
 
+
 ## Pass
 
 ```js
@@ -20,14 +22,21 @@ const foo = parts[3];
 const [,...rest] = parts;
 ```
 
-## Worth to know
 
-You might have to modify the built-in [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring) rule to be compatible with this one. In that case you can do it like this:
+## Note
 
-```js
+You might have to modify the built-in [`prefer-destructuring`](https://eslint.org/docs/rules/prefer-destructuring) rule to be compatible with this one:
+
+```json
 {
-  "rules": {
-    "prefer-destructuring": ["error", {"object": true, "array": false}]
-  }
+	"rules": {
+		"prefer-destructuring": [
+			"error",
+			{
+				"object": true,
+				"array": false
+			}
+		]
+	}
 }
 ```
--- a/docs/rules/no-unreadable-array-destructuring.md
+++ b/docs/rules/no-unreadable-array-destructuring.md
@@ -1,6 +1,6 @@
 # Disallow unreadable array destructuring
 
-Destructuring is very useful, but it can also make some code harder to read.
+Destructuring is very useful, but it can also make some code harder to read. This limits the abusive ignoring consecutive values when destructuring from array.
 
 ## Fail
 
@@ -8,8 +8,8 @@ Destructuring is very useful, but it can also make some code harder to read.
 const [,, foo] = parts;
 const [,,, foo] = parts;
 const [,,,, foo] = parts;
+const [,,...rest] = parts;
 ```
-
 
 ## Pass
 
@@ -17,4 +17,17 @@ const [,,,, foo] = parts;
 const [, foo] = parts;
 const [foo] = parts;
 const foo = parts[3];
+const [,...rest] = parts;
+```
+
+## Worth to know
+
+You might have to modify the built-in [prefer-destructuring](https://eslint.org/docs/rules/prefer-destructuring) rule to be compatible with this one. In that case you can do it like this:
+
+```js
+{
+  "rules": {
+    "prefer-destructuring": ["error", {"object": true, "array": false}]
+  }
+}
 ```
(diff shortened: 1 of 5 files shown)


rules/no-unreadable-array-destructuring.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const isCommaFollwedWithComma = (el, index, array) => {
```
jfmengels: There is a typo here:
isCommaFollwedWithComma --> isCommaFollowedWithComma

rules/no-unreadable-array-destructuring.js
```diff
@@ -0,0 +1,35 @@
...
+	if (el === null) {
+		return array[index + 1] === el;
```
jfmengels: I think it would be more explicit to change the condition to `array[index + 1] === null`

rules/no-unreadable-array-destructuring.js
```diff
@@ -0,0 +1,35 @@
...
+
+const create = context => {
+	return {
+		ArrayPattern(node) {
+			const {elements} = node;
+			if (!elements || elements.length === 0) {
+				return;
+			}
+
+			if (elements.some(isCommaFollwedWithComma)) {
```
jfmengels: I think you would get the same result, but faster if you changed this condition to `elements.find(isCommaFollwedWithComma)`

rules/no-unreadable-array-destructuring.js
```diff
@@ -0,0 +1,35 @@
...
+				context.report({
+					node,
+					message: 'Only one ignored value in series allowed in array destructuring.'
```
jfmengels: Proposal:
`Only one consecutive ignored value in series allowed in array destructuring.`
or
`Array destructuring may not contain consecutive ignored values`

rules/no-unreadable-array-destructuring.js
janowsiany: Why do you think it will be faster? `.some` returns on first truthy condition evaluation. It is also more expressive in this case as it reads better.

jfmengels: You're totally right. Misplaced my head there for some reason :sweat_smile: 
:+1:

docs/rules/no-unreadable-array-destructuring.md
```diff
@@ -0,0 +1,20 @@
+# Disallow unreadable array destructuring
+
+Destructuring is very useful, but it can also make some code harder to read.
```
sindresorhus: It should describe what it defines as "unreadable", meaning, what it limits. I know we show that in the examples below, but doesn't hurt to describe it.

test/no-unreadable-array-destructuring.js
```diff
@@ -0,0 +1,62 @@
...
+		'function foo([bar,,baz]) {}',
+		'function foo([bar,   ,     baz]) {}',
+		'function foo([bar,]) {}',
+		'function foo([bar,,]) {}',
+		'function foo([bar,, baz,, qux]) {}'
+	],
+	invalid: [
+		{
+			code: 'const [,, foo] = parts;',
+			errors: [{message: 'Array destructuring may not contain consecutive ignored values.'}]
```
sindresorhus: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/198#discussion_r228740873

rules/no-unreadable-array-destructuring.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const isCommaFollowedWithComma = (el, index, array) => {
```
sindresorhus: `el` => `element`

index.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
 				'unicorn/filename-case': ['error', {case: 'kebabCase'}],
 				'unicorn/no-abusive-eslint-disable': 'error',
 				'unicorn/no-process-exit': 'error',
+				'unicorn/no-unreadable-array-destructuring': 'error',
```
sindresorhus: Should be added to the end. https://github.com/sindresorhus/eslint-plugin-unicorn/pull/198#discussion_r228740711

sindresorhus:
Thanks for working on this, @janowsiany 🙌

sindresorhus:
The rule needs to be added to the bottom of the list here: https://github.com/sindresorhus/eslint-plugin-unicorn#rules

docs/rules/no-unreadable-array-destructuring.md
sindresorhus: I think we also should mention that you might have to modify the built-in `prefer-destructuring` rule to be compatible with this one. And how to modify it. Context: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/197#issuecomment-427459976

janowsiany:
@sindresorhus @MrHen thanks for review

sindresorhus:
Thank you, @janowsiany :)

readme.md
```diff
@@ -55,7 +55,8 @@ Configure it in `package.json`.
 			"unicorn/error-message": "error",
 			"unicorn/no-unsafe-regex": "off",
 			"unicorn/prefer-add-event-listener": "error",
-			"unicorn/no-console-spaces": "error"
+			"unicorn/no-console-spaces": "error",
+			"no-unreadable-array-destructuring": "error"
```
futpib: Isn't this line missing a "unicorn/" prefix?

sindresorhus: Yup, thanks for noticing: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/6bd6156ba1ddcd0675e7a8fee75f4bbe2b819f92

PR #212, opened 2019-01-21
Make the `no-abusive-eslint-disable` rule support shorthand syntax

buschtoens:
This change allows users to use the shorthand notation for rules, like:

```js
// eslint-disable-next-line @typescript-eslint/no-var-requires
```

If a package in a scope uses the name `eslint-plugin`, like [`@typescript-eslint/eslint-plugin`](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin), users can omit the `eslint-plugin` part.

### DIFF
--- a/rules/no-abusive-eslint-disable.js
+++ b/rules/no-abusive-eslint-disable.js
@@ -1,7 +1,7 @@
 'use strict';
 const getDocsUrl = require('./utils/get-docs-url');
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Great! I actually experienced this myself today too. Can you add a test https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/no-abusive-eslint-disable.js ?

buschtoens:
Sorry for forgetting 😅 

The tests have been updated and CI is green. 💚

PR #216, opened 2019-01-25
Add event clearing to `prefer-add-event-listener` rule

### LINKED ISSUE #163: The `prefer-add-event-listener` rule should not autofix event clearing
These should be reported, but not auto-fixed:

```js
foo.onclick = undefined;
```

```js
foo.onclick = null;
```

These need to be handled manually, as `removeEventListener` requires the listener function be able to unsubscribe.

medusalix:
Adds reporting functionality for these cases:

```js
foo.onclick = undefined;
```
```js
foo.onclick = null;
```

Resolves #163


sindresorhus:
Yeah, I agree. Just add a test to ensure it's ignored then.

@futpib Unless you know a good way or module that can help with this?

sindresorhus:

test/prefer-add-event-listener.js
```diff
@@ -26,10 +26,10 @@ const expectedBeforeUnloadWithReturnMessage = [
 ruleTester.run('prefer-add-event-listener', rule, {
 	valid: [
 		'foo.addEventListener(\'click\', () => {})',
+		'foo.removeEventListener(\'click\', onClick)',
 		'foo.onclick',
 		'foo.setCallBack = () => {console.log(\'foo\')}',
 		'setCallBack = () => {console.log(\'foo\')}',
-		'foo.onclick.bar = () => {}',
```
futpib: Why is this no longer valid?

medusalix: Oops, I think I might have deleted it by accident 😌.

medusalix:
Any news on the review process?

futpib:
I wonder about interaction between this and #218, could you also add a test for `onbeforeunload` clearing?

medusalix:
It seems like the tests are working 😃. I had to change the code a little bit, but it looks like the other test cases are still behaving as before 👍.

sindresorhus:
This looks good. Thank you, @medusalix! :)

PR #217, opened 2019-01-25
Fix ES2015 exports not counting an object as used

### LINKED ISSUE #215: no-unused-properties reports exported objects when they are also used in the file
no-unused-properties has found many unused properties - great rule. However in this situation it is reporting errors..

```
export const ITEM_A = {
    id: 'a',
    fails: true,
};

export const ITEM_B = {
    id: ITEM_A.id,
};
```

here, `fails` is reported. If I don't use `ITEM_A.id`, then nothing is reported.

futpib:
fix #215

### DIFF
--- a/rules/no-unused-properties.js
+++ b/rules/no-unused-properties.js
@@ -123,11 +123,19 @@ const create = context => {
 
 			const nextReferences = references
 				.map(reference => {
+					const {parent} = reference.identifier;
+
 					if (reference.init) {
+						if (
+							parent.type === 'VariableDeclarator' &&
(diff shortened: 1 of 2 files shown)


tz5514:
@sindresorhus When will this be released?
The latest version `7.1.0` didn't contain this PR. Thanks!

sindresorhus:
Done: https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v8.0.0

tz5514:
Huge thanks!

PR #218, opened 2019-01-25
Don't fix `onbeforeunload` unless it's handler does not return anything

### LINKED ISSUE #158: The `prefer-add-event-listener` rule should not apply to `onbeforeunload` that returns something
Meaning inline arrow functions or ones with a `return` statement.

This is because returning something changes the behavior:

```js
window.onbeforeunload = () => 'Custom message';

window.onbeforeunload = () => {
	return 'Custom message';
};
```

The above cannot be easily auto-converted to `addEventListener` as it would need to be turned into:

```js
window.addEventListener('beforeunload', event => {
	event.returnValue = 'Custom message';
});
```

Easier to just skip it then.

// @johnwu93

futpib:
fix #158

### DIFF
--- a/rules/prefer-add-event-listener.js
+++ b/rules/prefer-add-event-listener.js
@@ -7,6 +7,16 @@ const eventTypes = new Set(nestedEvents.reduce((accEvents, events) => accEvents.
 const getEventMethodName = memberExpression => memberExpression.property.name;
 const getEventTypeName = eventMethodName => eventMethodName.slice('on'.length);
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Hmm, even MDN recommends always using `addEventListener`:

> Typically, it is better to use window.addEventListener() and the beforeunload event, instead of onbeforeunload.

How about we still report on `window.onbeforeunload = () => 'Custom message';`, but don't auto-fix, and add a custom error message that tells the user to use `event.preventDefault(); event.returnValue = 'foo'`?

sindresorhus:
Perfect 👌

PR #222, opened 2019-01-26
Add `prefer-node-remove` rule

### LINKED ISSUE #213: Rule proposal: `prefer-node-remove`
Prefer [`node.remove()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove) over [`node.parentNode.removeChild(node)`](https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild). `removeChild` is the legacy way of doing it.

## Fail

```js
foo.parentNode.removeChild(foo);
this.parentNode.removeChild(this);
```

## Pass

```js
foo.remove();
this.remove();
```

medusalix:
Resolves #213

### DIFF
--- a/docs/rules/prefer-node-remove.md
+++ b/docs/rules/prefer-node-remove.md
@@ -1,0 +1,22 @@
+# Prefer `remove` over `removeChild`
+
+Enforces the use of, for example, `child.remove();` over `child.parentNode.removeChild(child);` for DOM nodes. The DOM function [`.remove()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove) is preferred over the indirect removal of an object with [`.removeChild()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild).
+
+This rule is fixable.
+
+
+## Fail
+
+```js
+foo.parentNode.removeChild(foo);
+this.parentNode.removeChild(this);
+```
+
+## Pass
+
+```js
+foo.parentNode.removeChild(bar),
+this.parentNode.removeChild(foo),
+foo.remove(),
+this.remove(),
+```
--- a/docs/rules/prefer-node-remove.md
+++ b/docs/rules/prefer-node-remove.md
@@ -1,4 +1,4 @@
-# Prefer `remove` over `removeChild`
+# Prefer `remove` over `parentNode.removeChild`
 
 Enforces the use of, for example, `child.remove();` over `child.parentNode.removeChild(child);` for DOM nodes. The DOM function [`.remove()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove) is preferred over the indirect removal of an object with [`.removeChild()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/removeChild).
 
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -54,6 +54,7 @@ module.exports = {
 				'unicorn/no-unreadable-array-destructuring': 'error',
 				'unicorn/no-unused-properties': 'off',
 				'unicorn/prefer-node-append': 'error',
+				'unicorn/prefer-node-remove': 'error',
```
sindresorhus: Should be added to the bottom.

docs/rules/prefer-node-remove.md
```diff
@@ -0,0 +1,22 @@
...
+this.parentNode.removeChild(this);
+```
+
+## Pass
+
+```js
+foo.parentNode.removeChild(bar),
+this.parentNode.removeChild(foo),
+foo.remove(),
+this.remove(),
```
sindresorhus: Use semicolon

medusalix:
Is there still something you'd want me to change? @sindresorhus

test/prefer-node-remove.js
```diff
@@ -0,0 +1,43 @@
...
+			errors: [{
+				message: 'Prefer `remove` over `removeChild`'
+			}]
+		},
+		{
+			code: 'parentNode.removeChild(this)',
+			output: 'this.remove()',
+			errors: [{
+				message: 'Prefer `remove` over `removeChild`'
+			}]
```
sindresorhus: Can you move the error array to a variable so it doesn't have to be repeated for each case?

test/prefer-node-remove.js
```diff
@@ -0,0 +1,43 @@
...
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true
+	}
+});
+
+ruleTester.run('prefer-node-remove', rule, {
+	valid: [
+		'foo.parentNode.removeChild(bar)',
+		'this.parentNode.removeChild(foo)',
```
sindresorhus: I think we should error on these too (line 13 and 14), but just don't auto-fix.

sindresorhus: Good point. I incorrectly thought that would change the semantics.

medusalix: Oh wow, I didn't even think of it from that point of view. That makes the whole rule a lot simpler 😄.

sindresorhus:
I just discovered the other day in a different project that there's something called [`.parentElement`](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentElement) (not `node`). We should support that too (also needs tests and docs).

docs/rules/prefer-node-remove.md
```diff
@@ -0,0 +1,26 @@
+# Prefer `remove` over `parentNode.removeChild` or `parentElement.removeChild`
```
sindresorhus:
```suggestion
# Prefer `remove` over `parentNode.removeChild` and `parentElement.removeChild`
```

rules/prefer-node-remove.js
```diff
@@ -0,0 +1,87 @@
...
+	if (identifier.type === 'ThisExpression') {
+		return 'this';
+	}
+
+	if (identifier.type === 'Identifier') {
+		return identifier.name;
+	}
+
+	return null;
+};
```
sindresorhus: Are any of these methods maybe generally useful and generic enough to be moved to a `util.js` file?

medusalix: Yes, I think `getMethodName` and `getCallerName` would generally be useful. `getArgumentName` only works for a single argument to a function, so I don't think that one would be used much.

futpib:
@medusalix ignoring `foo.parentNode['removeChild'](foo);` is okay.

rules/prefer-node-remove.js
sindresorhus: Would be cool if you could extract them in a separate PR.

sindresorhus:
Nice work, @medusalix 👍

PR #223, opened 2019-01-27
Add `prefer-text-content` rule

### LINKED ISSUE #156: Rule proposal: `prefer-textcontent`
Idea from: https://github.com/sindresorhus/refined-github/pull/1088#discussion_r168383196

Prefer `node.textContent` over `node.innerText`.

See: https://stackoverflow.com/a/35213639/64949

With ESLint, we can't know the type, but I guess we can safely assume `.innerText` refers to a DOM element, and if not, you can always disable it. Maybe it would be possible if `env` is `browser` and only validate then.

Thoughts?

alexzherdev:
Resolves #156.


rules/prefer-textcontent.js
```diff
@@ -0,0 +1,44 @@
...
+			const {property} = node;
+
+			if (property.type === 'Identifier' && !node.computed && property.name === 'innerText') {
+				// Handles node.innerText
+				context.report({
+					node,
+					message,
+					fix: fixer => fixer.replaceText(property, 'textContent')
+				});
+			} else if (property.type === 'Literal' && property.value === 'innerText') {
```
futpib: Could've used [`astUtils.getPropertyName(node)`](https://github.com/jfmengels/eslint-ast-utils#astutilsgetpropertynamenode) while also fixing `node["innerText"]` into `node.textContent`

alexzherdev: good point. I wasn't sure if we want to retain the existing syntax when autofixing.

futpib: I'm not sure either, @sindresorhus, what do you think?

futpib: My take would be that I don't care, since property access variant is never seen in the wild. The rule might as well ignore it completely.

sindresorhus: Yeah, we can just ignore it.

sindresorhus:

test/prefer-textcontent.js
```diff
@@ -0,0 +1,29 @@
...
+const error = {
+	ruleId: 'prefer-textcontent',
+	message: 'Prefer `textContent` over `innerText`.'
+};
+
+ruleTester.run('prefer-textcontent', rule, {
+	valid: [
+		'innerText;',
+		'node.textContent;',
+		'node[innerText];'
```
sindresorhus: Can you also add:

- `innerText = true;`
- `'node['innerText'];'`

test/prefer-textcontent.js
```diff
@@ -0,0 +1,29 @@
...
+	],
+	invalid: [
+		{
+			code: 'node.innerText;',
+			output: 'node.textContent;',
+			errors: [error]
+		}
+	]
```
sindresorhus: Can you also add an assignment case: `node.innerText = 'foo';`

sindresorhus:
I just realized that to be consistent with the naming of the other rules, this rule needs to be named `prefer-textcontent` => `prefer-text-content`.

sindresorhus:
Nice work, @alexzherdev 🙌

PR #224, opened 2019-01-27
Avoid false positives in the `new-for-builtins` rule

### LINKED ISSUE #122: False positive for the `new-for-builtins` rule
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/122)

Currently the plugin doesn't support variables that have the same name as the built-ins. 
For example:
```js
import { Map } from 'immutable'
const m = Map()
```
will report ```Use `new Map()` instead of `Map()`. (unicorn/new-for-builtins)``` which is not correct for Immutable.js

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $30.00)

### Submitted pull Requests
- [#224 Avoid false positives in new-for-builtins](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/224)
- [#540 Add `isShadowed` check to `new-for-builtins`](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/540)
---

### Tips

---
</details>

alexzherdev:
Fixes #122

### DIFF
--- a/rules/new-for-builtins.js
+++ b/rules/new-for-builtins.js
@@ -33,10 +33,35 @@ const disallowNew = new Set([
 	'Symbol'
 ]);
 
+function isVariableInScope(context, name) {
+	let scope = context.getScope();
+	let {variables} = scope;
+
+	while (scope.type !== 'global') {
+		scope = scope.upper;
+		variables = scope.variables.concat(variables);
+	}
+
+	if (scope.childScopes.length > 0) {
+		variables = scope.childScopes[0].variables.concat(variables);
+		if (scope.childScopes[0].childScopes.length > 0) {
+			variables = scope.childScopes[0].childScopes[0].variables.concat(variables);
+		}
+	}
+
+	return variables.some(variable => (
+		variable.name === name && variable.defs[0] && variable.defs[0].node
+	));
+}
+
 const create = context => {
 	return {
 		CallExpression: node => {
 			const {name} = node.callee;
+
+			if (isVariableInScope(context, name)) {
+				return;
+			}
 
 			if (enforceNew.has(name)) {
 				context.report({
@@ -48,6 +73,10 @@ const create = context => {
 		},
 		NewExpression: node => {
 			const {name} = node.callee;
+
+			if (isVariableInScope(context, name)) {
+				return;
+			}
 
 			if (disallowNew.has(name)) {
 				context.report({
--- a/test/new-for-builtins.js
+++ b/test/new-for-builtins.js
@@ -46,7 +46,33 @@ ruleTester.run('new-for-builtins', rule, {
 		'const foo = Boolean()',
 		'const foo = Number()',
 		'const foo = String()',
-		'const foo = Symbol()'
+		'const foo = Symbol()',
+		`
+			const {Map} = require('immutable');
+			const foo = Map();
+		`,
+		{
+			code: `
+				import {Map} from 'immutable';
+				const foo = Map();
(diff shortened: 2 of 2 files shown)


rules/new-for-builtins.js
```diff
@@ -33,11 +33,36 @@ const disallowNew = new Set([
 	'Symbol'
 ]);
 
+function isVariableInScope(context, name) {
```
futpib: So, if I got this right, this function is supposed to check whether the variable is defined by the user (returns `true` in that case), or the variable is implicitly defined in the global scope. Is this right?

If so, here is an alternative implementation idea, without having to collect all the variables into an array:

<p>
<details>
 <summary>Click to see, or don't click and try to figure it out :thinking: </summary>

Walk up the scopes (via `scope.upper`) while checking each scope's [`.set`](https://eslint.org/docs/developer-guide/scope-manager-interface#set) for the variable name. If you find it and the scope is not the global scope, the variable was defined by the user, otherwise it came from the global scope.

</details>
</p>


Also, if I did get the intention right, a better name for the function would be `isUserDefinedVariable`.

rules/new-for-builtins.js
```diff
@@ -33,11 +33,36 @@ const disallowNew = new Set([
...
+
+	while (scope.type !== 'global') {
+		scope = scope.upper;
+		variables = scope.variables.concat(variables);
+	}
+
+	if (scope.childScopes.length > 0) {
+		variables = scope.childScopes[0].variables.concat(variables);
+		if (scope.childScopes[0].childScopes.length > 0) {
+			variables = scope.childScopes[0].childScopes[0].variables.concat(variables);
```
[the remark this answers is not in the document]
alexzherdev: This is an adapted version of [eslint-plugin-react's detection](https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/util/variable.js#L35). I wasn't brave enough to change it, perhaps @yannickcr could shed some light.

sindresorhus:

sindresorhus:

rules/new-for-builtins.js
sindresorhus: And I think it should be moved to the `utils` directory.

sindresorhus:
@alexzherdev Still interested in finishing this? :)

rules/new-for-builtins.js
futpib: This is already implemented in https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/utils/resolve-variable-name.js

alexzherdev:
Yup! sorry for the delay, I do intend to get back to this.

rules/new-for-builtins.js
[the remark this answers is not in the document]
alexzherdev: @futpib This was added in https://github.com/yannickcr/eslint-plugin-react/commit/6a63a6c4b143d0148a71d96ef7144f0b134c0234 to work around an issue in a now ancient version of babel-eslint. I ran eslint-plugin-react tests against a few recent major versions of babel-eslint and this workaround doesn't seem necessary anymore.
But you're right, I'll try to refactor this.

sindresorhus:
@alexzherdev Ping

sindresorhus:
Closing for lack of activity: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/122#issuecomment-552803259

PR #225, opened 2019-01-29
Add `prefer-dataset` rule

### LINKED ISSUE #166: Rule proposal: `prefer-dataset`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/166)

Prefer the [`.dataset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset) property on DOM elements over using `setAttribute`.

We would detect `setAttribute` and that the string starts with `data-`.

## Fail

```js
element.setAttribute('data-unicorn', '🦄');
```

## Pass

```js
element.dataset.unicorn = '🦄';
```

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/166).

amedora:
Fix #166

### DIFF
--- a/rules/prefer-dataset.js
+++ b/rules/prefer-dataset.js
@@ -11,13 +11,9 @@ const getDataAttributeName = arg => {
 	return '';
 };
 
-const parseValueArgument = (context, arg) => {
-	return context.getSourceCode().getText(arg);
-};
+const parseValueArgument = (context, arg) => context.getSourceCode().getText(arg);
 
-const dashToCamelCase = str => {
-	return str.replace(/-([a-z])/g, s => s[1].toUpperCase());
-};
+const dashToCamelCase = string => string.replace(/-([a-z])/g, s => s[1].toUpperCase());
 
 const getReplacement = (context, node, memberExpression, propertyName) => {
 	const objectName = memberExpression.object.name;
@@ -33,9 +29,9 @@ const getReplacement = (context, node, memberExpression, propertyName) => {
 };
 
 const isBracketNotation = (context, callee) => {
-	const bracketOpen = context.getSourceCode().getFirstTokenBetween(callee.object, callee.property, {filter: token => {
-		return token.value === '[';
-	}});
+	const bracketOpen = context.getSourceCode().getFirstTokenBetween(callee.object, callee.property, {
+		filter: token => token.value === '[';
+	});
 
 	return bracketOpen !== null && bracketOpen.value === '[';
 };
@@ -62,7 +58,7 @@ const create = context => {
 				const replacement = getReplacement(context, node, callee, attributeName);
 				context.report({
 					node,
-					message: 'Prefer `dataset` over `setAttribute`',
+					message: 'Prefer `.dataset` over `setAttribute(…)`.',
 					fix: fixer => fixer.replaceText(node, replacement)
 				});
 			}
--- a/rules/prefer-dataset.js
+++ b/rules/prefer-dataset.js
@@ -1,0 +1,72 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const getMethodName = memberExpression => memberExpression.property.name;
+
+const getDataAttributeName = arg => {
+	if (arg.type === 'Literal') {
+		return (arg.value.match(/^data-(\w+)/) || ['', ''])[1];
+	}
+
+	return '';
+};
+
+const parseValueArgument = (context, arg) => {
+	return context.getSourceCode().getText(arg);
+};
+
+const getReplacement = (context, node, memberExpression, propertyName) => {
+	const objectName = memberExpression.object.name;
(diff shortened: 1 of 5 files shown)


rules/prefer-dataset.js
```diff
@@ -0,0 +1,53 @@
...
+
+	return `${objectName}.dataset.${propertyName} = ${value}`;
+};
+
+const create = context => {
+	return {
+		CallExpression(node) {
+			const {callee} = node;
+
+			if (callee.type === 'MemberExpression' && getMethodName(callee) === 'setAttribute') {
```
futpib: Code like `element[setAttribute]('data-unicorn', '🦄')` will "break" this check

amedora: oh, I got you.
I have no idea which one is better...
- just ignore bracket notation
- fix it to element[dataset].unicorn = '🦄';

sindresorhus: You have to ignore bracket notation. It’s very hard to impossible to reliably check this.

sindresorhus: You should add some tests for this case too.

sindresorhus:

sindresorhus:

test/prefer-dataset.js
```diff
@@ -0,0 +1,25 @@
...
+		'element.dataset[\'unicorn\'] = \'🦄\';',
+		'element[setAttribute](\'data-unicorn\', \'🦄\');',
+		'element.setAttribute(\'foo\', \'bar\');'
+	],
+	invalid: [
+		{
+			code: 'element.setAttribute(\'data-unicorn\', \'🦄\');',
+			errors: [{message: 'Prefer `dataset` over `setAttribute`'}],
+			output: 'element.dataset.unicorn = \'🦄\';'
+		}
```
MrHen: A few more test cases worth checking:

* `element.setAttribute(\'data-foo-bar\', \'zaz\')`
* `element.setAttribute(\'data-🦄\', \'🦄\')`
* `element.setAttribute(foo, bar)`
* `element.setAttribute(\'data-foo2\', \'bar\')`
* `element.setAttribute(\'data-foo:bar\', \'zaz\')`
* `element.setAttribute(\'data-foo\', /* comment */ \'bar\')`
* ```
element.setAttribute(
  \'data-foo\', // comment
  \'bar\', // comment
)
```

Note that the fix conversions should follow the description in the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset):

> The name of a custom data attribute in JavaScript is the name of the same HTML attribute but in camelCase and with no dashes, dots, etc.

I think this means `element.setAttribute('data-date-of-birth', '1960-10-03')` should be fixed to `element.dataset.dateOfBirth = '1960-10-03'`.

Comment (amedora) [reply test/prefer-dataset.js]: Thank you for your feedback :smiley:

File: test/prefer-dataset.js [hunk 3]
```diff
@@ -8,18 +9,67 @@ const ruleTester = avaRuleTester(test, {
...
+		},
+		{
+			code: 'element.setAttribute(\'data-foo-bar\', \'zaz\');',
+			errors,
+			output: 'element.dataset.fooBar = \'zaz\';'
+		},
+		{
+			code: 'element.setAttribute(\'data-foo\', /* comment */ \'bar\');',
+			errors,
+			output: 'element.dataset.foo = \'bar\';'
```
MrHen: I don't think we want to delete comments during fixing but it probably doesn't come up that often. @sindresorhus, what's your take on this?

1. Removing comments is okay
2. Comments should be preserved (may be non-trivial)
3. Do not fix if comments are detected

[I've done (3) in the past](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/305/files#diff-c717db2a17e1fd6455af92ec118c404eR35) but I think a bunch of different rules current do (1).

sindresorhus: @MrHen I don't think we should do `1.`, but we need to come up with a guideline that we can apply to all rules. Maybe we could make an utility that just extract all comments to above a node we give it? The comments can require manual formatting and movement. We just shouldn't remove them. Alternatively, `3.` if `2.` is too complex, but I think we could make something generic that we could apply to many rules. Would you be able to open a new issue about it? We probably need to review all existing rules for this and update the contributing guidelines.

sindresorhus:
Thanks, @amedora :)

sindresorhus:
Note https://github.com/sindresorhus/eslint-plugin-unicorn/commit/850b8a1eb3af5edf9c02c440672a8c293231429c. I missed that during the review.

test/prefer-dataset.js
MrHen: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/370

PR #226, opened 2019-01-29
Add `prefer-event-key` rule

### LINKED ISSUE #118: Prefer `event.key` over `event.keyCode`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/118)

For readability. Got the idea when doing https://github.com/wulkano/kap/commit/71be04f7411aa33d0b393c05077c619de2ab07be.

Should trigger when in an `addEventListener` callback.

## Fail

```js
foo.addEventListener('click', event => {
	if (event.keyCode === '27') {
		…
	}
});
```

## Pass

```js
foo.addEventListener('click', event => {
	if (event.key === 'Escape') {
		…
	}
});
```

More info: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

---

For the first implementation, we can just naively look for the same variable name as defined as the parameter. Usually `event` or `e`.

Later on, we could add code path support and support having the function outside the `addEventListener` call:

```js
const handler = event => {
	if (event.key === 'Escape') {
		…
	}
};

…

foo.addEventListener('click', handler);
```

---

@kevva @SamVerschueren Thoughts?

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/118).

ankeetmaini:
Fixes #118
- [x] add fixer code


ankeetmaini:
I think my approach isn't very good here, trying to first narrow down to `addEventListener` and then finding all the `keyCode` bits using `SourceCode`'s `getToken` API. Because querying under a Node is a little difficult and by doing `getTokens` I lose the parent-child link. This will make it unusable for destructured elements.

Following up with a better one where in I directly select `keyCode` identifier and then using `context.getAncestors()` to assert that they're infact inside an event callback.

futpib:
One foreseeable gotcha with `context.getAncestors()` is that a `.keyCode` that is syntactically inside an `addEventListener` callback is not exactly the same as a `.keyCode` on the callback's argument.

```js
const something = undefined;
element.addEventListener('keyup', event => {
  console.log(something.keyCode); // This `.keyCode` is "inside an event callback"
});
```

I think this can be correctly solved by targeting the callback function in the visitor, then getting it's scope and checking all the references to the first argument to see if they are `.keyCode`.

ankeetmaini:
I'll have the CallbackExpression and was in process to do the same thing! I'll double back shortly on how it goes :)

Also I assume we only want to target `keyup` and `keydown` events. I'll add that check as well.

futpib:
> Also I assume we only want to target `keyup` and `keydown` events. I'll add that check as well.

We might as well target all events, since accessing `keyCode` (and others) on non-`KeyboardEvent`s makes no sense, as it will always be `undefined`.

@sindresorhus Any thoughts?

@ankeetmaini If we indeed want to target only [`KeyboardEvent`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent)s, there is also `keypress`.

sindresorhus:

sindresorhus:
> We might as well target all events, since accessing keyCode (and others) on non-KeyboardEvents makes no sense, as it will always be undefined.

👍

ankeetmaini:
@futpib [this version shows my current approach](https://astexplorer.net/#/gist/d851aecda1ea05170d7797d2de24d04d/49ba94f129b2a9eddda35edd9107f3e0efce638b). It also takes care of the edge case you suggested. Will poke a bit more before making it final.

ankeetmaini:
I am happy with the current approach and have added a lot of tests for valid/invalid code. The rule also covers the destructuring cases too. Please take a look :)

The thing that I am not convinced on is the **fixing** part. How do we fix if the case is like

```js
bar.addEventListener('keyup', e => {
  sendThisToAnOtherModule(e.keyCode)
})
```

In this even if I try to fix it, it'll break the contract as the the receiving party might not accept a string representation of the key. And in other case like

```js
const keys = {
  'enter': 34
}
bar.addEventListener('keyup', e => {
  if (e.keyCode === keys.enter)  return;
})

```

I am not sure if we should try to fix as we might step over and break things. Please advise!

futpib:
I think best we can do is fix only comparisons against literals (like `event.keyCode === 27` => `event.key === 'Escape'`)

ankeetmaini:
Update: @futpib I've added the fixer code which takes care of direct if conditions.
```js
if (e.keyCode === 27) -> if (e.key === 'Escape')
```

but the case
```js
const {keyCode} = e;
if (keyCode === 27) // do something
```
needs to be looked into.

readme.md
```diff
@@ -110,8 +111,12 @@ Configure it in `package.json`.
 - [no-unused-properties](docs/rules/no-unused-properties.md) - Disallow unused object properties.
 - [prefer-node-append](docs/rules/prefer-node-append.md) - Prefer `append` over `appendChild`. *(fixable)*
 - [prefer-query-selector](docs/rules/prefer-query-selector.md) - Prefer `querySelector` over `getElementById`, `querySelectorAll` over `getElementsByClassName` and `getElementsByTagName`. *(partly fixable)*
+<<<<<<< HEAD
 - [prefer-node-remove](docs/rules/prefer-node-remove.md) - Prefer `remove` over `parentNode.removeChild` and `parentElement.removeChild`. *(fixable)*
 - [prefer-text-content](docs/rules/prefer-text-content.md) - Prefer `textContent` over `innerText`. *(fixable)*
+=======
```
futpib: Some git conflict leftovers

ankeetmaini: Fixed, sorry!

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,201 @@
...
+	81: 'q',
+	82: 'r',
+	83: 's',
+	84: 't',
+	85: 'u',
+	86: 'v',
+	87: 'w',
+	88: 'x',
+	89: 'y',
+	90: 'z'
```
futpib: You don't need a table for common printable characters, you can instead use `String.fromCharCode`. Let's learn from React's code for `key`, which is almost exactly what we need (except we don't need the `normalizeKey` thingy) https://github.com/facebook/react/blob/b87aabd/packages/react-dom/src/events/getEventKey.js

futpib:
@ankeetmaini Would be nice to handle destructurings too, but I think we can give it a pass, as other rules like this (`prefer-node-remove` (with respect to `parentNode`/`parentElement`) and `prefer-text-content`) don't fix it too.

Unless @sindresorhus disagrees.

ankeetmaini:
@futpib I'm planning to give `fixing-destructured-variables` a try over the weekend. Will come back with an update.

ankeetmaini:
Hi @futpib, this is ready for review now. Also I couldn't figure out a way to _fix_ destructured variables in `if` conditions.

I tried multiple ways, only one handler for all all keyCode, which etc (currently I target `Property` for destructured keys) but it threw as it was generating duplicate errors one for the key part and one for value
```js
const {keyCode} = event
```
Here this generates into two identifiers, where both the key and value of `keyCode` is same, but the tests couldn't figure out === equality and were failing.

Even if I fix this, when the above variable is referenced again I wasn't able to find out if it is resolved from the destructured variable, as the AST walk function is called separately for them.

I also tried using `context.getScope()` it returns all the references but I don't get the `parent` attribute (weird!) and hence wasn't able to assert if it's in an If condition.

index.js
```diff
@@ -56,7 +56,8 @@ module.exports = {
 				'unicorn/prefer-node-append': 'error',
 				'unicorn/prefer-query-selector': 'error',
 				'unicorn/prefer-node-remove': 'error',
-				'unicorn/prefer-text-content': 'error'
+				'unicorn/prefer-text-content': 'error',
+				'unicorn/prefer-key-over-key-code': 'error'
```
futpib: @sindresorhus Should this rule be named `prefer-key` or `prefer-event-key`, without the "over something else" part, like the other rules, or is the current name fine?

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,189 @@
...
+	context.report({
+		message: `Use key instead of ${node.name}. ${extraCitation}`,
+		node,
+		fix: fix(context, node)
+	});
+};
+
+const create = context => {
+	return {
+		'Identifier:matches([name=keyCode], [name=charCode], [name=which])': directAccessRule(
```
futpib: Why not inline the `directAccessRule` here, just like in all the other rules? This way readers won't have to jump back and forth to see which selector this visitor method is handling.

futpib:
> I also tried using `context.getScope()` it returns all the references but I don't get the `parent` attribute (weird!) and hence wasn't able to assert if it's in an If condition.

I guess you were looking for [`referece.identifier.parent`](https://eslint.org/docs/developer-guide/scope-manager-interface#identifier). But fixing destructuring was an optional goal anyway.

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,189 @@
...
+
+		default:
+			return null;
+	}
+};
+
+const extraCitation = 'See https://goo.gl/cRK532 for more info.';
+
+const util = context => {
+	const getNearestAncestorByType = nodes => type => {
```
futpib: I've pushed a failing test (https://github.com/sindresorhus/eslint-plugin-unicorn/pull/226/commits/1b7c37f261c8de5253a7b2fcb84889b4bfe79ab9), and I believe it is indicative of a problem with a general approach of finding the closest `CallExpression` and then assuming it's callback's argument refers to the same variable you got in the `Identifier` visitor.

You should use [ScopeManager](https://eslint.org/docs/developer-guide/scope-manager-interface) to properly check if the `e` from `e.keyCode` refers to the same variable as the `e` from `.addEventListener((e) => { ... })`.

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,189 @@
...
+		for (let i = nodes.length - 1; i >= 0; i--) {
+			if (nodes[i].type === type) {
+				return nodes[i];
```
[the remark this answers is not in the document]
ankeetmaini: I've totally removed this implementation. Now traversing the parent instead of relying on ancestors, which I thought was a little misleading in case I wanted to check the parents of another node in the current traversal.

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+		default:
+			return null;
+	}
+};
+
+const extraCitation = 'See https://goo.gl/cRK532 for more info.';
+
+const util = context => {
```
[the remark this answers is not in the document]
ankeetmaini: Yep, totally removed the util which was returning helpers. All the helpers are now defined at the top level!

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,189 @@
...
+		callExpressionNode && callExpressionNode.arguments[1]
+	);
+	return {
+		report,
+		event,
+		getNearestAncestorByType: getNearestAncestorByType(ancestors)
+	};
+};
+
+const directAccessRule = context => node => {
```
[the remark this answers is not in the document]
ankeetmaini: Correct! Removed and also inlined removing needless indirection

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,189 @@
...
+		for (let i = nodes.length - 1; i >= 0; i--) {
+			if (nodes[i].type === type) {
+				return nodes[i];
+			}
+		}
+	};
+
+	const report = reportError(context);
+	const ancestors = context.getAncestors();
+	const callExpressionNode = isInsideEventCallback(
```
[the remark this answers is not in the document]
ankeetmaini: I agree! Renamed it and also changed its implementation without depending on ancestors.

rules/prefer-key-over-key-code.js
ankeetmaini: Done!

rules/prefer-key-over-key-code.js
ankeetmaini: Yes, the the way I was doing was not correct. I'm now tracking the references and then correctly reporting on the intended node! I am sorry, I missed your comment before. Would've circled back earlier :)

ankeetmaini:
@futpib thanks a lot for a thorough review! I've fixed and updated the implementation slightly as per your comments. Let me know!

sindresorhus:

index.js
sindresorhus: Yes, it should be `prefer-event-key`.

docs/rules/prefer-key-over-key-code.md
```diff
@@ -0,0 +1,33 @@
+# Enforce the use of event.key instead of event.keyCode
```
sindresorhus:
```suggestion
# Prefer `event.key` over `event.keyCode`
```

Same with the description in the readme.

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,198 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const keys = ['keyCode', 'charCode', 'which'];
+const extraCitation = 'See https://goo.gl/cRK532 for more info.';
```
sindresorhus: This doesn't belong in the message. People can read about it in the rule docs instead.

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,198 @@
...
+	117: 'F6',
+	118: 'F7',
+	119: 'F8',
+	120: 'F9',
+	121: 'F10',
+	122: 'F11',
+	123: 'F12',
+	144: 'NumLock',
+	145: 'ScrollLock',
+	224: 'Meta'
```
sindresorhus: Quote the keys.

test/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,519 @@
...
+			code: `
+				foo.addEventListener('click', a => {
+					if (a.charCode === 27) {
+					}
+				});
+			`,
+			errors: [error('charCode')],
+			output: `
+				foo.addEventListener('click', a => {
+					if (a.key === 'Escape') {
```
sindresorhus: Add tests for more keys than just Escape. I'd like to see tests for the following `keyCode`'s:

- 13
- 38
- 190
- 221

rules/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,198 @@
...
+	return [
+		fixer.replaceText(node, 'key'),
+		fixer.replaceText(right, `'${keyCode}'`)
+	];
+};
+
+const create = context => {
+	const report = node => {
+		context.report({
+			message: `Use key instead of ${node.name}. ${extraCitation}`,
```
sindresorhus:
```suggestion
			message: `Use \`key\` instead of \`${node.name}\`.`,
```

test/prefer-key-over-key-code.js
```diff
@@ -0,0 +1,519 @@
...
+					const {charCode, a, b} = e;
+				});
+			`,
+			errors: [error('charCode')]
+		},
+		{
+			code: `
+				foo.addEventListener('click', e => {
+					const {a: charCode, a, b} = e;
+				});
```
sindresorhus: Can you also add a test with destructuring directly in the parameter?

```js
foo.addEventListener('click', ({keyCode}) => {

});
```

rules/prefer-event-key.js
```diff
@@ -0,0 +1,197 @@
...
+			return {};
+	}
+};
+
+const isPropertyOf = (node, eventNode) => {
+	return (
+		node &&
+		node.parent &&
+		node.parent.object &&
+		node.parent.object === eventNode
```
futpib: Please also check explicitly for node types (`node.type`, `node.parent.type`, etc.). This also applies to other checks above and below. This is necessary as multiple node types (both currently and in the future) may have a same property with different meaning (example: `object` property is defined on both `MemberExpression` and `WithStatement`, see [estree](https://github.com/estree/estree))

docs/rules/prefer-key-over-key-code.md
ankeetmaini: Done

test/prefer-key-over-key-code.js
ankeetmaini: Done!

rules/prefer-key-over-key-code.js
ankeetmaini: It throws `Unnecessarily quoted property 145 found.  quote-props`

test/prefer-key-over-key-code.js
ankeetmaini: Nice catch, I did not handle this case :D. Added now!

docs/rules/prefer-event-key.md
```diff
@@ -0,0 +1,37 @@
+# Enforce the use of `event.key` instead of `event.keyCode`
+
+Enforces the use of [`KeyboardEvent#key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) over [`KeyboardEvent#keyCode`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) which is deprecated. The `key` property is also more semantic and readable.
+
+
```
sindresorhus: Please describe that it's fixable and what is fixable. See the other docs for how to write it.

rules/prefer-event-key.js
sindresorhus: @ankeetmaini Is this done?

sindresorhus:
@ankeetmaini Can you fix the merge conflict and resolve the last feedback?

ankeetmaini:
@sindresorhus I am so sorry for not getting to it I'll update this shortly. Thanks for updating this PR. Will also address @futpib's comment. :)

rules/prefer-event-key.js
```diff
@@ -0,0 +1,222 @@
...
+	return [
+		fixer.replaceText(node, 'key'),
+		fixer.replaceText(right, `'${keyCode}'`)
+	];
+};
+
+const create = context => {
+	const report = node => {
+		context.report({
+			message: `Use \`key\` instead of \`${node.name}\``,
```
sindresorhus:
```suggestion
			message: `Use \`.key\` instead of \`.${node.name}\``,
```

rules/prefer-event-key.js
```diff
@@ -0,0 +1,222 @@
...
+const isPropertyOf = (node, eventNode) => {
+	return (
+		node &&
+		node.parent &&
+		node.parent.object &&
+		node.parent.object === eventNode
+	);
+};
+
+// Third argument is a condition function, as in passed to `Array#filter()`
```
sindresorhus:
```suggestion
// The third argument is a condition function, as one passed to `Array#filter()`
```

rules/prefer-event-key.js
ankeetmaini: done

ankeetmaini:
@sindresorhus all requested changes have been pushed!

sindresorhus:
// @MrHen

sindresorhus:
Looks good to me now. Thanks for contributing, @ankeetmaini :)

ankeetmaini:
Thanks so much! Yay!

PR #228, opened 2019-01-30
Add `no-default-parameter-options` rule

### LINKED ISSUE #208: Rule proposal: No default parameter options
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/208)

I have seen this pattern of using default parameters with an object as default options values:

```js
const foo = (options = {a: false}) => {
	// …
};
```

This looks ok, but it's easy to miss that it only works with one option. If the user passes in an object, the whole default object will be replaced.

So here, the `b` option default value will be lost:

```js
const foo = (options = {a: false, b: true}) => {
	// …
};

foo({a: true});
```

I have seen this mistake a lot in the wild, so would be good to have a rule for it.

Example: https://github.com/lukechilds/byte-range/blob/3d154a9af0f827dfbb44c5acf32029cf1a549ccb/src/index.js#L3

The correct way to handle this is to use object spreading:

```js
const foo = options => {
	options = {
		a: false,
		b: true,
		...options
	};
};

foo({a: true});
```

The rule should fail on parameters named `options` or `opts` that have a default parameters value that is an object literal. The exception is an empty object, which should be allowed.

## Fail

```js
const foo = (options = {a: false}) => {
	// …
};
```

```js
const foo = (opts = {a: false, b: true}) => {
	// …
};
```

## Pass

```js
const foo = options => {
	options = {
		a: false,
		b: true,
		...options
	};
};
```

```js
const foo = (options = {}) => {
	// …
};
```

Would be nice to auto-fix this, but I don't think it's very easy, so it's not a requirement to resolve this issue.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $82.00)

- [0maxxam0 0maxxam0](https://issuehunt.io/u/0maxxam0) ($2.00)

### Submitted pull Requests
- [#633 Add `disallow-objects-as-default` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/633)
---

### Tips

---
</details>

medusalix:
Resolves #208


medusalix:
@futpib I would really appreciate if you could help review this PR 😃.

test/no-default-parameter-options.js
```diff
@@ -0,0 +1,143 @@
...
+			errors: [error]
+		},
+		{
+			code: 'const foo = (options = {a: false, b: 1, c: "test", d: null}) => {};',
+			output: `const foo = (options) => {
+			options = {
+				a: false,
+b: 1,
+c: "test",
+d: null,
```
futpib: Please fix the indentation here

medusalix: This is necessary for the tests to work. Just like [this rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/no-unused-properties.js). But I'll see if I can come up with a solution that corrects the indentation.

futpib: Oh, I see that your fixer generates the `options = {...}` statement without indentation, and this is probably fine as `indent` rule would probably fix the indentation (but I have not checked this). But why are you referring to [`no-unused-properties` tests](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/no-unused-properties.js) then? It has no fixer and no tests with `output` key.

medusalix: Yeah, I'm not sure whether it makes sense to make the fixer more complicated and error prone when it's being auto-indented automatically. I referred to the other rule because I thought the indentation also mattered in that rule's tests (which apparently doesn't 😏).

I could probably count the amount of spaces before the first statement and append those spaces to the next line. However I'm still waiting for @sindresorhus take on the question I asked [here](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/208#issuecomment-465053487).

sindresorhus:
@medusalix Are you still interested in finishing this? See: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/208#issuecomment-576647974

medusalix:
@sindresorhus Yes, I am. Just to be clear that I understood everything correctly: You want me to split this into a `prefer-default-parameters` rule (with fixing) and a `disallow-objects-as-default` rule (without fixing), right?

sindresorhus:
Correct

PR #232, opened 2019-02-02
Add `import-style` rule

### LINKED ISSUE #211: Rule proposal: Don't import `util` Node.js core module directly
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/211)

This module is the perfect case for using destructuring / named imports, as it's just a bunch of unrelated functions.

## Fail

```js
const util = require('util');
```

```js
import util from 'util';
```

```js
import * as util from 'util';
```

## Pass

```js
const {promisify} = require('util');
```

```js
import {promisify} from 'util';
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Submitted pull Requests
- [#232 #211 - Add `no-direct-import` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/232)
- [#789 Add `import-style` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/789)
---

### Tips

</details>

gurrrung:
Fixed #211 

Things Done:
- Throws error on directly importing default modules like utils, lodash, etc with support for overriding the objects to add own modules.


sindresorhus:
> (I just need a little help on the todo part, can you give me a code pointer or any current rule which does this, I will come up with a solution.)

This rule accepts options: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/catch-error-name.js See the `schema` part. More here; https://eslint.org/docs/developer-guide/working-with-rules#options-schemas

test/no-direct-import.js
```diff
@@ -0,0 +1,38 @@
...
+		{
+			code: 'import util from \'util\';',
+			errors: [error]
+		},
+		{
+			code: 'import * as util from \'util\';',
+			errors: [error]
+		}
+	]
+});
```
sindresorhus: This needs a lot more tests, both valid and invalid ones. There are many more ways to import. For example:

- `import {promisify as foo} from \'util\';`
- `import {debuglog, promisify} from \'util\';`
- `import util, {promisify} from \'util\';`
- `const foo = myFunction(\'util\')`
- `const {promisify as foo} = require(\'util\');`

docs/rules/no-direct-import.md
```diff
@@ -0,0 +1,22 @@
+# Do not import modules directly
+
+Do not import the core Node.js modules directly.
```
sindresorhus: This is not correct. It doesn't apply to all core Node.js modules, just some, and not even just core modules.

readme.md
```diff
@@ -68,6 +68,7 @@ Configure it in `package.json`.
 			"unicorn/no-unsafe-regex": "off",
 			"unicorn/prefer-add-event-listener": "error",
 			"unicorn/no-console-spaces": "error",
+			"unicorn/no-direct-import": "error",
```
sindresorhus: Should be added to the bottom.

index.js
```diff
@@ -51,6 +51,7 @@ module.exports = {
 				'unicorn/prefer-add-event-listener': 'error',
 				'unicorn/prefer-exponentiation-operator': 'error',
 				'unicorn/no-console-spaces': 'error',
+				'unicorn/no-direct-import': 'error',
```
sindresorhus: Should be added to the bottom.

readme.md
```diff
@@ -104,6 +105,7 @@ Configure it in `package.json`.
 - [prefer-add-event-listener](docs/rules/prefer-add-event-listener.md) - Prefer `addEventListener` over `on`-functions. *(fixable)*
 - [prefer-exponentiation-operator](docs/rules/prefer-exponentiation-operator.md) - Prefer the exponentiation operator over `Math.pow()` *(fixable)*
 - [no-console-spaces](docs/rules/no-console-spaces.md) - Do not use leading/trailing space between `console.log` parameters. *(fixable)*
+- [no-direct-import](docs/rules/no-direct-import.md) - Do not import the core Node.js modules directly.
```
sindresorhus: Should be added to the bottom.

sindresorhus:
The rule should support enforcing destructuring and using default import, as commented in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/211#issuecomment-457173568 So I think the rule needs a better name. It can't have the `no-` prefix:

> we could also have a list for the inverse? For example, chalk should not be destructured, but people still do. Same with the core path module.

gurungrahul2:
> The rule should support enforcing destructuring and using default import, as commented in [#211 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/211#issuecomment-457173568) So I think the rule needs a better name. It can't have the `no-` prefix:
> 
> > we could also have a list for the inverse? For example, chalk should not be destructured, but people still do. Same with the core path module.

how about prefer-default-import ? @sindresorhus

futpib:
How about `import-specifier-style` or `import-destructuring`?

gurungrahul2:
@sindresorhus

sindresorhus:
> How about `import-specifier-style` or `import-destructuring`?

The `specifier` is the right side of the import statement. The braces in import is also not destructuring.

How about just `import-style` or `import-default-or-named`?

gurungrahul2:
Another problem is how can I work for the inverse part, can I take an input from user where he will input something like `` import-style: inverse`` and those module names will be linted in inverse way ?

gurungrahul2:
> > How about `import-specifier-style` or `import-destructuring`?
> 
> The `specifier` is the right side of the import statement. The braces in import is also not destructuring.
> 
> How about just `import-style` or `import-default-or-named`?

Import-style sounds good to me,

futpib:
> The `specifier` is the right side of the import statement. The braces in import is also not destructuring.

Specifier is the left-hand side. In case of a `require` call it can instead be a destructring. `import-style` is ok

sindresorhus:
> Specifier is the left-hand side.

I just Google'd "javascript import specificer" and all the sources claim it's the right part, like: https://nodejs.org/api/esm.html#esm_url_based_paths

sindresorhus:
Alright, let's go with `import-style.`

sindresorhus:
> Another problem is how can I work for the inverse part, can I take an input from user where he will input something like ` import-style: inverse` and those module names will be linted in inverse way ?

The rule should accepts an options-object like this:

```js
{
	defaultExport: [
		'path',
		'chalk'
	],
	namedExport: [
		'util',
		'lodash',
		'underscore'
	]
}
```

With the above being the default.

sindresorhus:
@futpib Or do you think we should do it like this so it's easier to override individual entries?

```js
{
	defaultExport: [
		path: true
		chalk: true
	],
	namedExport: [
		util: true,
		lodash: true
		underscore: true
	]
}
```

futpib:
> I just Google'd "javascript import specifier" and all the sources claim it's the right part, like: [nodejs.org/api/esm.html#esm_url_based_paths](https://nodejs.org/api/esm.html#esm_url_based_paths)

I went with what astexplorer and [estree](https://github.com/estree/estree/blob/master/es2015.md#importspecifier) call import specifier, and it's the same in [the standard](https://www.ecma-international.org/ecma-262/#prod-ImportSpecifier). Ok, this is too confusing for a rule name.

futpib:
> Or do you think we should do it like this so it's easier to override individual entries?

Yes, I prefer the object variant. Would be consistent with what I chose in #237

sindresorhus:
Alright, let's go with the object variant.

gurungrahul2:
Ok thanks, I will come up with a commit soon.

gurungrahul2:
I'm stuck on how to pass the configurations, I am working in AST Explorer How do I pass options here? @futpib @sindresorhus

futpib:
@gurungrahul2 Not sure if you can. But you can hardcode them for testing and remove when you are ready to commit the code.

gurungrahul2:
You mean I can write a test as this ?
`{
			code: 'try {} catch (skipErr) {}',
			options: [
				{
					defaultExport: {
		                                    path: true,
		                                    chalk: true
	                                 }
		                 }
			]
}`

test/no-direct-import.js
gurrrung: > * `const {promisify as foo} = require(\'util\');`
It think it should be `const {promisify: foo} = require(\'util\');`.

gurrrung:
What would having something like
```
defaultExport: {
		mymodule: false
	}
```
mean?
Should `mymodule` be destructed or false means let it be imported anyway?
@sindresorhus @futpib

futpib:
@gurrrung We are basically modelling a set in JSON here. `defaultExport` would be a set of module names for which only the default export should be used. Only value `true` should mean that the module is in the set.

Examples:
* `defaultExport: { util: true }` - `util` is restricted to default export (`util` is in the `defaultExport` set).
* `defaultExport: {}` - `util` is not restricted (`util` is not in the `defaultExport` set)
* `defaultExport: { util: false }` - same as above, `util` is not restricted (`util` is not in the `defaultExport` set)

The benefit of this design is that users can override the build-in defaults. We could merge the user-supplied option with the default one (for example with `defaultsDeep`). So if we had `{ util: true, path: true }` hardcoded, and a user (who wants to lift the restriction off of `util` module) passed `{ util: false }`, the result would be `{ util: false, path: true }`, which result in `util` not being restricted, but all the other hardcoded defaults (here, only `path`) would still apply.

sindresorhus:
@gurrrung Still interested in finishing this?

gurrrung:
Yes, I'm on it.

gurrrung:
Why would Integration tests blow up. :confused:

sindresorhus:
@fisker Would you be able to help review this?

test/import-style.js
```diff
@@ -0,0 +1,88 @@
...
+		{
+			code: 'import {something} from \'path\';',
+			errors: [buildError({moduleName: 'path', type: 'namedImport'})]
+		},
+		{
+			code: 'import util, {promisify} from \'util\';',
+			errors: [buildError({moduleName: 'util', type: 'defaultImport'})]
+		},
+		{
+			code: 'const foo = myFunction(\'util\')',
```
fisker: Why this error?

fisker:
Should we also check `export ... from ...`?

```js
export {chunk} from 'utils';
export * as default from 'utils';
export utils from 'utils';
export {default as default} from 'lodash';
export {default as lodash} from 'lodash';
export * from 'lodash';
```

test/import-style.js
gurrrung: Oh! Thanks for the review. I was a little busy in academic work. Yeah, it shouldn't be error, I'll add the check to throw error only in case of `require()`.

gurrrung:
> What is the result of
> 
> ```js
> const promisify = require('utils').promisify;
> ```
> 
> and dynamic import
> 
> ```js
> (async () => {
>   const lodash = (await import('lodash')).default;
> })();
> 
> (async () => {
>   const {default: lodash} = await import('lodash');
> })();
> 
> (async () => {
>   const {chunk} = await import('lodash');
> })();
> ```
> 
> we should add them to tests
that's brilliant. I really couldn't thing of this. Can you give me more insight like what should be the behaviour?

sindresorhus:
> Can you give me more insight like what should be the behaviour?

@fisker ^

sindresorhus:
@gurrrung Still interested in finishing this?

fisker:
@gurrrung

```js
const promisify = require('utils').promisify;
```

We need add test, make sure this pass.

----

```js
(async () => {
  const lodash = (await import('lodash')).default;
})();

(async () => {
  const {default: lodash} = await import('lodash');
})();

(async () => {
  const {chunk} = await import('lodash');
})();
```

I'm not sure should we support dynamic import?

----

@sindresorhus 

`export ... from` and `import()` support? You should make the call.

I suggest we should support `export ... from`, `import()` maybe a little hard , we can ignore that for now.

sindresorhus:
> I'm not sure should we support dynamic import?

Nah, not worth the effort. For all our rules, we can stay with the top-level imports.

> I suggest we should support export ... from,

Agreed

sindresorhus:
Closing for lack of activity.

PR #233, opened 2019-02-02
Add `no-unnecessary-polyfills` rule

### LINKED ISSUE #36: Rule proposal: `no-unnecessary-polyfills`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/36)

A rule that examines the `engine` argument of `package.json`, and keeps a list of poly/ponyfills and which Node version makes them obsolete.

For instance, if you're using `object-assign`, but only targeting Node 4 and higher, just use native `Object.assign`. Many others.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $120.00)

- [issuehunt issuehunt](https://issuehunt.io/u/issuehunt) ($120.00)

### Submitted pull Requests
- [#233 Add `no-unnecessary-polyfills` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/233)
- [#1717 Add `no-unnecessary-polyfills` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/1717)
---

### Tips

</details>

doniyor2109:
This is initial version of `no-unnecessary-polyfills`.

Resolves #36.

Here some todos to complicate:

 - [x] Read `package.json` from installed project
 - [x] Add more polyfills
 - [x] Add optional engine version via options
 - [x] Support for `imports`

### DIFF
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -1,9 +1,6 @@
 # Enforce the use of built-in methods instead of unnecessary polyfills
 
-Takes node version from `package.json` and warns to use supported built in methods
-instead of using unnecessary polyfills.
-
-The desired name is configurable, but defaults to `error`.
+This rules helps to use existing methods instead of using extra polyfills. 
 
 ## Fail
 
@@ -23,7 +20,7 @@ const objectAssign = require('object-assign');
 ## Pass
 
 package.json
-```json5
+```json
 {
 	"engines": {
 		"node": "<4.0.0"
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -11,7 +11,7 @@ package.json
 ```json
 {
 	"engines": {
-		"node": ">8.0.0"
+		"node": ">=8"
 	}
 }
 
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -34,3 +34,12 @@ package.json
 ```js
 const assign = require("object-assign"); // passes as Object.assign is not supported
 ```
+
+## Options
+
+By default it takes node version from `package.json`. However you can override
+target node version:
+
+```js
+"unicorn/no-unnecessary-polyfills": ["error", {"targetVersion": "6"}]
+```
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -8,7 +8,7 @@ The desired name is configurable, but defaults to `error`.
 ## Fail
 
 package.json
-```json5
+```json
 {
 	"engines": {
 		"node": ">8.0.0"
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -1,4 +1,4 @@
-# Enforce to use built in method instead of unnecessary polyfills
+# Enforce the use of built-in methods instead of unnecessary polyfills
 
 Takes node version from `package.json` and warns to use supported built in methods
 instead of using unnecessary polyfills.
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -1,0 +1,36 @@
+# Enforce to use built in method instead of unnecessary polyfills
+
+Takes node version from `package.json` and warns to use supported built in methods
+instead of using unnecessary polyfills.
+
+The desired name is configurable, but defaults to `error`.
+
+## Fail
+
+package.json
+```json5
+{
+	"engines": {
+		"node": ">8.0.0"
+	}
+}
+
+```
+```js
+const assign = require("object-assign"); // warns to use built in Object.assign
+```
+
+## Pass
+
+package.json
+```json5
+{
+	"engines": {
+		"node": "<4.0.0"
+	}
+}
+
+```
+```js
+const assign = require("object-assign"); // passes as Object.assign is not supported
+```
--- a/docs/rules/no-unnecessary-polyfills.md
+++ b/docs/rules/no-unnecessary-polyfills.md
@@ -17,7 +17,7 @@ package.json
 
 ```
 ```js
-const assign = require("object-assign"); // warns to use built in Object.assign
+const objectAssign = require('object-assign');
 ```
 
 ## Pass
--- a/rules/no-unnecessary-polyfills.js
+++ b/rules/no-unnecessary-polyfills.js
@@ -1,122 +1,115 @@
-'use strict';
-const semver = require('semver');
-const readPkgUp = require('read-pkg-up');
-const builtIns = require('@babel/preset-env/data/built-ins');
-const getDocsUrl = require('./utils/get-docs-url');
+'use strict'
+const semver = require('semver')
+const readPkgUp = require('read-pkg-up')
+const compatTable = require('@babel/preset-env/data/built-ins.json')
+const getDocsUrl = require('./utils/get-docs-url')
+const upperfirst = require('lodash.upperfirst')
+const camelcase = require('lodash.camelcase')
 
 function isRequireCall(node) {
-	return node.callee.name === 'require';
+  return node.callee.name === 'require'
 }
 
 function getVersionFromPkg(cwd) {
-	const pkg = readPkgUp.sync({cwd});
-	return pkg && pkg.pkg.engines && pkg.pkg.engines.node;
+  const pkg = readPkgUp.sync({ cwd })
+  return pkg && pkg.pkg.engines && pkg.pkg.engines.node
 }
 
-const compatTable = Object.keys(builtIns).reduce((current, feature) =>
-	Object.assign(current, {[feature.split('.').slice(1).join('.')]: builtIns[feature]})
-, {});
-
 function isValidVersion(version) {
-	return /^[\d.]+$/.test(version.trim());
+  return /^[\d.]+$/.test(version.trim())
+}
+
+const polyfills = {
+  'es6.promise': [ 'promise-polyfill' ],
+  'es7.promise.finally': [ 'p-finally' ],
 }
 
 const polyfillMap = Object.keys(compatTable).reduce((current, name) => {
-	const polyfills = [];
-	const parts = name.split('.');
-	const allParts = name.split(/[.-]/);
+  let [ ecmaVersion, constructorName, methodName = '' ] = name.split('.')
 
-	function dotted(parts) {
-		return parts.join('.');
-	}
+  constructorName = `(${constructorName}|${camelcase(constructorName)})`
+  methodName = `(${methodName}|${camelcase(methodName)})`
 
-	function dashed(parts) {
-		return parts.join('-');
-	}
+  const prefixes = '(mdn-polyfills|polyfill-|core-js(-pure)?/features/|)'
+  const suffixes = '(-polyfill|)'
+  const delimiter = '(\\.|-|\.prototype\.|/|)'
 
-	function joined(parts) {
-		return parts.join('');
-	}
+  const polyfill = polyfills[ name ]
 
-	function addPrototype(parts) {
-		const clone = parts.slice();
-		clone.splice(1, 0, 'prototype');
-		return clone;
-	}
+  current.push({
+    feature: name,
+    polyfillRegex: new RegExp(`^${prefixes}(` + `${ecmaVersion}${delimiter}${constructorName}${delimiter}${methodName}` + '|' + `${constructorName}${delimiter}${methodName}` + '|' + `${ecmaVersion}${delimiter}${constructorName}` + '|' + methodName +  '|' + (polyfill ? `(${polyfill.join('|')})` : '') + `)${suffixes}$`, 'i'),
+  })
 
-	function allVariants(parts) {
-		return [
-			dotted(parts),
-			joined(parts),
-			dashed(parts),
-			dotted(addPrototype(parts)),
-			dashed(addPrototype(parts)),
-			joined(addPrototype(parts))
-		];
-	}
+  return current
+}, [])
 
-	polyfills.push(name);
+function formatErrorMessage(featureName) {
+  const [ _ecmaVersion, namespace, method ] = featureName.split('.')
 
-	polyfills.push(...allVariants(parts));
-	polyfills.push(...allVariants(allParts));
-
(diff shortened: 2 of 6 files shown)


doniyor2109:
@sindresorhus Should we allow setting node version via options?

sindresorhus:
Yes, default to package.json, but allow overriding.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,122 @@
...
+				node,
+				message: `Use built in ${polyfill.feature}`
+			});
+		}
+	}
+}
+
+const create = context => {
+	const options = context.options[0];
+	const targetVersion = (options && options.targetVersion) || getVersionFromPkg(context.getFilename());
```
bfred-it: What if it isn't defined anywhere? Should this rule be disabled?

doniyor2109: If there is no target version then it simply does not apply rule.

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,45 @@
+# Enforce to use built in method instead of unnecessary polyfills
```
sindresorhus:
```suggestion
# Enforce the use of built-in methods instead of unnecessary polyfills
```

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,45 @@
...
+
+Takes node version from `package.json` and warns to use supported built in methods
+instead of using unnecessary polyfills.
```
sindresorhus: This could be written better.

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,45 @@
...
+
+The desired name is configurable, but defaults to `error`.
```
sindresorhus: ?

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Fail
+
+package.json
+```json5
```
sindresorhus:
```suggestion
```json
```

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,45 @@
...
+{
+	"engines": {
+		"node": ">8.0.0"
```
sindresorhus:
```suggestion
		"node": ">=8"
```

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,45 @@
...
+	}
+}
+
+```
+```js
+const assign = require("object-assign"); // warns to use built in Object.assign
```
sindresorhus:
```suggestion
const objectAssign = require('object-assign');
```

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,122 @@
...
+	return {
+		CallExpression: node => {
+			if (targetVersion && isRequireCall(node)) {
+				const arg0 = node.arguments[0];
```
sindresorhus: There's no value in making this a variable.

test/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,85 @@
...
+		},
+		{
+			code: 'require("array-from")',
+			options: [{targetVersion: '>7 <8'}],
+			errors: [{message: 'Use built in array.from'}]
+		},
+		{
+			code: 'require("array.prototype.every")',
+			options: [{targetVersion: '~4.0.0'}],
+			errors: [{message: 'Use built in array.every'}]
```
sindresorhus: The message should be:

```
Use the built-in `Array#every`.
```

test/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,85 @@
...
+		},
+		{
+			code: 'require("array-find-index")',
+			options: [{targetVersion: '>=4 <5.2 || >6.0.0'}],
+			errors: [{message: 'Use built in array.find-index'}]
+		},
+		{
+			code: 'require("array-find-index")',
+			options: [{targetVersion: '4'}],
+			errors: [{message: 'Use built in array.find-index'}]
```
sindresorhus: `array.find-index` => `Array#findIndex`

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,122 @@
...
+				processRule(context, node, moduleName, targetVersion);
+			}
+		},
+		ImportDeclaration: node => targetVersion && processRule(context, node, node.source.value, targetVersion)
+	};
+};
+
+module.exports = {
+	create,
+	meta: {
```
sindresorhus: Add the type and schema. See:

- https://github.com/sindresorhus/eslint-plugin-unicorn/commit/610a9014ab6a7076eaebf1990a143f15668ccba2
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/247

rules/no-unnecessary-polyfills.js
sindresorhus: You should od the early return for no target version here instead of in the AST walker.

test/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,85 @@
...
+		}
+	]
+});
```
sindresorhus: Can you add test for these polyfills:

- https://www.npmjs.com/package/promise-polyfill
- https://www.npmjs.com/package/es6-symbol
- https://www.npmjs.com/package/p-finally
- https://www.npmjs.com/package/code-point-at
- https://www.npmjs.com/package/util.promisify
- https://www.npmjs.com/package/object.getownpropertydescriptors
- https://www.npmjs.com/package/object.fromentries
- https://www.npmjs.com/package/string.prototype.padstart

sindresorhus:
While your automatic detection is cool, I think you should also go through the most popular polyfills and hard-code them here to ensure we correctly detect those: https://www.npmjs.com/search?q=keywords%3Apolyfill&ranking=popularity (The first few pages)

sindresorhus:

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,122 @@
...
+	return current;
+}, []);
+
+function processRule(context, node, moduleName, targetVersion) {
+	const polyfill = polyfillMap.find(({polyfills}) => polyfills.includes(moduleName));
+
+	if (polyfill) {
+		const feature = compatTable[polyfill.feature];
+		const supportedNodeVersion = semver.valid(semver.coerce(feature.node));
+		const validRangeTargetVersion = semver.validRange(targetVersion).replace('=', '');
```
coreyfarrell: semver 5.7.0 introduces a `minVersion` function.  I think this should be used here:

```js
const validTargetVersion = semver.minVersion(targetVersion);
if (validTargetVersion && semver.lte(supportedNodeVersion, validTargetVersion)) {
  // report error.
}
```

test/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,85 @@
...
+			errors: [{message: 'Use built in object.assign'}]
+		},
+		{
+			code: 'require("object-assign")',
+			options: [{targetVersion: '8'}],
+			errors: [{message: 'Use built in object.assign'}]
+		},
+		{
+			code: 'require("array-from")',
+			options: [{targetVersion: '>7 <8'}],
```
coreyfarrell: This does not actually match any ranges. `>7` is treated as `>=8.0.0` by semver and the range `>=8.0.0 <8` cannot possibly match any version.

I think this should be moved to the `valid` list, `require("array-from")` should be allowed for this range.

I've opened a bug at npm/node-semver#275 to document this edge case of semver ranges.

doniyor2109:
> Can you fix the merge conflict?

Sure. I will resolve all conflicts and feedbacks ASAP (by weekends)

sindresorhus:
@doniyor2109 Ping

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
+const semver = require('semver');
+const readPkgUp = require('read-pkg-up');
+const builtIns = require('@babel/preset-env/data/built-ins');
```
zloirock: It's obsolete data, use `core-js-compat`.

doniyor2109: I could not find node versions in `core-js-compat`

test/no-unnecessary-polyfills.js
doniyor2109: Currently, it works with JS features. I did not figure out where to get node compat libs version for `util.promisify`

rules/no-unnecessary-polyfills.js
zloirock: @doniyor2109 what do you mean? https://www.npmjs.com/package/core-js-compat

test/no-unnecessary-polyfills.js
sindresorhus: You can find compat data about Node.js in its docs and https://github.com/mysticatea/eslint-plugin-node/blob/master/lib/rules/no-unsupported-features/node-builtins.js

> I did not figure out where to get node compat libs version for util.promisify

You don't need a package for this. You can easily find the information yourself.

sindresorhus:
TODO:
- [ ] https://github.com/sindresorhus/eslint-plugin-unicorn/pull/233#discussion_r319752058
- [ ] https://github.com/sindresorhus/eslint-plugin-unicorn/pull/233#issuecomment-478382861
- [ ] https://github.com/sindresorhus/eslint-plugin-unicorn/pull/233#discussion_r270684293

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,111 @@
+'use strict';
+const semver = require('semver');
+const readPkgUp = require('read-pkg-up');
+const compatTable = require('@babel/preset-env/data/built-ins.json');
+const upperfirst = require('lodash.upperfirst');
+const camelcase = require('lodash.camelcase');
+const getDocsUrl = require('./utils/get-docs-url');
+
+function isRequireCall(node) {
```
sindresorhus: This method is misleading as it doesn't actually check it's a method. Better to just inline it where it's used.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,111 @@
...
+function formatErrorMessage(featureName) {
+	const [, namespace, method] = featureName.split('.');
+
+	return `Use the built-in \`${method ? `${upperfirst(namespace)}#${camelcase(method)}` : `${upperfirst(namespace)}`}\`.`;
+}
+
+function processRule(context, node, moduleName, targetVersion) {
+	const polyfill = polyfillMap.find(({polyfillRegex}) => polyfillRegex.test(moduleName));
+
+	if (polyfill) {
```
sindresorhus: Do an early-return instead.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,111 @@
...
+	const options = context.options[0];
+	const targetVersion = (options && options.targetVersion) || getVersionFromPkg(context.getFilename());
+
+	if (!targetVersion) {
+		return {};
+	}
+
+	return {
+		CallExpression: node => {
+			if (isRequireCall(node)) {
```
sindresorhus: Do an early-return instead.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,111 @@
...
+}];
+
+module.exports = {
+	create,
+	meta: {
+		schema,
+		type: 'suggestion',
+		docs: {
+			url: getDocsUrl(__filename)
+		}
```
sindresorhus: Follow the order used in other rules. You're also missing the

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,42 @@
...
+const assign = require("object-assign"); // passes as Object.assign is not supported
+```
+
+## Options
+
+By default it takes node version from `package.json`. However you can override
+target node version:
+
+```js
+"unicorn/no-unnecessary-polyfills": ["error", {"targetVersion": "6"}]
```
sindresorhus: I think the option should be called `nodeVersion` and you need to document that it can be a semver range. The example here is also wrong as `6` will only match exactly Node.js `6.0.0`, not `6.0.1`.

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,42 @@
...
+```json
+{
+	"engines": {
+		"node": "<4.0.0"
+	}
+}
+
+```
+```js
+const assign = require("object-assign"); // passes as Object.assign is not supported
```
sindresorhus: Single-quotes.

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,42 @@
...
+const objectAssign = require('object-assign');
+```
+
+## Pass
+
+package.json
+```json
+{
+	"engines": {
+		"node": "<4.0.0"
```
sindresorhus: Nobody would use a semver range like this here.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,111 @@
...
+	const pkg = readPkgUp.sync({cwd});
+	return pkg && pkg.pkg.engines && pkg.pkg.engines.node;
+}
+
+const polyfills = {
+	'es6.promise': ['promise-polyfill'],
+	'es7.promise.finally': ['p-finally']
+};
+
+const polyfillMap = Object.keys(compatTable).reduce((current, name) => {
```
sindresorhus: Just use a `for-of` loop. `reduce()` just makes the code harder to read.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,113 @@
...
+};
+
+const schema = [
+	{
+		type: 'object',
+		properties: {
+			targetVersion: {
+				type: 'string'
+			}
+		}
```
sindresorhus: You need to set `additionalProperties: false`.

sindresorhus:
Closing for lack of activity. Better let someone else give it a shot: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/36

rules/no-unnecessary-polyfills.js
omril1: or `.map()`

PR #234, opened 2019-02-02
Add check that ensures new rules have docs and tests

doniyor2109:
Checks every rule has test and documentation


sindresorhus:
Good idea! 👌

PR #235, opened 2019-02-02
 Add `excludedPackages` option to the `prefer-add-event-listener` rule

### LINKED ISSUE #174: The `prefer-add-event-listener` rule should not apply to Koa's `app.context.onerror`

medusalix:
Resolves #174


rules/prefer-add-event-listener.js
```diff
@@ -2,6 +2,8 @@
 const getDocsUrl = require('./utils/get-docs-url');
 const domEventsJson = require('./utils/dom-events.json');
 
+const excludedPackages = new Set(['koa', 'sax']);
```
futpib: Should we add an option to override or extend this set?

rules/prefer-add-event-listener.js
```diff
@@ -72,11 +75,31 @@ const create = context => {
 			codePathInfo = codePathInfo.upper;
 		},
 
+		Program() {
+			isDisabled = false;
```
futpib: Why not just initialize the variable to `false`?

medusalix: Oh, I thought those variables were bound to the rule, so I wanted to reset `isDisabled` at the start of a new file 🤦‍♂️. That's what reviews are for 😄.

rules/prefer-add-event-listener.js
medusalix: Yeah, we could add a configuration option, but I think we should keep these values as a default if no options are specified.

sindresorhus:
@medusalix I think it should be an options-object, so we're able to add additional options in the future:

```js
'prefer-add-event-listener': ['error', {
	excludedPackages: ['foo']
}]
```

sindresorhus:
It also needs to be documented.

sindresorhus:
@futpib Looks good to you?

PR #237, opened 2019-02-03
Add `prevent-abbreviations` rule

### LINKED ISSUE #169: Rule proposal: Prevent abbreviations
Using complete words result in more readable code. Not everyone knows all your abbreviations. You only write code once, but it's read many times.

For a start, we could go with these:

- `err` => `error`
- `cb` => `callback`
- `opts` => `options`
- `str` => `string`
- `obj` => `object`
- `num` => `number`
- `val` => `value`
- `e` => `event`/`error`
- `evt` => `event`
- `el` => `element`
- `req` => `request`
- `res` => `response`/`result`
- `btn` => `button`
- `msg` => `message`
- `len` => `length`
- `env` => `environment`
- `dev` => `development`
- `prod` => `production`
- `tmp` => `temporary`
- `arg` => `argument`
- `args` => `arguments`
- `tbl` => `table`
- `db` => `database`
- `ctx` => `context`
- `mod` => `module`

Also `i` => `index`, but I have a feeling, that it's going to be too controversial.

Suggestions welcome for additional ones. Also looking for people's thoughts on this rule.

futpib:
Fixes #169


docs/rules/prevent-abbreviations.md
```diff
@@ -0,0 +1,75 @@
+# Prevent abbreviations
+
+Using complete words result in more readable code. Not everyone knows all your abbreviations. You only write code once, but it's read many times.
```
alexzherdev:
```suggestion
Using complete words results in more readable code. Not everyone knows all your abbreviations. You only write code once, but it's read many times.
```

docs/rules/prevent-abbreviations.md
```diff
@@ -0,0 +1,75 @@
...
+
+Default replacements are available [here](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/prevent-abbreviations.js#L7).
+
+This rule is fixable for unambiguous abbreviations (that is abbreviations with exactly one replacement defined).
```
alexzherdev:
```suggestion
This rule is fixable for unambiguous abbreviations (that is, abbreviations with exactly one replacement defined).
```

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,244 @@
...
+};
+
+const getVariableReplacements = (replacements, variable) => {
+	const variableNameReplacements = replacements.get(variable.name);
+
+	if (!variableNameReplacements) {
+		return [];
+	}
+
+	return [...variableNameReplacements.keys()]
```
alexzherdev: Just curious, is `keys()` not an array here?

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,244 @@
...
+		const problem = {
+			node: definition.name,
+			message: formatMessage(definition.name, variableReplacements)
+		};
+
+		if (variableReplacements.length === 1) {
+			const [replacement] = variableReplacements;
+			const captureAvoidingReplacement = avoidCapture(replacement, scopes);
+
+			problem.fix = fixer => [].concat(
```
alexzherdev: Is `[].concat` preferable to spreading out the two arrays into one?

rules/prevent-abbreviations.js
futpib: No, it's an iterator: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/keys

sindresorhus:
I tried running this on the [AVA](https://github.com/avajs/ava) codebase.

```
❯ xo api.js
can't resolve reference #/definitions/abbreviations from id #
```

```
❯ xo profile.js
can't resolve reference #/definitions/abbreviations from id #
AssertionError [ERR_ASSERTION]: Fix objects must not be overlapped in a report.
    at mergeFixes (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/util/report-translator.js:150:9)
    at normalizeFixes (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/util/report-translator.js:180:16)
    at args (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/util/report-translator.js:278:18)
    at Object.report (/Users/sindresorhus/dev/oss/xo/node_modules/eslint/lib/linter.js:720:41)
    at checkVariable (/Users/sindresorhus/dev/oss/eslint-plugin-unicorn/rules/prevent-abbreviations.js:192:11)
    at checkVariables (/Users/sindresorhus/dev/oss/eslint-plugin-unicorn/rules/prevent-abbreviations.js:196:19)
    at checkScope (/Users/sindresorhus/dev/oss/eslint-plugin-unicorn/rules/prevent-abbreviations.js:204:3)
    at checkChildScopes (/Users/sindresorhus/dev/oss/eslint-plugin-unicorn/rules/prevent-abbreviations.js:200:21)
```

rules/utils/avoid-capture.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+const resolveVariableName = require('./resolve-variable-name');
+
+const indexifyName = (name, index) => index === 1 ? name : name + index;
+
+const someScopeHasVariableName = (name, scopes) => scopes.some(scope => resolveVariableName(name, scope));
+
+module.exports = (name, scopes) => {
```
sindresorhus: Can you add doc comment on what this does and what problem it solves? For posterity.

rules/utils/resolve-variable-name.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = (name, scope) => {
```
sindresorhus: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237/files#r256275987

futpib:
I wonder why this did not come up in the integration tests run.

futpib:
Fixed the integration test and the error.

@sindresorhus Should I cherry-pick the integration test fix into a separate PR?

sindresorhus:
In AVA:

```
~/dev/oss/ava
❯ xo lib/fork.js --fix

  lib/fork.js:53:8
  ✖  53:8  Parsing error: Binding arguments in strict mode

  1 error
```

sindresorhus:
In `~/dev/oss/ava/bench/run.js`, it did the following auto-fix, which is not safe:

screen shot 2019-02-15 at 14 39 51

It should instead turn `err,` into `err: error,` and report about the `err,` property so I can manually fix it.

sindresorhus:
I would recommend running `xo --fix` in the AVA project and then run `npm run test:tap` to ensure the tests still pass. That's a good way to validate the changes.

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,255 @@
...
+	db: {
+		database: true
+	},
+	ctx: {
+		context: true
+	},
+	mod: {
+		module: true
+	}
+};
```
sindresorhus: What do you think about including `prop` => `property`?

sindresorhus:
Should we also report on properties (obviously not auto-fix them)? For example, `foo.err` should be `foo.error`.

Hmm, or maybe we should only report when creating objects and not using them?

```js
const foo = {
	err: new Error();
};
```

sindresorhus:
Instead of converting, for example, `mod` to `module2`, because `module` is taken, how do you feel about doing `mod` => `module_`?

rules/prevent-abbreviations.js
futpib: :ok_hand:

futpib:
> Should we also report on properties (obviously not auto-fix them)? For example, `foo.err` should be `foo.error`.
> Hmm, or maybe we should only report when creating objects and not using them?

Unlike variable, property names can be a part of an interface, so users may be forced to use certain property names by third party code. We could add another rule for property/method names and set it to `warn`.

> Instead of converting, for example, `mod` to `module2`, because `module` is taken, how do you feel about doing `mod` => `module_`?

I prefer `module_`, and that's what I do usually write manually when renaming a variable just to prevent shadowing another variable.

What should the new general renaming scheme be?
* `module_`
* `module__`
* `module___`

Also, if we go with `_`s, we better update catch-error-name too for consistency https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6f7869846371ea32053b5a2a5922087db8326298/test/catch-error-name.js#L76-L80

test/prevent-abbreviations.js
```diff
@@ -0,0 +1,400 @@
...
+				return g.apply(this, arguments) + args;
+			}
+		`,
+
+		// Renaming would change the API of the module
+		'export const err = {}',
+		'export let err = {}',
+		'export var err = {}',
+		'export function err() {}',
+		'export class err {}'
```
futpib: @sindresorhus I'm not sure what to do with these cases.

* Fix to this:
```js
function error() {}
export {error as err}
```
* Not report
* Report but not fix

What do you think?

sindresorhus:
> Unlike variable, property names can be a part of an interface, so users may be forced to use certain property names by third party code. We could add another rule for property/method names and set it to `warn`.

You could argue the same about the `camelcase` rule in ESLint though. The fix there is to use a disable-comment. I find `warn` pretty useless as it's too easy to ignore and won't be visible when run in a CI. How about we add an option of whether to check property/method names (`true` by default)? And obviously not auto-fix it.

sindresorhus:
> What should the new general renaming scheme be?
> * `module_`
> * `module__`
> * `module___`

👍 I can't imagine more than one `_` coming up in practice so seems fine.

> Also, if we go with `_`s, we better update catch-error-name too for consistency

👍

test/prevent-abbreviations.js
sindresorhus: > * Report but not fix

👍

test/integration/test.js
```diff
@@ -15,7 +15,6 @@ const packages = new Map([
 	['np', 'https://github.com/sindresorhus/np'],
 	['ora', 'https://github.com/sindresorhus/ora'],
 	['p-map', 'https://github.com/sindresorhus/p-map'],
-	['detect-indent', 'https://github.com/sindresorhus/detect-indent'],
```
sindresorhus: Why remove this?

futpib: Fixtures like this one cause parse errors when integrating test runs eslint https://github.com/sindresorhus/detect-indent/blob/master/fixture/fifty-fifty-space-first.js

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,369 @@
...
+		scopes.some(scope => scope.isStrict);
+};
+
+const anotherNameMessage = 'A more descriptive name will do too.';
+
+const formatMessage = (discouragedName, replacements) => {
+	const message = [];
+
+	if (replacements.length === 1) {
+		message.push(`This variable should be named \`${replacements[0]}\`.`);
```
sindresorhus: I think we should mention the original variable name to make it faster to understand what needs changing.

How about this?

```
The variable `evt` should be named `event`.
```

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,369 @@
...
+	} else {
+		const replacementsText = replacements
+			.map(replacement => `\`${replacement}\``)
+			.join(', ');
+
+		message.push('Please rename this variable.');
```
sindresorhus: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237/files#r257567522

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,369 @@
...
+	};
+
+	return {
+		'Program:exit'() {
+			checkScope(context.getScope());
+		}
+	};
+};
+
+const schema = [{
```
sindresorhus: If you upgrade to latest ESLint, we can now set default values in the schema. Not sure whether that's useful here. Just wanted to mention it. https://github.com/sindresorhus/eslint-plugin-unicorn/issues/247

docs/rules/prevent-abbreviations.md
```diff
@@ -0,0 +1,81 @@
...
+			"cmd": {
+				"command": true
+			},
+			"e": {
+				"event": false
+			}
+		}
+	}
+]
+```
```
sindresorhus: @futpib What do you think about starting to use JS config syntax for our examples? The JSON looks so ugly, and most people use JS config anyway.

docs/rules/prevent-abbreviations.md
```diff
@@ -0,0 +1,81 @@
+# Prevent abbreviations
+
+Using complete words results in more readable code. Not everyone knows all your abbreviations. You only write code once, but it's read many times.
+
+Default replacements are available [here](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/prevent-abbreviations.js#L7).
+
+This rule is fixable for abbreviations with exactly one replacement defined.
+
+
```
sindresorhus: Can you add a short mention that this rule can be useful for more than just preventing abbreviations if they use the `extendDefaults` option. Could be used by companies to replace terms, disallowed words, etc.

sindresorhus:
I have cherry-picked https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237/commits/2d361f76762977bc29047710c3f5c2fcc6cf0f0e into `master`.

rules/prevent-abbreviations.js
futpib: We'll have to up the eslint version in `peerDependencies`, let's not do this in this PR.

futpib:
@sindresorhus Should I add special support for PascalCased identifiers? Like if you define `err` => `error` replacement, you get `Err` => `Error` working automatically? Or should users have to define such replacements explicitly (no changes to the code required)?

sindresorhus:
> Should I add special support for PascalCased identifiers? Like if you define `err` => `error` replacement, you get `Err` => `Error` working automatically?

👍 Yes, that would be useful.

docs/rules/prevent-abbreviations.md
futpib: Ok, but can we defer this to a separate issue too? This would require editing other rules docs.

sindresorhus:
@futpib Is this done?

futpib:
@sindresorhus Yes, I think so.

sindresorhus:
I spent some time trying this out on various project and going through the source of popular libraries to see what kind of abbreviations are common. I think we should add these too:

arr => array
ret => returnValue
retval => returnValue
retVal => returnValue
ext => extension
exts => extensions
lib => library
dir => directory
dirs => directories
ref => reference
refs => references
pkg => package_
sep => separator
doc => document_
docs => documents
elem => element
src => source
dest => destination
prev => previous
rel => relative / related
conf => config

It's important to add these now as after the next release, we would have to do a major release every time we add another abbreviation.

sindresorhus:
We have also not considered whether we should check part of a variable. For example, I found a lot of variables where part of it was `dir`:

cacheDir => cacheDirectory
tmpDir => tempDirectory
tempDir => tempDirectory
projectDir => projectDirectory
baseDir => baseDirectory
snapshotDir => snapshotDirectory
testDir => testDirectory
contextRef => contextReference

Same applies to `fooVal` and `fooProp`, `fooObj`, and some others.

How should we handle that?

---

If you feel it's out of scope for this PR, I can open an issue instead. Just thought I would mention it.

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,486 @@
...
+	},
+	ctx: {
+		context: true
+	},
+	mod: {
+		module: true
+	},
+	prop: {
+		property: true
+	}
```
sindresorhus: @futpib We should probably add `props` here too? My concern is that it would cause a lot of error when using React, but I guess React users could disable it. Thoughts?

rules/prevent-abbreviations.js
```diff
@@ -0,0 +1,486 @@
...
+const getDocsUrl = require('./utils/get-docs-url');
+const avoidCapture = require('./utils/avoid-capture');
+
+const pascalCase = str => upperfirst(camelCase(str));
+const isPascalCase = str => str === pascalCase(str);
+
+const defaultReplacements = {
+	err: {
+		error: true
+	},
```
sindresorhus: We should probably support `errs` => `errors` too.

rules/prevent-abbreviations.js
futpib: It would come up in React code, but would mostly be fixable (will mostly report variable names), since you never assign directly to `this.props` and rarely have a prop named `props` (you may want that if you define a component that accepts another component as a prop, like in [redux-form](https://redux-form.com/8.1.0/docs/api/field.md/#-code-props-object-code-optional-)). But the variable name fixing would result in a IMO weird-looking (but correct) code like this:

```js
const Foo = properies => <Bar><Buz {...properies} /></Bar>;
```
```js
class Foo extends Component {
  constructor(properies) {
    super(properies);
    this.state = { /* ... */ };
  }
}
```

But, if we implement your `fooProp` idea https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237#issuecomment-467748063, we will report [`getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) (and some other methods for older React versions). This comes back to the "part of the interface" point from https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237#issuecomment-464188799

When both React and Vue use "props", how bad really is it? It's almost like a separate word now (or is a word if we consider meanings other than "properties").

But I don't really have an opinion on this. We could ask in the original issue #169 

`attr`/`attrs` also comes to mind.

futpib:
> dir => directory
Also direction

> ref => reference
May also cause problems with React

> rel => relative / related
Also relationship (as in html's `a[rel]`, `link[rel]`)

sindresorhus:
> Also direction

👍 

> May also cause problems with React

Let's try it and see.

> Also relationship (as in html's `a[rel]`, `link[rel]`)

👍

rules/prevent-abbreviations.js
sindresorhus: > But, if we implement your `fooProp` idea [#237 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237#issuecomment-467748063), we will report [`getDerivedStateFromProps`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) and `defaultProps` (and some other methods for older React versions).

Yeah, that's a problem. We should probably only implement that for variables.

> When both React and Vue use "props", how bad really is it? It's almost like a separate word now (or is a word if we consider meanings other than "properties").

Couldn't React/Vue projects just disable it though? I would personally prefer using `property` and `properties` in non-React code. React/Vue code needs to adjust other ESLint rules anyway.

> `attr`/`attrs` also comes to mind.

Yes, we should have these. I don't think that conflicts with anything.

futpib: > We should probably only implement that for variables.

I'm a little hesitant about this. If we allow abbreviated property name, but not variable name, on every destructuring you have to convert from abbreviated to full name, and generally keep both names in mind.

futpib:
`dev` also came up in `stdDev` in ava, I also came up with `device`. Are `deviation` and `device` worth adding?

docs/rules/prevent-abbreviations.md
futpib: I use `xo` key in `package.json` for configuration though, so it's in json.

sindresorhus:
When it's just `dev`, it always means `development`, so I don't think we should add more possible meanings for that when it's not combined. We could add a replacement for `stdDev` => `standardDeviation`. Or just add a hard-coded rule to skip `dev` replacement when it's `stdDev`.

futpib:
@sindresorhus This is now complete with wordwise replacements and `whitelist` option.

sindresorhus:
```
lib/fork.js:12:36
  ✖    12:36   The property NODE_ENV should be named nodeEnvironment. A more descriptive name will do too.                                                                         unicorn/prevent-abbreviations

```

A lot of Node.js devs are going to hit this one. Can you add `NODE_ENV` to the whitelist? Or maybe ignore anything on `process.env` as it's going to be upper-snake-case.

futpib:
@sindresorhus  I'd rather ignore uppercased names, that is what `camelcase` rule does. Are you fine with this?

sindresorhus:
Yes

sindresorhus:
Noticed one more thing, it doesn't preserve prefixed `_` in the suggestion:

```
  lib/test.js:105:2
  ✖   105:2    The property _throwsArgStart should be named throwsArgumentStart. A more descriptive name will do too.                                                              unicorn/prevent-abbreviations
```

The suggestion there should be `_throwsArgumentStart`.

sindresorhus:
> Also `direction`

Can we drop this? It causes so much code to not be auto-fixable, while I have personally never seen `dir` being used as a abbreviation for `direction`. I'd rather be pragmatic about it.

sindresorhus:
Should we really rename default imports? Yes, it's more readable, but also weird that it doesn't match the package name. Thoughts?

screen shot 2019-03-08 at 21 12 31

futpib:
I'm really not sure about any of this.

Default import does not seem much different from named import. This renaming is equally weird to me:
```diff
-import { prop } from 'ramda';
+import { prop as property } from 'ramda';
```

Which is analogous to an object destructuring

```diff
-const { prop } = require('ramda');
+const { prop: property } = require('ramda');
```

or 

```diff
-const { prop } = foo;
+const { prop: property } = foo;
```

So I think if default import is to be allowed, named imports and variable names that are defined in object destructuring should be allowed too.

In case abbreviated property name came from the same project that is being linted, it will be reported wherever the object is created (or property assigned, or name exported).

I'm not sure which way it should be, but I think rule should be the same for all cases I mentioned (default import, named import and object destructuring).

sindresorhus:
I agree it should be the same for all kinds of imports/requires.

Since we already have the logic for handling imports, it would be too bad to throw it away, but I also don't really think it's a good default. So how about we put it behind an option that is off by default?

futpib:
> I agree it should be the same for all kinds of imports/requires.

Do you agree about destructuring in general? That is, by default these should pass too:

```js
const {prop} = foo;
```
```js
const f = ({prop}) => prop;
```

sindresorhus:
Yes, they should instead be renamed at the source, if possible.

futpib:
@sindresorhus I'd like to change the rules around `arguments` to be the same as with taken variable names and reserved words. That is instead of allowing `args`, report it and rename it to `arguments_`. What do you think?

sindresorhus:
> That is instead of allowing args, report it and rename it to arguments_. What do you think?

👍

futpib:
Done, once again :laughing:

sindresorhus:
And now it's perfect! 👌

Thanks for your very hard work and perseverance on this one, @futpib. The result is fantastic.

leo-celebrate

sindresorhus:
@futpib You have to submit the PR URL to https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/169 to claim the bounty ;)

PR #241, opened 2019-02-07
Add `meta.type` to rules

### LINKED ISSUE #202: Support the new rule `meta.type` property
https://eslint.org/blog/2018/11/eslint-v5.9.0-released#the-fix-type-option
https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

stroncium:
- Added `meta.type` to rules
- Added test to check if fixable rules got types
- Updated new-rule.md

Fixes #202

### DIFF
--- a/docs/new-rule.md
+++ b/docs/new-rule.md
@@ -15,7 +15,7 @@ Use the [`astexplorer` site](https://astexplorer.net) with the `espree` parser a
 
 - Go to the `test` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then write some tests before starting to implement the rule.
 - Go to the `rules` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then start implementing the new rule logic.
-- Add `meta.type` to rule as per [ESLint docs](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics).
+- Add the correct [`meta.type`](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics) to the rule.
 - Go to the `docs/rules` directory and duplicate the `no-hex-escape.md` file and rename it to the name of your rule. Then write some documentation.
 - Add the rule to the *bottom* of:
 	- [The recommended config](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/352ba4a0291f9210ca5c8e2e61c7e3ad14028e77/index.js#L19)
--- a/docs/new-rule.md
+++ b/docs/new-rule.md
@@ -15,7 +15,7 @@ Use the [`astexplorer` site](https://astexplorer.net) with the `espree` parser a
 
 - Go to the `test` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then write some tests before starting to implement the rule.
 - Go to the `rules` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then start implementing the new rule logic.
-- Add `meta.type` to rule.
+- Add `meta.type` to rule as per [ESLint docs](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics).
 - Go to the `docs/rules` directory and duplicate the `no-hex-escape.md` file and rename it to the name of your rule. Then write some documentation.
 - Add the rule to the *bottom* of:
 	- [The recommended config](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/352ba4a0291f9210ca5c8e2e61c7e3ad14028e77/index.js#L19)
--- a/docs/new-rule.md
+++ b/docs/new-rule.md
@@ -15,6 +15,10 @@ Use the [`astexplorer` site](https://astexplorer.net) with the `espree` parser a
 
 - Go to the `test` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then write some tests before starting to implement the rule.
 - Go to the `rules` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then start implementing the new rule logic.
+- Don't forget to add `meta.type` to rule(especially if it's fixable). Possible types are:
+  - `problem` - the rule flags a potential problem with your code.
+  - `suggestion` - the rule flags something that could be done in a better or preferred way.
+  - `layout` - the rule flags changes limited to whitespace, semicolons, commas, or parentheses.
 - Go to the `docs/rules` directory and duplicate the `no-hex-escape.md` file and rename it to the name of your rule. Then write some documentation.
 - Add the rule to the *bottom* of:
 	- [The recommended config](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/352ba4a0291f9210ca5c8e2e61c7e3ad14028e77/index.js#L19)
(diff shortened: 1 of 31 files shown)


stroncium:
Oh, by the way, I've made quite a lot of pull requests, but I still plan to do at least some minor refactoring after functionality is verified. I also plan to finish the remaining tasks from issuehunt. I've looked around the code and it seems the style goes left and right in every single piece of code. So, is there some official guidelines/request regarding that?

I myself prefer to write compact code with minimal comments, though some complex parts might get hard to understand from a glance(we're talking AST manipulations here after all), but if needed I can tweak the style the way maintainer(s?) seem fitting.

test/package.js
```diff
@@ -21,3 +21,21 @@ test('Every rule is defined in index file', async t => {
...
+
+	const files = await pify(fs.readdir)('rules');
+	const ruleFiles = files.filter(file => path.extname(file) === '.js');
+
+	for (const file of ruleFiles) {
+		const name = path.basename(file, '.js');
+		const rule = index.rules[name];
+
+		t.true(rule.meta !== null, `${name} got no meta`);
+		if (rule.meta.fixable) {
```
sindresorhus: I think we should require `meta.type` for all rules, even if they're not fixable. They might become fixable in the future and we might also use the data for other things, like categorization.

test/package.js
```diff
@@ -21,3 +21,21 @@ test('Every rule is defined in index file', async t => {
 	t.is(Object.keys(index.configs.recommended.rules).length, ruleFiles.length,
 		'There are more exported rules in the recommended config than rule files.');
 });
+
+test('Every fixable rule got valid meta.type', async t => {
+	const validTypes = ['problem', 'suggestion', 'layout'];
+
+	const files = await pify(fs.readdir)('rules');
+	const ruleFiles = files.filter(file => path.extname(file) === '.js');
```
sindresorhus: This is duplicated from the other rule and could just be moved to the top-level of the file and reused.

test/package.js
```diff
@@ -21,3 +21,21 @@ test('Every rule is defined in index file', async t => {
 	t.is(Object.keys(index.configs.recommended.rules).length, ruleFiles.length,
 		'There are more exported rules in the recommended config than rule files.');
 });
+
+test('Every fixable rule got valid meta.type', async t => {
```
sindresorhus: `got` => `has`

test/package.js
```diff
@@ -21,3 +21,21 @@ test('Every rule is defined in index file', async t => {
...
+	const validTypes = ['problem', 'suggestion', 'layout'];
+
+	const files = await pify(fs.readdir)('rules');
+	const ruleFiles = files.filter(file => path.extname(file) === '.js');
+
+	for (const file of ruleFiles) {
+		const name = path.basename(file, '.js');
+		const rule = index.rules[name];
+
+		t.true(rule.meta !== null, `${name} got no meta`);
```
sindresorhus: I don't think it's correct to check `null`, unless I'm mistaken, it would be `undefined` when not specified.

test/package.js
```diff
@@ -21,3 +21,21 @@ test('Every rule is defined in index file', async t => {
...
+		if (rule.meta.fixable) {
+			t.true(rule.meta.type !== null, `${name} got no meta.type`);
```
sindresorhus: `got` => `has`

And same here with the `null` check

docs/new-rule.md
```diff
@@ -15,6 +15,10 @@ Use the [`astexplorer` site](https://astexplorer.net) with the `espree` parser a
 
 - Go to the `test` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then write some tests before starting to implement the rule.
 - Go to the `rules` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then start implementing the new rule logic.
+- Don't forget to add `meta.type` to rule(especially if it's fixable). Possible types are:
+  - `problem` - the rule flags a potential problem with your code.
+  - `suggestion` - the rule flags something that could be done in a better or preferred way.
+  - `layout` - the rule flags changes limited to whitespace, semicolons, commas, or parentheses.
```
sindresorhus: We don't have to document this. Just link to the ESLint docs: https://eslint.org/docs/developer-guide/working-with-rules#rule-basics

sindresorhus:
> I've looked around the code and it seems the style goes left and right in every single piece of code. So, is there some official guidelines/request regarding that?

Not exactly sure what you mean. Can you link to some examples of what you're referring to? "Official guidelines" for what exactly?

test/package.js
stroncium: Generally speaking, meta in rule seems to be mandatory, at least from the docs, so even a check for `undefined` shouldn't be needed as it will fail later anyhow. However, this guards us from reading property of `null` on the next line and outputs proper error instead of gibberish.

stroncium:
> Not exactly sure what you mean. Can you link to some examples of what you're referring to? "Official guidelines" for what exactly?

Comment density, code illustrative side(in a sense of being a good guide for new coming contributors), idiomatic vs readable vs expressive vs fast code, levels of DRY. But it's ok if there is no premade answers to those questions, a couple of reviews will be enough.

test/package.js
stroncium: Though I can remove it of course, it's that accessing by possibly null references makes me itchy.

docs/new-rule.md
```diff
@@ -15,10 +15,7 @@ Use the [`astexplorer` site](https://astexplorer.net) with the `espree` parser a
 
 - Go to the `test` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then write some tests before starting to implement the rule.
 - Go to the `rules` directory and duplicate the `no-hex-escape.js` file and rename it to the name of your rule. Then start implementing the new rule logic.
-- Don't forget to add `meta.type` to rule(especially if it's fixable). Possible types are:
-  - `problem` - the rule flags a potential problem with your code.
-  - `suggestion` - the rule flags something that could be done in a better or preferred way.
-  - `layout` - the rule flags changes limited to whitespace, semicolons, commas, or parentheses.
+- Add `meta.type` to rule.
```
sindresorhus: As commented in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/241#discussion_r254770269, this should link to the ESLint docs.

test/package.js
```diff
@@ -21,3 +25,16 @@ test('Every rule is defined in index file', async t => {
...
+
+test('Every rule has valid meta.type', t => {
+	const validTypes = ['problem', 'suggestion', 'layout'];
+
+	for (const file of ruleFiles) {
+		const name = path.basename(file, '.js');
+		const rule = index.rules[name];
+
+		t.true(rule.meta !== null && rule.meta !== undefined, `${name} has no meta`);
+		t.true(rule.meta.type !== null, `${name} has no meta.type`);
```
sindresorhus: This should also test again `undefined` too.

sindresorhus: Actually, it would be better to reverse it, and test against the expected value instead:
```suggestion
		t.is(typeof rule.meta.type, 'string', `${name} has no meta.type`);
```

Same with the above.

stroncium: Oh yeah, sorry, linter played a a bad joke with men. As of testing for type, it's tested to be one of values in the next line anyhow, (which includes same type).

docs/new-rule.md
stroncium: Ok, added.
Though at this point it is getting a bit stupid, as this link is the first line of the whole document. I intended this to be a quick helper to returning contributors who didn't know about the new meta.type property so they can fix code without going anywhere, and now it is just a duplicate of part of docs with a duplicate of link to said docs.

docs/new-rule.md
stroncium: Ok, added.
Though at this point it is getting a bit stupid, as this link is the first line of the whole document. I intended this to be a quick helper to returning contributors who didn't know about the new meta.type property so they can fix code without going anywhere, and now it is just a duplicate of part of docs with a duplicate of link to said docs.

PR #242, opened 2019-02-07
Add `prefer-event-key` rule, Fixes #118

### LINKED ISSUE #118: Prefer `event.key` over `event.keyCode`

stroncium:

### DIFF
--- a/docs/rules/prefer-event-key.md
+++ b/docs/rules/prefer-event-key.md
@@ -1,0 +1,28 @@
+# Prefer `event.key` over `event.keyCode`
+
+Enforces the use of `event.key` over `event.keyCode` for clarity.
+
+
+## Fail
+
+```js
+foo.addEventListener('keydown', event => {
+  if (event.keyCode === 27) {
+  }
+});
+
+foo.addEventListener('keydown', event => {
+  if (event.keyCode !== 27) {
+  }
+});
+```
+
+
+## Pass
+
+```js
+foo.addEventListener('keydown', event => {
+  if (event.key === 'Escape') {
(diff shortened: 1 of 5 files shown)


ankeetmaini:
There's already a PR open for this #226

sindresorhus:
@stroncium #226 was first, so I went with that one, but I would really appreciate it if you could take a look and see if there's anything that could be improved.

PR #243, opened 2019-02-07
Add `prefer-better-name` Solves #169

stroncium:
I took a freedom to rethink the concept a bit to make the rule more general and usable by more people. It should also help testing various rules before finalizing default rule sets.

Initial rules implemented in `default` rule set, community suggestions implemented in `extended` rule set, but it is also fully configurable for highly specific and differently opinionated use cases.

I also used the name `prefer-better-name` as it seems to suit the rule better.
However, if needed the rule can be reverted to original specification in #169 by changing a couple of names.

### DIFF
--- a/docs/rules/prefer-better-name.md
+++ b/docs/rules/prefer-better-name.md
@@ -9,7 +9,7 @@ Configuration is done by 2 options `baseRuleset` and `rules`:
 ```js
 {
   "unicorn/prefer-better-name": ["error", {
-    "baseRuleset": "default", // optional, possible values are `default` and `extended`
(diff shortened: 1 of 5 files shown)


sindresorhus:
Closing this in favor of #237 (which was already open when this PR was submitted). Sorry for your wasted time on this.

@stroncium I would encourage you to open an issue if you see anything that could be improved or should be changed with the `prevent-abbreviation` rule ;)

stroncium:
@sindresorhus In my case, I use it to actually enforce abbreviations, so in terms of `prevent-abbreviation` I have nothing to add. =D

PR #244, opened 2019-02-11
Add `no-zero-fractions` rule

### LINKED ISSUE #207: Rule proposal: No zero fractions
There are no difference between `1` and `1.0` in JS. For consistency, the former should be preferred.

## Fail

```js
console.log(1.0 + 5.0);
console.log(3.00000000);
```

## Pass

```js
console.log(1 + 5);
console.log(3);
console.log(7.0000001);
```

stroncium:
Added removing dangling dots too, as there is also no difference in JS between `1.` and `1`.

(Sorry, while rebasing to upstream I managed to trigger some weird github rule, which automatically closed previous pull request.)

---

Fixes #207

### DIFF
--- a/docs/rules/no-zero-fractions.md
+++ b/docs/rules/no-zero-fractions.md
@@ -1,6 +1,8 @@
 # Disallow number literals with zero fractions or dangling dots
 
-Disallows number literals with zero fractions or dangling dots, as there is no difference in JS between one having it and one not having it.
+There is no difference in JS between e.g. `1`, `1.0` and `1.`, so prefer the former for consistency.
+
+This rule is fixable.
 
 ## Fail
 
--- a/docs/rules/no-zero-fractions.md
+++ b/docs/rules/no-zero-fractions.md
@@ -1,6 +1,6 @@
 # Disallow number literals with zero fractions or dangling dots
 
-There is no difference in JS between e.g. `1`, `1.0` and `1.`, so prefer the former for consistency.
+There is no difference in JavaScript between, for example, `1`, `1.0` and `1.`, so prefer the former for consistency.
 
 This rule is fixable.
 
--- a/docs/rules/no-zero-fractions.md
+++ b/docs/rules/no-zero-fractions.md
@@ -1,0 +1,25 @@
+# Disallow number literals with zero fractions or dangling dots
+
+Disallows number literals with zero fractions or dangling dots, as there is no difference in JS between one having it and one not having it.
+
+## Fail
+
+```js
+const foo = 1.0;
+const foo = -1.0;
+const foo = 123123123.0;
+const foo = 1.;
+```
+
+
+## Pass
+
+```js
+const foo = 1;
+const foo = -1;
+const foo = 123123123;
+const foo = 1.1;
+const foo = -1.1;
+const foo = 123123123.4;
+const foo = 1e3;
+```
--- a/docs/rules/no-zero-fractions.md
+++ b/docs/rules/no-zero-fractions.md
@@ -11,6 +11,8 @@ const foo = 1.0;
 const foo = -1.0;
 const foo = 123123123.0;
 const foo = 1.;
+const foo = 123.111000000;
+const foo = 123.00e20;
 ```
 
 
--- a/readme.md
+++ b/readme.md
@@ -64,7 +64,8 @@ Configure it in `package.json`.
 			"unicorn/prefer-query-selector": "error",
 			"unicorn/prefer-node-remove": "error",
 			"unicorn/prefer-text-content": "error",
-			"unicorn/no-for-loop": "error"
+			"unicorn/no-for-loop": "error",
+			"unicorn/no-zero-fractions": "off"
 		}
 	}
 }
@@ -104,6 +105,7 @@ Configure it in `package.json`.
 - [prefer-node-remove](docs/rules/prefer-node-remove.md) - Prefer `remove` over `parentNode.removeChild` and `parentElement.removeChild`. *(fixable)*
 - [prefer-text-content](docs/rules/prefer-text-content.md) - Prefer `textContent` over `innerText`. *(fixable)*
 - [no-for-loop](docs/rules/no-for-loop.md) - Do not use a `for` loop that can be replaced with a `for-of` loop. *(fixable)*
+- [no-zero-fractions](docs/rules/no-zero-fractions.md) - Disallow number literals with zero fractions or dangling dots. *(fixable)*
 
 
 ## Recommended config
--- a/readme.md
+++ b/readme.md
@@ -65,7 +65,7 @@ Configure it in `package.json`.
 			"unicorn/prefer-node-remove": "error",
 			"unicorn/prefer-text-content": "error",
 			"unicorn/no-for-loop": "error",
-			"unicorn/no-zero-fractions": "off"
+			"unicorn/no-zero-fractions": "error"
 		}
 	}
 }
--- a/rules/no-zero-fractions.js
+++ b/rules/no-zero-fractions.js
@@ -1,18 +1,35 @@
 'use strict';
 const getDocsUrl = require('./utils/get-docs-url');
 
-const message = 'Zero fraction or dangling dot in number.';
+const MSG_ZERO_FRACTION = 'Zero fraction in number.';
+const MSG_DANGLING_DOT = 'Dangling dot in number.';
 
-const regexpEndsWithZeroFraction = /\.0*$/;
+// Groups:
+// 1 - integer part
+// 2 - dangling dot or dot with zeroes
+// 3 - dot with digits except last zeroes
+// 4 - scientific notation
+const RE_DANGLINGDOT_OR_ZERO_FRACTIONS = /^([+-]?[0-9]*)(?:(\.0*)|(\.[0-9]*[1-9])0+)(e[+-]?[0-9]+)?$/;
+
 const create = context => {
 	return {
 		Literal: node => {
-			if (typeof node.value === 'number' && node.value % 1 === 0 && regexpEndsWithZeroFraction.test(node.raw)) {
-				context.report({
-					node,
-					message,
-					fix: fixer => fixer.replaceText(node, `${node.value}`)
-				});
+			if (typeof node.value === 'number'){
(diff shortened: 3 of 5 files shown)


sindresorhus:

readme.md
```diff
@@ -101,7 +102,11 @@ Configure it in `package.json`.
 - [prefer-node-append](docs/rules/prefer-node-append.md) - Prefer `append` over `appendChild`. *(fixable)*
 - [prefer-query-selector](docs/rules/prefer-query-selector.md) - Prefer `querySelector` over `getElementById`, `querySelectorAll` over `getElementsByClassName` and `getElementsByTagName`. *(partly fixable)*
 - [prefer-node-remove](docs/rules/prefer-node-remove.md) - Prefer `remove` over `parentNode.removeChild` and `parentElement.removeChild`. *(fixable)*
+<<<<<<< HEAD
```
christophehurpeau: seems a merge went wrong ?

stroncium: --_-- after 10s of merges it finally got me

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const message = 'Zero fraction or dangling dot in number.';
```
christophehurpeau: Maybe the message can be improve by suggesting how to fix the problem ?

stroncium: If you have some particular message in mind, please suggest. Otherwise, I think it's pretty self describing, plus it's a fixable error, so expected to be fixed automatically for codebases with fixing set up.

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,30 @@
...
+
+const regexpEndsWithZeroFraction = /\.0*$/;
+const create = context => {
+	return {
+		Literal: node => {
+			if (typeof node.value === 'number' && node.value % 1 === 0 && regexpEndsWithZeroFraction.test(node.raw)) {
```
futpib: Will this also catch `1.10`?

stroncium: This implementation is specifically limited to only work with integer numbers. It also doesn't react to numbers in scientific notation.
I just did it the way it was said in original issue, but if my interpretation isn't correct, I can add both other cases. Honestly speaking, I can't make a judgement about extent of consistency we want here(mostly because while `1.0` and `1` are equal in JS, they actually have different behavior, while `1.10` and `1.1` are really always the same as far as I know).

sindresorhus: > Will this also catch 1.10?

Yes, it should handle this too. Sorry, missed that in the original issue.

test/no-zero-fractions.js
```diff
@@ -0,0 +1,61 @@
...
+			output: 'const foo = 123123123',
+			errors: [error]
+		},
+		{
+			code: 'const foo = 1.',
+			output: 'const foo = 1',
+			errors: [error]
+		}
+	]
+});
```
sindresorhus: Some more tests I'd like to see:

- `const foo = 0.1 + 1;`
- `foo(.010);`

rules/no-zero-fractions.js
sindresorhus: Messages usually don't contain the solution if they're auto-fixable. I do think the message could be more actionable. How about:

> Don't use a zero fraction or dangling dot in the number.

sindresorhus: This made me think. We should probably have separate error messages for these two cases to make it clearer at least:

> Don't use a zero fraction in the number.

> Don't use a dangling dot in the number.

test/no-zero-fractions.js
stroncium: added some

rules/no-zero-fractions.js
stroncium: Added that, also added scientific notation `1.10e123` => `1.1e123`.

stroncium:
@sindresorhus For the record, that regexp became much less readable when I satisfied linter's request to change `[0-9]`s to `\d`s.

stroncium:
@sindresorhus If we change requirement on inclusion of rules in readme and defaults to alphabetical sort, it will remove most of merge conflicts.

test/no-zero-fractions.js
sindresorhus: No, it's good now.

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,48 @@
...
+// 2 - dangling dot or dot with zeroes
+// 3 - dot with digits except last zeroes
+// 4 - scientific notation
+const RE_DANGLINGDOT_OR_ZERO_FRACTIONS = /^([+-]?\d*)(?:(\.0*)|(\.\d*[1-9])0+)(e[+-]?\d+)?$/;
+
+const create = context => {
+	return {
+		Literal: node => {
+			if (typeof node.value === 'number') {
+				const m = RE_DANGLINGDOT_OR_ZERO_FRACTIONS.exec(node.raw);
```
sindresorhus: Can you use a more descriptive variable name than `m`?

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,48 @@
...
+				if (m !== null) {
+					const isDanglingDot = m[2] === '.';
+					context.report({
+						node,
+						message: isDanglingDot ? MSG_DANGLING_DOT : MSG_ZERO_FRACTION,
+						fix: fixer => {
+							let wantedStr = (m[2] === undefined) ? m[1] + m[3] : m[1];
```
sindresorhus: `wantedStr` => `wantedString`

docs/rules/no-zero-fractions.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Fail
+
+```js
+const foo = 1.0;
+const foo = -1.0;
+const foo = 123123123.0;
+const foo = 1.;
+```
+
```
sindresorhus: Can you include `123.11100000000` too?

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const MSG_ZERO_FRACTION = 'Don\'t use a zero fraction in the number.';
```
sindresorhus: `MSG` => `MESSAGE`

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,48 @@
...
+const MSG_DANGLING_DOT = 'Don\'t use a dangling dot in the number.';
+
+// Groups:
+// 1 - integer part
+// 2 - dangling dot or dot with zeroes
+// 3 - dot with digits except last zeroes
+// 4 - scientific notation
+const RE_DANGLINGDOT_OR_ZERO_FRACTIONS = /^([+-]?\d*)(?:(\.0*)|(\.\d*[1-9])0+)(e[+-]?\d+)?$/;
```
sindresorhus: Can you add a TODO comment about using named capture groups when we target Node.js 10?

rules/no-zero-fractions.js
```diff
@@ -0,0 +1,48 @@
...
+
+const create = context => {
+	return {
+		Literal: node => {
+			if (typeof node.value === 'number') {
+				const m = RE_DANGLINGDOT_OR_ZERO_FRACTIONS.exec(node.raw);
+				if (m !== null) {
```
sindresorhus: Can you destructure `m` here into 4 variables with readable names?

sindresorhus:
> For the record, that regexp became much less readable when I satisfied linter's request to change [0-9]s to \ds.

I think that's subjective. I personally find `\d` much more readable.

sindresorhus:
> If we change requirement on inclusion of rules in readme and defaults to alphabetical sort, it will remove most of merge conflicts.

I'd rather deal with the merge conflicts. They are easy to resolve in the GitHub UI. If I require alphabetic sorting, I have to ensure people follow that too (I know from experience that people *do* struggle with following that).

rules/no-zero-fractions.js
stroncium: Considering this is high performance code, I would suggest not doing this just yet as named capture groups require 1 more object allocation than numbered which introduces non-trivial slowdown.
Microbench(and that is expected to be much more allocator-friendly than real application is almost 3 times slower: https://jsperf.com/named-capture-groups

stroncium:
> I'd rather deal with the merge conflicts. They are easy to resolve in the GitHub UI. If I require alphabetic sorting, I have to ensure people follow that too (I know from experience that people do struggle with following that).

@sindresorhus We already have tests against inclusion of new rules into index and defaults, can be easily changed to enforce alphabetical sorting.

rules/no-zero-fractions.js
sindresorhus: That's not really relevant. We're not switching yet. It's going to be a long time until we can target Node.js 10, and by that time V8 will be much more optimized.

sindresorhus:
> We already have tests against inclusion of new rules into index and defaults, can be easily changed to enforce alphabetical sorting.

Sure, happy to switch if you want to do the work of automatically enforcing it. It needs to be enforced in 3 places: `index.js`, the example in the readme, and the list of rules in the readme. It should be done in a separate PR though :)

sindresorhus:
This is looking great. Thanks for working on this, @stroncium 🙌

rules/no-zero-fractions.js
stroncium: As far as I understand, the interface is standardized already, and it requires 2 objects(unless object will reference itself, but enforcing compatibility with older versions might be a problem then, and optimizations aren't usually the goal of JS crowd anyway, so it would be a big surprise it gets implemented this way anywhere), and additional allocation will always slow down things a lot, so it's safe to say it might get faster but won't ever be up to speed with numbered groups.

sindresorhus: New features are always slower when first added, however, the V8 team has a good track record of optimizing features over time.

stroncium: One can't fight standard tho, and standard requires 2 objects. Memory allocation is among most important programming problems for 40 years already, and it's hard to expect any breakthroughs here.

Also, on a second thought I don't think it's even possible to use self-referencing object, as standard seem to implicit both numbered and named groups should be enumerable.

PR #246, opened 2019-02-13
Fix integration travis task not failing unless unit tests fail too

futpib:
The old script is equivalent to 
```bash
([ "$INTEGRATION" == true ] && npm run integration) || npm test
```
which will exit with 0 whenever `npm test` exists with 0.

PS: I bet this question has played a part in this: https://stackoverflow.com/questions/3953645/ternary-operator-in-bash

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,6 @@ matrix:
     - node_js: 10
       env: INTEGRATION=true
 script:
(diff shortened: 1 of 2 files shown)


sindresorhus:
Good catch! Ugh, I hate shell scripting so much.

PR #253, opened 2019-03-04
 Fix `no-for-loop` breaking loops that mutate index variable in body 

### LINKED ISSUE #252: The `no-for-loop` rule doesn't preserve `let`
```js
const foo = input => {
	for (let i = 0; i < input.length; i++) {
		const c = input[i];
		i++;
		console.log(i, c);
	}

	return input;
};
```

Is auto-fixed into:

```js
const foo = input => {
	for (const [i, c] of input.entries()) {
		i++;
		console.log(i, c);
	}

	return input;
};
```

Which is a runtime error, since `i` is now `const`, but it cannot be because of `i++`.

---

// @futpib

futpib:
Fixes #252

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -228,6 +228,12 @@ const isIndexVariableUsedElsewhereInTheLoopBody = (indexVariable, bodyScope) => 
 	return inBodyReferences.length > 1;
 };
 
+const isIndexVariableAssignedToInTheLoopBody = (indexVariable, bodyScope) => {
+	return indexVariable.references
+		.filter(reference => scopeContains(bodyScope, reference.from))
+		.some(inBodyReference => inBodyReference.isWrite());
+};
+
 const someVariablesLeakOutOfTheLoop = (forStatement, variables, forScope) => {
 	return variables.some(variable => {
 		return !variable.references.every(reference => {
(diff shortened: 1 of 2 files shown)


rules/no-for-loop.js
```diff
@@ -220,19 +220,22 @@ const nodeContains = (ancestor, descendant) => {
...
 	// Any reference besides that should retain the index variable.
 	return inBodyReferences.length > 1;
 };
 
-const someIdentifiersLeakOutOfTheLoop = (forStatement, identifiers, forScope, bodyScope) => {
-	return identifiers.some(name => {
-		const variable = resolveIdentifierName(name, bodyScope);
+const isIndexVariableAssignedToInTheLoopBody = (indexVariable, bodyScope) => {
+	return indexVariable.references
+		.filter(reference => scopeContains(bodyScope, reference.from))
```
alexzherdev: Not sure if worth combining the two conditions?
```js
return indexVariable.references
	.some(reference => scopeContains(bodyScope, reference.from) && reference.isWrite());
```

futpib: I'm afraid you'll have to make an argument for it. Performance difference is negligible, readability is fine both ways.

PR #258, opened 2019-03-20
Fix `catch-error-name` fixer not renaming all references in `.catch`

### LINKED ISSUE #255: `catch-error-name` fixer bug
`catch-error-name` fixer bug

error code like

```js
p.catch(err => doSomething(err))
```

will fix like this

```js
p.catch(error => doSomething(err))
```

the param `err` in catch function body is not replaced

 I noticed there is a bug #107 , but please make this rule not fixable for now please, this breaks code

futpib:
fixes #255

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -43,6 +43,8 @@ const create = context => {
 		caughtErrorsIgnorePattern: '^_$'
 	}, context.options[0]);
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for fixing. You need to submit your PR here to claim the bounty: https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/255

PR #260, opened 2019-03-23
Alphabetically sort the rules and enforce that

stroncium:
Added rules to enforce alphabetical sorting and rule presence in readme and index
added missing rules in readme
sorted rules


stroncium:
integration test error inherited from master

test/package.js
```diff
@@ -24,6 +34,48 @@ test('Every rule is defined in index file', t => {
 		'There are more exported rules than rule files.');
 	t.is(Object.keys(index.configs.recommended.rules).length, ruleFiles.length,
 		'There are more exported rules in the recommended config than rule files.');
+
+	testSorted(t, Object.keys(index.configs.recommended.rules), 'configs.recommended.rules');
```
futpib: Could also use [`sort-keys`](https://eslint.org/docs/rules/sort-keys) rule.

sindresorhus: Seems like overkill to setup ESLint to run programmatically just for this one thing.

sindresorhus:
Perfect :)

test/package.js
futpib: I meant to suggest `/* eslint-enable sort-keys */`

sindresorhus: Ah ok. Yeah, that would work.

PR #282, opened 2019-04-17
Add `no-keyword-prefix` rule

### LINKED ISSUE #251: Rule Proposal: Forbid naming variables beginning with "new"
Forbid variables being named with `new` because it looks like the `new` keyword. `newFoo` and `new Foo()` are fairly similar when scanning code quickly (especially without syntax highlighting) and `newFoo` is already a pretty terrible variable name.

Fail:

* `const newFoo = 'foo';`

Pass:

* `const fooNew = 'foo';`
* `const foo = new Foo();`

----

A similar argument could be made for `class` and a few other keywords but it seems like `newFoo` is the most common scenario I've been bumping into recently.

MrHen:
Fixes #251.

This is an initial PR to judge interest. As I mentioned in the issue, I was planning on writing this rule anyway and figured I would submit it to unicorn instead of building a one-off package.

I don't expect this to be a popular rule -- it's merely solving a nitpick I've had on my existing team's code style.

There is some remaining work to do but I didn't want to polish it without knowing if the PR would be accepted.

* [x] Document configuration options
* [x] Update rule listing in readme
* [x] Update recommended config
* [x] Update recommended config in readme
* [x] Resolve TODOs (mostly missing tests)
* [x] Split up the `Identifier` function into smaller, easier to digest chunks


sindresorhus:
While I won't use this rule personally, I'm happy to host it here if you can commit to maintain it. I think it should be `"off"` by default in the recommended config though.

docs/rules/no-keyword-prefix.md
```diff
@@ -0,0 +1,20 @@
...
+
+
+## Pass
+
+```js
+const foo = 'foo';
+const _newFoo = 'foo';
+const new_foo = 'foo';
+const fooNew = 'foo';
+```
```
sindresorhus: Would it make sense to let users configure prefixes to disallow? To make the rule more generally useful.

MrHen: It wouldn't be hard to add. I liked having "keyword" in the rule name because that was my reasoning for what should be disallowed so then I limited its usage to the keywords used in a camel case context. But having it be customizable is fine.

(On the other hand, I doubt many other people would are going to use the rule...)

sindresorhus: Yeah, let's keep it simple for now. I can't think of any use-cases for making it configurable.

MrHen:
@sindresorhus This PR is now ready for review.

Relatedly, I'm not sure what your preferred process is for reviewing / merging PRs on your repositories. I've been assuming you want to review every PR and will decide when to merge them. Is that correct?

sindresorhus:
> Relatedly, I'm not sure what your preferred process is for reviewing / merging PRs on your repositories. I've been assuming you want to review every PR and will decide when to merge them. Is that correct?

It doesn't have to be me, but it's always useful to get a second set of eyes on the code.

Issue #278, opened 2019-04-08
Auto-fix for prefer-includes removes too many/wrong parenthesis


voxpelli:
Input:

```javascript
const forceLatest = ((url.parse(req.url).query || '').indexOf('latest') !== -1);
```
```javascript
return (req.originalUrl || '').indexOf(page.path) !== -1;
```

Expected output:

```javascript
const forceLatest = (url.parse(req.url).query || '').includes('latest');
```
```javascript
return (req.originalUrl || '').includes(page.path);
```

Output from auto-fixer:

```javascript
const forceLatest = (url.parse(req.url).query || ''.includes('latest'));
```
```javascript
return req.originalUrl || ''.includes(page.path);
```

sindresorhus:
// @PrBart

PrBart:
> // @PrBart

i will try to fix it soon.

PR #284, opened 2019-04-18
Add `prefer-flat-map` rule

### LINKED ISSUE #204: Rule proposal: `prefer-flat-map`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/204)

Prefer `array.flatMap(…)` over doing `array.flat().map(…)`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/204).

MrHen:
Basic check with a fixer. Fixes #204.

Converts ```[1,2,3].map(i => [i]).flat();``` to ```[1,2,3].flatMap(i => [i]);```

----

There are two scenarios that are a bit more complicated:

```
[1,2,3].map(i => [i]).foo().flat();
```

If the `map` and `flat` are not adjacent in the call chain than the rule will ignore it. It wouldn't be terribly hard to detect this case if desired but I don't think we should auto-fix it.

```
let foo = [1,2,3].map(i => [i]);
foo = foo.flat();
```

The complexity around this scenario is _much_ higher. It would require a totally different type of implementation.

### DIFF
--- a/rules/prefer-flat-map.js
+++ b/rules/prefer-flat-map.js
@@ -1,0 +1,65 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const MESSAGE_ID = 'preferFlatMap';
+
+const isFlat = node => {
+	return (
+		node.type === 'CallExpression' &&
+		node.callee.type === 'MemberExpression' &&
+		node.callee.property.type === 'Identifier' &&
+		node.callee.property.name === 'flat'
+	);
+};
+
+const isMap = node => {
+	return (
+		node.type === 'CallExpression' &&
+		node.callee.type === 'MemberExpression' &&
+		node.callee.property.type === 'Identifier' &&
+		node.callee.property.name === 'map'
+	);
+};
+
+const report = (context, nodeFlat, nodeMap) => {
+	context.report({
+		node: nodeFlat,
+		messageId: MESSAGE_ID,
+		fix: fixer => {
+			return [
+				fixer.removeRange([nodeMap.end, nodeFlat.end]),
+				fixer.replaceText(nodeMap.callee.property, 'flatMap')
+			];
+		}
+	});
+};
+
+const create = context => ({
+	CallExpression: node => {
+		if (!isFlat(node)) {
+			return;
+		}
+
+		const parent = node.callee.object;
+
+		if (!isMap(parent)) {
+			return;
+		}
+
+		report(context, node, parent);
+	}
+});
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocsUrl(__filename)
+		},
+		fixable: 'code',
+		messages: {
+			[MESSAGE_ID]: 'Use `.flatMap()`, rather than `.map(...).flat()`.'
+		}
+	}
+};
--- a/rules/prefer-flat-map.js
+++ b/rules/prefer-flat-map.js
@@ -22,13 +22,25 @@ const isMap = node => {
 };
 
 const report = (context, nodeFlat, nodeMap) => {
+	// Location will be:
+	//   map(...).flat();
+	//                 ^
+	const flatEnd = nodeFlat && nodeFlat.end;
+
+	// Location will be:
+	//   map(...).flat();
+	//          ^
+	const mapEnd = nodeMap && nodeMap.end;
+
+	const mapProperty = nodeMap && nodeMap.callee && nodeMap.callee.property;
+
 	context.report({
 		node: nodeFlat,
 		messageId: MESSAGE_ID,
 		fix: fixer => {
 			return [
-				fixer.removeRange([nodeMap.end, nodeFlat.end]),
-				fixer.replaceText(nodeMap.callee.property, 'flatMap')
+				fixer.removeRange([mapEnd, flatEnd]),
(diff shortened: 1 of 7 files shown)


rules/prefer-flat-map.js
```diff
@@ -0,0 +1,77 @@
...
+	});
+};
+
+const create = context => ({
+	CallExpression: node => {
+		if (!isFlat(node)) {
+			return;
+		}
+
+		const parent = node && node.callee && node.callee.object;
```
futpib: Please check explicitly if `node.callee` is `MemberExpression`

MrHen: That check happens in a helper: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/284/files#diff-5b3bb9237e1bb62a17dd58af5ed9c0afR18

futpib: But `node.callee` is not passed to `isMap`, right?

MrHen: Oh, linked the wrong helper function. https://github.com/sindresorhus/eslint-plugin-unicorn/pull/284/files#diff-5b3bb9237e1bb62a17dd58af5ed9c0afR9

The entire rule returns early if it isn't working on `.flat()` -- it can't get to the `isMap` check unless `node` is a `MemberExpression`.

futpib: Oh, right, I see. But this line could be simplified then:
```js
const parent = node.callee.object;
```

MrHen:
@futpib I just pushed a bunch of fixes due to issues like https://github.com/sindresorhus/eslint-plugin-unicorn/issues/278. There were two nasty ones hiding here:

```
let bar = ([1,2,3].map(i => [i]).flat();

let bar = [1,2,3].map(i => {
  return [i];
}) // comment
.flat();
```

Both required massive changes to the guts of the `report` function. I might have gone overkill on the comments but was having trouble keeping track of the various nodes / tokens in play.

MrHen:
@futpib @sindresorhus Sorry for the ping but I wanted to double check that there was nothing else expected of me for this PR.

test/prefer-flat-map.js
```diff
@@ -0,0 +1,105 @@
...
+			errors: [error]
+		},
+		{
+			code: 'const bar = [1,2,3].sort().map(i => [i]).flat()',
+			output: 'const bar = [1,2,3].sort().flatMap(i => [i])',
+			errors: [error]
+		},
+		{
+			code: 'const bar = (([1,2,3].map(i => [i]))).flat()',
+			output: 'const bar = (([1,2,3].flatMap(i => [i])))',
```
futpib: I see that the fix range logic is really tricky, can you also add tests for different spacing patterns like these?

```js
array
	.map(...)
	.flat()
```

```js
array.map(...)
	.flat()
```

```js
array . map( x => y ) . flat () // 🤪
```

futpib:
> You might want to take another look at the code. I might have gone overkill on the comments but was having trouble keeping track of the various nodes / tokens in play. Are there any helper functions nearby that can clean this up a bit?

I don't know if there are any, but we could make an util for that and reuse it in both #287 and here. I'd also like to see it tested on typescript's generics syntax `[].flatMap<any>(...)`.

Not sure which PR should implement this, but this seems to be the best solution technically.

rules/prefer-flat-map.js
```diff
@@ -0,0 +1,140 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const MESSAGE_ID = 'preferFlatMap';
+
+const isFlat = node => {
```
sindresorhus:
```suggestion
const isFlatMethod = node => {
```

Same with the below.

rules/prefer-flat-map.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+const isMap = node => {
+	return (
+		node.type === 'CallExpression' &&
+		node.callee.type === 'MemberExpression' &&
+		node.callee.property.type === 'Identifier' &&
+		node.callee.property.name === 'map'
+	);
+};
```
sindresorhus: These two methods are almost identical. Could be DRY'd up into a `isMethodNamed('map')` and moved to util.

MrHen:
> Not sure which PR should implement this, but this seems to be the best solution technically.

@futpib I am willing to make the util but I think it makes sense to merge one PR and then use the other to create the util. Or maybe merge both and then I can send out a dedicated PR for the util. Either way, let me know what you prefer.

----

@sindresorhus I made the recommended changes. Please re-review at your leisure.

futpib:
> Or maybe merge both and then I can send out a dedicated PR for the util.

Ok

sindresorhus:
> I'd also like to see it tested on typescript's generics syntax [].flatMap<any>(...).

I agree, but I think adding the TypeScript parser could be a separate PR where some TS fixtures like generics are added to other rules too.

sindresorhus:
Looks good 👍

PR #288, opened 2019-04-23
Add `prefer-string-slice` rule

### LINKED ISSUE #65: `prefer-string-slice` rule
I prefer `.slice()` over `.substr()` and `.substring()`. `.slice()` has clearer behavior and has a counterpart with arrays. Also better to be consistent. Anything you do with `.substr()` or `.substring()` can be done with `.slice()`.

Read more: http://stackoverflow.com/questions/2243824/what-is-the-difference-between-string-slice-and-string-substring

I'm aware we can't do type checking, but I think we can just assume a string when we see `.substr()` and `.substring()`. Worth the risk I think.
## 

Thoughts?

futpib:
This is primarily a request for feedback and a showcase of [futpib/eslint-template-visitor](https://github.com/futpib/eslint-template-visitor) (no meaningful readme there for now).

Take a look at the [initial rule implementation](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/288/files#diff-65f4c1557386dc75ca3f6a1d7f0c5f8c). More advanced features like spread matching (would be required to properly match multiple `.substr` arguments) are missing, but the basic API is there.

I aim at eventually removing most of the comparison boilerplate from the rules, such as this, for example: 
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6bccd4176f5552d5e2ec785a2dae0835b4a8cd4d/rules/no-console-spaces.js#L16-L19

Let me know if this is worthwhile and would be accepted for use in this (or maybe other) project. If so, I'll continue working on it using `prefer-string-slice` as the driving use-case for now. Also feel free to suggest alternative API or features you'd like to see or anything else.

Cc: @sindresorhus @SamVerschueren @jfmengels @MrHen @stroncium (and anybody else who comes by) 

Could eventually fix #65.

### DIFF
--- a/docs/rules/prefer-string-slice.md
+++ b/docs/rules/prefer-string-slice.md
@@ -1,0 +1,19 @@
+# Prefer `String#slice` over `String#substr` and `String#substring`
+
+[`String#substr()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) and [`String#substring()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) are the two lesser known legacy ways to slice a string. It's better to use [`String#slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) as a more popular option with clearer behaviour that has a consistent [`Array` counterpart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice).
+
+This rule is fixible when no arguments are passed to the method.
+
+## Fail
+
+```js
+foo.substr(1, 2);
+foo.substring(1, 3);
+```
+
+
+## Pass
+
+```js
+foo.slice(1, 3);
+```
--- a/docs/rules/prefer-string-slice.md
+++ b/docs/rules/prefer-string-slice.md
@@ -1,8 +1,9 @@
 # Prefer `String#slice()` over `String#substr()` and `String#substring()`
 
-[`String#substr()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) and [`String#substring()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring) are the two lesser known legacy ways to slice a string. It's better to use [`String#slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) as a more popular option with clearer behaviour that has a consistent [`Array` counterpart](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice).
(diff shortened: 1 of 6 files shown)


rules/prefer-string-slice.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+const eslintTemplateVisitor = require('eslint-template-visitor');
+const getDocsUrl = require('./utils/get-docs-url');
+
+const templates = eslintTemplateVisitor();
+
+const objectVariable = templates.variable();
+
+const substrCallTemplate = templates.template`${objectVariable}.substr()`;
```
sindresorhus: Would be cool to be able to specify the method dynamically too and then have the be able to automatically generate the fixer by saying what we want to replace.

```suggestion
const substrCallTemplate = templates.template`${objectVariable}.${method}()`;
```

I guess it would be much more complicated if the arguments needs to change, but in this case it shouldn't be too hard.

I like how easy [jscodeshift](https://github.com/facebook/jscodeshift) makes it to create Codemods: https://github.com/avajs/ava-codemods/blob/1b3f183d58834e5b9f24340ba8563efa5378d30e/lib/error-to-iferror.js#L7 You might find some inspiration in it.

rules/prefer-string-slice.js
```diff
@@ -0,0 +1,36 @@
...
+	const sourceCode = context.getSourceCode();
+
+	return templates.visitor({
+		[substrCallTemplate](node) {
+			const objectNode = substrCallTemplate.context.getMatch(objectVariable);
+
+			context.report({
+				node,
+				fix: fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + '.slice()'),
+				message: 'Prefer `String#slice` over `substr`.'
```
sindresorhus:
```suggestion
				message: 'Prefer `String#slice()` over `String#substr()`.'
```

rules/prefer-string-slice.js
sindresorhus: You might also find https://github.com/jfmengels/enhance-visitors useful. We use it heavily in `eslint-plugin-ava`.

docs/rules/prefer-string-slice.md
```diff
@@ -0,0 +1,18 @@
+# Prefer `String#slice` over `String#substr` and `String#substring`
```
sindresorhus:
```suggestion
# Prefer `String#slice()` over `String#substr()` and `String#substring()`
```

sindresorhus:
> Let me know if this is worthwhile and would be accepted for use in this (or maybe other) project. 

It makes it a lot more readable. Go for it.

sindresorhus:
@futpib Do you still intend to finish this? No rush, just wondering whether it's worth keeping it open.

rules/prefer-string-slice.js
futpib: Yes, this would be nice to have, but this does not fit well withing `eslint-template-visitor`. Maybe I could provide this with another package with an API like this:

```js
fix: fixer => jscodeshiftFixer(fixer, node).findMethodCalls('substr').renameTo('slice')
```

sindresorhus:
Nice work! The template stuff definitely made the rule simpler.

PR #293, opened 2019-05-02
Make the `import-index` rule handle more cases

### LINKED ISSUE #292: Make 'import-index' handle more cases
### Case 1

```js
const m = require('../.');
```
should be changed to
```js
const m = require('..');
```
### Case 2
```js
const m = require('./.');
```
should should be changed to
```js
const m = require('.');
```

henri5:
fixes #292 

`import-index` rule disallows `'../.'` and `'./.'`

### DIFF
--- a/rules/import-index.js
+++ b/rules/import-index.js
@@ -1,7 +1,7 @@
 'use strict';
 const getDocsUrl = require('./utils/get-docs-url');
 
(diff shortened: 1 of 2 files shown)


amedora:
LGTM

sindresorhus:
Thanks :)

Issue #272, opened 2019-03-31
`no-for-loop` rule bug with `@typescript-eslint/parser`


SimonSiefke:
I'm runnning into an issue with the `no-for-loop` rule. Here is the code:
```javascript
const links = document.querySelectorAll('link')
for (let i = 0; i < links.length; i++) {
  const link = links[i]
}
```

The ESLint config:
```javascript
module.exports = {
  parser: '@typescript-eslint/parser',
  plugins: ['@typescript-eslint', 'unicorn'],
  rules: {
    'unicorn/no-for-loop': 'error',
  },
}
```
However, it is working fine without `@typescript-eslint/parser`.

And here is the stacktrace from VSCode:
```
[Error - 11:07:51 PM] ESLint stack trace:
[Error - 11:07:51 PM] TypeError: Expected `index` to be a number.
Occurred while linting /home/simon/Documents/self-tuts/live-server/unicorn.bug.js:2
    at SourceCode.getLocFromIndex (/home/simon/Documents/self-tuts/live-server/node_modules/eslint/lib/util/source-code.js:436:19)
    at getRemovalRange (/home/simon/Documents/self-tuts/live-server/node_modules/eslint-plugin-unicorn/rules/no-for-loop.js:171:28)
    at Object.ForStatement.problem.fix.fixer [as fix] (/home/simon/Documents/self-tuts/live-server/node_modules/eslint-plugin-unicorn/rules/no-for-loop.js:308:25)
    at normalizeFixes (/home/simon/Documents/self-tuts/live-server/node_modules/eslint/lib/util/report-translator.js:176:28)
    at args (/home/simon/Documents/self-tuts/live-server/node_modules/eslint/lib/util/report-translator.js:278:18)
    at Object.report (/home/simon/Documents/self-tuts/live-server/node_modules/eslint/lib/linter.js:721:41)
    at ForStatement (/home/simon/Documents/self-tuts/live-server/node_modules/eslint-plugin-unicorn/rules/no-for-loop.js:313:12)
    at listeners.(anonymous function).forEach.listener (/home/simon/Documents/self-tuts/live-server/node_modules/eslint/lib/util/safe-emitter.js:45:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/home/simon/Documents/self-tuts/live-server/node_modules/eslint/lib/util/safe-emitter.js:45:38)
```

SimonSiefke:
[...]

PR #294, opened 2019-05-02
Apply `escape-case` to regex literal and remove transformation of `\c` escape on string literal

### LINKED ISSUE #273: Apply `escape-case` to regex literal sequences
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/273)

Thanks for the excellent set of rules!

I'm wondering if you might allow `escape-case` to be applied to regex literal sequences as well as strings:

## Fail

```js
const foo = /\xa9/;
const foo = /\ud834/;
const foo = /\u{1d306}/u;
const foo = /\ca/;
```

## Pass

```js
const foo = /\xA9/;
const foo = /\uD834/;
const foo = /\u{1D306}/u;
const foo = /\cA/;
```

(Incidentally, while these examples I adopted are suitable for my case, are you sure you want to be supporting them in strings, given that `\c` does not relate to control characters in strings?)

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/273).

JLHwung:
This PR addresses issue and feature request in #273. I've made separate commit on this.

Although the owner suggested using https://github.com/mysticatea/regexpp, I decide not to use it for performance considerations -- There would be many regular expression in a mature codebase and we can not afford the extra regular expression parse + replace cycle only to find whether there is any escapable sequence.

The original `escapeWithLowercase` pattern should work good enough on regular expression under most situation. I have minimally modified the code to support slightly difference escape syntax between string literal and RegExp literal/objects.

Fixes #273

### DIFF
--- a/rules/escape-case.js
+++ b/rules/escape-case.js
@@ -34,7 +34,7 @@ const findLowercaseEscape = value => {
 
 	let escapeNodePosition;
 	visitRegExpAST(ast, {
-/**
+		/**
 Record escaped node position in regexpp ASTNode. Returns undefined if not found.
 @param {ASTNode} node A regexpp ASTNode. Note that it is of different type to the ASTNode of ESLint parsers
 @returns {undefined}
@@ -109,10 +109,13 @@ const create = context => {
 			const matches = node.value.raw.match(escapeWithLowercase);
 
 			if (matches && matches[2].slice(1).match(hasLowercaseCharacter)) {
+				// Move cursor inside the head and tail apostrophe
+				const start = node.range[0] + 1;
+				const end = node.range[1] - 1;
 				context.report({
 					node,
 					message,
-					fix: fixer => fixer.replaceTextRange([node.start, node.end], fix(node.value.raw, escapeWithLowercase))
+					fix: fixer => fixer.replaceTextRange([start, end], fix(node.value.raw, escapeWithLowercase))
 				});
 			}
 		}
--- a/rules/escape-case.js
+++ b/rules/escape-case.js
@@ -1,7 +1,7 @@
 'use strict';
 const getDocsUrl = require('./utils/get-docs-url');
 
-const escapeWithLowercase = /((?:^|[^\\])(?:\\\\)*)\\(x[a-f\d]{2}|u[a-f\d]{4}|u\{(?:[a-f\d]{1,})\}|c[a-z])/;
+const escapeWithLowercase = /((?:^|[^\\])(?:\\\\)*)\\(x[a-f\d]{2}|u[a-f\d]{4}|u\{(?:[a-f\d]{1,})\})/;
 const hasLowercaseCharacter = /[a-z]+/;
 const message = 'Use uppercase characters for the value of the escape sequence.';
 
--- a/rules/escape-case.js
+++ b/rules/escape-case.js
@@ -6,8 +6,8 @@ const {
 
 const getDocsUrl = require('./utils/get-docs-url');
 
-const escapeWithLowercase = /((?:^|[^\\])(?:\\\\)*)\\(x[a-f\d]{2}|u[a-f\d]{4}|u\{(?:[a-f\d]{1,})\})/;
-const escapePatternWithLowercase = /((?:^|[^\\])(?:\\\\)*)\\(x[a-f\d]{2}|u[a-f\d]{4}|u\{(?:[a-f\d]{1,})\}|c[a-z])/;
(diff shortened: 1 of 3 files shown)


futpib:
A slightly off-topic nitpick.

> Although the owner suggested using mysticatea/regexpp, I decide not to use it for performance considerations -- There would be many regular expression in a mature codebase and we can not afford the extra regular expression parse + replace cycle only to find whether there is any escapable sequence.

Would it really be that slow? We could eventually cache the result and reuse it in other regexp-related rules (or maybe even across projects). Even if it were substantially slower, we may value correctness and maintainability over it, especially for a linter (unless it was **really** slow).

Also a mature codebase with a lot of unique regular expressions sounds like a nightmare.

I'm **not** saying this PR should not be accepted, only that the argument against `regexpp` is not convincing.

MrHen:
I agree with @futpib. Performance issues should be addressed once they have been confirmed. Does the slower approach add 0.5 seconds for every 1,000 regex statements? 1 minute for every 10 regex statements?

JLHwung:
> we may value correctness and maintainability over speed, especially for a linter (unless it was really slow).

Make sense, I can spare some time to try on the `regexpp` approach.

JLHwung:
@futpib Upgraded the branch. The new lowercase detection approach is

1. use `regexpp` to parse regular expression literal
2. tap on `onCharacterLeave` handler, check if it is the first occurrence of escaped lowercase sequence
3. report if invalid

The fixer approach is
1. use `regexpp` to parse regular expression literal
2. tap on `onCharacterLeave` handler, check if it is the first occurrence of lowercase sequence, record the position of escaped sequence in the regular expression
3. replace the original escaped sequence with the one applied with the string escape fixer
4. return the replaced string as the fixed regular expression literal

sindresorhus:
@JLHwung Can you give the PR a proper title that describes what it fixes?

rules/escape-case.js
```diff
@@ -17,6 +23,48 @@ const fix = value => {
...
+	visitRegExpAST(ast, {
+		onCharacterLeave(node) {
+			if (escapeNodePosition) {
+				return;
+			}
+
+			const matches = node.raw.match(escapePatternWithLowercase);
+
+			if (matches && matches[2].slice(1).match(hasLowercaseCharacter)) {
+				escapeNodePosition = [node.start, node.end];
```
sindresorhus: You cannot use `.start` and `.end`. See: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/272

sindresorhus:
I pushed some minor formatting tweaks: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/294/commits/38f91bb879549f0e861bb9b4823a1b0d11fe6708

rules/escape-case.js
```diff
@@ -30,7 +81,18 @@ const create = context => {
 				context.report({
 					node,
 					message,
-					fix: fixer => fixer.replaceTextRange([node.start, node.end], fix(node.raw))
+					fix: fixer => fixer.replaceTextRange([node.start, node.end], fix(node.raw, escapeWithLowercase))
```
sindresorhus: Use `replaceText(node` instead of `replaceTextRange[node.start, node.end]`

rules/escape-case.js
```diff
@@ -17,6 +23,51 @@ const fix = value => {
 	return value;
 };
 
+/**
+Find the `[start, end]` position of the lowercase escape sequence in a regular expression literal ASTNode.
+
+@param {string} value - String representation of a literal ASTNode.
+@returns {number[] | null} The `[start, end]` pair if found, or null if not.
```
sindresorhus: I would prefer returning `undefined` instead of `null`.

JLHwung:
Thanks for the review comments. I am sorry that I am taking vacation off my laptop for a couple days. I will come back and revise on the next week. Thank you.

rules/escape-case.js
JLHwung: Here is different to `fixer.replaceTextRange([node.start, node.end])` as the `node` here is actually the `ASTNode` from `regexpp`, while other `node` is the AST node from eslint parsers. 

As `start` and `end` is [official properties](https://github.com/mysticatea/regexpp/blob/v2.0.1/src/ast.ts#L66) of a `ASTNode` in `regexpp`, I think we can leave it as-is, but I can add a comment here as a reminder.

JLHwung:
Regards to the feedback, I have pushed fixes and they are ready to review.

rules/escape-case.js
sindresorhus: Ah ok. I missed that. It's ok to leave it. My mistake.

sindresorhus:
Thanks for working on this, @JLHwung 🙌

PR #296, opened 2019-05-22
Add `no-nested-ternary` rule

### LINKED ISSUE #50: A better `no-nested-ternary` rule
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/50)

Since it was denied in ESLint: https://github.com/eslint/eslint/issues/3480

I think it would be better with a new rule that only errors on nested ternaries that are not wrapped in parens. For example: `parens-wrapped-nested-ternary` (a mouthful, but very descriptive).
## 

Thoughts?

> There is a $82.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/50).

karaggeorge:
Fixes #50

Extends the existing eslint rule but allows one-level nesting if it's wrapped in parens

Notes:
- Wasn't sure what the best name should be.
- First time writing an eslint rule, so let me know if it needs edits
- Tested it against most of the examples in the issue and it seemed to work

Not sure about the case where both sides are ternary. Should it report two errors, one for each? Current rule only reports one for the whole expression, but if it's one-level nested, we could provide a fix for both by wrapping it in parens (current implementation)


docs/rules/no-nested-ternary.md
```diff
@@ -0,0 +1,20 @@
...
+var foo = i > 5 ? i < 100 ? true : false : true;
+var foo = i > 5 ? true : (i < 100 ? true : (i < 1000 ? true : false));
+```
+
+
+## Pass
+
+```js
+var foo = i > 5 ? (i < 100 ? true : false) : true;
+var foo = i > 5 ? (i < 100 ? true : false) : (i < 100 ? true : false);
```
sindresorhus: Use `const` for these.

docs/rules/no-nested-ternary.md
```diff
@@ -0,0 +1,20 @@
+# Disallow nested ternary expressions
+
+Nesting ternary expressions can make code more difficult to understand.
+
+One level of nesting is allowed as long as the nested ternary is wrapped in parens
+
```
sindresorhus: You need to mention how it's an improved version of the built-in ESLint rule (and link to it) and what it improves.

Also, at the bottom, recommend disabling the built-in rule and show how to do that.

sindresorhus: In the readm, you marked it as "partly fixable", so would be good to explain what is and what is not fixable.

test/no-nested-ternary.js
```diff
@@ -0,0 +1,47 @@
...
+
+const errors = [{ruleId: 'my-rule'}];
+
+ruleTester.run('new-error', rule, {
+	valid: [
+		'var foo = i > 5 ? true : false;',
+		'var foo = i > 5 ? true : (i < 100 ? true : false);',
+		'var foo = i > 5 ? (i < 100 ? true : false) : true;',
+		'var foo = i > 5 ? (i < 100 ? true : false) : (i < 100 ? true : false);',
+		'var foo = i > 5 ? true : (i < 100 ? FOO(i > 50 ? false : true) : false);'
```
sindresorhus: Use `const`.

test/no-nested-ternary.js
```diff
@@ -0,0 +1,47 @@
...
+			code: 'var foo = i > 5 ? i < 100 ? true : false : i < 100 ? true : false;',
+			output: 'var foo = i > 5 ? (i < 100 ? true : false) : (i < 100 ? true : false);',
+			errors: [{ruleId: 'my-rule', column: 19}, {ruleId: 'my-rule', column: 44}]
+		},
+		{
+			code: 'var foo = i > 5 ? true : i < 100 ? true : false;',
+			output: 'var foo = i > 5 ? true : (i < 100 ? true : false);',
+			errors
+		}
+	]
```
sindresorhus: Would be good to take a look at the built-in ESLint rule and its test and see if there are cases that we don't handle. tl;dr Add more tests.

rules/no-nested-ternary.js
```diff
@@ -0,0 +1,54 @@
...
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocsUrl(__filename)
+		},
+		fixable: 'code',
+		schema: []
```
sindresorhus: No point in having `scheme` here when it's empty.

test/no-nested-ternary.js
karaggeorge: They only had two of each. I added the ones that were still valid in the list

docs/rules/no-nested-ternary.md
karaggeorge: I tried to reword the readme. Let me know if that makes more sense

karaggeorge:
@sindresorhus this should be good for another look

test/no-nested-ternary.js
```diff
@@ -0,0 +1,58 @@
...
+import avaRuleTester from 'eslint-ava-rule-tester';
+import rule from '../rules/no-nested-ternary';
+
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true
+	}
+});
+
+const errors = [{ruleId: 'my-rule'}];
```
sindresorhus: Huh? What is `my-rule`?

---

@MrHen Seems like the `ruleId` property is not correctly checked in the tests...

sindresorhus: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/368

sindresorhus:
Nice work, @karaggeorge 👌

Issue #250, opened 2019-02-27
Make the `no-for-loop` rule handle more cases


sindresorhus:
See: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/49#issuecomment-457832099

IssuehuntBot:
@issuehunt has funded $80.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/55832243/issues/250) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/55832243/issues/250).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

PR #297, opened 2019-05-22
Extend functionality of `no-for-loop`

### LINKED ISSUE #286: `no-for-loop` false negative
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/286)

```js
/* eslint unicorn/no-for-loop: 2 */
for (let i = 0; i < arr.length; i += 1) {
  console.log(arr[i])
}
```

**Expected**:
It reports.
**Actual**:
It does not report.

This is a subtask of https://github.com/sindresorhus/eslint-plugin-unicorn/issues/250 .

> There is a $60.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/286).

karaggeorge:
Fixes #286 

Works on part of #250 

## Summary
- Will track for variable declaration to `arrayName[indexName]` anywhere in the body
- Will fix even if the declaration is complex : `let x = 1, y = arr[i], z = 13;`
- Replaces all instances of `arr[i]` with the element name
- Will not report if `arr[i]` is set to anything or if the array variable is used any other way than to read `arr[i]` since that might have different effects in a regular `for` loop versus a `for..of`

- Moved the case of #252 to be completely ignored instead of not fixed since this is valid example of that, that can't be moved to `for..of`: 
```
for (let i = 0;i < arr.length;i++) {
  let el = arr[i];
  if (el.skipNext) {
    i++;
  }
  console.log(el);
}
```

So now, any time the index is set to in the loop body, the rule will ignore it

## Question

When there is this case: 
```
/* eslint unicorn/no-for-loop: 2 */
for (let i = 0; i < arr.length; i += 1) {
  console.log(arr[i])
}
```

it can be fixed with 
```
/* eslint unicorn/no-for-loop: 2 */
for (const value of arr) {
  console.log(value)
}
```

But the name `value` is arbitrary since the code doesn't provide a name itself. Do we want to keep it `value`? Or not fix? Or provide an option?

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -1,6 +1,7 @@
 'use strict';
 const getDocsUrl = require('./utils/get-docs-url');
 
+const defaultElementName = 'element';
 const isLiteralValue = value => node => node && node.type === 'Literal' && node.value === value;
 const isLiteralZero = isLiteralValue(0);
 const isLiteralOne = isLiteralValue(1);
@@ -311,7 +312,7 @@ const create = context => {
 					const shouldGenerateIndex = isIndexVariableUsedElsewhereInTheLoopBody(indexVariable, bodyScope, arrayIdentifierName);
 
 					const index = indexIdentifierName;
-					const element = elementIdentifierName || 'value';
+					const element = elementIdentifierName || defaultElementName;
 					const array = arrayIdentifierName;
 
 					const replacement = shouldGenerateIndex ?
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -326,7 +326,7 @@ const create = context => {
 						], replacement),
 						...arrayReferences.map(reference => {
 							if (reference === elementReference) {
-								return;
+								return null;
 							}
 
 							return fixer.replaceText(reference.identifier.parent, element);
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -325,12 +325,12 @@ const create = context => {
 						], replacement),
 						...arrayReferences.map(reference => {
 							if (reference === elementReference) {
-								return null;
+								return;
 							}
 
 							return fixer.replaceText(reference.identifier.parent, element);
 						}),
-						elementNode ? fixer.removeRange(getRemovalRange(elementNode, sourceCode)) : null
+						elementNode && fixer.removeRange(getRemovalRange(elementNode, sourceCode))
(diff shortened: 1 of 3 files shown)


sindresorhus:
> But the name value is arbitrary since the code doesn't provide a name itself. Do we want to keep it value? Or not fix? Or provide an option?

Lets call it `element`. It should be fixable. I don't see the need to customize that as the naming depends on the context. User's should just give it a proper name afterwards.

rules/no-for-loop.js
```diff
@@ -304,9 +323,15 @@ const create = context => {
 							node.init.range[0],
 							node.update.range[1]
 						], replacement),
-
-						fixer.removeRange(getRemovalRange(elementVariableDeclaration, sourceCode))
-					];
+						...arrayReferences.map(reference => {
+							if (reference === elementReference) {
+								return null;
```
sindresorhus:
```suggestion
								return;
```

rules/no-for-loop.js
```diff
@@ -304,9 +323,15 @@ const create = context => {
...
+							}
+
+							return fixer.replaceText(reference.identifier.parent, element);
+						}),
+						elementNode ? fixer.removeRange(getRemovalRange(elementNode, sourceCode)) : null
```
sindresorhus: I prefer using `undefined`.

```suggestion
						elementNode && fixer.removeRange(getRemovalRange(elementNode, sourceCode))
```

rules/no-for-loop.js
karaggeorge: This is caught by XO:

`329:9   Arrow function expected a return value.  array-callback-return`

sindresorhus: Right, you can `return undefined;` then.

karaggeorge: 750aa1c

sindresorhus:
@golopot Hehe. You were seconds too late.

@karaggeorge Can you handle that?

karaggeorge:
@sindresorhus @golopot Yeah, that's weird. I'll fix and open a new PR for it 😄

karaggeorge:
@sindresorhus @golopot Ended up being an issue of getting the references in a scope (doesn't include child scopes). Fixed here: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/303

PR #302, opened 2019-05-30
Add `expiring-todo-comments` rule

### LINKED ISSUE #238: Rule proposal: Expiring TODO comments
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/238)

For example, `// TODO [2018-05-01]: Do this thing` would be reported from that date and onwards. The date format should be `YYYY-MM-DD` ([ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)).

Would also be nice to support semver, for example, `// TODO [>=2]: Do this thing`, which would make it report from version 2 and later.

> There is a $100.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/238).

lubien:
Fixes #238

* Due date check.
* Own package version check.
* Dependency package version check.
* Dependency inclusion checks.

I've explored the related issue and picked up some features the seem to the to the liking of those who participated.

If anything let me know so I can work around it.


docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,56 @@
+# Add conditions to TODO comments to make them error
+
+Makes possible to pass arguments to TODO and FIXME comments to trigger errors.
+
+Currently supporting:
+
+* `[YYYY-MM-DD]` to define a due date.
+* `[> 1]` or `[>= 2]` to expire at some version (from package.json).
```
sindresorhus: I think we should require there to be no whitespace here to reduce ambiguity. Making it stricter also leaves the door open for other conditions in the future.
```suggestion
* `[>1]` or `[>=2]` to expire at some version (from package.json).
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,56 @@
...
+* `[+package]` or `[-package]` to expire when you add/remove a package.
+* `[package > 1]` or `[package >= 2]` to expire when a package hits some version.
```
sindresorhus: This should use the npm install syntax for versions.
```suggestion
* `[package@>1]` or `[package@>=2]` to expire when a package hits some version.
```

lubien:
About the build: apparently [no issue is related to my rule](https://api.travis-ci.org/v3/job/538820440/log.txt) and even when I run tests again without my commit it doesn't pass too. 

About the change requests. I'll work on that right now!

docs/rules/expiring-todo-comments.md
lubien: Agreed

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,56 @@
+# Add conditions to TODO comments to make them error
+
+Makes possible to pass arguments to TODO and FIXME comments to trigger errors.
+
+Currently supporting:
+
+* `[YYYY-MM-DD]` to define a due date.
```
itaisteinherz: What about supporting date ranges (e.g. `[startDate-endDate]`)? It could be useful in some cases.

sindresorhus: > It could be useful in some cases.

What cases?

sindresorhus:
> About the build: apparently no issue is related to my rule and even when I run tests again without my commit it doesn't pass too.

Fixed: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/91efcba204d63f1fd9838e3866a7a21e1cbe82fa

sindresorhus:
@lubien Note: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/238#issuecomment-461078819

> If this option is enabled, should it always warn for TODOs without a date / version?
>> Yes, it would have to replace the functionality of no-warning-comments as it conflicts.

You need to document that the user has to disable the built-in TODO rule.

And this rule needs to handle what the built-in rule does. I think there's a way to import the built-in rule and use that. https://github.com/typescript-eslint/typescript-eslint does that to some rules.

docs/rules/expiring-todo-comments.md
lubien: Fixed that. Working on the next one.

docs/rules/expiring-todo-comments.md
itaisteinherz: I'm not sure, but others might come up with some 😅

Anyways, the rule looks good to me overall (aside for the suggestions you made above and other minor implementation improvements).

docs/rules/expiring-todo-comments.md
lubien: Fixed that one now

lubien:
> And this rule needs to handle what the built-in rule does. I think there's a way to import the built-in rule and use that. [typescript-eslint/typescript-eslint](https://github.com/typescript-eslint/typescript-eslint) does that to some rules.

I'll look into that

lubien:
> I think there's a way to import the built-in rule and use that.

In a nutshell:

```js
const baseRule = require('eslint/lib/rules/no-warning-comments');
// ...

const rules = baseRule(context);

return {
	Program() {
		rules.Program();
		const comments = sourceCode.getAllComments();
		comments.filter(token => token.type !== 'Shebang').forEach(processComment);
	}
};
```

But it does make this scenario happen:

```js
  AssertionError [ERR_ASSERTION] (AssertionError) {
    actual: 2,
    code: 'ERR_ASSERTION',
    expected: 1,
    generatedMessage: false,
    operator: 'strictEqual',
    message: `Should have 1 error but had 2: [ { ruleId: 'expiring-todo-comments',␊
        severity: 1,␊
        message: 'Unexpected \\'todo\\' comment.',␊
        line: 1,␊
        column: 1,␊
        nodeType: 'Line',␊
        endLine: 1,␊
        endColumn: 30 },␊
      { ruleId: 'expiring-todo-comments',␊
        severity: 1,␊
        message: 'You have a TODO that past due date 2000-01-01',␊
        line: 1,␊
        column: 1,␊
        nodeType: null,␊
        messageId: 'expiredTodo',␊
        endLine: 1,␊
        endColumn: 30 } ]`,
  }

```

When I force-run the default no-warn-comments as my rule eslint takes it's errors as mine's.

I'm a little bit lost onto what do next over this information.

lubien:
> > If this option is enabled, should it always warn for TODOs without a date / version?

> Yes, it would have to replace the functionality of no-warning-comments as it conflicts.

I think I got a better grasp of the situation now. I'll do some tinkering to make the default rule as fallback to mine's.

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,285 @@
...
+	const sourceCode = context.getSourceCode();
+	const comments = sourceCode.getAllComments();
+	const unnusedComments = comments
+		.filter(token => token.type !== 'Shebang')
+		.filter(processComment);
+
+	// This is highly dependable on eslint's no-warning-comments implementation.
+	// What I do is patch the parts I know the rule will use, the getAllComments.
+	// Since I have priority, I leave only the comments that I didn't use to it.
+	const fakeContext = {...context, getSourceCode() {
```
lubien: @sindresorhus I'd love your comment on this approach.

Unlike typescript-eslint [approach](https://github.com/typescript-eslint/typescript-eslint/blob/4edf0d71ef0a8a3370165e804dd693033d6207bc/packages/eslint-plugin/src/rules/no-unused-vars.ts#L42) where they could extend the visitor functions, I had not such ideal scenario as the default rule read nothing but the context.

lubien:
I might have to setup this repo eslint rule to check for dates if I wanna make this test work. Any recommendation onto how I should do it @sindresorhus ?

docs/rules/expiring-todo-comments.md
lubien: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/302/commits/4955129775d322f71103a429fd702ca3b5ede7be

docs/rules/expiring-todo-comments.md
lubien: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/302/commits/024c8657e923a30832cc9c82fc87628ec9952fde

elliot-nelson:
Seems pretty cool for projects that only have a master branch.

For projects with legacy branches (ie you maintain and sometimes release minor versions on older major branches), you likely end up in this scenario where the bug is long since fixed in master but not in other branches that are security-patch-only.

(I don't have a good fix for that, other than not using it in that kind of project I suppose!)

sindresorhus:
@lubien Can you add a note about the above ⬆️ ? So people are careful if they have that workflow.

lubien:
@sindresorhus sure. Also I realized I forgot to document about disabling the default rule, will do both.

yakov116:
Would adding a specific branch name be an option? (Ignore it if it's not the specified branch)

lubien:
> Would adding a specific branch name be an option? (Ignore it if it's not the specified branch)
> Side note : thank you so much for working on this, it will really improve my work flow!

You welcome :D

If @sindresorhus is ok with it I can work on this. Perhaps option disabled by default then you can specify an array of branches for this to work on.

yakov116:
@futpib 	how?

futpib:
@yakov116 by changing xo or eslint [`rules` option](https://github.com/xojs/xo#rules)

yakov116:
@lubien 	sorry for going off topic. I tried google and reading the docs and I dont see a branch option.

Unless you mean that I should change the config on that branch?

futpib:
Yes, that's what I mean. If you want different branches linted differently, in my opinion, they should simply have different linter configs.

Otherwise, imagine if a rule was to work only on `master` and a branch `feature` had a violation of that rule that did not show up in CI. Later you merge the `feature` branch into the `master`, and the `master` now does not pass the CI check. (Both branches were "green" before merge, but `master` became "red" after the merge)

yakov116:
@futpib Good point

lubien:
// @sindresorhus 

Just did a rebase to fix the package.json conflict.

Sorry for the @ since you're busy but I this has been done for a while and the revisions seems done.

Anything I should go for now?

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,357 @@
...
+	const rules = baseRule.create(fakeContext);
+
+	function processComment(comment) {
+		const parsed = parseTodoWithArgs(comment.value, options);
+
+		if (!parsed) {
+			return true;
+		}
+
+		// Count if there where valid props
```
MrHen:
```suggestion
		// Count if there are valid props
```

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,357 @@
...
+			dependencies = [],
+			engines = []
+		} = parsed;
+
+		if (dates.length > 1) {
+			uses++;
+			context.report({
+				node: null,
+				loc: comment.loc,
+				messageId: 'avoidMultipleDates',
```
MrHen: Can you move the message ids to constants at the top of the file? E.g., `const MESSAGE_ID_MULTIPLE_DATES = 'avoidMultipleDates';`

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,357 @@
...
+		messages: {
+			avoidMultipleDates: 'Avoid using multiple expiration dates for TODO {{ expirationDates }}',
+			expiredTodo: 'You have a TODO that past due date {{ expirationDate }}',
+			reachedPackageVersion: 'You have a TODO that past due package version {{ comparison }}',
+			avoidMultiplePackageVersions: 'Avoid asking multiple package versions for TODO {{ versions }}',
+			havePackage: 'You have a TODO that is deprecated since you installed {{ package }}',
+			dontHavePackage: 'You have a TODO that is deprecated since you uninstalled {{ package }}',
+			versionMatches: 'You have a TODO match for version for package {{ comparison }}',
+			engineMatches: 'You have a TODO match for engine version {{ comparison }}'
+		},
```
MrHen: A few phrasing tweaks:

```
		messages: {
 			avoidMultipleDates: 'Avoid using multiple expiration dates in TODO: {{ expirationDates }}',
 			expiredTodo: 'There is a TODO that is past due date: {{ expirationDate }}',
 			reachedPackageVersion: 'There is a TODO that is past due package version: {{ comparison }}',
 			avoidMultiplePackageVersions: 'Avoid using multiple package versions in TODO: {{ versions }}',
 			havePackage: 'There is a TODO that is deprecated since you installed: {{ package }}',
 			dontHavePackage: 'There is a TODO that is deprecated since you uninstalled: {{ package }}',
 			versionMatches: 'There is a TODO match for package version: {{ comparison }}',
 			engineMatches: 'There is a TODO match for engine version: {{ comparison }}'
 		},
```

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+		  * TODO [2200-12-12]: Way
+		  */`
+	],
+	invalid: [
+		// Commented as the ignoreDatesOnPR setting make it not fail
+		// {
+		// 	code: '// TODO [2000-01-01]: too old',
+		// 	errors: [expiredTodoError('2000-01-01')],
+		// 	output: '// TODO [2000-01-01]: too old'
+		// },
```
MrHen: I don't see much point in taking this if this rule doesn't work. I think this is the primary use case.

If the only issue its hat the integration tests have comments that would no longer be valid then we should fix up the integration tests.

rules/expiring-todo-comments.js
lubien: Changed. Thanks.

rules/expiring-todo-comments.js
lubien: I've missed this, thanks for pointing out. Now fixed.

rules/expiring-todo-comments.js
lubien: Added.

test/expiring-todo-comments.js
lubien: What happens is that there's a default option called `ignoreDatesOnPR` that's `true` to prevent unwanted TODOs breaking PRs around:

> Isn't it better that master breaks than the PR submitter having to deal with it?

https://github.com/sindresorhus/eslint-plugin-unicorn/issues/238#issuecomment-497278040

For the commented section show the proper error in my PR this setting should be made `false`.

For now I removed the code, if that's the desired scenario.

docs/rules/expiring-todo-comments.md
lubien: I hope you don't mind I mark this a solved.

test/expiring-todo-comments.js
MrHen: You can set the `ignoreDatesOnPR` flag to `true` for one particular testcase. [An example from a different rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/no-keyword-prefix.js#L155):

```
		{
			code: 'foo.newFoo = 2;',
			options: [{checkProperties: true}],
			errors: [errorNew]
		},
```

There does needs to be a test covering the outdated date scenario.

lubien: Thank you very much, now it's on the PR

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,99 @@
...
+ * TODO [2200-12-25]: Way
+ */
+```
+
+
+## Options
+
+### ignoreDatesOnPR
+
+Disables time checks during pull requests. Default: `true`.
```
sindresorhus: This should make it clear **why** it's beneficial to disable the time checks on pull requests.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,98 @@
+# Add conditions to TODO comments to make them error
+
+Makes it possible to pass arguments to TODO and FIXME comments to trigger errors.
+
+This rule also implements [`eslint/no-warning-comments`](https://eslint.org/docs/rules/no-warning-comments) behavior so you should disable that in favor of using this.
+
+For projects with legacy branches, such as long term supported older versions, there might be situations where you fix TODOs in the master branch but on the other branches you leave it missing. Choose carefully which projects you would want this feature.
```
sindresorhus: This should be further down. The rule should be explained first. I also thing this paragraph could be clearer written. It doesn't explain **why** it's problematic for legacy branches.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,98 @@
...
+
+Supported conditions:
```
sindresorhus: Before listing the conditions, you should better explain what exactly this rule does and what problems it solves.

sindresorhus: Generally, the docs here could be improved a lot. And include more specifics.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,104 @@
...
+ * TODO [2200-12-25]: Yet
+ * TODO [2200-12-25]: Another
+ * TODO [2200-12-25]: Way
+ */
+```
+
+
+## Options
+
+### ignoreDatesOnPR
```
sindresorhus:
```suggestion
### ignoreDatesOnPullRequests
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,108 @@
...
+
+For projects with legacy branches, such as long term supported older versions, there might be situations where you fix TODOs in the master branch but on the other branches you leave it missing. Choose carefully which projects you would want this feature.
+
+Supported conditions:
+
+- `[YYYY-MM-DD]` to define a due date.
+- `[>1]` or `[>=2]` to expire at some version (from package.json). No whitespace.
+- `[+package]` or `[-package]` to expire when you add/remove a package.
+- `[package@>1]` or `[package@>=2]` to expire when a package hits some version. No whitespace.
+- `[engines node>8]` or `[engines node>=8]` to expire when package bump supported engines. Don't use spaces between the comparison.
```
sindresorhus: I think each of the conditions here should mention some use-cases for it. It might not be apparent to users what they could use them for.

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,390 @@
...
+					type: 'string'
+				}
+			},
+
+			ignoreDatesOnPR: {
+				type: 'boolean'
+			}
+		}
+	}
+];
```
sindresorhus: Set the default values in the schema here instead of in the code.

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,390 @@
...
+	// Currently being ignored as integration tests pointed
+	// some TODO comments have [random data like this]
+	return {
+		type: 'unknowns',
+		value: argString
+	};
+}
+
+function reachedDate(past) {
+	const now = new Date().toISOString().substring(0, 10);
```
sindresorhus: `substring` => `slice`

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,390 @@
...
+			[MESSAGE_ID_VERSION_MATCHES]:
+				'There is a TODO match for package version: {{ comparison }}',
+			[MESSAGE_ID_ENGINE_MATCHES]:
+				'There is a TODO match for engine version: {{ comparison }}'
+		},
+		schema
+	}
+};
+
+function parseTodoWithArgs(str, {terms}) {
```
sindresorhus: Don't use abbreviations.
```suggestion
function parseTodoWithArguments(string, {terms}) {
```

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,390 @@
...
+	const TODO_RE = new RegExp(`[${terms.join('|')}][\\s\\S]*\\[([^}]+)\\]`, 'i');
+	const result = TODO_RE.exec(str);
+
+	if (!result) {
+		return false;
+	}
+
+	const rawArgs = result[1];
```
sindresorhus:
```suggestion
	const rawArguments = result[1];
```

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+});
+
+const avoidMultipleDatesError = expirationDates => ({
+	ruleId: 'expiring-todo-comments',
+	message: `Avoid using multiple expiration dates in TODO: ${expirationDates}`
+});
+
+const havePackageError = pkg => ({
+	ruleId: 'expiring-todo-comments',
+	message: `There is a TODO that is deprecated since you installed: ${pkg}`
```
sindresorhus: The messages should include the actual TODO message too.

```
// TODO [2200-12-12]: This part
```

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+		},
+		{
+			code: '// TODO [+read-pkg]: when you install `read-pkg`',
+			errors: [havePackageError('read-pkg')],
+			output: '// TODO [+read-pkg]: when you install `read-pkg`'
+		},
+		{
+			code: '// TODO [-popura]: when you uninstall `popura`',
+			errors: [dontHavePackageError('popura')],
+			output: '// TODO [-popura]: when you uninstall `popura`'
```
sindresorhus: Don't specify the `output`. We're not fixing anything.

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+			output: '// TODO [2000-01-01]: too old'
+		},
+		{
+			code: '// TODO [2200-12-12, 2200-12-12]: Multiple dates',
+			errors: [avoidMultipleDatesError('2200-12-12, 2200-12-12')],
+			output: '// TODO [2200-12-12, 2200-12-12]: Multiple dates'
+		},
+		{
+			code: '// TODO [>1]: if your package.json version is > 1',
+			errors: [reachedPackageVersionError('> 1')],
```
sindresorhus:
```suggestion
			errors: [reachedPackageVersionError('>1')],
```

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+		{
+			code: '// TODO [engines node >=1]: Default rule if space.',
+			errors: [noWarningCommentError('todo')],
+			output: '// TODO [engines node >=1]: Default rule if space.'
+		},
+		{
+			code: '// TODO [engines node>= 1]: Default rule if space.',
+			errors: [noWarningCommentError('todo')],
+			output: '// TODO [engines node>= 1]: Default rule if space.'
+		}
```
sindresorhus: Needs more tests with combination of conditions.

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+			errors: [noWarningCommentError('todo')],
+			output: '// TODO [semver @>=1]: Default rule if space.'
+		},
+		{
+			code: '// TODO [engines node >=1]: Default rule if space.',
+			errors: [noWarningCommentError('todo')],
+			output: '// TODO [engines node >=1]: Default rule if space.'
+		},
+		{
+			code: '// TODO [engines node>= 1]: Default rule if space.',
```
sindresorhus: This should fail loud and not silently, and inform the user what they did wrong.

test/expiring-todo-comments.js
```diff
@@ -0,0 +1,173 @@
...
+			errors: [versionMatchesError('read-pkg >= 5.1.1')],
+			output: '// TODO [read-pkg@>=5.1.1]: when `read-pkg` version is >= 5.1.1'
+		},
+		{
+			code: '// TODO [> 1]: Default rule if space',
+			errors: [noWarningCommentError('todo')],
+			output: '// TODO [> 1]: Default rule if space'
+		},
+		{
+			code: '// TODO [semver>1]: Default rule if no @.',
```
sindresorhus: This should fail loud and not silently, and inform the user what they did wrong.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,108 @@
...
+
+Makes it possible to pass arguments to TODO and FIXME comments to trigger errors.
+
+This rule also implements [`eslint/no-warning-comments`](https://eslint.org/docs/rules/no-warning-comments) behavior so you should disable that in favor of using this.
+
+For projects with legacy branches, such as long term supported older versions, there might be situations where you fix TODOs in the master branch but on the other branches you leave it missing. Choose carefully which projects you would want this feature.
+
+Supported conditions:
+
+- `[YYYY-MM-DD]` to define a due date.
```
sindresorhus: Should mention that the dates are UTC.

docs/rules/expiring-todo-comments.md
sindresorhus: I think we should make this:

```suggestion
- `[engines:node>8]` or `[engines:node>=8]` to expire when package bump supported engines. Don't use spaces between the comparison.
```

So completely remove whitespace from the conditions.

lubien: Should whitespaces around `:` be valid?

1. `[engines : node>8]`
2. `[engines: node>8]`
3. `[engines :node>8]`
4. Multiple whitespaces

rules/expiring-todo-comments.js
lubien: I'm kinda lost on this one, can you give me an example? Currently in this repo I found:

I'm doing like this:
```js
// catch-error-name.js
const create = context => {
	const options = {
		name: 'error',
		caughtErrorsIgnorePattern: '^_$',
		...context.options[0]
	};
```

```js
// filename-case.js
const create = context => {
	const chosenCases = getChosenCases(context);

// ...
function getChosenCases(context) {
	const option = context.options[0] || {};
```

```js
// no-keyword-prefix.js
const create = context => {
	const options = prepareOptions(context.options[0]);
```

```js
// prefer-add-event-listener.js
const create = context => {
	const options = context.options[0] || {};
```

This method sounds better:
```js
// prevent-abbreviations.js
const create = context => {
	const options = prepareOptions(context.options[0]);
```

After going to the site os JSON Schema I got into [this example with default](https://json-schema.org/understanding-json-schema/structuring.html?highlight=default#using-id-with-ref) I got this:

```js

const schema = [
	{
		type: 'object',
		properties: {
			terms: {
				type: 'array',
				default: ['todo', 'fixme', 'xxx'],
				items: {
					type: 'string'
				}
			},
			ignoreDatesOnPullRequests: {
				type: 'boolean',
				default: true
			}
		},
	}
];
```

But doesn't feel right since tests get `[]` as ctx.options and fail unless I handle on code. Any tip?

MrHen: I've also seen issues with `default` during test runs. I haven't dug into it yet but it's possible that the schema resolution is being skipped during tests.

test/expiring-todo-comments.js
lubien: Should I drop support for whitespaces here?

rules/expiring-todo-comments.js
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

test/expiring-todo-comments.js
lubien: Done

test/expiring-todo-comments.js
lubien: Done

test/expiring-todo-comments.js
lubien: Done

test/expiring-todo-comments.js
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Using it now

rules/expiring-todo-comments.js
lubien: Done

rules/expiring-todo-comments.js
lubien: Done

test/expiring-todo-comments.js
lubien: Done

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,108 @@
...
+- `[>1]` or `[>=2]` to expire at some version (from package.json). No whitespace.
+- `[+package]` or `[-package]` to expire when you add/remove a package.
+- `[package@>1]` or `[package@>=2]` to expire when a package hits some version. No whitespace.
+- `[engines node>8]` or `[engines node>=8]` to expire when package bump supported engines. Don't use spaces between the comparison.
+
+Info:
+- Multiple conditions can be separated by comma: `[YYYY-MM-DD, +foo-package, engines node>=8, bar-package@>=2]`.
```
sindresorhus: Don't forget to keep the docs in sync with changes.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,108 @@
...
+// FIXME [2200-12-25]: Too long... Can you feel it?
+
+// TODO (lubien) [2200-12-12]: Too long... Can you feel it?
+// FIXME [2200-12-25] (lubien): Too long... Can you feel it?
+
+// TODO [-read-pkg]: We actually use this. If we remove this package I'll error.
+// TODO [+popura]: I think we wont need a broken package.
+
+// TODO [semver@>1000]: Welp hopefully we wont get at that.
+// TODO [semver@>=1000]: Welp hopefully we wont get at that.
```
sindresorhus: Use another package than `semver` for all the examples here. It can cause confusion and some might think it just indicates that it's a semver version, not the package. Use `lodash` instead. It's known by everyone.

docs/rules/expiring-todo-comments.md
sindresorhus: Having the word `or` here makes it sound like only those two semver notation are supported.

sindresorhus:
For consistency, `[engines:node>=8]` should be `[engines:node@>=8]`.

sindresorhus:
I'm also not happy with the `engines` keyword. It's should at minimum be singular, but the word `runtime` might be better. Would be nice to think of some alternatives to the syntax here too. // @futpib @yakov116 @MrHen Any ideas?

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,108 @@
+# Add conditions to TODO comments to make them error
```
sindresorhus: Whether it *errors* or *warns* is up to the user, so the title needs some improvement.

rules/expiring-todo-comments.js
sindresorhus: We should upgrade to ESLint 6, maybe we're lucky and it's been fixed already: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/326

rules/expiring-todo-comments.js
```diff
@@ -0,0 +1,470 @@
...
+			[MESSAGE_ID_AVOID_MULTIPLE_PACKAGE_VERSIONS]:
+				'Avoid using multiple package versions in TODO: {{versions}}. {{message}}',
+			[MESSAGE_ID_HAVE_PACKAGE]:
+				'There is a TODO that is deprecated since you installed: {{package}}. {{message}}',
+			[MESSAGE_ID_DONT_HAVE_PACKAGE]:
+				'There is a TODO that is deprecated since you uninstalled: {{package}}. {{message}}',
+			[MESSAGE_ID_VERSION_MATCHES]:
+				'There is a TODO match for package version: {{comparison}}. {{message}}',
+			[MESSAGE_ID_ENGINE_MATCHES]:
+				'There is a TODO match for engine version: {{comparison}}. {{message}}',
```
sindresorhus: We only support `node`, so we should be more explicit in the message.
```suggestion
				'There is a TODO match for Node.js version: {{comparison}}. {{message}}',
```

MrHen:
> I'm also not happy with the `engines` keyword. It's should at minimum be singular, but the word `runtime` might be better. Would be nice to think of some alternatives to the syntax here too. // @futpib @yakov116 @MrHen Any ideas?

`engines` is what [`package.json`](https://docs.npmjs.com/files/package.json#engines) calls it. I think the pluralization is what makes it seem awkward. What about `package-engine`?

docs/rules/expiring-todo-comments.md
lubien: Sure. I was wrapping up code before going for the docs.

docs/rules/expiring-todo-comments.md
lubien: Actually it is.

On my first draft I used also '=', '<' and '<='. Is this desirable?

rules/expiring-todo-comments.js
lubien: How about `npm` as engine?

Should we report when we receive something like `iojs` then?

futpib:
I'm fine with `engines` as it makes it clear that this is a check against the `engines` field of `package.json`.

test/expiring-todo-comments.js
lubien: Done

MrHen:
I do prefer `engine` over `engines` since the check is only for a particular engine.

sindresorhus:
Ok. Let’s go with `engine` (singular).

lubien:
> Ok. Let’s go with `engine` (singular).

Ok.

I'm about to finish my new draft of the docs and I'll push before tackling the new code reviews so you guys can take a look beforehand.

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

lubien:
> For consistency, `[engines:node>=8]` should be `[engines:node@>=8]` (don't forget to update all docs and examples)

Done.

> Ok. Let’s go with `engine` (singular).

Done

docs/rules/expiring-todo-comments.md
yakov116: @lubien 	Looks like you by mistake undid your change

lubien: Oh. Might have happened while I switched machines. Just a sec.

lubien: If anything just let me know please :)

docs/rules/expiring-todo-comments.md
yakov116: There are some Grammer mistakes but will wait for @sindresorhus to review first.

lubien: I tried to use a grammar checking tool but I knew we'd miss some, sorry 😋

docs/rules/expiring-todo-comments.md
sindresorhus: Nah, we can keep it simple. I would just make it clear with a sentence, that while it looks like we support all of semver (which I just assumed when seeing semver syntax), we only support ...

rules/expiring-todo-comments.js
sindresorhus: > How about npm as engine?

We don't support that right now and I don't really see the point in supporting it at all.

> Should we report when we receive something like iojs then?

No, just ignore it. Nobody is using iojs anymore. Not even worth reporting as invalid.

docs/rules/expiring-todo-comments.md
```diff
@@ -1,67 +1,232 @@
...
-This rule also implements [`eslint/no-warning-comments`](https://eslint.org/docs/rules/no-warning-comments) behavior so you should disable that in favor of using this.
+From [ESLint's documentation](https://eslint.org/docs/rules/no-warning-comments):
 
-For projects with legacy branches, such as long term supported older versions, there might be situations where you fix TODOs in the master branch but on the other branches you leave it missing. Choose carefully which projects you would want this feature.
+> Developers often add comments to code which is not complete or needs review.
 
-Supported conditions:
+TODO comments are useful when a piece of code needs some work.
+Unfortunately these can be easily forgotten as it's easy to forget to
+track them, leaving dangling tasks to be found at later random moments.
```
sindresorhus: Don't hard-wrap.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+
+From [ESLint's documentation](https://eslint.org/docs/rules/no-warning-comments):
+
+> Developers often add comments to code which is not complete or needs review.
+
+TODO comments are useful when a piece of code needs some work.
+Unfortunately these can be easily forgotten as it's easy to forget to
+track them, leaving dangling tasks to be found at later random moments.
+
+With this rule a TODO can have right from the beginning a condition
```
sindresorhus:
```suggestion
With this rule, a TODO can have a condition right from the beginning
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+to define it's lifespan. When the condition is met, ESLint will take care
```
sindresorhus:
```suggestion
to define its lifespan. When the condition is met, ESLint will take care
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+of reporting that there's work to be done.
+
+This rule also define that **there must be no TODO comment without conditions** so that you should take more care before simply adding
```
sindresorhus:
```suggestion
This rule also defines that **there must be no TODO comment without conditions** so that you should take more care before simply adding
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+tasks with no life expectancy. For more information read the
+section [`eslint/no-warning-comments`](#disallow-warning-comments-no-warning-comments) below.
+
+Conditions quick overview:
+
+- Expire after a **specific date**.
+- Expire when **package.json** reaches a specific **version**.
```
sindresorhus:
```suggestion
- Expire when **your package** (package.json) reaches a specific **version**.
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+- Expire when a package.json **`engines`** property reaches a specific **version**.
+- Expire when you **install/uninstall** a specific **package**.
+- Expire when a **package** reaches a specific **version**.
+
+## Conditions
+
+### Expiry Date
+
+Using a date as condition, a TODO will only work as longs as this date
```
sindresorhus:
```suggestion
Using a date as condition, a TODO will only work as long as this date
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+// TODO [2019-11-15]: Refactor this code before the sprint ends.
+// TODO (@lubien) [2019-07-18]: When John deliver his code. I can reuse it.
+// TODO [2019-08-10] I must refactor this for sure before I deliver.
+```
+
+Dates are [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Dates) and
+use [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) timezone.
+
+Multiple dates on the same TODO are not valid and will be reported.
+There's no meaning in having two dates as the closest one will always be
```
sindresorhus:
```suggestion
There's no point in having two dates as the closest one will always be
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+the one to report.
```
sindresorhus:
```suggestion
the one that is reported.
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+```
+
+## Disallow Warning Comments (no-warning-comments)
+
+This rule implements [`eslint/no-warning-comments`](https://eslint.org/docs/rules/no-warning-comments).
+
+The sole difference is that first we check for **valid conditions**
+to apply this rule. If no valid condition is met we fallback to
+`eslint/no-warning-comments` and you'll see something like
+`Unexpected 'todo' comment.`
```
sindresorhus: I'm wondering if we should add a setting to disable this. Imagine you would want condition TODO comment to error, but normal TODO comments to just warn. Unfortunately, there's no way for rules to force a report to be a warning. So the best we can do is to allow disabling it, so users can use the `eslint/no-warning-comments` rule in addition with a warning level.

readme.md
```diff
@@ -84,6 +85,7 @@ Configure it in `package.json`.
 - [custom-error-definition](docs/rules/custom-error-definition.md) - Enforce correct `Error` subclassing. *(fixable)*
 - [error-message](docs/rules/error-message.md) - Enforce passing a `message` value when throwing a built-in error.
 - [escape-case](docs/rules/escape-case.md) - Require escape sequences to use uppercase values. *(fixable)*
+- [expiring-todo-comments](docs/rules/expiring-todo-comments.md) - Add conditions to TODO comments to make them error.
```
sindresorhus: This should be in sync with the docs title.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+// TODO [>=1]: If your package.json version is >= 1
+// TODO [>1, >2]: Multiple package versions won't work
+
+// TODO [+already-have-pkg]: Since we already have it, this reports.
+// TODO [-we-dont-have-this-package]: Since we don't have, trigger report.
+
+// TODO [read-pkg@>1]: When `read-pkg` version is > 1 don't forget to do this
+// TODO [read-pkg@>=5.1.1]: When `read-pkg` version is >= 5.1.1 don't forget to do that
+
+// TODO [engine:@node>=8]: Whoops we are already supporting it!
```
sindresorhus:
```suggestion
// TODO [engine:node@>=8]: Whoops we are already supporting it!
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+no stray TODOs on your code, you should strive for best pratices. Don't just put
+TODOs and leave then live forever. Define conditions to justify the presence of
+warning comments.
+
+With that in mind, **you should disable** that ESLint rule in favor of this one
+as you will get the same behavior and more.
+
+## Legacy Branches
+
+Although this rule works just fine when you maintain a master branch it gets trickier
```
sindresorhus:
```suggestion
Although this rule works just fine when you maintain a `master` branch, it gets trickier
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+when you have legacy branches that happen to have unclosed TODOs.
+
+Assume you maintain a master branch at a version such as 10 and always keep working on it
```
sindresorhus:
```suggestion
Imagine you maintain a `master` branch at a version such as 10 and always keep working on it
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+Dates are [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Dates) and
+use [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) timezone.
+
+Multiple dates on the same TODO are not valid and will be reported.
+There's no meaning in having two dates as the closest one will always be
+the one to report.
+
+### Package Version
+
+If you're working on a NPM package, you should have a root level package.json
```
sindresorhus:
```suggestion
If you're working on a npm package, you should have a root level package.json
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+with a [`version`](https://docs.npmjs.com/files/package.json#version) field
+as specified from NPM itself. This condition takes place under this value.
```
sindresorhus:
```suggestion
as specified by npm. This condition takes place under this value.
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+
+By assigning a condition to compare a package.json `version` the TODO will
```
sindresorhus:
```suggestion
By assigning a condition to compare a package.json `version`, the TODO will
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,286 @@
...
+// TODO [engine:node@>=8]: We can use async/await now.
+// FIXME [engine:node@>=20.0.0] Hey, node can use import/export now, we should refactor.
+// XXX @lubien [engine:npm@>4.0]: Now npm have [feature] we should use it.
+```
+
+Argument versions should be [semver](https://semver.org/) compatible such as: `1.2.45`, `5.3`, `1`.
+
+### Dependency Presence
+
+As a developer using NodeJS you should be know to package.json
```
sindresorhus:
```suggestion
As a developer using Node.js, you should be know to package.json
```

Not sure what you're trying to say here.

docs/rules/expiring-todo-comments.md
lubien: How about: `allowWarningComments`? Defaults to true?

docs/rules/expiring-todo-comments.md
lubien: Rephrased

> As a Node.js developer, you probably already know about package.json dependencies and devDependencies.

lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

readme.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

rules/expiring-todo-comments.js
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

docs/rules/expiring-todo-comments.md
lubien: Done

lubien:
Fixed reviews.

Added block comment support.

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,262 @@
...
+// TODO [-vue-function-api]: When we remove `vue-function-api` we should refactor this.
+// FIXME [+read-pkg] If we use this package we don't need to use this function below.
+// XXX @lubien [+react, -jquery]: We can use react for this widget instead of JQuery.
+```
+
+This condition is really useful when the code should be looked upon changes in dependencies that are known to happen such as `vue-function-api` that will eventually be part of `vue` itself.
+
+### Dependency Version
+
+Another way to make conditions over dependencies is to look upon versions. This is useful when developers know that further versions of dependencies make changes that impact on their code.
```
yakov116:
```suggestion
Another way to make conditions over dependencies is to look upon versions. This is useful when developers know that changes in further versions of dependencies will impact their code.
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,262 @@
...
+ * TODO [2200-12-25]: Another
+ * TODO [2200-12-25]: Way
+ */
+```
+
+## Options
+
+### ignoreDatesOnPullRequests
+
+Disables time checks during pull requests.
```
yakov116:
```suggestion
Disables `Expiry Date` checks during pull requests.
```

docs/rules/expiring-todo-comments.md
```diff
@@ -0,0 +1,262 @@
...
+
+Sometimes developers may send [Pull Requests](https://help.github.com/en/articles/about-pull-requests) right on time when TODO expiry dates are triggered. This will mean that their code would fail to pass linting and would cause a false positive.
```
yakov116:
```suggestion
Sometimes developers may send [Pull Requests](https://help.github.com/en/articles/about-pull-requests) at a time when TODO expiry dates are triggered. This means that their code would fail to pass linting, causing a false positive.
```

docs/rules/expiring-todo-comments.md
MrHen: This shouldn't use backticks since `Expiry Date` is not code.

lubien:
Thanks @yakov116 you reminded me to document `allowWarningComments`.

I've already made your changes just in case sindresorhus approves.

brettz9:
While @sindresorhus [mentioned](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/238#issuecomment-507033720) adding jsdoc support in a separate PR (item #8 on the list), since it seems you have started work on something that is very similar (your multiline comment support), I thought I'd mention the following about how jsdoc has somewhat different expectations. If you were not trying to support jsdoc, then nevermind, but just mentioning in case you were.

jsdoc tags must have a `@`, and ["todo"](https://jsdoc.app/tags-todo.html) (possibly lower case only) is the official name, and I believe only a single `@todo` is expected for multiline to-dos, or otherwise jsdoc may display it as though each line is a separate to-do. (Also, I might mention that jsdoc does require two asterisks at the beginning of the comment block containing its tags (such as `@todo`), though I don't think there is any reason to require that on your end).

So instead of this format:

```js
/*
 * TODO [2200-12-12]: Yet
 * TODO [engine:node@>=100]: Another
 * TODO [+popura]: Way
*/
```

...jsdoc could perhaps support something like this instead:

```js
/**
 * @todo [2200-12-12]: Yet
 * [engine:node@>=100]: Another
 * [+popura]: Way
*/
```

or perhaps:

```js
/**
 * @todo [2200-12-12][engine:node@>=100][+popura]: Yet
 * Another
 * Way
*/
```

lubien:
> If you were not trying to support jsdoc

I wasn't, actually 😅. I was just fixing something I had a valid test for ever since first draft but never had actually made it work and just realized it yesterday.

But sure, we can make it work with @todo on block comments if that's something wanted.

brettz9:
Per @sindresorhus , and if you feel so inspired, in a subsequent PR/issue then perhaps...

lubien:
> Per @sindresorhus , and if you feel so inspired, in a subsequent PR/issue then perhaps...

I'd love to help improve/maintain this feature

rules/expiring-todo-comments.js
lubien: I'm trying once again now that it's fixed. Already rebased. Will post result soon.

lubien: Added schema defaults except for array. [Apparently they don't use `default` on array on eslint](https://github.com/eslint/eslint/blob/d058a8639a2ef8837974ed2b516c7ce8a79418d6/lib/rules/new-cap.js)

yakov116:
@sindresorhus 	are you able to review this?

yakov116:
Bump

sindresorhus:
Sorry for the long delay. I've been on vacation and had limited computer time.

sindresorhus:
Thanks for all your hard work on this @lubien 🙌 It turned out great 🎉

lubien:
No problem about delay, I also know you had tons of repos to maintain so I was expecting this.

I'm just glad I had this chance to work with you @all

PR #309, opened 2019-06-01
Add `consistent-function-scoping` rule

### LINKED ISSUE #230: Rule proposal: Consistent function scoping
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/230)

A function definition should be placed as close to the top level scope as possible without breaking it's captured values.

## Fail

```js
export function create(context) {
  // does not capture anything from the scope, can be moved to the outer scope
  function isSpecial(node) {
    return node.type === 'Special';
  }

  return isSpecial;
}
```

## Pass

```js
function isSpecial(node) {
  return node.type === 'Special';
}

export function create(context) {
  return isSpecial;
}
```

```js
export function create(context) {
  function isSpecial(node) {
    return node.type === 'Special' && context.isSpecial(node);
  }

  return isSpecial;
}
```

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/230).

MrHen:
Closes #230.

Covers functions and arrow functions. Does NOT cover things like for loops.

### DIFF
--- a/docs/rules/consistent-function-scoping.md
+++ b/docs/rules/consistent-function-scoping.md
@@ -1,6 +1,6 @@
-# Prefer consistent function scoping
+# Move function definitions to the highest possible scope
 
-A function definition should be placed as close to the top-level scope as possible without breaking its captured values.
+A function definition should be placed as close to the top-level scope as possible without breaking its captured values. This improves readability and allows JavaScript engines to better optimize performance.
 
 
 ## Fail
@@ -40,5 +40,20 @@ export function doFoo(foo) {
   }
 
   return doBar;
+}
+```
+
+## Limitations
+
+This rule does not detect or remove extraneous code blocks:
+
+```js
+function doFoo(foo) {
+  {
+		function doBar(bar) {
+			return bar;
+		}
+	}
+  return foo;
 }
 ```
--- a/docs/rules/consistent-function-scoping.md
+++ b/docs/rules/consistent-function-scoping.md
@@ -7,12 +7,12 @@ A function definition should be placed as close to the top-level scope as possib
 
 ```js
 export function doFoo(foo) {
-  // Does not capture anything from the scope, can be moved to the outer scope
-  function doBar(bar) {
-    return bar === 'bar';
-  }
+	// Does not capture anything from the scope, can be moved to the outer scope
+	function doBar(bar) {
+		return bar === 'bar';
+	}
 
-  return doBar;
+	return doBar;
 }
 
 function doFoo(foo) {
@@ -27,21 +27,22 @@ function doFoo(foo) {
 
 ```js
 function doBar(bar) {
-  return bar === 'bar';
+	return bar === 'bar';
 }
 
 export function doFoo(foo) {
-  return doBar;
+	return doBar;
 }
 
 export function doFoo(foo) {
-  function doBar(bar) {
-    return bar === 'bar' && foo.doBar(bar);
-  }
+	function doBar(bar) {
+		return bar === 'bar' && foo.doBar(bar);
+	}
 
-  return doBar;
+	return doBar;
 }
 ```
+
 
 ## Limitations
 
@@ -49,11 +50,12 @@ This rule does not detect or remove extraneous code blocks inside of functions:
 
 ```js
 function doFoo(foo) {
-  {
+	{
 		function doBar(bar) {
 			return bar;
 		}
 	}
-  return foo;
+
+	return foo;
 }
 ```
--- a/docs/rules/consistent-function-scoping.md
+++ b/docs/rules/consistent-function-scoping.md
@@ -45,7 +45,7 @@ export function doFoo(foo) {
 
 ## Limitations
 
-This rule does not detect or remove extraneous code blocks:
+This rule does not detect or remove extraneous code blocks inside of functions:
 
 ```js
 function doFoo(foo) {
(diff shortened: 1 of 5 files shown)


rules/consistent-function-scoping.js
```diff
@@ -0,0 +1,131 @@
...
+	return checkReferences(scope, parentScope, scopeManager);
+}
+
+const create = context => {
+	const sourceCode = context.getSourceCode();
+	const {scopeManager} = sourceCode;
+
+	return {
+		ArrowFunctionExpression(node) {
+			const valid = checkNode(node, scopeManager);
```
ankeetmaini: this looks so neat and readable!

rules/consistent-function-scoping.js
```diff
@@ -0,0 +1,131 @@
...
+
+	if (parentNode.type === 'VariableDeclarator') {
+		parentNode = parentNode.parent;
+	}
+
+	if (parentNode.type === 'VariableDeclaration') {
+		parentNode = parentNode.parent;
+	}
+
+	if (parentNode.type === 'BlockStatement') {
```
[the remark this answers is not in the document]
MrHen: Hm, yeah that's a good test case. What should happen is that it will get reported because you can safely move `doBar` up one scope.

The way this works is:

1. Check for function / arrow function
2. Check immediate parent for a block statement and skip it
3. Starting at first un-skipped ancestor, recursively check all descendants for a reference to a variable in the ancestor scope
4. If none found, warn

In the test case you posted, it _should_ correctly mark the extra block statement as a pointless scope.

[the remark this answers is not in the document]
ankeetmaini: I tried this out in ast explorer. It didn't fire the lint rule https://astexplorer.net/#/gist/92ef00cc145f6f92dec0268d64ca4a03/ea7ebb21a6ad6c218795719dfbaedd633e4e97dc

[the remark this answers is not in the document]
MrHen: Looks like `scopeManager.acquire` won't return a scope for the extra block statement. I'm not entirely sure why but I can probably fix it today or tomorrow.

Most likely the easiest fix is similar to what you recommended: Run a loop until we find a scope, hit the global scope or run out of ancestors.

Good catch. 👍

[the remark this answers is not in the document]
MrHen: @ankeetmaini Alright, I took at look at fixing this and it turns out it is non-trivial. Trying to add a loop to track back up more than one parent caused all sorts of problems with things like `for` loops, variables defined within the blocks, functions used in an assignment and nested blocks already at root.

@sindresorhus, @futpib how much do either of you care about about detecting things inside of empty block scopes?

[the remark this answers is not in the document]
sindresorhus: > how much do either of you care about about detecting things inside of empty block scopes?

I don't care much. I think I've only used an empty block a couple of times in the past years. Just note the limitation in the docs and if someone cares they can add support for it.

docs/rules/consistent-function-scoping.md
```diff
@@ -0,0 +1,44 @@
+# Prefer consistent function scoping
```
sindresorhus: Can you come up with something more descriptive that doesn't just repeat the rule name?

rules/consistent-function-scoping.js
```diff
@@ -0,0 +1,131 @@
...
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocsUrl(__filename)
+		},
+		messages: {
+			[MESSAGE_ID_ARROW]: 'Move arrow function to outer scope.',
```
sindresorhus:
```suggestion
			[MESSAGE_ID_ARROW]: 'Move arrow function to the outer scope.',
```

test/consistent-function-scoping.js
```diff
@@ -0,0 +1,264 @@
...
+		outdent`
+			function doFoo(foo = 'foo') {
+				function doBar(bar) {
+					return foo + bar;
+				}
+			}
+		`,
+		outdent`
+			function doFoo() {
+				var foo = 'foo';
```
sindresorhus: Nitpick, but use `const` instead of `var` everywhere.

docs/rules/consistent-function-scoping.md
```diff
@@ -0,0 +1,44 @@
...
+
+A function definition should be placed as close to the top-level scope as possible without breaking its captured values.
+
```
sindresorhus: Can you mention that this makes the code more readable, and it can also have a performance benefit as JS engines can better optimize.

MrHen: If you happen to have a source for the JS optimization part I can also add a link. I hadn't heard that before.

MrHen:
... I'm not sure how but the `for` loop case is now being caught correctly. Updated tests to match but it makes me a little nervous. 😕

docs/rules/consistent-function-scoping.md
sindresorhus: Sure. There's the obvious overhead of recreating the function for each call: https://stackoverflow.com/a/81329/64949 There's also a function size limit where the JIT optimizer bails out: https://ponyfoo.com/articles/javascript-performance-pitfalls-v8#optimization-limit

sindresorhus:
This looks good 👌

PR #312, opened 2019-06-02
prefer-flat-map: Use outdent for multiline test cases

MrHen:
Follow up from https://github.com/sindresorhus/eslint-plugin-unicorn/pull/284#issuecomment-484586021. I found a library that will automatically clean up indents in template strings which makes these test cases _much_ easier to read.


sindresorhus:
IMHO, this is how template literals should have worked in the first place. Such a missed opportunity...

PR #313, opened 2019-06-03
custom-error-definition: Cover exports name

### LINKED ISSUE #190: custom-error-definition rule does not validate against exported name
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/190)

The following passes linting:

```js
exports.fooError = class FooError extends Error {
    constructor(error) {
        super(error);
        this.name = 'FooError';
    }
};
```

The name used in `exports.fooError` should match the name used in `this.name = 'FooError';`.

A commented out test was added as part of #188 to cover this scenario.

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

### Tips

</details>

MrHen:
Fixes #190.

I cheated a bit for this one. If the rule is turned on it will force the error name to be correct so the rule only has to check for a mismatch on the `exports` identifier in the case where the right hand side is an error class.

There was another interesting case that I did NOT cover:

```js
	constructor(error) {
		super(error);
		this.name = 'FooError';
	}
};
exports.fooError = FooError;
```

It would theoretically be possible to find an error definition from earlier in the same file. If we want this scenario covered, then I can add it in.

### DIFF
--- a/rules/custom-error-definition.js
+++ b/rules/custom-error-definition.js
@@ -2,6 +2,8 @@
 const upperfirst = require('lodash.upperfirst');
 const getDocsUrl = require('./utils/get-docs-url');
 
+const MESSAGE_ID_INVALID_EXPORT = 'invalidExport';
+
 const nameRegexp = /^(?:[A-Z][a-z\d]*)*Error$/;
 
 const getClassName = name => upperfirst(name).replace(/(error|)$/i, 'Error');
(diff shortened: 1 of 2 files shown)


sindresorhus:
> It would theoretically be possible to find an error definition from earlier in the same file. If we want this scenario covered, then I can add it in.

Nah. Not worth the effort since we'll have ESM in Node.js soon.

PR #317, opened 2019-06-04
Do not warn if `.flat()` has more than one parameter or if the one parameter is not `1`

### LINKED ISSUE #315: `prefer-flat-map` fixer bug
`Array#flat` accepts argument `depth`, but flatMap not.

The fixer is not considering this `depth`, codes like

```js
const foo = [[1]].map(i => [i]).flat(2)
```

should not be reported.

MrHen:
Fixes #315.

A follow up question is if we should warn / fix on `.flat(1)` which is default flattening behavior but I wanted to get the fix out quickly.

### DIFF
--- a/rules/prefer-flat-map.js
+++ b/rules/prefer-flat-map.js
@@ -98,6 +98,10 @@ const create = context => ({
 			return;
 		}
 
+		if (node.arguments.length > 0) {
+			return;
+		}
+
 		const parent = node.callee.object;
 
 		if (!isMethodNamed(parent, 'map')) {
--- a/rules/prefer-flat-map.js
+++ b/rules/prefer-flat-map.js
@@ -98,7 +98,11 @@ const create = context => ({
 			return;
 		}
 
-		if (node.arguments.length > 0) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Yes, we should report and fix `.flat(1)`.

fisker:
and also `undefined` ? maybe also `"1"`? 

according to , https://tc39.github.io/proposal-flatMap/#sec-Array.prototype.flat 

Any value can `ToInteger` equals to 1, all the same

sindresorhus:
@fisker Nah. No one would actually pass in that. Not worth caring about.

fisker:
make sense

fisker:
on second thought,  most people also not pass `1`, how about not check arguments at all, instead a new rule to check `flat` argument ?

sindresorhus:
> on second thought, most people also not pass 1, how about not check arguments at all, instead maybe make a new rule to check flat argument ?

A rule that checks arguments for built-ins in general sounds useful, but would be very fragile without type information. Better to just use TypeScript then.

PR #321, opened 2019-06-11
prefer-node-append: Only fix when expression is not used

### LINKED ISSUE #319: `prefer-node-append` fixer bug
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/319)

```js
const canvas = document.body.appendChild(document.createElement('canvas'))
```

```js
const canvas = document.body.append(document.createElement('canvas'))
```

This can't be fixed, `ParentNode.append()` has no return value

https://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $30.00)

### Tips

</details>

MrHen:
Fixes #319.

There are three possible approaches here and I didn't know exactly which was best:

1. Never fix
2. Only fix when parent node type is in a whitelist
3. Only fix when parent node type is not in a blacklist

This PR uses approach 3 but approach 2 would be a more conservative approach. Before I go digging up all sorts of variants of the pattern I figured I'd confirm the blacklist approach is the way we wanted to go.

Also, I chose to leave the warning enabled even when the fixer is disabled. Should we not even warn if the return value is being used?

### DIFF
--- a/rules/prefer-node-append.js
+++ b/rules/prefer-node-append.js
@@ -3,16 +3,30 @@ const getDocsUrl = require('./utils/get-docs-url');
 
 const getMethodName = memberExpression => memberExpression.property.name;
 
+const ignoredParentTypes = [
+	'VariableDeclarator',
+	'CallExpression'
+];
+
 const create = context => {
 	return {
 		CallExpression(node) {
-			const {callee} = node;
+			const {
+				callee,
+				parent
+			} = node;
 
 			if (callee.type === 'MemberExpression' && getMethodName(callee) === 'appendChild') {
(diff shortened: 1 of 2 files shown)


sindresorhus:
I think your approach `3` is good.

---

> Also, I chose to leave the warning enabled even when the fixer is disabled. Should we not even warn if the return value is being used?

We should still warn, even if we don't fix.

The user could still change:

```js
const canvas = document.body.appendChild(document.createElement('canvas'))
```

to

```js
const canvas = document.createElement('canvas');
document.body.append(canvas);
```

manually.

PR #323, opened 2019-06-15
prefer-flat-map: Report for common concat pattern

### LINKED ISSUE #311: A case for `prefer-flat-map`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/311)

I think it will be way greater rule if it will be able to identify legacy (before `.flat` got supported) code and suggest to refactor to `flatMap`. I believe one of the most common scenario was use of `[].concat` with array spread:

```js
[].concat(...someArray.map(fn))
// is the same
someArray.flatMap(fn)
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

### Tips

</details>

MrHen:
Closes #311.

Tried something different for this rule. If we don't want to fix it, then all we need to do is detect the particular pattern. I chose to use a fairly complicated selector but the selector was too long to reasonably read as a single string.

### DIFF
--- a/docs/rules/prefer-flat-map.md
+++ b/docs/rules/prefer-flat-map.md
@@ -9,6 +9,7 @@ This rule is fixable.
 
 ```js
 [1, 2, 3].map(i => [i]).flat();
+[].concat(...foo.map((i) => i))
 ```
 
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you also update the docs with a concat example?

PR #325, opened 2019-06-20
Add `consistent-destructuring` rule

### LINKED ISSUE #248: Rule proposal: Consistent destructuring
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/248)

If a variable (or a property, but probably not a call or complex expression) is destructured somewhere, there is rarely good reason to use it's properties instead of the destructured variables. 

```jsx
const {size, color} = this.props;

return <Foo size={size}>
  <Bar color={color}>
    <Buz shape={this.props.shape} />
  </Bar>
</Foo>;
```

A later editor of the code can simply miss the fact that `this.props` is destructured at the beginning of a large enough block and use the wanted value directly, leading to this inconsistency.

Alternative rule name suggestions are welcome.

## Fail
```js
const {a} = foo;
console.log(a, foo.b); 
```

```js
const {a} = foo;
console.log(foo.a); 
```

## Pass
```js
const {a} = foo;
console.log(a, foo.b()); 
```
```js
const {a, b} = foo;
console.log(a, b);
```
```js
console.log(foo.a, foo.b);
```

> There is a $80.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/248).

medusalix:
Resolves #248

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -113,7 +113,6 @@ Configure it in `package.json`.
 - [custom-error-definition](docs/rules/custom-error-definition.md) - Enforce correct `Error` subclassing. *(fixable)*
 - [empty-brace-spaces](docs/rules/empty-brace-spaces.md) - Enforce no spaces between braces. *(fixable)*
 - [error-message](docs/rules/error-message.md) - Enforce passing a `message` value when creating a built-in error.
-- [error-message](docs/rules/error-message.md) - Enforce passing a `message` value when throwing a built-in error.
 - [escape-case](docs/rules/escape-case.md) - Require escape sequences to use uppercase values. *(fixable)*
 - [expiring-todo-comments](docs/rules/expiring-todo-comments.md) - Add expiration conditions to TODO comments.
 - [explicit-length-check](docs/rules/explicit-length-check.md) - Enforce explicitly comparing the `length` property of a value. *(partly fixable)*
--- a/readme.md
+++ b/readme.md
@@ -35,6 +35,7 @@ Configure it in `package.json`.
 		"rules": {
 			"unicorn/better-regex": "error",
 			"unicorn/catch-error-name": "error",
+			"unicorn/consistent-destructuring": "error",
 			"unicorn/consistent-function-scoping": "error",
 			"unicorn/custom-error-definition": "off",
 			"unicorn/empty-brace-spaces": "error",
@@ -107,10 +108,12 @@ Configure it in `package.json`.
 
 - [better-regex](docs/rules/better-regex.md) - Improve regexes by making them shorter, consistent, and safer. *(fixable)*
 - [catch-error-name](docs/rules/catch-error-name.md) - Enforce a specific parameter name in catch clauses. *(fixable)*
+- [consistent-destructuring](docs/rules/consistent-destructuring.md) - Use destructured variables over properties. *(partly fixable)*
 - [consistent-function-scoping](docs/rules/consistent-function-scoping.md) - Move function definitions to the highest possible scope.
 - [custom-error-definition](docs/rules/custom-error-definition.md) - Enforce correct `Error` subclassing. *(fixable)*
 - [empty-brace-spaces](docs/rules/empty-brace-spaces.md) - Enforce no spaces between braces. *(fixable)*
 - [error-message](docs/rules/error-message.md) - Enforce passing a `message` value when creating a built-in error.
+- [error-message](docs/rules/error-message.md) - Enforce passing a `message` value when throwing a built-in error.
 - [escape-case](docs/rules/escape-case.md) - Require escape sequences to use uppercase values. *(fixable)*
 - [expiring-todo-comments](docs/rules/expiring-todo-comments.md) - Add expiration conditions to TODO comments.
 - [explicit-length-check](docs/rules/explicit-length-check.md) - Enforce explicitly comparing the `length` property of a value. *(partly fixable)*
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -46,8 +46,8 @@ const create = context => {
 		ignore: [],
 		...context.options[0]
 	};
-	const {name: expectedName, ignore: ignore_} = options;
-	const ignore = ignore_.map(
+	const {name: expectedName} = options;
+	const ignore = options.ignore.map(
 		pattern => pattern instanceof RegExp ? pattern : new RegExp(pattern, 'u')
 	);
 	const isNameAllowed = name =>
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -46,8 +46,8 @@ const create = context => {
 		ignore: [],
 		...context.options[0]
 	};
-	const {name: expectedName} = options;
-	const ignore = options.ignore.map(
+	const {name: expectedName, ignore: ignore_} = options;
+	const ignore = ignore_.map(
 		pattern => pattern instanceof RegExp ? pattern : new RegExp(pattern, 'u')
 	);
 	const isNameAllowed = name =>
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -134,12 +134,12 @@ const create = context => {
 			context.report({
 				node,
 				messageId: MESSAGE_ID,
-				data: {
-					expression,
-					property: newMember
-				},
 				suggest: [{
 					messageId: MESSAGE_ID_SUGGEST,
+					data: {
+						expression,
+						property: newMember
+					},
 					fix: destructuredMember ?
 						fixer => [fixer.replaceText(node, newMember)] :
 						fixer => [
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -1,0 +1,68 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
+
+const fix = (fixer, source, objectPattern, memberExpression) => {
+	const member = source.getText(memberExpression.property);
+	const {properties} = objectPattern;
+	const declarations = properties.map(property => source.getText(property));
+	const fixings = [fixer.replaceText(memberExpression, member)];
+
+	// Member is not already destructured
+	if (!declarations.includes(member)) {
+		const lastProperty = properties[properties.length - 1];
+
+		if (lastProperty) {
+			fixings.push(fixer.insertTextAfter(lastProperty, `, ${member}`));
+		} else {
+			fixings.push(fixer.replaceText(objectPattern, `{${member}}`));
+		}
+	}
+
+	return fixings;
+};
+
+const create = context => {
+	const source = context.getSourceCode();
+	const declarations = new Map();
+
+	return {
+		VariableDeclarator(node) {
+			const {id, init} = node;
+
+			if (id.type === 'ObjectPattern' && init.type !== 'Literal') {
+				declarations.set(source.getText(init), id);
+			}
+		},
+		MemberExpression(node) {
+			const {parent, object} = node;
+
+			// Ignore member function calls
+			if (parent.type === 'CallExpression' && parent.callee === node) {
+				return;
+			}
+
+			const objectPattern = declarations.get(source.getText(object));
+
+			if (objectPattern) {
+				const isNested = parent.type === 'MemberExpression';
+
+				context.report({
+					node,
+					message: 'Use destructured variables over properties.',
+					fix: isNested ? null : fixer => fix(fixer, source, objectPattern, node)
+				});
+			}
+		}
+	};
+};
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocsUrl(__filename)
+		},
+		fixable: 'code'
+	}
+};
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -64,8 +64,11 @@ const create = context => {
 		MemberExpression(node) {
 			const {parent, object, property} = node;
 
-			// Ignore member function calls
-			if (parent.type === 'CallExpression' && parent.callee === node) {
+			// Ignore member function calls and delete expressions
+			if ((parent.type === 'CallExpression' &&
+				parent.callee === node) ||
+				(parent.type === 'UnaryExpression' &&
+				parent.operator === 'delete')) {
 				return;
 			}
 
@@ -91,7 +94,7 @@ const create = context => {
 				memberVariables.delete(variable.name);
 			}
 
-			const isDefined = memberVariables.has(member);
+			const isDefined = memberVariables.has(member) || member === 'arguments';
 			const isNested = parent.type === 'MemberExpression';
 
 			context.report({
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -23,33 +23,82 @@ const fix = (fixer, source, objectPattern, memberExpression) => {
 	return fixings;
 };
 
+const isInParentScope = (child, parent) => {
+	while (child) {
+		if (child === parent) {
+			return true;
+		}
+
+		child = child.upper;
+	}
+
+	return false;
+};
+
+const getVariablesInScope = scope => {
+	const variables = new Set();
+
+	while (scope) {
+		for (const variable of scope.variables) {
+			variables.add(variable.name);
+		}
+
+		scope = scope.upper;
+	}
+
+	return variables;
+};
+
 const create = context => {
 	const source = context.getSourceCode();
 	const declarations = new Map();
 
 	return {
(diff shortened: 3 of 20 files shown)


rules/consistent-destructuring.js
```diff
@@ -0,0 +1,68 @@
...
+
+const create = context => {
+	const source = context.getSourceCode();
+	const declarations = new Map();
+
+	return {
+		VariableDeclarator(node) {
+			const {id, init} = node;
+
+			if (id.type === 'ObjectPattern' && init.type !== 'Literal') {
```
MrHen: You can put both of these in the [selector](https://eslint.org/docs/developer-guide/selectors). Instead of `VariableDeclarator` it will be something like `'VariableDeclarator[id.type="ObjectPattern"][init.type!="Literal"]`.

rules/consistent-destructuring.js
```diff
@@ -0,0 +1,68 @@
...
+			}
+
+			const objectPattern = declarations.get(source.getText(object));
+
+			if (objectPattern) {
+				const isNested = parent.type === 'MemberExpression';
+
+				context.report({
+					node,
+					message: 'Use destructured variables over properties.',
```
MrHen: Use [`messageId`](https://eslint.org/docs/developer-guide/working-with-rules#messageids) instead of a `message`. See [prefer-flat-map](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6f10722c2397d9734375cc5165a3675e07752982/rules/prefer-flat-map.js#L62) for an example.

test/consistent-destructuring.js
```diff
@@ -0,0 +1,160 @@
...
+			console.log(a);`
+		),
+		invalidTestCase(
+			`const {a: b, c} = foo;
+			console.log(foo.d);`,
+			`const {a: b, c, d} = foo;
+			console.log(d);`
+		)
+	]
+});
```
MrHen: Some additional test cases:

----
```
const { a } = foo;
console.log(
  'foo',
  foo.b
);
```
----
```
const { a } = foo;
console.log(
  'foo', // comment
  foo.b // comment
);
```
----
Cannot use normal fix:
```
const { a } = foo;
const b = 'bar';
console.log(foo.b);
```
----
Should not fix because the destructuring is in a different scope:
```
const foo = { a: 1, b: 2};
function alfa() {
  const { a } = foo;
}
function beta() {
  console.log(foo.b);
}
```
----
Not sure what should happen here:
```
const { a } = foo;
const { b } = foo;
console.log(foo.c);
```
----
```
let foo = { a: 1 };
const { a } = foo;
foo = { a: 2 };
console.log(foo.a);
```
----

rules/consistent-destructuring.js
medusalix: Yeah, I just thought it improved readability to have these in a separate `if` statement.

medusalix:
@MrHen I added all of your suggested test cases, though I'm still not sure how to tackle this one:

```js
let foo = {a: 1};
const {a} = foo;
foo = {a: 2};
console.log(foo.a);
```

MrHen:
> I'm still not sure how to tackle this one

There needs to be a check to see if a variable in that scope already exists. If there is one, either (a) use [avoid-capture](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/utils/avoid-capture.js) to pick a name not being used or (b) chose not to fix it at all.

I lean toward (b) because this is probably an unusual situation.

medusalix:
I'm not talking about the case where the variable is already defined and therefore can't be destructured. I'm concerned about the situtation where the destructured member is changed after being destructured (The last test case you suggested). If I apply my fix to this scenario, the semantics of the code will change.

MrHen:
>  I'm concerned about the situtation where the destructured member is changed after being destructured.

Oh, I see. I wasn't even originally worried about that but now I get the issue. A really common example of this would be getters:

```
class Foo {
  let _a = 'foo';

  set a(value) {
    this._a = value;
  }

  get a() {
    return this._a;
  }
}

const foo = new Foo();
const { a } = foo;
foo.a = 'bar';
console.log(a, foo.a);
```

The smallest example would just be:

```
const foo = {a: 1};
const {a} = foo;
foo.a = 2;
console.log(a, foo.a);
```

I'm not sure how to deal with this, either. @sindresorhus, @futpib, either of you have any thoughts / insights?

sindresorhus:
The [`prefer-destructuring`](https://eslint.org/docs/rules/prefer-destructuring) rule has a fixer, so not sure how it handles it. @medusalix Can you take a look at its source and tests and see whether it handles this or just ignores it?

medusalix:
@sindresorhus The `prefer-destructuring` rule is apparently only tailored to the case of variable assignments. All of the failing tests I've tried passed that rule without any warning at all.

sindresorhus:
I think for that case we can use the [ESLint suggestions API](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/441) instead of a fixer. Then the user can decide to manually apply the suggestion.

medusalix:
I have now switched to the suggestions API as suggested and it seems to be working great. It catched a bunch of problems in all kinds of different rules, that's why the lint task is failing.

test/consistent-destructuring.js
```diff
@@ -0,0 +1,344 @@
+import test from 'ava';
+import avaRuleTester from 'eslint-ava-rule-tester';
+import {outdent} from 'outdent';
+import rule from '../rules/consistent-destructuring';
+
+const ruleTester = avaRuleTester(test, {
+	env: {
+		es6: true
+	}
+});
```
sindresorhus, line 10:
```suggestion
const ruleTester = avaRuleTester(test, {
	parserOptions: {
		ecmaVersion: 2020
	}
});
```

fisker:
@medusalix Don't try to fix lint, We can do review first.

fisker:
Did you merge the latest master? Failed integration test should has a link link to the source code.

fisker:
Integration tests need to be fixed,  you can use [this](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/08274f9603092070083642f2a6be25bc16d72c23/test/prefer-number-properties.js#L240) as an example, find the code add to tests.

rules/consistent-destructuring.js
```diff
@@ -0,0 +1,140 @@
...
+
+			// Ignore member function calls and delete expressions
+			if (
+				(parent.type === 'CallExpression' &&
+				parent.callee === node) ||
+				(parent.type === 'UnaryExpression' &&
+				parent.operator === 'delete')
+			) {
+				return;
+			}
```
fisker, line 65:
```js
MemberExpression
:not(
	CallExpression > MemberExpression.callee,
	UnaryExpression[operator="delete"] > MemberExpression.argument,
)
```

Want try this selector? It can do the check

rules/consistent-destructuring.js
```diff
@@ -0,0 +1,140 @@
...
+
+			context.report({
+				node,
+				messageId: MESSAGE_ID,
+				data: {
+					expression,
+					property: newMember
+				},
+				suggest: [{
+					messageId: MESSAGE_ID_SUGGEST,
```
fisker, line 114: `suggest` need pass `data`, I suggest do a actual messages test.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/11a1e9c560beb3fe8c2f58f81bd7d50971a5ff37/test/no-fn-reference-in-iterator.js#L209

rules/consistent-destructuring.js
```diff
@@ -0,0 +1,140 @@
...
+					fix: destructuredMember ?
+						fixer => [fixer.replaceText(node, newMember)] :
+						fixer => [
+							fixer.replaceText(node, newMember),
+							fixDestructuring(fixer, objectPattern, member, newMember)
```
fisker, line 119: Are you sure this is right? If multiple `MemberExpression`, will it update the `VariableDeclarator` again?

rules/consistent-destructuring.js
```diff
@@ -0,0 +1,140 @@
...
+
+	return {
+		[declaratorSelector]: node => {
+			declarations.set(source.getText(node.init), {
+				scope: context.getScope(),
+				variables: context.getDeclaredVariables(node),
+				objectPattern: node.id
+			});
+		},
+		MemberExpression: node => {
```
fisker, line 54: You didn't filter the computed `MemberExpression`,

```js
				const {a} = foo;
				console.log(foo[a]);
```

test/consistent-destructuring.js
```diff
@@ -0,0 +1,344 @@
...
+
+const ruleTester = avaRuleTester(test, {
+	parserOptions: {
+		ecmaVersion: 2020
+	}
+});
+
+const invalidTestCase = (code, correctCode) => {
+	return {
+		code,
```
fisker, line 14:
```suggestion
		code,
		output: code,
```

Make sure output didn't change.

test/consistent-destructuring.js
```diff
@@ -0,0 +1,344 @@
...
+		invalidTestCase(
+			outdent`
+				const {a} = foo;
+				console.log(foo.a);
+			`,
+			outdent`
+				const {a} = foo;
+				console.log(a);
+			`
+		),
```
fisker, line 118:
```js
		invalidTestCase({
			code: outdent`
				const {a} = foo;
				console.log(foo.a);
			`,
			correctCode: outdent`
				const {a} = foo;
				console.log(a);
			`
		}),
```

Easier to read, and we can add more options later.

readme.md
```diff
@@ -92,7 +93,8 @@ Configure it in `package.json`.
 ## Rules
 
 - [better-regex](docs/rules/better-regex.md) - Improve regexes by making them shorter, consistent, and safer. *(fixable)*
-- [catch-error-name](docs/rules/catch-error-name.md) - Enforce a specific parameter name in catch clauses. *(fixable)*
+- [catch-error-name](docs/rules/catch-error-name.md) - Enforce a specific parameter name in catch clauses.
```
fisker, line 96: Missing text when merging.

test/consistent-destructuring.js
```diff
@@ -344,6 +344,26 @@ ruleTester.run('consistent-destructuring', rule, {
...
+				console.log(c);
+			`
+		}),
+		invalidTestCase({
+			code: outdent`
+				const {a, ...b} = foo;
+				console.log(foo.b);
+			`,
+			correctCode: outdent`
+				const {a, b: b_, ...b} = foo;
```
fisker, line 364: You can't do this,  value of `b` changed.

If contains `RestElement`, it should not fixable, and should be correct code.

medusalix: Except for cases where the member is already destructured, right?

rules/consistent-destructuring.js
medusalix: It only fixes simple cases where the parent is not another `MemberExpression` (doesn't fix if `node.parent.type === 'MemberExpression'`), so I think it works correctly.

test/consistent-destructuring.js
fisker: I don't think we need check destructing with RestElement at all, it' not changable. It's correct code.

`const {a, ...b} = foo;` no better/equal way.

We need ignore it unless `b` is destructing again.


# Valid no error

```js
const {a, ...b} = foo;
alert(foo.c)
```
# Invalid

```js
const {a, ...b} = foo;
const {d}=b;
alert(b.c)
```

rules/consistent-destructuring.js
fisker: I mean 

```js
const {a} =foo;
alert(foo.b);
alert(foo.b);
```

test/consistent-destructuring.js
fisker: Sorry, I fogot we need check `foo.a` , I'll review tomorrow, on cellphone.

medusalix: > const {a, ...b} = foo; no better/equal way.

Yeah, except for the case where somebody uses `foo.a`. We can fix that to `a` without changing the object expression at all.

rules/consistent-destructuring.js
medusalix: It will give you two separate suggestions that will both add `b` to the destructured members and rename the member expression.

test/consistent-destructuring.js
```diff
@@ -0,0 +1,390 @@
...
+			suggestions: [{
+				messageId: 'consistentDestructuringSuggest',
+				output: suggestion
+			}]
+		}))
+	};
+};
+
+ruleTester.run('consistent-destructuring', rule, {
+	valid: [
```
futpib, line 39: Should we also test that function calls and other complex expressions are not reported by this rule?

For example, this

```js
const { foo } = array.pop();
const { bar } = array.pop();
```

cannot be fixed to this

```js
const { foo, bar } = array.pop();
```

sindresorhus: ^ @medusalix

medusalix: @futpib This rule only reports member expressions on already destructured objects, so your example will be fine in any case. The selector ignores any member expressions on function calls and there are already test cases for that :)

futpib: Here is a test I think should be valid, but it is reported:
```js
const { a } = foo().bar;
console.log(a, foo().bar.b);
```

medusalix:
Destructurings of function calls inside member expression should be ignored now. Might need some additional tests.

sindresorhus:
> Might need some additional tests.

Can you add some?

sindresorhus:
Friendly bump :)

medusalix:
I've added tests for various kinds of expressions (as defined by the `ESTree` specification) and discovered a bug in my `VariableDeclarator` parsing. I've decided to add a check to ignore complex expressions (anything that's not a `MemberExpression` or `Identifier`) which fixed those problems.

sindresorhus (approved):

medusalix:
@fisker Would you mind giving this a review before merging?

fisker:
I'm sorry, I'll take a look tomorrow.

fisker:
I'm really sorry, that I didn't read all the code, but I pushed some tests, I think they should be valid?

fisker:
Took a deep think about this rule, I think it's impossible to make this rule work.

```js
const foo = bar;
function unicorn(){
  const {a} = bar;
  // Modify `foo`, but actually modifying `bar`
  bar.a = 1;
  console.log(bar.a)
}
```

medusalix:
> Took a deep think about this rule, I think it's impossible to make this rule work.

Yeah, there are some tricky edge cases that this rule does not handle well. But I thought we agreed on the suggestions API to make sure the rule doesn't try to fix those situations automatically. See [this](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/325#issuecomment-593924147) comment and my previous concerns.

fisker:
Even without auto-fix, this rule is going to be annoying.

To be honest, I don't think we should  add this rule.

@sindresorhus @futpib WDYT?

sindresorhus:
@fisker I think this rule will work fine in most real-world code. The code examples I've seen in this thread are very awkward (like https://github.com/sindresorhus/eslint-plugin-unicorn/pull/325#issuecomment-715963186) and I don't think they represent real-world code. How about we try it out and if it turns out to be annoying, we can turn it off by default? It's actually solving a real problem for me as a lot of PR submitters forget to use destructuring, so I need to comment on it all the time.

sindresorhus:
> but I pushed some tests, I think they should be valid?

@fisker The tests seems to be failing.

fisker:
> The tests seems to be failing.

They should not, 

For example 

```js
const {a} = foo; foo.a++;
```

It's updating `foo.a`, we can't use `a` to update it.

sindresorhus:
@medusalix Tests are still failing.

medusalix:
@sindresorhus Tests should be good now (except for `lint`). I've managed to fix *some* of the tests and removed the remaining ones.

sindresorhus:
Can you try to fix the lint issues? They might reveal some problems in the rule.

medusalix:
> Can you try to fix the lint issues? They might reveal some problems in the rule.

@sindresorhus I've managed to fix all the reported lint issues, so it does indeed look like the rule is working correctly.

rules/catch-error-name.js
```diff
@@ -46,8 +46,8 @@ const create = context => {
 		ignore: [],
 		...context.options[0]
 	};
-	const {name: expectedName} = options;
-	const ignore = options.ignore.map(
+	const {name: expectedName, ignore: ignore_} = options;
```
sindresorhus, line 49: I don't think it should enforce destructuring if it conflicts with an existing identifier, like here.

sindresorhus (approved):

PR #329, opened 2019-06-23
Add `import-path-order` rule

### LINKED ISSUE #324: Rule proposal: `import-path-order`
I can't get what I want out of either ESLint's [sort-imports](https://eslint.org/docs/rules/sort-imports) _or_ [import/order](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) and it looks like both rules are stuck in a world of too many configuration options and debate. What I want feels simple, though:

* Sort all import lines by path
* Relative paths (`../foo`) are sorted after absolute paths (`foo`)
* `../../foo` is before `../foo`

I don't think this rule would be compatible with either of the above linked rules and I don't think it would need to be.

----

Fail:

```js
import * as bar from 'bar.js';
import * as foo from 'foo.js';
import {alpha, beta} from 'alpha.js';
import {delta, gamma} from 'delta.js';
import a from 'baz.js';
import b from 'qux.js';
```

----

```js
import w from './foo';
import x from '../foo';
import y from '../../foo';
import z from 'foo';
```

Pass:

```js
import {alpha, beta} from 'alpha.js';
import * as bar from 'bar.js';
import a from 'baz.js';
import {delta, gamma} from 'delta.js';
import * as foo from 'foo.js';
import b from 'qux.js';
```

----

```js
import z from 'foo';
import y from '../../foo';
import x from '../foo';
import w from './foo';
```

MrHen:
Closes #324.

First look at `import-path-order`. Remaining work:

* [x] Documentation
* [x] Fixer
* [x] `allow-blank-lines` configuration setting
* [x] Turn off alphabetization by default
* [x] Add grouping by package word
* [x] Fix `parts` error message label
* [x] Don't mix `a/b` and `a-b` parts together


docs/rules/import-path-order.md
```diff
@@ -0,0 +1,139 @@
...
+* Built-in packages (`fs`, `path`, etc.)are sorted first
+* Relative paths (`../foo`) are sorted after absolute paths (`foo`)
+* `../../foo` is before `../foo`
+* Paths with the same nesting are sorted alphabetically
+
+## Fail
+
+```js
+var b = require('b');
+var a = require('a');
```
sindresorhus: Use `const`

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,139 @@
+# Organize imports by path order
+
+Imports and requires should be organized together and sorted by relative path depth.
+
+The scope of this rule is deliberately simple. Both [sort-imports](https://eslint.org/docs/rules/sort-imports) and [import/order](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/order.md) exist for more complicated and configurable patterns.
+
+This rule will:
+
+* Sort import lines by path
+* Built-in packages (`fs`, `path`, etc.)are sorted first
```
sindresorhus:
```suggestion
* Built-in modules (`fs`, `path`, etc.) are sorted first
```

rules/import-path-order.js
```diff
@@ -0,0 +1,236 @@
+'use strict';
+
+const coreModules = require('resolve/lib/core');
```
sindresorhus: I have a better list for this here: https://github.com/sindresorhus/builtin-modules and https://github.com/sindresorhus/is-builtin-module

Can you also add `electron` as a built-in so it's sorted before third-party packages?

rules/import-path-order.js
```diff
@@ -0,0 +1,236 @@
...
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocsUrl(__filename)
+		},
+		messages: {
+			[MESSAGE_ID_BLANKLINES]: 'Group imports together',
+			[MESSAGE_ID_DEPTH]: 'Relative paths should be sorted by depth',
+			[MESSAGE_ID_GROUP]: '{{earlier}} imports should come before {{later}} imports',
+			[MESSAGE_ID_ORDER]: 'Imports should be sorted alphabetically'
```
sindresorhus: We didn't discuss this in #324, or I must have missed it. Sorting alphabetically sounds nice in theory, but it creates a lot of annoyance in practice for very little actual benefit.

test/import-path-order.js
```diff
@@ -0,0 +1,284 @@
...
+				var a = require('a');
+
+				var b = require('b');
+			`,
+			options: [{
+				allowBlankLines: true
+			}]
+		},
+		outdent`
+			var a = require('a');
```
sindresorhus: Use `const` instead of `var` for everything here.

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,139 @@
...
+// Built-in packages should come first
+import fs from 'fs';
+import a from 'a';
+```
+
+## Options
+
+### `allowBlankLines`
+
+`allowBlankLines` will force all imports to be organized together. It defaults to `true`.
```
sindresorhus: I think this should be `false` by default.

rules/import-path-order.js
MrHen: The primary reason I like it is because it puts imports from the same package next to each other. I don't actually care much about the alphabetization; it's just the fastest proxy for doing this:

```
import flat from 'lodash/flat';
import map from 'lodash/map';
import React from 'react';
```

Instead of this:

```
import flat from 'lodash/flat';
import React from 'react';
import map from 'lodash/map';
```

The latter drives me bonkers and I do care pretty strongly about it.

rules/import-path-order.js
MrHen: Ha, why am I not surprised you have a package for this. 🦄

rules/import-path-order.js
sindresorhus: Why not just `import {flat, map} from 'lodash';`? I guess the answer is tree-shaking, but I think that also works for named imports.

Even if not, i think it would be better to just place same-named imports together without forcing full alphabetization.

MrHen: In the case of `lodash`, yeah, tree-shaking is the answer. `lodash` is particularly notorious for [not tree-shaking](https://github.com/lodash/lodash/issues?q=is%3Aissue+tree+shaking+is%3Aclosed).

In addition, some libraries split up their common imports into separate packages:

```
import { withRouter } from "react-router";
import { BrowserRouter as Router, Route, Link } from "react-router-dom";
```

----

```
import { makeStyles } from '@material-ui/core/styles';
import Button from '@material-ui/core/Button';
import { AccessAlarm, ThreeDRotation } from '@material-ui/icons';
```

---

I see no real disadvantages to keeping imports alphabetized other than it's mildly annoying. Once the fixer is in place, I don't think it will trip people up too much.

If it really bothers you, I can try to add something that keeps imports clumped by common path part (e.g., `@material-ui` and `core`.) It won't work for things like `react-router` and `react-router-dom` but I can live with that.

sindresorhus: > I can try to add something that keeps imports clumped by common path part (e.g., @material-ui and core.) It won't work for things like react-router and react-router-dom but I can live with that.

Group modules that have the same super-string? So if you have `react`, then `react-router` and `react-dom` will be grouped with it. That would be generally useful for many other things too, like `electron` projects; `electron`, `electron-util`, `electron-context-menu`, etc.

MrHen: I see four plausible choices. In descending order of complexity:

1. Alphabetize
2. Group identical path segments next to each other (e.g., `@material-ui/core` and `@material-ui/icons`)
3. Group identical path segments _and_ group initial package "word" (e.g., `electron` and `electron-util`)
4. Group identical path segments _and_ group package words (e.g., `foo-bar-baz` would need to be near `foo-bar-zaz`)

(4) smells overly complex and something that will have weird edge cases.

(3) seems fine other than "initial word" is arbitrary. Should `url-join` be grouped next to `url-parse` even though they are technically unrelated? If the description of the rule was (1) than I don't think anyone would be surprised but the intent behind (3) is that libraries are grouped.

(2) is a reasonable compromise.

(1) is my preference because it is simple and covers nearly all of the benefits from the other three options. I don't think anyone would be surprised by it. We can always make it a configuration option, too. (I was planning on adding a case-insensitive switch anyway so we can just make it an `"alphabetize"` key with `"case-sensitive"`, `"case-insensitive"` and `"off"` values.)

sindresorhus: > Should url-join be grouped next to url-parse even though they are technically unrelated?

I don't think that will happen a lot and it kinda makes sense to though, as both deal with URL. It will however catch a lot of valid cases.

> (4) smells overly complex and something that will have weird edge cases.

Agreed.

---

The problem with 2 is that it doesn't catch common cases. It only catches scoped packages, but the majority of packages today are still top-level packages with a prefix, like `react` and `react-dom`.

So personally, I would go with 2 and 3 on by default, and ability to opt into 1 for people that want it very strict.

MrHen: > So personally, I would go with 2 and 3 on by default, and ability to opt into 1 for people that want it very strict.

Alright; I'll see what I can do.

MrHen:
@sindresorhus Alright, finally got around to this. Sorry for the delay.

I'll probably take another pass to clean up the code a bit since it got way more complicated than I was hoping. It should do everything you expect now. A couple notes:

* Comments, as always, are hell to auto-fix so I was very defensive and turned off the fixer if I noticed a comment before, after or between the lines that needed fixing.
* I chose to fix if-and-only-if a line detects that the line above it is wrong. This should eventually resolve but it might take multiple fixing passes to finish. (This is already done by `--fix`. It will only impact someone manually doing fixes in their IDE.)
* I am terrible at naming configuration options. If you have a better idea for `parts`, I'm all ears.
* The part grouping logic is completely distinct from the rest of the logic in the rule. For the sake of readability, it might be better to move it to a separate rule.

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,272 @@
...
+
+Type: `string`
+
+This adds additional ordering constraints for imports that have the same relative source.
+
+Options are:
+- `case-sensitive` - Orders alphabetically, case-sensitive
+- `case-insensitive` - Orders alphabetically, case-insensitive
+- `parts` - Groups imports with similar parts next to each other *(default)*
+- `off` - No additional ordering
```
sindresorhus: What if you want `case-sensitive` **and** `parts` ordering? Not clear if that's possible right now.

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+```js
+/* eslint unicorn/import-path-order: ["error", {comparator: 'off'}] */
+import c from 'a-c';
+import b from 'b';
+import a from 'a';
+```
+
+### `partsRegex`
```
sindresorhus: Do we really need this option? What use-cases are there for it?

rules/import-path-order.js
```diff
@@ -0,0 +1,473 @@
...
+		name,
+		group: GROUP_ABSOLUTE,
+		depth: 0,
+		parts
+	};
+}
+
+function getInvalidBlankLinesReport(nodePrev, nodeNext, context) {
+	if (nodePrev === null) {
+		return null;
```
sindresorhus: Can you use `undefined` everywhere instead of `null`?

https://github.com/sindresorhus/meta/issues/7

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+```js
+/* eslint unicorn/import-path-order: ["error", {allowBlankLines: false}] */
+import a from 'a';
+const foo = a.foo; // Non-import statements will fail
+import b from 'b';
+```
+
+### `comparator`
```
sindresorhus: `relativeSourceOrdering`?

`comparator` makes me think it's a function that receives two arguments to compare, like `.sort()`.

sindresorhus:
> Comments, as always, are hell to auto-fix so I was very defensive and turned off the fixer if I noticed a comment before, after or between the lines that needed fixing.

👍 

> I chose to fix if-and-only-if a line detects that the line above it is wrong. This should eventually resolve but it might take multiple fixing passes to finish. (This is already done by --fix. It will only impact someone manually doing fixes in their IDE.)

👍 

> I am terrible at naming configuration options. If you have a better idea for parts, I'm all ears.

`parts` is fine. Alternatively, `segments`.

docs/rules/import-path-order.md
MrHen: They aren't really compatible. Are you trying to figure out how to get `parts` to work with a case-sensitive comparison?

Maybe "case" is a bad name. It really means "alphabetical". If you want "alphabetical" than you won't need "parts".

docs/rules/import-path-order.md
MrHen: I habitually expose anything that configures how code works. We can hide it if you want but the code needs to know what part separators are so I put it in the config.

The use case would be if someone doesn't want `/` or `-` to matter (or perhaps wants to also treat `_` as a separator.)

docs/rules/import-path-order.md
MrHen: Yeah, `comparator` is not a great name. I could live with just `ordering`. `relativeSourceOrdering` seems verbose.

rules/import-path-order.js
MrHen: Oh, yeah, I keep forgetting that you prefer it that way.

docs/rules/import-path-order.md
sindresorhus: `ordering` then

docs/rules/import-path-order.md
sindresorhus: I generally only expose things when I can think of valid use-cases that many will have, otherwise I wait until someone requests it.

> The use case would be if someone doesn't want / or - to matter

If that's the only use-case, there's probably a simpler way to expose it. Not really sure if even that is valid though. I cannot think of when I would want to individually control them.

> or perhaps wants to also treat _ as a separator.

`_` is not a valid package name, so it would only apply to local filenames, and I cannot really think of any good reason using that.

MrHen: Alright, I'll hide the option. It's trivial to reintroduce if someone asks for it.

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,272 @@
...
+
+Type: `string`
+
+This adds additional ordering constraints for imports that have the same relative source.
+
+Options are:
+- `case-sensitive` - Orders alphabetically, case-sensitive
+- `case-insensitive` - Orders alphabetically, case-insensitive
+- `parts` - Groups imports with similar parts next to each other *(default)*
```
sindresorhus: I think we should better explain what `parts` entails.

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,272 @@
...
+- `off` - No additional ordering
+
+#### Fail
+
+```js
+/* eslint unicorn/import-path-order: ["error", {comparator: 'case-sensitive'}] */
+import a from 'a';
+import B from 'B';
```
sindresorhus: Package names has to be lowercase, so would be better to use local paths as example here.

docs/rules/import-path-order.md
sindresorhus: Never mind. I was overthinking it.

docs/rules/import-path-order.md
```diff
@@ -0,0 +1,272 @@
...
+
+### `comparator`
+
+Type: `string`
+
+This adds additional ordering constraints for imports that have the same relative source.
+
+Options are:
+- `case-sensitive` - Orders alphabetically, case-sensitive
+- `case-insensitive` - Orders alphabetically, case-insensitive
```
sindresorhus: Let's call these `alphabetical-case-insensitive`. I know it's verbose, but better to be clear.

rules/import-path-order.js
```diff
@@ -0,0 +1,473 @@
...
+	if (prev.name === next.name) {
+		return 0;
+	}
+
+	if (prev.name < next.name) {
+		return -1;
+	}
+
+	return 1;
+}
```
fisker: how about String#localeCompare

MrHen:
Closing as I've been pulled away from JavaScript / OSS development due to IRL responsibilities.

@sindresorhus Thanks for letting me contribute to the repository but I can't commit any time to helping out until at least Q3 2020. If you want to remove me as a contributor feel free.

sindresorhus:
Don’t worry. This is open source. We do it for fun and when we have time. IRL is more important. I really appreciate all you have done 🙌. You are welcome back any time. I’m not going to remove you as a maintainer.

Issue #247, opened 2019-02-16
Use JSON schema `default` in rules


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2430%20Rewarded-%237E24E3.svg" />](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/247)

Instead of manual `Object.assign`. See: https://github.com/eslint/eslint/commit/550de1e611a1e9af873bcb18d74cf2056e8d2e1b

Note this though: https://github.com/eslint/eslint/commit/de5cbc526b30405e742b35d85d04361529d49ed4

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $30.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://oss.issuehunt.io/u/issuehunt) ($30.00)

### Submitted pull Requests
- [#2622 Support `Float16Array`](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/2622)
---

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

IssuehuntBot:
@issuehunt has funded $30.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/55832243/issues/247) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/55832243/issues/247).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/users/44827199/repositories) on IssueHunt to raise funds.

sindresorhus:
There seems to be something weird with using `default` in the ESLint rule schema: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/332

One task for this issue is to try to identify if this is an ESLint bug and if so report it there.

GMartigny:
[...]

PR #332, opened 2019-07-05
Use defaults in schema

MrHen:
Draft for #247.

@sindresorhus, before I keep going down this path I wanted to make sure this approach is what you wanted. The `default` part of the schema might not be doing what you thought it should do. It doesn't look like it will apply the defaults automatically in a majority of cases.

If the rule is configured with the `default` in the schema and does _not_ have the fallback during option resolution than the following eslint configuration:

```json
{
  "env": {
      "browser": true,
      "es6": true
  },
  "parser": "babel-eslint",
  "parserOptions": {
      "ecmaVersion": 2018,
      "sourceType": "module"
  },
  "rules": {
      "unicorn/catch-error-name": "error"
  },
  "plugins": [
      "unicorn"
  ]
}
```

Will _not_ fail on the following code:

```js
try {
  console.log('foo');
} catch (foo) {
  console.log(foo); // should fail
}
```

The following code _will_ correctly fail:

```js
/* eslint unicorn/catch-error-name: ["error", { "name": "error" }] */

try {
  console.log('foo');
} catch (foo) {
  console.log(foo); // The catch parameter should be named `error`
}
```


sindresorhus:
If this is true, it sounds more like an ESLint bug than limitation.

The known limitation is that `default` does not work for unnamed options or in `oneOf`, `anyOf`, or `not.` Otherwise, it should work.

MrHen:
Yeah, I thought so too but it doesn't seem to be applying. Maybe I have an old version of ESLint hanging around? Or there is a condition on the schema default not being met. I can spend some time digging into what's going on but it might take a while.

In the mean time, I don't think I can remove the fallback without breaking the rule.

golopot:
It appears to be that if the option is not provided, `context.options[0]` will remained `undefined` instead of an object with defaults.

sindresorhus:
Try rebasing from master. Maybe https://github.com/sindresorhus/eslint-plugin-unicorn/commit/bd9e00bf57b590ed5282557e17cb5eef77a18163 helped.

MrHen:
I have not been able to get the defaults to apply. I have `eslint@6.0.1` installed:

```
npm ls eslint
eslint-plugin-unicorn@9.1.1 /Users/adambabcock/Source/eslint-plugin-unicorn
├── UNMET PEER DEPENDENCY eslint@6.0.1 
└─┬ xo@0.24.0
  └── eslint@5.16.0 

npm ERR! peer dep missing: eslint@^5.5.0, required by eslint-ava-rule-tester@3.0.0
```

The peer dep warning is odd but should be harmless. I am running tests using `npx ava test/catch-error-name.js` and the tests fail because the default settings are not being picked up.

I probably won't be able to spend too much more time digging into this. If anyone else wants to carry the torch, be my guest.

sindresorhus:
Closing for now. It's not very important and not worth you wasting your time on it :)

https://github.com/sindresorhus/eslint-plugin-unicorn/issues/247#issuecomment-552798134

Issue #267, opened 2019-03-28
Should the `prevent-abbreviations` rule work on filenames?


fisker:
[<img alt="Issuehunt badges" src="https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/267/badge.svg" />](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/267)

maybe a `checkFilenames` option

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/267).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

sindresorhus:
Yes, I think that would be useful.

IssuehuntBot:
@issuehunt has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/55832243/issues/267) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/55832243/issues/267).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

PR #336, opened 2019-07-06
Add `checkFilenames` option to the `prevent-abbreviations` rule

fisker:
fixes: #267

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -1,10 +1,12 @@
 'use strict';
+const path = require('path');
 const astUtils = require('eslint-ast-utils');
 const defaultsDeep = require('lodash.defaultsdeep');
 const toPairs = require('lodash.topairs');
 const camelCase = require('lodash.camelcase');
 const kebabCase = require('lodash.kebabcase');
 const upperfirst = require('lodash.upperfirst');
+const snakeCase = require('lodash.snakecase');
 
 const getDocsUrl = require('./utils/get-docs-url');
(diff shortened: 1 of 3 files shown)


rules/prevent-abbreviations.js
```diff
@@ -178,6 +180,16 @@ const defaultWhitelist = {
 	stdDev: true
 };
 
+const getCase = string => {
+	for (const fn of [camelCase, kebabCase, snakeCase, upperfirst]) {
```
futpib: Shouldn't this use `pascalCase` instead of `upperfirst`?

test/prevent-abbreviations.js
```diff
@@ -911,6 +926,17 @@ ruleTester.run('prevent-abbreviations', rule, {
...
+		},
+		// `checkFilenames` option
+		{
+			code: 'foo();',
+			filename: 'err/http-err.js',
+			errors: createErrors()
+		},
+		{
+			code: 'foo();',
+			filename: 'http-err.js',
```
futpib: Could we also add a test for each kind of case from `getCase`? And also a test with an underscore (`_helper.js`) and dot (`.babelrc.js`) prefixes and a dot separator (`next.config.js`)

rules/prevent-abbreviations.js
fisker: will send another PR to fix it

test/prevent-abbreviations.js
fisker: will add more tests

PR #337, opened 2019-07-07
Refactor the `prevent-abbreviations` rule

fisker:
What I have done:

1, refact `getNameReplacements`, It should handle more kinds of strings, hope @futpib don't mind
2, update test fix #336 review

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -232,12 +232,12 @@ const getWordReplacements = (word, replacements) => {
 	return wordReplacement.length > 0 ? wordReplacement.sort() : [];
 };
 
-const getReplacementsFromCombined = (combined, length = Infinity) => {
-	const total = combined.reduce((total, {length}) => total * length, 1);
+const getReplacementsFromCombinations = (combinations, length = Infinity) => {
+	const total = combinations.reduce((total, {length}) => total * length, 1);
 
 	const samples = Array.from({length: Math.min(total, length)}, (_, sampleIndex) => {
 		let indexLeft = sampleIndex;
-		return combined.reduceRight((words, replacements) => {
+		return combinations.reduceRight((words, replacements) => {
 			const {length} = replacements;
 			const replacementIndex = indexLeft % length;
 			indexLeft -= replacementIndex;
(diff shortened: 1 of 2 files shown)


test/prevent-abbreviations.js
```diff
@@ -43,6 +43,34 @@ const createErrors = message => {
 	return [error];
 };
 
+const anotherNameMessage = 'A more descriptive name will do too.';
+
+const formatMessage = (discouragedName, replacements, nameTypeText, replacementsLimit = 3) => {
```
futpib: Copy-pasting the tested code (this is literally [`formatMessage` from `rules/prevent-abbreviations.js`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/4c3d61c04318f75480bb6ab30bf948fa1643d550/rules/prevent-abbreviations.js#L363)) diminishes the purpose of testing.

I suggest instead we have a few tests with expected messages hardcoded in the test (at least one for each code path to keep the coverage), and have a bunch of tests without the exact message (with `ruleId` only) to avoid unnecessary repetitiveness.

test/prevent-abbreviations.js
```diff
@@ -200,64 +228,62 @@ ruleTester.run('prevent-abbreviations', rule, {
 	invalid: [
 		noFixingTestCase({
 			code: 'let e',
-			errors: createErrors('Please rename the variable `e`. Suggested names are: `error`, `event`. A more descriptive name will do too.')
+			errors: createErrors(formatMessage('e', ['event', 'error'], 'variable'))
```
futpib: Let's keep the old `errors` value here as per this comment https://github.com/sindresorhus/eslint-plugin-unicorn/pull/337/files#r300863993

test/prevent-abbreviations.js
fisker: the reason I copied this function is replacements orders really annoying, so I  sort the result. and I  want test the replacements. maybe we can do a unit test for `getNameReplacements`

fisker: removed

test/prevent-abbreviations.js
fisker: restored

rules/prevent-abbreviations.js
```diff
@@ -231,146 +217,105 @@ const prepareOptions = ({
...
+			]
+		};
 	}
 
-	return [...variableNameReplacements.keys()]
-		.filter(name => variableNameReplacements.get(name))
-		.map(originalIsInPascalCase ? pascalCase : name => name)
-		.sort();
-};
+	let options = [[]];
```
fisker: this can be quick loop from 0 to limit, I will figure out later

fisker: It's `cartesian product`, I've done this before but forgot how to calculate

futpib: Why is changing `getWordByWordReplacements` even necessary here?

fisker: I want have a function can get replacements only by word, no matter what cases, file names is more complicated than variable names

fisker:
@futpib ready for review

sindresorhus:
Thanks, @fisker 👍

PR #331, opened 2019-07-05
Update ESLint to v6

fisker:
Fixes #326

PR #352, opened 2019-09-06
Add utils `quoteString`

fisker:
this PR fix some problems in #331 , improves readability


futpib:
Can we use `@babel/types` and `@babel/generator` for a more general solution?

I think code generation via string concatenation should be avoided unless absolutely necessary.

```js
const t = require('@babel/types');
const { default: generate } = require('@babel/generator');

console.log(generate(t.regExpLiteral("Whomst'd've")));
```

PR #353, opened 2019-09-06
Add `BigInt64Array` and `BigUint64Array` to the `new-for-builtins` rule

fisker:


sindresorhus:
Can you check if any of the other rules needs these types or support for `BigInt` / `1n`?

PR #354, opened 2019-09-06
Refactor: use `map-obj` to add `unicorn/` prefix to recommended config rules

fisker:
I'm not sure about this PR, should we?


sindresorhus:
Nah, not worth it. I think it's fine to be explicit here.

PR #356, opened 2019-09-06
Add `isBigInt` to `prefer-type-error`

fisker:


sindresorhus:
Unrelated to this, but seems like we should just have a list of types and generate most of the `tcIdentifiers`. We should also support `is.bigInt` and others, not just `isBigInt`.

PR #362, opened 2019-09-09
Add `prefer-modern-dom-apis` rule

### LINKED ISSUE #344: Rule proposal:  `prefer-modern-dom-apis`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/344)

There more methods like `appendChild` and `removeChild`, for which we already have rules. 
They've all been replaced with better versions:

- [insertBefore](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore) -> [before](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/before)
- [replaceChild](https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild) -> [replaceWith](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith)
- [insertAdjacentText](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentText) -> [before](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/before), [after](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/after), [append](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/append), [prepend](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/prepend)
- [insertAdjacentElement](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement) -> [before](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/before), [after](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/after), [append](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/append), [prepend](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/prepend)

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

### Submitted pull Requests
- [#362 Add `prefer-modern-dom-apis` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/362)
---

### Tips

---
</details>

mhatvan:
Fixes #344.

The rule can be tested in this gist here: https://astexplorer.net/#/gist/338d492213d95b40122046fb29abd1bc/bb5030be4689f651524402efad8d24d2b310d588

The links in the description for `prepend()` and `append()` from MDN are returning 404 by the way, they should be updated.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#344: Rule proposal:  `prefer-modern-dom-apis`](https://issuehunt.io/repos/55832243/issues/344)
---

</details>


rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+const getDocsUrl = require('./utils/get-docs-url');
+
+const checkForReplaceChildOrInsertBefore = (context, node) => {
+	const getArgumentName = args => {
+		if (args.type === 'Identifier') {
+			return args.name;
+		}
+
+		return null;
+	};
```
MrHen: Move this helper function out of `checkForReplaceChildOrInsertBefore`. There may be a suitable util here: https://github.com/sindresorhus/eslint-plugin-unicorn/tree/master/rules/utils. Perhaps `is-method-named`?

Also, `args` is not a very good argument name. Perhaps `nodeArgument`?

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+
+	const forbiddenIdentifierNames = new Map([
+		['replaceChild', 'replaceWith'],
+		['insertBefore', 'before']
+	]);
+
+	const identifierName = node.callee.property.name;
+
+	// Only handle methods that exist in forbiddenIdentifierNames
+	if (forbiddenIdentifierNames.has(identifierName)) {
```
MrHen: I prefer the return-early when the condition is being used as a guard.

```
	if (!forbiddenIdentifierNames.has(identifierName)) {
		return;
	}
```

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+			return args.name;
+		}
+
+		return null;
+	};
+
+	const forbiddenIdentifierNames = new Map([
+		['replaceChild', 'replaceWith'],
+		['insertBefore', 'before']
+	]);
```
MrHen: Move this to a constant variable outside of `checkForReplaceChildOrInsertBefore`.

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+		if (args.type === 'Identifier') {
+			return args.name;
+		}
+
+		if (args.type === 'Literal') {
+			return args.raw;
+		}
+
+		return null;
+	};
```
MrHen: Same comment as before. This should be moved to the parent scope.

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+
+	const positionReplacers = new Map([
+		['beforebegin', 'before'],
+		['afterbegin', 'prepend'],
+		['beforeend', 'append'],
+		['afterend', 'after']
+	]);
```
MrHen: Same comment as before. This should be moved to the parent scope as a constant.

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+
+	const identifierName = node.callee.property.name;
+
+	if (
+		identifierName === 'insertAdjacentText' ||
+		identifierName === 'insertAdjacentElement'
+	) {
```
MrHen: Same comment as before. Use return-early pattern.

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,121 @@
...
+		});
+	}
+};
+
+const create = context => {
+	return {
+		CallExpression(node) {
+			if (
+				node.callee.type === 'MemberExpression' &&
+				node.arguments.length === 2
```
MrHen: You can include these checks in the selector pattern for the rule. Not required but in case you hadn't seen it yet, here are the docs: https://eslint.org/docs/developer-guide/selectors

test/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,129 @@
...
+			errors: [
+				{
+					message:
+						'Prefer `referenceNode.after(newNode)` over `referenceNode.insertAdjacentElement("afterend", newNode)`.'
+				}
+			],
+			output: 'referenceNode.after(newNode);'
+		}
+	]
+});
```
MrHen: Add tests for the following scenarios:

```
referenceNode.insertAdjacentElement(foo, bar);

referenceNode.insertAdjacentElement(
    "afterend",
    newNode
);

referenceNode.insertAdjacentElement( // inline comments
    "afterend", // inline comments
    newNode,  // inline comments
); // inline comments

parentNode.replaceChild(
    newChildNode,
    oldChildNode
);

parentNode.replaceChild( // inline comments
    newChildNode, // inline comments
    oldChildNode // inline comments
);

parentNode.insertBefore(alfa, beta).insertBefore(charlie, delta);
```

readme.md
```diff
@@ -142,19 +140,17 @@ See the [ESLint docs](http://eslint.org/docs/user-guide/configuring#extending-co
...
-
 ## Maintainers
 
 - [Sindre Sorhus](https://github.com/sindresorhus)
 - [Sam Verschueren](https://github.com/SamVerschueren)
 - [futpib](https://github.com/futpib)
 - [Adam Babcock](https://github.com/MrHen)
 
-###### Former
+### Former
```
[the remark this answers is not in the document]
mhatvan: In this case, these were intended, manual changes because markdownlint threw me a couple of warnings e.g. https://github.com/DavidAnson/markdownlint/blob/v0.16.0/doc/Rules.md#md001

But I am fine with reverting back to how it was before, however ;)

rules/prefer-modern-dom-apis.js
mhatvan: Do you mean the global scope?
Why would I move it outside when it is only used in `checkForReplaceChildOrInsertBefore`?
The more scoped the better imo

rules/prefer-modern-dom-apis.js
mhatvan: Good point, changing it in the upcoming commit.

rules/prefer-modern-dom-apis.js
mhatvan: `is-method-named` would not help me, since i cannot say what the argument name will be.
There are multiple other rules with `const args = node.arguments;` and their own scoped helper functions, e.g: `prefer-node-remove` in the code already, but I am also fine with renaming it ;)

test/prefer-modern-dom-apis.js
mhatvan: Cool, thanks for these test scenario suggestions!
So I guess the inline comments should be preserved? How can I do that?

readme.md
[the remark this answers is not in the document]
MrHen: I'm okay making the changes in a separate PR.

test/prefer-modern-dom-apis.js
MrHen: It's a total pain. If you want, you can disable the fixer if there are comments in the node. Take a look at `getCommentsInside` or `commentsExistBetween`.

https://eslint.org/docs/developer-guide/working-with-rules#contextgetsourcecode

rules/prefer-modern-dom-apis.js
MrHen: We recently introduced a rule that will enforce this behavior. See the discussion here: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/230.

The short version is that there isn't any significant benefit to scoping the function and a couple minor benefits to code readability, maintainability and performance if you move it out.

test/prefer-modern-dom-apis.js
mhatvan: `referenceNode.insertAdjacentElement(foo, bar);` -> this is incorrect syntax, I would not touch it and not modify it and add it to the valid array, correct?

`parentNode.insertBefore(alfa, beta).insertBefore(charlie, delta);` -> how to handle this?

MrHen: `referenceNode.insertAdjacentElement(foo, bar);` should be just as valid as `referenceNode.insertAdjacentElement("beforebegin", newNode);`. It's just using a variable instead of a literal.

----

`parentNode.insertBefore(alfa, beta).insertBefore(charlie, delta);` cannot be fixed properly because the return value is being used. A similar test would be `var foo = parentNode.insertBefore(alfa, beta);`

mhatvan:
@MrHen can you take another look at the changes you requested? I implemented your recommendations a while ago.

MrHen:
@mhatvan Sorry, I'm actually in the middle of changing employers right now. I probably won't get a chance to look at it in the next couple weeks.

readme.md
```diff
@@ -8,14 +8,12 @@ You might want to check out [XO](https://github.com/xojs/xo), which includes thi
 
 [**Propose or contribute a new rule ➡**](.github/contributing.md)
 
-
```
fisker: restore blank lines in this file please

readme.md
```diff
@@ -8,14 +8,12 @@ You might want to check out [XO](https://github.com/xojs/xo), which includes thi
...
 ## Install
 
-```
+```bash
```
fisker: restore this,since there is `$` inside

index.js
```diff
@@ -13,9 +13,7 @@ module.exports = {
 				ecmaVersion: 2020,
 				sourceType: 'module'
 			},
-			plugins: [
-				'unicorn'
-			],
+			plugins: ['unicorn'],
```
fisker:
```suggestion
			plugins: [
				'unicorn'	
			],
```

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,136 @@
+'use strict';
+const getDocsUrl = require('./utils/get-docs-url');
```
fisker: rebase this PR, 
run `yarn lint rules/prefer-modern-dom-apis.js --fix`
you'll get an error here, and will be fixed

fisker:
since [`childNode.replaceWith()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith) and [`Node.replaceChild()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild) returns different value

also `ChildNode.before()`

[`ChildNode.before()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/before)
[`Node.insertBefore()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore)
[`Element.insertAdjacentElement()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement)

so, maybe you want do a similar check like https://github.com/sindresorhus/eslint-plugin-unicorn/pull/321

readme.md
fisker: there is `$` inside

rules/prefer-modern-dom-apis.js
fisker: sorry this will not fix currectly, after rebase, rename it to `getDocumentationUrl`

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d638e5477d7240ea777dcaf249d0121a5a1910eb/rules/catch-error-name.js#L4

readme.md
mhatvan: More than one blank line is bad practice, especially since it doesn't affect the layout at all.

readme.md
mhatvan: wouldn't it be better to have it conforming to markdownlint? 

https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md040

mhatvan:
@fisker I pushed some code adaptions in the newest commit.

fisker:
what about this?

> since [`childNode.replaceWith()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith) and [`Node.replaceChild()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/replaceChild) returns different value
> 
> also `ChildNode.before()`
> 
> [`ChildNode.before()`](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/before)
> [`Node.insertBefore()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore)
> [`Element.insertAdjacentElement()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement)
> 
> so, maybe you want do a similar check like #321

mhatvan:
> so, maybe you want do a similar check like #321
you linked the complete PR without specifying what you exactly want me to change, I need more details here.

fisker:
> you linked the complete PR without specifying what you exactly want me to change, I need more details here.

@mhatvan I mean some methods returns different result, so we need check if the return value is used.

docs/rules/prefer-modern-dom-apis.md
```diff
@@ -0,0 +1,52 @@
+# Prefer modern DOM APIs
+
+Enforces the use of:
+
+- `childNode.replaceWith(newNode);` over `parentNode.replaceChild(newNode, oldNode)`;
+- `referenceNode.before(newNode);` over `parentNode.insertBefore(newNode, referenceNode);`
+- `referenceNode.before("text");` over `referenceNode.insertAdjacentText("beforebegin", "text");`
+- `referenceNode.before(newNode);` over `referenceNode.insertAdjacentElement("beforebegin", newNode);`
```
sindresorhus: Can you linkify all these to their MDN page?

readme.md
sindresorhus:
```suggestion
```console
```

works better

mhatvan:
@sindresorhus I implemented the changes you requested, can we get this merged now?

I still don't understand @fisker feedback:
> childNode.replaceWith() and Node.replaceChild() returns different value

I implemented it according to the examples in MDN and they return the expected result.

sindresorhus:
@mhatvan `Node.replaceChild()` returns a value. If the user actually uses this value, we cannot safely auto-fix with with `ChildNode.replaceWith()` as it doesn't return any value. Can you handle this and add a test for that?

mhatvan:
@sindresorhus Sorry, I missed that you already left a comment a while ago. I pushed another commit which implements the requested changes and added additional tests.

fisker:
@mhatvan I think we should still  report an error, just can't fix

sindresorhus:
>@mhatvan I think we should still  report an error, just can't fix

👍

mhatvan:
@sindresorhus @fisker feedback implemented, I pushed another commit ;)

sindresorhus:
Looks good. Thanks for contributing 👌

sindresorhus:
Looks good. Thanks for contributing 👌

fregante:
Awesome!

rules/prefer-modern-dom-apis.js
```diff
@@ -0,0 +1,158 @@
...
+	['insertBefore', 'before']
+]);
+
+const isPartOfVariableAssignment = nodeParentType => {
+	if (nodeParentType === 'VariableDeclarator' || nodeParentType === 'AssignmentExpression') {
+		return true;
+	}
+
+	return false;
+};
```
fisker: I don't think this is enough, 

simple one:

```js
console.log(referenceNode.insertAdjacentElement("beforebegin", newNode));
``

PR #378, opened 2019-09-18
`consistent-function-scoping`: Do not fail on same-scoped function references

### LINKED ISSUE #375: consistent-function-scoping false positive calling same-scoped function
Running ESLint 6.4.0 with eslint-plugin-unicorn 11.0.0 and "extends": "plugin:unicorn/recommended" on a file with the following reduced example code:

```js
'use strict';

module.exports = function recordErrors(eventEmitter, stateArgument) {
  function onError(error) {
    stateArgument.inputError = error;
  }
  function onError2(error) {
    onError(error);
  }

  eventEmitter.once('error', onError2);
};
```
gives the following output:

    /tmp/consistent-function-scoping-issue/issue2.js
      7:3  error  Move function to the outer scope  unicorn/consistent-function-scoping
    
I believe this is incorrect since `onError2` can not be moved to the global scope, since it calls `onError`, which is only defined in the function scope.

Thanks,
Kevin

Note: Split from #374 per https://github.com/sindresorhus/eslint-plugin-unicorn/issues/374#issuecomment-532348080

MrHen:
Fixes #375.

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -38,15 +38,38 @@ function checkReferences(scope, parent, scopeManager) {
 			return true;
 		}
 
+		// This check looks for neighboring function definitions
+		const hitIdentifier = variable.identifiers.some(identifier => {
+			// Only look at identifiers that live in a FunctionDeclaration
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for the quick fix 👌

PR #381, opened 2019-09-18
Update `filename-case` rule to handle complicated filename

### LINKED ISSUE #343: False error on kebab-case files
I have file with name like [id].tsx

and getting following error:

```
/pages/[id].tsx
  1:1  error  Filename is not in kebab case. Rename it to `id.tsx`  unicorn/filename-case
```

fisker:
Update `filename-case` to validate only `a-z`, `0-9`, `-`, `_` and `$`.

fixes #343

### DIFF
--- a/rules/filename-case.js
+++ b/rules/filename-case.js
@@ -4,12 +4,15 @@ const camelCase = require('lodash.camelcase');
 const kebabCase = require('lodash.kebabcase');
 const snakeCase = require('lodash.snakecase');
 const upperfirst = require('lodash.upperfirst');
+
 const getDocsUrl = require('./utils/get-docs-url');
+const cartesianProductSamples = require('./utils/cartesian-product-samples');
 
 const pascalCase = string => upperfirst(camelCase(string));
 const numberRegex = /(\d+)/;
 const PLACEHOLDER = '\uFFFF\uFFFF\uFFFF';
 const PLACEHOLDER_REGEX = new RegExp(PLACEHOLDER, 'i');
+const isIgnoredChar = char => !/^[a-z0-9-_$]$/i.test(char);
 
 function ignoreNumbers(fn) {
 	return string => {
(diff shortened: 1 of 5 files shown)


test/filename-case.js
```diff
@@ -90,7 +90,8 @@ ruleTester.run('filename-case', rule, {
 		testManyCases('src/foo/fooBar.js', {camelCase: true}),
 		testManyCases('src/foo/FooBar.js', {kebabCase: true, pascalCase: true}),
 		testManyCases('src/foo/___foo_bar.js', {snakeCase: true, pascalCase: true}),
-		testCaseWithOptions('src/foo/bar.js')
+		testCaseWithOptions('src/foo/bar.js'),
+		testCase('src/foo/[bar].js', 'camelCase')
```
sindresorhus: Can you add some more test-cases? For example:

- `src/foo/{bar}.js`

rules/filename-case.js
```diff
@@ -4,12 +4,15 @@ const camelCase = require('lodash.camelcase');
...
 const upperfirst = require('lodash.upperfirst');
+
 const getDocsUrl = require('./utils/get-docs-url');
+const cartesianProductSamples = require('./utils/cartesian-product-samples');
 
 const pascalCase = string => upperfirst(camelCase(string));
 const numberRegex = /(\d+)/;
 const PLACEHOLDER = '\uFFFF\uFFFF\uFFFF';
 const PLACEHOLDER_REGEX = new RegExp(PLACEHOLDER, 'i');
+const isIgnoredChar = char => !/^[a-z0-9-_$]$/i.test(char);
```
sindresorhus:
```suggestion
const isIgnoredChar = char => !/^[a-z\d-_$]$/i.test(char);
```

rules/filename-case.js
```diff
@@ -4,12 +4,15 @@ const camelCase = require('lodash.camelcase');
 const kebabCase = require('lodash.kebabcase');
 const snakeCase = require('lodash.snakecase');
 const upperfirst = require('lodash.upperfirst');
+
```
sindresorhus:
```suggestion
```

sindresorhus:
Can you also update the docs about this?

test/filename-case.js
fisker: done

fisker:
All request changes done.

sindresorhus:
Looks good :)

PR #382, opened 2019-09-18
`consistent-function-scoping`: Ignore anything containing JSX

### LINKED ISSUE #376: consistent-function-scoping breaks down on React HOCs
Given this code:

```js
function withSomething() {
  return function withSomethingFactory(WrappedComponent) {
    function WithSomething(props) {
      // HOC does something cool
      return <WrappedComponent {...props} />;
    }

    return WithSomething;
  };
}
```

The rule fails on line 3. However, `WrappedComponent` is being referenced from the parent as a JSX element.

MrHen:
Fixes #376.
Depends on #378.

I could not find a quick way to include JSX elements in the reference detection. It looks like the scope utilities don't recognize JSX.

We can circle back around to this later but I wanted to get the bugs cleared out first.

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -125,7 +125,7 @@ const create = context => {
 	const {scopeManager} = sourceCode;
 
 	const reports = [];
-	let jsx = false;
+	let hasJsx = false;
 
 	return {
 		ArrowFunctionExpression: node => {
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -30,9 +30,6 @@ function checkReferences(scope, parent, scopeManager) {
 		}
 
 		const hitDefinitions = variable.defs.some(definition => {
-			console.log('CCC.2', {
-				definition,
-			});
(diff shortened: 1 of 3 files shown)


rules/consistent-function-scoping.js
```diff
@@ -121,30 +121,48 @@ const create = context => {
...
+			} else {
+				reports.push({
+					node,
+					messageId: MESSAGE_ID_FUNCTION
+				});
+			}
+		},
+		JSXElement: () => {
+			// Turn off this rule if we see a JSX element because scope
+			// references does not include JSXElement nodes.
```
MrHen: Copying [comment](https://github.com/MrHen/eslint-plugin-unicorn/pull/1#discussion_r325481128) from @sindresorhus:

> Can you open an ESLint issue about this afterwards?

Looks like there was a similar issue opened here: https://github.com/eslint/eslint/issues/3622

I can try getting something like that hooked up. Ideally, if someone who uses `eslint-plugin-unicorn` also had `eslint-plugin-react` then it would "just work". I'm not sure if we can realistically test for that, though.

Issue #367, opened 2019-09-14
`prevent-abbreviations` is too eager


dasZGFz:
This code:
```js
const image = new Image();
image.src = '';
```
outputs this as an error:
```
The property src should be named source. A more descriptive name will do too.  unicorn/prevent-abbreviations
```

My current workaround is to disable eslint for that line but I wonder if there's a way to fix it in the plugin.

garyking:
sindresorhus's own package, [execa](https://github.com/sindresorhus/execa), uses `env` as an option, yet this rule reports an error for that abbreviation, too. So I agree that sometimes the rule is too aggressive. It shouldn't report on arguments that we pass to a function, or values returned from a function. Since if those were functions that we created ourselves and were in our own scripts, then those occurrences would already be reported.

Issue #271, opened 2019-03-28
`prevent-abbreviations` default option is not good


fisker:
`prevent-abbreviations` is really good, I have tried to lint my old scripts.

after a few test, I notice that

`checkDefaultAndNamespaceImports` default value is `false`, but normally change the name of `default and namespace` import has no side effect, this should set to `true`, I think.

`checkProperties` default value is `true`, but normally when export or return a object, the prop name can't just rename without api changes, I think this should set to `false` by default.

fisker:
This is what I'm using, feel better when linting, no need change api (including internal using props)

```js
    'unicorn/prevent-abbreviations': [
      'error',
      {
        checkDefaultAndNamespaceImports: true,
        checkProperties: false,
      },
    ],
```

futpib:
Could you please split this into two separate issues (one for each default we've chosen)?

futpib:
Some reasons behind `checkDefaultAndNamespaceImports: true` are outlined in the discussion starting from here https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237#issuecomment-470942156

Regarding `checkProperties`, we currently ignore names exported via `export` keyword. Maybe we should ignore `module.exports` too.

fisker:
thanks for explaining `checkDefaultAndNamespaceImports`.

for `checkProperties`, I don't think `export` / `module.exports` is good enough, how do you know `return` is not a kind of "exporting", maybe just for another function to use, or handwriting `umd` or `iife` mod. 

since those two options can easily overwrite, I'm going to keep my own choices.

sobolevn:
I am facing this issue too. Since the new release linter checks options that I do not have control over. That's sad. 

Failing build: https://travis-ci.org/wemake-services/wemake-vue-template/builds/513997750
Code: https://github.com/wemake-services/wemake-vue-template/blob/master/template/nuxt.config.ts#L93

This is a default `Nuxt` function which works with the default API:

```js
{
  'build': {
      // Error here on `isDev`:
      extend (config, { isDev, isClient }) { ... }
  }
}
```

fisker:
https://github.com/xojs/xo/blob/c55f962a961b5b00b4b4239a7253a2e874b0e700/config/plugins.js#L40

sindresorhus:
> Regarding checkProperties, we currently ignore names exported via export keyword. Maybe we should ignore module.exports too.

👍

sindresorhus:
[...]

PR #387, opened 2019-09-22
Run recommended config against source code

fisker:
Shouldn't we run recommended configs on our own repo? And our own source code should follow recommended rules.

This is a prototype, maybe move to integration test


sindresorhus:
Makes sense to me. You need to add `configs` to `files` in package.json.

fisker:
Just a idea, maybe a better way, will do it later

fisker:
This PR is almost done.

Some source code not following recommended configs, most of them are fixed, some questions:

1) `prevent-abbreviations` default `checkProperties: true` is annoying, I override it for now, maybe we should consider to change that, #271 #367

2) `prevent-abbreviations` also complain about `get-docs-url.js` filename, I can rename it to fix that, and I suggest we do this to follow `recommended` config, just lots of files, should I do that?

3) some comments fails on `expiring-todo-comments` rule, I can add `eslint-disable` but `xo` will reports another error

4) `consistent-function-scoping` seems buggy?

/cc @sindresorhus

fisker:
/cc @sindresorhus 

Please have a look at the problems. 
I need your opinion on them to continue.

sindresorhus:
1. As commented in the linked issue, let's change the default.
2. Yes, rename it.
3. Fixed by https://github.com/sindresorhus/eslint-plugin-unicorn/pull/400
4. Disable it for now.

docs/new-rule.md
```diff
@@ -24,5 +24,6 @@ Use the [`astexplorer` site](https://astexplorer.net) with the `espree` parser a
 	*(The description should be the same as the heading of the documentation file).*
 - Run `$ npm test` to ensure the tests pass.
 - Run `$ npm run integration` to run the rules against real projects to ensure your rule does not fail on real-world code.
+- Run `$ npm run lint` to run the rules against this repository to ensure code in the repository are following your rule.
```
sindresorhus: Why a separate step? It should just be part of `npm test`.

sindresorhus:
Can you fix the merge conflict

docs/new-rule.md
fisker: I was thinking, when some one making a new rule, they can do this after their PR is reviewed, new rules might requires other code changes

sindresorhus: Good point. Ok.

fisker:
`lint.js` can't pass xo, because I'm not using ava?

rules/catch-error-name.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 const astUtils = require('eslint-ast-utils');
 const avoidCapture = require('./utils/avoid-capture');
-const getDocsUrl = require('./utils/get-docs-url');
+const getDocumentsUrl = require('./utils/get-documents-url');
```
futpib: I think `docs` here stands for `documentation` rather than for `documents`.

fisker: So, I should rename them again....

futpib: I recommend using [`rg`](https://github.com/BurntSushi/ripgrep) and `perl` for batch replacements.

```bash
rg 'getDocumentsUrl' -l | xargs perl -i -pe 's/getDocumentsUrl/getDocumentationUrl/g'
rg 'get-documents-url' -l | xargs perl -i -pe 's/get-documents-url/get-documentation-url/g'
```

fisker: done.

fisker:
@futpib any idea why CI fails?

rules/catch-error-name.js
fisker: @futpib I did this with vscode `replace in file`(not sure what this called, using other lang), after I saw you comment, remember me that sindresorhus has a new cli [`replace-in-files-cli`](https://github.com/sindresorhus/replace-in-files-cli)

fisker:
How the same code success on my CI? 

https://travis-ci.com/fisker/eslint-plugin-unicorn/builds/131622273

.travis.yml
```diff
@@ -3,11 +3,17 @@ node_js:
   - '12'
   - '10'
   - '8'
+cache:
+  npm: false
```
fisker: CI finally passed, should I keep this?

sindresorhus: Sure, for now.

PR #389, opened 2019-09-24
Generate docs url automaticly

fisker:
This will makes new rules a little easier to write, but may breaks direct import from `rules` dir

Motivation to do this is `get-docs-url.js` filename is against our `prevent-abbreviations` rule(#387), to fix this we have to change the path in every rules file

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,9 +1,22 @@
 'use strict';
 const path = require('path');
 const importModules = require('import-modules');
+const mapObject = require('map-obj');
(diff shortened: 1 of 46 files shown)


MrHen:
I'm not a big fan of this kind of thing. It's already trivial code and auto-generated code is hard to read, maintain and debug.

fisker:
was trying to avoid replacments in #387 , since I've replaced them all, closing this

PR #390, opened 2019-09-24
Add rule meta test

fisker:


fisker:
sorry didn't see test in `package.js`

PR #397, opened 2019-09-26
Handle lack of package.json in `expiring-todo-comments` rule

lubien:
@garyking came with this scenario of a packageless project. For this my choice would be not to allow package related triggers to work at least until a proper package.json file is set up.

Even tho the original issue says eslint 5, this happens also in version 6.

Fixes #394

PR #399, opened 2019-09-27
Rename variables on expiring-todo-comments

lubien:
See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/397#discussion_r328431028

// @fisker


lubien:
There are other `pkg*` variables. Would it be ok if I update them to `package*` too?

fisker:
> There are other `pkg*` variables. Would it be ok if I update them to `package*` too?

we can do this later, #387 will fix it. thank you

PR #405, opened 2019-10-03
Test if CI will fail

fisker:
Just checking if the CI will fail


fisker:
Should be fixed by #406

PR #408, opened 2019-10-03
Change `prefer-query-selector` rule description to avoid confusion

### LINKED ISSUE #342: Description of `prefer-query-selector` is misleading WRT “consistency”
The description of the [`prefer-query-selector`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/fb5a50ce95f85d958db83bdad3a642ed2cbb732b/docs/rules/prefer-query-selector.md) rule states:

> **Prefer `.querySelectorAll()` over `.getElementsByClassName()` and `.getElementsByTagName()`**
> It's better to be consistent.

"Consistency" has nothing to do here. These methods aren't synonymous with each other: `.getElementsByTagName()` and `.getElementsByClassName()` return an [`HTMLCollection`](https://mdn.io/HTMLCollection) object, which updates dynamically as elements are added or removed from the DOM:

~~~js
const pages = document.getElementsByClassName("page");
console.log(pages.length); // 1

const newPage = Object.assign(document.createElement("div"), {className: "page"});
document.body.appendChild(newPage);
console.log(pages.length) // 2
~~~

Conversely, [`.querySelectorAll()`](https://mdn.io/querySelectorAll) returns a static [`NodeList`](https://mdn.io/NodeList) which does **not** reflect changes to the DOM:

~~~js
const live = document.getElementsByTagName("div");
const dead = document.querySelectorAll("div");
live.length === 3;
dead.length === 3;

document.body.appendChild(document.createElement("div"));
live.length === 4;
dead.length === 3;
~~~

The rule — or at least its description — should be amended to acknowledge this.

hkulur:
Removed the word consistent, which seemed to cause confusion.

Fix #342

### DIFF
--- a/docs/rules/prefer-query-selector.md
+++ b/docs/rules/prefer-query-selector.md
@@ -1,6 +1,6 @@
 # Prefer `.querySelector()` over `.getElementById()`, `.querySelectorAll()` over `.getElementsByClassName()` and `.getElementsByTagName()`
 
-It's better to be use one method to query DOM elements.
+It's better to use a single method to query DOM elements.
 
 This rule is partly fixable.
 
--- a/docs/rules/prefer-query-selector.md
+++ b/docs/rules/prefer-query-selector.md
@@ -1,6 +1,6 @@
 # Prefer `.querySelector()` over `.getElementById()`, `.querySelectorAll()` over `.getElementsByClassName()` and `.getElementsByTagName()`
 
-It's better to use a single method to query DOM elements.
+It's better to use the same method to query DOM elements.
 
 This rule is partly fixable.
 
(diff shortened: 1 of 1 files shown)


docs/rules/prefer-query-selector.md
```diff
@@ -1,6 +1,6 @@
 # Prefer `.querySelector()` over `.getElementById()`, `.querySelectorAll()` over `.getElementsByClassName()` and `.getElementsByTagName()`
 
-It's better to be consistent.
+It's better to be use one method to query DOM elements.
```
fisker: I'm not a native speaker, but this sentence is not right to me

hkulur: My bad. Fixed it now.

docs/rules/prefer-query-selector.md
```diff
@@ -1,6 +1,6 @@
 # Prefer `.querySelector()` over `.getElementById()`, `.querySelectorAll()` over `.getElementsByClassName()` and `.getElementsByTagName()`
 
-It's better to be consistent.
+It's better to use a single method to query DOM elements.
```
fisker:
```suggestion
It's better to use the same method to query DOM elements.
```

fisker: Can you commit suggestion

PR #411, opened 2019-10-08
Improve `prefer-includes` rule

### LINKED ISSUE #410: prefer-includes fix breaks utility library usage (eg. lodash)
`prefer-includes` rule autofix incorrectly applies to usages of utility libraries like lodash.

For example:

```js
let arr = [1, 2, 3];

if (_.indexOf(arr, 1) !== -1) {
    console.log("true");
}
```

Becomes
```js
let arr = [1, 2, 3];

if (_.includes(arr)) {
    console.log("true");
}
```

After the fix is applied, which is semantically different and leads to bugs.

fisker:
1. Ignore `indexOf` more than 2 arguments
2. Ignore `_.indexOf`, fixes #410
3. keep `Array/String.indexOf` method second argument [`fromIndex`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf#Parameters)
4. strip `fromIndex` that is literal `0`

### DIFF
--- a/rules/prefer-includes.js
+++ b/rules/prefer-includes.js
@@ -5,7 +5,7 @@ const isMethodNamed = require('./utils/is-method-named');
 // Ignore {_,lodash,underscore}.indexOf
 const ignoredVariables = new Set(['_', 'lodash', 'underscore']);
 const isNegativeOne = (operator, value) => operator === '-' && value === 1;
-const isIgnoredTarget = node => node.type === 'Identifier' && ignoredVariables.has(node.name)
+const isIgnoredTarget = node => node.type === 'Identifier' && ignoredVariables.has(node.name);
 
 const report = (context, node, target, argumentsNodes) => {
 	const sourceCode = context.getSourceCode();
--- a/rules/prefer-includes.js
+++ b/rules/prefer-includes.js
@@ -2,7 +2,10 @@
 const getDocsUrl = require('./utils/get-docs-url');
 const isMethodNamed = require('./utils/is-method-named');
 
+// Ignore {_,lodash,underscore}.indexOf
+const ignoredVariables = new Set(['_', 'lodash', 'underscore']);
 const isNegativeOne = (operator, value) => operator === '-' && value === 1;
+const isIgnoredTarget = node => node.type === 'Identifier' && ignoredVariables.has(node.name)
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
> Ignore _.indexOf, fixes #410

Can you also ignore `lodash.indexOf` and `underscore.indexOf`?

fisker:
> > Ignore _.indexOf, fixes #410
> 
> Can you also ignore `lodash.indexOf` and `underscore.indexOf`?

done

fisker:
forgot test ...

fisker:
@sindresorhus 

I just realize that `lodash` also has `includes`, so we can still replace `indexOf` with `includes`, but `underscore` don't has `includes` but `contains`

1, should I replace `lodash.indexOf` with `lodash.includes`?
2, should I replace `underscore.indexOf` with `underscore.contains`, rule name is `prefer-includes`?
3, still ignore `_.indexOf`?

sindresorhus:
I would ignore them all. Yes, Lodash has `.includes()` today, but it might remove it in the future since it's already in the language now. Better to just leave it.

fisker:
Then this PR is ready

rules/prefer-includes.js
```diff
@@ -26,29 +32,40 @@ const create = context => ({
...
+		if (!isMethodNamed(left, 'indexOf')) {
+			return;
+		}
+
+		const target = left.callee.object;
 
-			if (right.type === 'UnaryExpression') {
-				const {argument} = right;
+		// Ignore {_,lodash,underscore}.indexOf
+		if (target.type === 'Identifier' && ['_', 'lodash', 'underscore'].includes(target.name)) {
```
MrHen: This shouldn't be hardcoded here. Either (a) convert it to a blacklist configuration property with these values as a default or (b) move the values to a constant at the top of the file.

fisker: will fix it tomorrow with (b) 

I think it should be whitelist not blacklist, right?

MrHen: The phrase "ignore" is less ambiguous. Maybe something like `ignoreWords`.

fisker: Good idea, going to get up and fix it now

sindresorhus:
Nice work on improving this rule 👌

PR #416, opened 2019-10-11
Small tweaks

fisker:
I was trying to make `prefer-includes.js` more readable, but found something else I can do.

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -65,7 +65,7 @@ const create = context => {
 				problem.fix = fixer => {
 					const fixings = [fixer.replaceText(node, expectedName)];
 
-					const variables =	scopeManager.getDeclaredVariables(scopeNode);
+					const variables = scopeManager.getDeclaredVariables(scopeNode);
 					for (const variable of variables) {
 						if (variable.name !== node.name) {
 							continue;
--- a/rules/utils/is-literal-value.js
+++ b/rules/utils/is-literal-value.js
@@ -1,0 +1,2 @@
+'use strict';
+module.exports = value => node => node && node.type === 'Literal' && node.value === value;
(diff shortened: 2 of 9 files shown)


rules/catch-error-name.js
```diff
@@ -65,7 +65,7 @@ const create = context => {
 				problem.fix = fixer => {
 					const fixings = [fixer.replaceText(node, expectedName)];
 
-					const variables =	scopeManager.getDeclaredVariables(scopeNode);
+					const variables = scopeManager.getDeclaredVariables(scopeNode);
```
fisker: How did this sneak in

MrHen: White space rules probably aren't looking for tabs inside of lines. They only check leading whitespace.

But definitely suspicious...

rules/utils/is-literal-value.js
```diff
@@ -0,0 +1,2 @@
+'use strict';
+module.exports = value => node => node && node.type === 'Literal' && node.value === value;
```
MrHen: I don't think the utility function should be automatically curried. It should just take both value and node as parameters since you will always need both to do the valid comparison.

Personally, the way I'd handle this is to make a base utility for checking node type + value:

```
module.exports = (node, { type, value }) => {
    if (!node) {
        return false;
    }

    if (node.type !== type) {
        return false;
    }

    return node.value !== value;
}
```

Then use that to create a specific literal check:

```
module.exports = (node, { value }) => {
    return is-node-match(node, { type: 'Literal', value });
}
```

But really, the farther you go down this path the more you just want a deep object comparison utility.

fisker: this code is  from  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/a2469ade9d2349fc74b9ed0ce67024cb3404b6b8/rules/no-for-loop.js#L5

I was  trying to  write

```js
const isLiteralValue = (node, value) => node && node.type === 'Literal' && node.value === value;

const isLiteralZero = node  => isLiteralValue(node, 0)
```

but curry seems easier to use, wish we can  use [Partial Application]https://github.com/tc39/proposal-partial-application) much easier to use 

```js
const isLiteralZero = isLiteralValue(?, 0)
```

Issue #415, opened 2019-10-10
Rule proposal: `prefer-negative-index` for things like `foo.slice`


MrHen:
See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/409/files#diff-1e364de46e721b05ccf3d616d4248ebdR37

Functions that support negative indexing should use that feature instead of calculating from `length`.

## Fail
```
foo.slice(1, foo.length - 1);
```

## Pass
```
foo.slice(1, -1);
```

## Functions

* [Array.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Syntax) but NOT [String.prototype.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Syntax)
* [Array.prototype.indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf#Syntax) but NOT [String.prototype.indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/indexOf#Syntax)
* [Array.prototype.lastIndexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/lastIndexOf#Syntax) but NOT [String.prototype.lastIndexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf#Syntax)
* [Array.prototype.slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Syntax) and [String.prototype.slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice#Syntax)
* [Array.prototype.splice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#Syntax)

fisker:
It's good to have  this rule to prevent this accident happen.

Can't remember any other method use negative index except `substr`(should not use at all)

MrHen:
I added a few examples of where this is supported. Surprisingly, `Array` and `String` have different support.

sindresorhus:
Good idea!

fisker:
test on chrome

includes

```js
'foobar'.includes('b', -5) // true
[...]

PR #417, opened 2019-10-11
Add `prefer-negative-index` rule

fisker:
Fixes: #415


docs/rules/prefer-negative-index.md
```diff
@@ -0,0 +1,23 @@
+# Prefer `-n` over `.length - n` for `slice` and `splice`
```
sindresorhus:
```suggestion
# Prefer negative index over `.length - index` for `String#slice()` and `String#splice()`
```

In the readme too.

docs/rules/prefer-negative-index.md
```diff
@@ -0,0 +1,23 @@
...
+
+Prefer negative index over calculating from `length` for `slice` and `splice`
```
sindresorhus:
```suggestion
Prefer negative index over calculating from `.length` for `String#slice()` and `String#splice()`
```

And link to MDN for those methods.

rules/prefer-negative-index.js
```diff
@@ -0,0 +1,307 @@
...
+				'Int16Array',
+				'Uint16Array',
+				'Int32Array',
+				'Uint32Array',
+				'Float32Array',
+				'Float64Array',
+				'BigInt64Array',
+				'BigUint64Array'
+				// 'Blob'
+				// 'File'
```
sindresorhus: Why are these commented out?

rules/prefer-negative-index.js
```diff
@@ -0,0 +1,307 @@
...
+		return lengthMemberExpression;
+	}
+};
+
+const getRemovalRange = (node, sourceCode) => {
+	let before = sourceCode.getTokenBefore(node);
+	let after = sourceCode.getTokenAfter(node);
+
+	let [start] = node.range;
+	let [, end] = node.range;
```
sindresorhus: Use `[0]` and `[1]` instead of `start`/`end`, as the latter are not official properties.

rules/prefer-negative-index.js
```diff
@@ -0,0 +1,307 @@
...
+			.map(index => getRemoveAbleNode(target, argumentsNodes[index]))
+			.filter(Boolean);
+
+		if (removeAbleNodes.length === 0) {
+			return;
+		}
+
+		context.report({
+			node,
+			message: `Prefer \`-n\` over \`.length - n\` for \`${method}\``,
```
sindresorhus:
```suggestion
			message: `Prefer negative index over length minus index for \`${method}\`.`,
```

rules/prefer-negative-index.js
```diff
@@ -0,0 +1,307 @@
...
+			fix(fixer) {
+				const sourceCode = context.getSourceCode();
+				return removeAbleNodes.map(
```
sindresorhus:
```suggestion
				return removableNodes.map(
```

rules/prefer-negative-index.js
fisker: I found them when I search `slice`, but I guess nobody use `Blob.prototype.slice.call(...)`

sindresorhus: Yeah, let's drop those two.

fisker: should I remove them? or add a comment and keep them

sindresorhus: Hmm, we can keep them commented out with a comment on why they're commented out, so we don't try to reintroduce them in the future.

rules/prefer-negative-index.js
```diff
@@ -0,0 +1,303 @@
...
+		return lengthMemberExpression;
+	}
+};
+
+const getRemovalRange = (node, sourceCode) => {
+	let before = sourceCode.getTokenBefore(node);
+	let after = sourceCode.getTokenAfter(node);
+
+	let start = node.range[0];
+	let end = node.range[1];
```
sindresorhus:
```suggestion
	let [start, end] = node.range;
```

Same with the below.

fisker: I forgot why I seperate them... maybe mistake

rules/prefer-negative-index.js
fisker: suggestion how to express?

rules/prefer-negative-index.js
fisker: assign can't do this, different ranges

rules/prefer-negative-index.js
```diff
@@ -22,6 +22,7 @@ const methods = new Map([
 				'Float64Array',
 				'BigInt64Array',
 				'BigUint64Array'
+				// {Blob,File}#slice is not used generally
```
sindresorhus: > `{Blob,File}#slice()` are not generally used

rules/prefer-negative-index.js
sindresorhus:
```js
let [start, end] = [1, 2];
console.log(start, end);
//=> 1  2
```

fisker: I know this, already did

rules/prefer-negative-index.js
```diff
@@ -0,0 +1,303 @@
...
+
+	while (hasParentheses) {
+		hasParentheses =
+			(before.type === 'Punctuator' && before.value === '(') &&
+			(after.type === 'Punctuator' && after.value === ')');
+		if (hasParentheses) {
+			before = sourceCode.getTokenBefore(before);
+			after = sourceCode.getTokenAfter(after);
+			start = before.range[1];
+			end = after.range[0];
```
fisker: i mean those two line cant combine

sindresorhus: It was not clear that you were referring to this as you commented on the other code.

sindresorhus:
Nice 👌

PR #422, opened 2019-10-18
Change some options for `prevent-abbreviations` rule

fisker:
The PR will make this code bellow report an error by default

```js
```

without breaking https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237#issuecomment-470942156

```js
import tempWrite from 'temp-write';
```

and

```js
import tempWrite from './node_modules/temp-write';
```

moduleId starts with `.`  `/` ~`@/`~ and without `node_modules` is considered `internal-module`.

The `internal` naming and `internal-module` logic may need discuss.

//cc @futpib @sindresorhus

### DIFF
--- a/docs/rules/prevent-abbreviations.md
+++ b/docs/rules/prevent-abbreviations.md
@@ -203,7 +203,7 @@ Default: `'internal'`
 By default, the following code will be reported:
 
 ```js
-import {prop} from './ramda';
+import {prop} from './foo';
 ```
 
 With this set to `true`, the following code will be reported:
--- a/docs/rules/prevent-abbreviations.md
+++ b/docs/rules/prevent-abbreviations.md
@@ -156,14 +156,14 @@ Pass `"extendDefaultWhitelist": false` to override the default `whitelist` compl
 
 ### checkDefaultAndNamespaceImports
 
-Type: `"internal"` | `boolean`<br>
-Default: `"internal"`
+Type: `'internal' | boolean`<br>
+Default: `'internal'`
 
-- `"internal"` check variables declared in default or namespace import, **but only internal module**.
-- `true` check variables declared in default or namespace import.
-- `false` disable check variables declared in default or namespace import.
+- `'internal'` - Check variables declared in default or namespace import, **but only for internal modules**.
+- `true` - Check variables declared in default or namespace import.
+- `false` - Don't check variables declared in default or namespace import.
 
-The following code will be reported.
+By default, the following code will be reported:
 
 ```js
 import * as err from './err';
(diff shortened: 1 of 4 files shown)


futpib:
What is the meaning of a module name starting with `@/`?

fisker:
@ is webpack alias, means project root

fisker:
reminds me there is also a `~`, or  maybe we do  external check, any thoughts?

futpib:
Webpack aliases can be absolutely arbitrary and using `@` as an alias for the root or the project is not something I've seen used. I don't think we should support something that is not even a widespread convention (and definitely not a standard).

Anyway, what is the issue you are trying to solve with this PR?

fisker:
I want imports report error name is using abbreviation,  but https://github.com/sindresorhus/eslint-plugin-unicorn/pull/237#issuecomment-470942156 don't want check npm packages import, so I made this PR to only target `internal-module`s resolve things like example I wrote in the comment above , and this https://github.com/sindresorhus/eslint-plugin-unicorn/blob/dbf4b7996a1aba6deef51916a5435092c130b34e/test/get-documentation-url.js#L2 It's not reported because it's `import`  .

fisker:
@futpib @sindresorhus Are you ok with this change?

sindresorhus:
> Webpack aliases can be absolutely arbitrary and @ as an alias for the root of the project is not something I've seen used. I don't think we should support something that is not even a widespread convention (and definitely not a standard).

👍

docs/rules/prevent-abbreviations.md
```diff
@@ -156,12 +156,28 @@ Pass `"extendDefaultWhitelist": false` to override the default `whitelist` compl
...
+```js
+import * as err from './err';
+```
+
+```js
+import err from '/err';
+```
+
+```js
+const err = require('@/err');
```
sindresorhus: I don't think we should support `@/`.

docs/rules/prevent-abbreviations.md
```diff
@@ -177,12 +193,20 @@ const err = require('err');
...
-Pass `"checkShorthandImports": true` to check variables declared in shorthand import.
+- `'internal'` - Check variables declared in shorthand import, **but only for internal modules**.
+- `true` - Check variables declared in shorthand import.
+- `false` - Don't check variables declared in default shorthand import.
 
-With this set to `true` the following code will be reported.
+By default, the following code will be reported:
+
+```js
+import {prop} from './ramda';
```
sindresorhus:
```suggestion
import {prop} from './foo';
```

It's confusing to use a word that is a known package name.

sindresorhus:
> Are you ok with this change?

Yes

PR #424, opened 2019-10-28
Extend fixers for `prefer-string-slice`

voxpelli:
Currently barely any `substr()` or `substring()` calls can be automatically fixed.

This PR makes pretty much all `substring()` calls fixable and makes many `substr()` fixable.

### DIFF
--- a/rules/prefer-string-slice.js
+++ b/rules/prefer-string-slice.js
@@ -52,7 +52,11 @@ const create = context => {
 				if (firstArgument === '0') {
 					slice = [firstArgument, secondArgument];
 				} else if (argumentNodes[0].type === 'Literal') {
-					slice = [firstArgument, firstArgument + ' + ' + secondArgument];
+					if (argumentNodes[1].type === 'Literal' && typeof argumentNodes[1].value === 'number' && typeof argumentNodes[0].value === 'number') {
+						slice = [argumentNodes[0].value, argumentNodes[0].value + argumentNodes[1].value];
+					} else {
+						slice = [firstArgument, firstArgument + ' + ' + secondArgument];
+					}
 				}
 			}
 
--- a/rules/prefer-string-slice.js
+++ b/rules/prefer-string-slice.js
@@ -13,6 +13,8 @@ const substringCallTemplate = templates.template`${objectVariable}.substring(${a
 const create = context => {
 	const sourceCode = context.getSourceCode();
 
+	const getPossiblyWrappedText = (objectNode) => objectNode.type === 'LogicalExpression' ? `(${sourceCode.getText(objectNode)})` : sourceCode.getText(objectNode);
+
 	return templates.visitor({
 		[substrCallTemplate](node) {
 			const objectNode = substrCallTemplate.context.getMatch(objectVariable);
@@ -27,14 +29,14 @@ const create = context => {
 			const secondArg = argumentNodes[1] ? sourceCode.getText(argumentNodes[1]) : undefined;
 
 			if (argumentNodes.length === 0) {
-				problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + '.slice()');
+				problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + '.slice()');
 			} else if (argumentNodes.length === 1) {
-				problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + `.slice(${firstArg})`);
+				problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + `.slice(${firstArg})`);
 			} else if (argumentNodes.length === 2) {
 				if (firstArg === '0') {
-					problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + `.slice(${firstArg}, ${secondArg})`);
+					problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + `.slice(${firstArg}, ${secondArg})`);
 				} else if (argumentNodes[0].type === 'Literal') {
-					problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + `.slice(${firstArg}, ${firstArg} + ${secondArg})`);
+					problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + `.slice(${firstArg}, ${firstArg} + ${secondArg})`);
 				}
 			}
 
@@ -54,11 +56,11 @@ const create = context => {
 			const secondArg = argumentNodes[1] ? sourceCode.getText(argumentNodes[1]) : undefined;
 
 			if (argumentNodes.length === 0) {
-				problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + '.slice()');
+				problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + '.slice()');
 			} else if (argumentNodes.length === 1) {
-				problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + `.slice(${firstArg})`);
+				problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + `.slice(${firstArg})`);
 			} else if (argumentNodes.length === 2) {
-				problem.fix = fixer => fixer.replaceText(node, sourceCode.getText(objectNode) + `.slice(${firstArg}, ${secondArg})`);
+				problem.fix = fixer => fixer.replaceText(node, getPossiblyWrappedText(objectNode) + `.slice(${firstArg}, ${secondArg})`);
 			}
 
 			context.report(problem);
--- a/rules/prefer-string-slice.js
+++ b/rules/prefer-string-slice.js
@@ -13,7 +13,9 @@ const substringCallTemplate = templates.template`${objectVariable}.substring(${a
 const getNumericValue = node => {
 	if (node.type === 'Literal' && typeof node.value === 'number') {
 		return node.value;
-	} else if (node.type === 'UnaryExpression' && node.operator === '-') {
+	}
+
+	if (node.type === 'UnaryExpression' && node.operator === '-') {
 		return 0 - getNumericValue(node.argument);
 	}
 };
@@ -61,9 +63,9 @@ const create = context => {
 			}
 
 			if (slice) {
-				const objectText = objectNode.type === 'LogicalExpression'
-					? `(${sourceCode.getText(objectNode)})`
-					: sourceCode.getText(objectNode);
+				const objectText = objectNode.type === 'LogicalExpression' ?
+					`(${sourceCode.getText(objectNode)})` :
+					sourceCode.getText(objectNode);
 
 				problem.fix = fixer => fixer.replaceText(node, `${objectText}.slice(${slice.join(', ')})`);
 			}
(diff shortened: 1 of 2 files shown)


voxpelli:
I've realized that I need to do some more changes to this

voxpelli:
Done, now it should be a more correct auto fixer

rules/prefer-string-slice.js
```diff
@@ -23,10 +39,27 @@ const create = context => {
 				message: 'Prefer `String#slice()` over `String#substr()`.'
 			};
 
-			const canFix = argumentNodes.length === 0;
+			const firstArgument = argumentNodes[0] ? sourceCode.getText(argumentNodes[0]) : undefined;
+			const secondArgument = argumentNodes[1] ? sourceCode.getText(argumentNodes[1]) : undefined;
+
+			let slice;
```
fisker: this can be an array and concat with `.slice(` and `)`

fisker:
some `NaN` tests?

rules/prefer-string-slice.js
```diff
@@ -41,10 +74,46 @@ const create = context => {
...
+
+			if (argumentNodes.length === 0) {
+				slice = '.slice()';
+			} else if (argumentNodes.length === 1 && firstNumber !== undefined) {
+				slice = `.slice(${Math.max(0, firstNumber)})`;
+			} else if (argumentNodes.length === 1 && isLengthProperty(argumentNodes[0])) {
+				slice = `.slice(${firstArgument})`;
+			} else if (argumentNodes.length === 1) {
```
fisker: merge this logic in one `argumentNodes.length === 1` block should be more readable

voxpelli:
Fixed most of your feedback @fisker, thanks!

> some `NaN` tests?

Could you give me an example of what kind of `NaN` tests you're thinking of?

voxpelli:
> If indexStart is greater than indexEnd, then the effect of substring() is as if the two arguments were swapped

This handled wherever it can be handled and in cases where it can't be handled, no fix is provided.

> If start is a positive number, the index starts counting at the start of the string. Its value is capped at str.length.
> If start is a negative number, the index starts counting from the end of the string. Its value is capped at -str.length.
> If length is undefined, substr() extracts characters to the end of the string.

This is very similar to `.slice()` except that `.slice()` has an end index instead of a length. Adding the startvalue, whenever possible, to the length solves that and is something I've done. When that isn't possible no fix is provided.

fisker:
> Could you give me an example of what kind of `NaN` tests you're thinking of?

sorry, I thought too much, not really needed

fisker:
@voxpelli I just send a PR to your fork, I did't change any your logic, should be more readable https://github.com/voxpelli/eslint-plugin-unicorn/pull/1

voxpelli:
@fisker I actually prefer my version, but I wouldn't want it to block a merger of this, so whoever merges it can pick :)

voxpelli:
@fisker How do we best proceed?

rules/prefer-string-slice.js
```diff
@@ -10,6 +10,24 @@ const argumentsVariable = templates.spreadVariable();
 const substrCallTemplate = templates.template`${objectVariable}.substr(${argumentsVariable})`;
 const substringCallTemplate = templates.template`${objectVariable}.substring(${argumentsVariable})`;
 
+const getNumericValue = node => {
+	if (node.type === 'Literal' && typeof node.value === 'number') {
+		return node.value;
+	}
+
+	if (node.type === 'UnaryExpression' && node.operator === '-') {
+		return 0 - getNumericValue(node.argument);
```
fisker: this  0 should  remove

rules/prefer-string-slice.js
```diff
@@ -10,6 +10,24 @@ const argumentsVariable = templates.spreadVariable();
...
+	}
+};
+
+const isLengthProperty = node => (
```
fisker: I don't think we  want this case, just  number  and  other

rules/prefer-string-slice.js
fisker: rename to `sliceArguments`

rules/prefer-string-slice.js
```diff
@@ -23,10 +41,33 @@ const create = context => {
...
+			const secondArgument = argumentNodes[1] ? sourceCode.getText(argumentNodes[1]) : undefined;
+
+			let slice;
+
+			if (argumentNodes.length === 0) {
+				slice = [];
+			} else if (argumentNodes.length === 1) {
+				slice = [firstArgument];
+			} else if (argumentNodes.length === 2) {
+				if (firstArgument === '0') {
```
fisker: should be literal zero, we  don't want handle `"0"`, utils should have a function d this

rules/prefer-string-slice.js
```diff
@@ -23,10 +41,33 @@ const create = context => {
...
+					slice = [firstArgument, secondArgument];
+				} else if (argumentNodes[0].type === 'Literal') {
+					if (argumentNodes[1].type === 'Literal' && typeof argumentNodes[1].value === 'number' && typeof argumentNodes[0].value === 'number') {
+						slice = [argumentNodes[0].value, argumentNodes[0].value + argumentNodes[1].value];
+					} else {
+						slice = [firstArgument, firstArgument + ' + ' + secondArgument];
```
fisker: I  think this should be  literal and number, image this might be `"1" + "2"`

rules/prefer-string-slice.js
```diff
@@ -41,10 +82,48 @@ const create = context => {
...
+			const firstArgument = argumentNodes[0] ? sourceCode.getText(argumentNodes[0]) : undefined;
+			const secondArgument = argumentNodes[1] ? sourceCode.getText(argumentNodes[1]) : undefined;
+
+			const firstNumber = argumentNodes[0] ? getNumericValue(argumentNodes[0]) : undefined;
+
+			let slice;
+
+			if (argumentNodes.length === 0) {
+				slice = [];
+			} else if (argumentNodes.length === 1) {
```
fisker: too much in this  branch  just  literal an negative , number, otherwise use  math.max

rules/prefer-string-slice.js
```diff
@@ -23,10 +41,33 @@ const create = context => {
...
+				slice = [firstArgument];
+			} else if (argumentNodes.length === 2) {
+				if (firstArgument === '0') {
+					slice = [firstArgument, secondArgument];
+				} else if (argumentNodes[0].type === 'Literal') {
+					if (argumentNodes[1].type === 'Literal' && typeof argumentNodes[1].value === 'number' && typeof argumentNodes[0].value === 'number') {
```
fisker: this branch can also merge with 0, same thing

fisker: sorry, its not.

rules/prefer-string-slice.js
fisker: also, do we really want recury this?  just number and negative number should enough

sindresorhus: > this 0 should remove

It's usually useful to explain *why* if it's not immediately obvious.

rules/prefer-string-slice.js
voxpelli: The text representation of an argument node can only ever be `=== '0'` if it's a literal zero, no?

rules/prefer-string-slice.js
voxpelli: Is it the `isLengthProperty()` that you are critical against @fisker?

I ran these modifications on a large internal project at my company and I ran into `.length` properties being used often enough for me to think it merits a plausible conversion.

rules/prefer-string-slice.js
voxpelli: Why should the `0` be removed?

rules/prefer-string-slice.js
voxpelli: I found this useful when applying this on a large project of mine. Can you explain why you don't want it? Because you think it will likely result in invalid fixes? Because you think the use case is too specific to merit the complexity of it? Something else?

rules/prefer-string-slice.js
fisker: sorry, I was not explaining clear, this is negative number, right? 

`0 - 1` is the same as `-1`

am I missing anything?

rules/prefer-string-slice.js
fisker: yes, it's useful in some case,  because it's
 always  >=0, so no need compare with 0, but in js word maybe too many cases to handle , like `Set#size()` , `Math.abs` also works, we don't want to handle all of them. so why .length special?

I wounder what  @sindresorhus think

rules/prefer-string-slice.js
fisker: let's wait for @sindresorhus

rules/prefer-string-slice.js
fisker: sorry, my comment is based on `const firstArgument = argumentNodes[0]`(I changed it in my refactor, but forgot it's not changed here), why we  need `sourceCode.getText`? Isn't `argumentNodes[0]` easier to use?

rules/prefer-string-slice.js
voxpelli: It's a positive number `Literal` wrapped in a `UnaryExpression` with a `-` operator, so the `0 -` part is to convert the positive number into its correct negative value. I do think I have added tests for that?

rules/prefer-string-slice.js
voxpelli: `.length` is what's used on strings and these methods are oriented on strings, so not that unlikely that they reference the length of a string. Eg. writing `.slice('foo'.length)` can be more easy to understand than writing `.slice(3)` and similarly eg:

```javascript
if (bar.startsWith(SOME_PREFIX)) {
  bar = bar.slice(SOME_PREFIX.length);
}
```

rules/prefer-string-slice.js
fisker: I know that `-1` is parsed as `UnaryExpression`and `operator:-` and `argument:1`

so `return 0 - node.argument` equals to `return - node.argument`.

I am get confusing, am I wrong?

rules/prefer-string-slice.js
voxpelli: As almost every branch used the value of `firstArgument` I decided to calculate them in  shared space to make the code more easy to read and more DRY.

As `firstArgument` then was already calculated I considered `firstArgument === '0'` to be preferable over `isLiteralNumber(argumentNodes[0]) && argumentNodes[0].value === 0` as it's shorter and simpler.

fisker: thanks for explaining this, this make sense now, but I think it makes the logic little hard to understand. you want use those functions? 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/9e0db0b1385c72695d157f2d00bbc6a8ffdbbfc6/rules/prefer-includes.js#L3

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/9e0db0b1385c72695d157f2d00bbc6a8ffdbbfc6/rules/prefer-includes.js#L10

rules/prefer-string-slice.js
fisker: could you rename this variable? more specific

rules/prefer-string-slice.js
sindresorhus: I think it covers a common use-case and that we should keep this.

rules/prefer-string-slice.js
sindresorhus: I'm fine with having special logic for `.length`.

sindresorhus:
Thanks for working on this, @voxpelli 👌

PR #431, opened 2019-11-08
Add `ignore` option to the `filename-case` rule

### LINKED ISSUE #395: allow using `lower` suffix for `pascalCase` in `filename-case` rule
For example:
```
FooBar.test.js
```
Why?
1. Some tools use/recommend `lower` notation for part after `.`, for example `jest` use lower `test` (`FooBar.test.js`) in examples for class testing.
2. Sometimes you have vendor libraries (not in node_modules) and they use notation `Timer.min.js` (rare use case) 
3. Code for multiple platform, for example: `MyClass.web.js`, `MyClass.android.js` (react/react-native/other same tools)

Or maybe option to configure this

evilebottnawi:
fixes #395

### DIFF
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -35,6 +35,8 @@ Characters in the filename except `a-z`, `A-Z`, `0-9`, `-`, `_` and `$` are igno
 
 ### case
 
+Type: `string`
+
 You can set the `case` option like this:
 
 ```js
@@ -47,6 +49,8 @@ You can set the `case` option like this:
 ```
 
 ### cases
+
+Type: `{[type: string]: string}`
 
 You can set the `cases` option to allow multiple cases:
 
@@ -64,14 +68,30 @@ You can set the `cases` option to allow multiple cases:
 
 ### ignore
 
-An array of RegExp to ignore. It’s set to [] by default. If provided, it must be an Array.
+Type: `Array<string | RegExp>`\
+Default: `[]`
+
+Filenames to ignore.
+
+Sometimes you may have in non-standard file names a project, so using this option allow to ignore them.
+
+For example:
+
+- vendor packages which are not published and was copy-pasted
+- ignore some of files when you use [eslint-plugin-markdown](https://github.com/eslint/eslint-plugin-markdown), for example `README.md`
+- some tools may require special names for some files
 
 ```js
 "unicorn/filename-case": [
 	"error",
 	{
 		"case": "kebabCase",
-		"ignore": ["FOOBAR\\.js", '^(B|b)az', '\\.SOMETHING\\.js']
+		"ignore": [
+			"FOOBAR\\.js",
+			"^(B|b)az",
+			"\\.SOMETHING\\.js",
+			/^vendor/i
+		]
 	}
 ]
 ```
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -88,9 +88,9 @@ For example:
 	{
 		"case": "kebabCase",
 		"ignore": [
-			"FOOBAR\\.js",
+			"^FOOBAR\\.js$",
 			"^(B|b)az",
-			"\\.SOMETHING\\.js",
+			"\\.SOMETHING\\.js$",
 			/^vendor/i
 		]
 	}
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -73,6 +73,8 @@ Default: `[]`
 
 Filenames to ignore.
 
+When string is given, it's interpreted as a regular expressions inside a string, don't forget what you must escape special characters (For example `.`, `[`,`]`, `(`, `)` and etc) by putting two backslashes in front of them (to ignore `[id].js` file, you should use `"\\[id\\]\\.js"`).
+
 Sometimes you may have non-standard filenames in a project. This option lets you ignore those files.
 
 For example:
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -33,6 +33,8 @@ Characters in the filename except `a-z`, `A-Z`, `0-9`, `-`, `_` and `$` are igno
 
 ## Options
 
+### case
+
 You can set the `case` option like this:
 
 ```js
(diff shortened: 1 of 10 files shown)


fisker:
I think you we need regexp support, 

for examples `MyClass.web.js`, `MyClass.android.js` from #395 we can use `/*\.(web|android)\.js/`, is regexp allowed in eslint config?

docs/rules/filename-case.md
```diff
@@ -57,3 +61,17 @@ Or set the `cases` option to allow multiple cases:
...
+### ignore
+
+An array of files to ignore. It’s set to [] by default. If provided, it must be an Array.
+
+```js
+"unicorn/filename-case": [
+	"error",
+	{
+		"case": "kebabCase",
+		"ignore": ["CHANGELOG.md", 'README.md']
```
fisker: should use js file for example, maybe just `foo.js` `bar.js`

rules/filename-case.js
```diff
@@ -56,19 +56,17 @@ const cases = {
...
-@param {unknown} context
+@param {object} config
 @returns {string[]} The chosen cases.
 */
-function getChosenCases(context) {
-	const option = context.options[0] || {};
-
-	if (option.case) {
-		return [option.case];
+function getChosenCases(config) {
```
fisker: should name `options` 

elint calls `options`

https://eslint.org/docs/rules/yoda#options

and other rules use it as options

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/cdbb9ffa57d7c927e07c0a44c6aab8e243a626fd/rules/catch-error-name.js#L34

evilebottnawi: In other rules it is `config` https://github.com/eslint/eslint/blob/master/lib/rules/no-magic-numbers.js#L56

docs/rules/filename-case.md
evilebottnawi: let's do it

rules/filename-case.js
evilebottnawi: i think it is something strange in eslint code base, let's rename to options :+1:

evilebottnawi:
@fisker done

evilebottnawi:
strange problem with integration tests, what tips will be?

fisker:
something in `got` codebase can't parse

fisker:
@sindresorhus I'm updating dependencies

test/filename-case.js
```diff
@@ -92,7 +92,89 @@ ruleTester.run('filename-case', rule, {
 		testManyCases('src/foo/___foo_bar.js', {snakeCase: true, pascalCase: true}),
 		testCaseWithOptions('src/foo/bar.js'),
 		testCase('src/foo/[fooBar].js', 'camelCase'),
-		testCase('src/foo/{foo_bar}.js', 'snakeCase')
+		testCase('src/foo/{foo_bar}.js', 'snakeCase'),
+		testCaseWithOptions(undefined, undefined, [
+			{case: 'kebabCase', ignore: ['FOOBAR\\.js']}
```
sindresorhus: It should allow specifying regex literals. Most people use JS configs, so it's nicer to also be able to specify literals.

docs/rules/filename-case.md
```diff
@@ -57,3 +65,24 @@ Or set the `cases` option to allow multiple cases:
...
 ]
 ```
+
+### ignore
+
+Type: `RegExp[]`\
+Default: `[]`
+
+Filenames to ignore.
+
```
sindresorhus: Could be useful to mention some real-world use-cases for this option, like done in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/395

test/filename-case.js
evilebottnawi: @sindresorhus just allow use `/foobar\.js/` as regexp, right? When i use string notation `foobar\\.js` it is also regexp?

sindresorhus: Allow both both regex literal and reges in a string.

evilebottnawi:
@sindresorhus @fisker done

rules/filename-case.js
```diff
@@ -144,7 +142,15 @@ function englishishJoinWords(words) {
...
 const create = context => {
-	const chosenCases = getChosenCases(context);
+	const options = context.options[0] || {};
+	const chosenCases = getChosenCases(options);
+	const ignore = (options.ignore || []).map(item => {
+		if (item instanceof RegExp) {
+			return item;
+		}
+
+		return new RegExp(item, 'u');
```
fisker: Shouldn't we escape the string? otherwise

```js
new RegExp('[fooBar].js').test('f.js')
// -> true
```

test/filename-case.js
fisker: is `regex in a string` a good idea? maybe confusing people

docs/rules/filename-case.md
```diff
@@ -68,11 +68,19 @@ You can set the `cases` option to allow multiple cases:
 
 ### ignore
 
-Type: `RegExp[]`\
+Type: `Array<string | RegExp>`\
 Default: `[]`
 
 Filenames to ignore.
```
sindresorhus: Can you explicitly say that if a string is given, it's interpreted as a regex inside a string.

rules/filename-case.js
sindresorhus: Huh? The string support is meant to allow regex in a string, for use with JSON config, where you cannot use regex literal.

fisker: right, forgot

fisker:
so to ignore `[id].js` . I have to pass `^\[id\]\.js$` right?

sindresorhus:
Correct. It might be worth adding a tip about that to the docs, as a lot of user miss how to correctly escape regexes.

evilebottnawi:
@sindresorhus done, English is not my native so I can make a mistake somewhere :smile:

docs/rules/filename-case.md
```diff
@@ -73,6 +73,8 @@ Default: `[]`
...
 
+When string is given, it's interpreted as a regular expressions inside a string, don't forget what you must escape special characters (For example `.`, `[`,`]`, `(`, `)` and etc) by putting two backslashes in front of them (to ignore `[id].js` file, you should use `"\\[id\\]\\.js"`).
```
fisker: no `^` and `$` ?

evilebottnawi: @fisker done

PR #435, opened 2019-11-12
Support pre-release versions in `expiring-todo-comments` rule

### LINKED ISSUE #432: Allow pre-release todo comments
As a developer, I want to be able to use todo comments with major release expiration while developing a pre-release version so I can progressively fix those todos without CI errors.

I am currently working on a 1.0.0-beta.1 version, and I want to deprecate a method, indicating that it will be removed in the official release (1.0.0). I want to have a `TODO` reminder that the method should be removed on `1.0.0`. However, the `unicorn/expiring-todo-comments` uses `semver.coerce`, which converts the `1.0.0-beta.1` to `1.0.0` when testing the `[>=1.0.0]` rule, which results in a linting error.

It would be a great improvement to expose a configuration to support pre-release todos.
Here would be, in my own opinion, some valid solutions for the plugin configuration.

```
"unicorn/expiring-todo-comments": [
        "error",
        {
                "allowPreRelease": true
        }
]

// or

"unicorn/expiring-todo-comments": [
        "error",
        {
                "allowPreRelease": [ // will use coercion for "rc" releases
                        "alpha"
                        "beta"
                ]
        }
]
```

## Pass
```
// package.json
{ "version": "1.0.0-beta.1" }

// my-file.js
// TODO [>=1.0.0] Something must be removed
```

## Fail
```
// package.json
{ "version": "1.0.0-beta.1" }

// my-file.js
// TODO [>=1.0.0-alpha.1]
```

lubien:
Fixes #432

Refactor tryToCoerceVersion

Don't coerce valid semver tags on expiring-todo-comments.

I'm not sure **where** to test this. Should I export (`exports.tryToCoerceVersion`) the function and test along with the rule itself (test/expiring-todo-comments.js)? @sindresorhus any idea?

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
 		"semver": "^6.3.0"
 	},
 	"devDependencies": {
+		"@lubien/fixture-beta-package": "^1.0.0-beta.1",
 		"ava": "^2.4.0",
 		"babel-eslint": "^10.0.3",
 		"chalk": "^2.4.2",
--- a/rules/expiring-todo-comments.js
+++ b/rules/expiring-todo-comments.js
@@ -27,8 +27,8 @@ const packageDependencies = {
 };
 
 const DEPENDENCY_INCLUSION_RE = /^[+|-]\s*@?[\S+]\/?\S+/;
-const VERSION_COMPARISON_RE = /^(@?[\S+]\/?\S+)@(>|>=)([\d]+(\.\d+){0,2})/;
-const PKG_VERSION_RE = /^(>|>=)([\d]+(\.\d+){0,2})\s*$/;
+const VERSION_COMPARISON_RE = /^(@?[\S+]\/?\S+)@(>|>=)([\d]+(\.\d+){0,2}(-[0-9a-z-]+(\.[0-9a-z-]+)*)?(\+[0-9a-z-]+(\.[0-9a-z-]+)*)?)/i;
+const PKG_VERSION_RE = /^(>|>=)([\d]+(\.\d+){0,2}(-[0-9a-z-]+(\.[0-9a-z-]+)*)?(\+[0-9a-z-]+(\.[0-9a-z-]+)*)?)\s*$/;
 const ISO8601_DATE = /(\d{4})-(\d{2})-(\d{2})/;
 
 function parseTodoWithArguments(string, {terms}) {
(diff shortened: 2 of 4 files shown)


sindresorhus:
Can you update the docs to explicitly show that prerelease versions are supported?

sindresorhus:
I would use a fixture instead. Integration tests are usually better. Like the one in the issue:

```
// package.json
{ "version": "1.0.0-beta.1" }

// my-file.js
// TODO [>=1.0.0] Something must be removed
```

lubien:
Sure I can do it later today.

Just be clear I'd have to create a fixture repo and put on test/integration/test.js like the others right?

sindresorhus:
I was thinking: Publish an empty package with a prerelease version, add it as a dev dependency here, and just add another test here: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1bc47a0715584a2948665325e08aecc5721bb6a4/test/expiring-todo-comments.js#L222-L225

package.json
```diff
@@ -50,6 +50,7 @@
 		"semver": "^6.3.0"
 	},
 	"devDependencies": {
+		"@lubien/fixture-beta-package": "^1.0.0-beta.1",
```
fisker: why dont we  use a `file:` package , easier to maintain

rules/expiring-todo-comments.js
```diff
@@ -158,9 +158,37 @@ function reachedDate(past) {
...
+	if (semver.valid(version)) {
+		return version;
+	}
+
 	try {
-		return semver.coerce(version);
+		// Try to semver.parse a perfect match while semver.coerce tries to fix errors
+		// But coerce can't parse pre-releases.
+		return semver.parse(version) || semver.coerce(version);
 	} catch (_) {
```
fisker:
```suggestion
	} catch {
```

fisker: can we can use this?

lubien: > can we can use this?

<details>
<summary>As engine says we support node 8, not yet</summary>

```
~/dev/eslint-plugin-unicorn:compare-pre-release-todo-conditions lubien@joestar λ nvm install 8
Downloading and installing node v8.16.2...
Downloading https://nodejs.org/dist/v8.16.2/node-v8.16.2-linux-x64.tar.xz...
############################################################################# 100,0%
Computing checksum with sha256sum
Checksums matched!
Now using node v8.16.2 (npm v6.4.1)
~/dev/eslint-plugin-unicorn:compare-pre-release-todo-conditions lubien@joestar λ npx ava test/expiring-todo-comments.js 


  ✖ No tests found in test/expiring-todo-comments.js

  1 uncaught exception

  Uncaught exception in test/expiring-todo-comments.js


  /home/lubien/dev/eslint-plugin-unicorn/rules/expiring-todo-comments.js:192
        } catch {
                ^

  SyntaxError: Unexpected token {
```

</details>

package.json
fisker: Is it possible to  mock one  during test?

lubien: Personally I haven't heard anything of sort. Tho for me this is pretty much all I need since I could handle all known cases so far.

rules/expiring-todo-comments.js
fisker: okay

PR #436, opened 2019-11-13
import-index: Fix for `require()`

garyking:
When using ESLint to lint code while typing, if you type `require()` with the `import-index` rule enabled, then it will throw an error.

This fixes that.


fisker:
can you check `prevent-abbreviations`?

garyking:
I haven't noticed any issues with `prevent-abbreviations` and `require()`, so it doesn't seem like any changes need to be made to those rule, or its tests.

fisker:
thanks

PR #440, opened 2019-11-19
Add `prefer-trim-start-end` rule

### LINKED ISSUE #439: Rule proposal: `prefer-trim-start-end` evaluating
Prefer `trimStart` over `trimLeft` (alias) and `trimEnd` over `trimRight` (alias)

Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimstart
Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trimend

Maybe better name for rule

fisker:
Fixes #439


sindresorhus:
Looks great 👌

PR #442, opened 2019-11-26
Fix `custom-error-definition` to support constructors without a body

pat841:
Using this example:

```typescript
export class CustomError extends Error {
  type: string;
  text: string;
  reply?: any;
  constructor(type: string, text: string, reply?: any);
}
```

The error `TypeError: Cannot read property 'body' of null` gets thrown for the line `const constructorBody = constructorBodyNode.body;`.

We can fix this by returning early if no constructor body is available.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
 	},
 	"devDependencies": {
 		"@lubien/fixture-beta-package": "^1.0.0-beta.1",
+		"@typescript-eslint/parser": "^2.9.0",
 		"ava": "^2.4.0",
 		"babel-eslint": "^10.0.3",
 		"chalk": "^2.4.2",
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you add a test?

mf-pherlihy:
> Can you add a test?

I tried but unfortunately the tests dont support the Typescript syntax.

The primary issue is `constructor(type: string, text: string, reply?: any);` not having a constructor body. This is supported by Typescript but throws `A fatal parsing error occurred: Parsing error: Unexpected token ;` in a non-TS environment.

futpib:
@mf-pherlihy @sindresorhus I took the liberty of adding a typescript test, check it out.

Issue #443, opened 2019-11-27
`prevent-abbreviations` fixer bug on deconstructing


fisker:
just run `prevent-abbreviations` against a large code base, found two fixer bug

input:

```js
const foo = ({num}) => num;
const bar = num => ({num});
```

fix:

```js
const foo = ({number}) => number;
const bar = number => ({number});
```

expect:

```js
const foo = ({num: number}) => number;
const bar = number => ({num: number});
```

fisker:
just found #330 is part of this

fisker:
only report when using `babel-eslint` parser

fisker:
another bug, only on `babel-eslint` parser

```js
function unicorn(unicorn) {
  const {prop = {}} = unicorn;
  return prop;
}
```

fix to 

```js
function unicorn(unicorn) {
  const {property = {}} = unicorn;
  return property;
}
```

should be

```js
function unicorn(unicorn) {
  const {prop: property = {}} = unicorn;
  return property;
}
```

PR #444, opened 2019-11-27
Fix `prevent-abbreviations` fixer bug

### LINKED ISSUE #330: `prevent-abbreviations` should not try to fix variable names when they're being destructured
First off, I think this is a great library with a lot of excellent rules and really appreciate all the work on it.

I was working on implementing some of the rules for my team and I ran an autofix and saw this change in my git diff:

image

In this case, it would definitely introduce a bug since there 's no 'development' variable being passed.

While I think this autofix is generally pretty safe, i think it might be good to check if the current scope includes destructuring and avoid autofixing.

I can look into working on a PR for this myself (i've been interested in learning how to work on eslint rules), just wanted to point this out and see if i can help out.

fisker:
fixes: #443 and fixes #330

I've located the problem, in `isShorthandPropertyIdentifier` check function, In `babel-eslint` parent.key is not reference of identifier, add a extra check to make it pass

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -364,26 +364,25 @@ const shouldFix = variable => {
 	return !variableIdentifiers(variable).some(isExportedIdentifier);
 };
 
+const isIdentifierKeyOfNode = (identifier, node) =>
+	node.key === identifier ||
+	// In `babel-eslint` parent.key is not reference of identifier
+	(
+		node.key.type === identifier.type &&
+		node.key.name === identifier.name
+	);
+
 const isShorthandPropertyIdentifier = identifier => {
 	return identifier.parent.type === 'Property' &&
 		identifier.parent.shorthand &&
-		(
-			identifier.parent.key === identifier ||
-			// In `babel-eslint` parent.key is not reference of identifier
-			(
-				identifier.parent.key.type === identifier.type &&
-				identifier.parent.key.name === identifier.name &&
-				identifier.parent.key.start === identifier.start &&
-				identifier.parent.key.end === identifier.end
-			)
-		);
+		isIdentifierKeyOfNode(identifier, identifier.parent);
 };
 
 const isAssignmentPatternShorthandPropertyIdentifier = identifier => {
 	return identifier.parent.type === 'AssignmentPattern' &&
 		identifier.parent.left === identifier &&
 		identifier.parent.parent.type === 'Property' &&
-		identifier.parent.parent.key === identifier &&
+		isIdentifierKeyOfNode(identifier, identifier.parent.parent) &&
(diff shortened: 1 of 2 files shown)


rules/prevent-abbreviations.js
```diff
@@ -366,8 +366,17 @@ const shouldFix = variable => {
...
+		identifier.parent.shorthand &&
+		(
+			identifier.parent.key === identifier ||
+			// In `babel-eslint` parent.key is not reference of identifier
+			(
+				identifier.parent.key.type === identifier.type &&
+				identifier.parent.key.name === identifier.name &&
+				identifier.parent.key.start === identifier.start &&
+				identifier.parent.key.end === identifier.end
+			)
```
fisker: I don't have much experience on those `AST`s, I'm not sure if this is good

fisker:
also there is a few other functions in this file compares `parent.key === identifier`, I'm not sure those logic requires modification

rules/prevent-abbreviations.js
```diff
@@ -366,8 +366,17 @@ const shouldFix = variable => {
 
 const isShorthandPropertyIdentifier = identifier => {
 	return identifier.parent.type === 'Property' &&
-		identifier.parent.key === identifier &&
-		identifier.parent.shorthand;
+		identifier.parent.shorthand &&
+		(
+			identifier.parent.key === identifier ||
+			// In `babel-eslint` parent.key is not reference of identifier
```
sindresorhus: Can you open an issue about this on `babel-eslint`? The point of that package is to normalize the AST.

rules/prevent-abbreviations.js
```diff
@@ -366,8 +366,17 @@ const shouldFix = variable => {
...
+			(
+				identifier.parent.key.type === identifier.type &&
+				identifier.parent.key.name === identifier.name &&
+				identifier.parent.key.start === identifier.start &&
```
sindresorhus: You should not rely on `.start` and `.end`: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/318

rules/prevent-abbreviations.js
fisker: you are right, we might not do this check, I've found other cases also broken, let me have a look at their repo first

sindresorhus: I’m ok with working around it, but we should always also try to get it fixed upstream.

sindresorhus: Add a code comment with a link to the Babel-eslint issue.

fisker:
i just confirmed `(Property)Node.key.parent ==== Node`, so it's maybe not `babel-eslint` bug, let me dig more deeper

Issue #446, opened 2019-11-29
(v14): false-positive "Invalid regular expression: Lone quantifier brackets"


merlinnot:
Hi! I noticed a bug in the latest release with some regular expressions, as this one:

```typescript
/\{\{verificationUrl\}\}/gu
```

It makes ESLint crash (not report an error, actually crash).

PR #451, opened 2019-11-29
Update `regex-shorthand` to ignore regex with `u` flag

### LINKED ISSUE #447: regex-shorthand suggesting invalid regular expression
I have this error
```
error  /(\s|\.|@|_|-)/u can be optimized to /([\s-.@_])/u  unicorn/regex-shorthand
```

But the regex `/([\s-.@_])/u` is invalid, as it is trying to say a range between `\s` and `.` (`\s-.`).

If I manually change it to `/[\s.@_-]/u`, which is correct, I get:
```
error  /[\s.@_-]/u can be optimized to /[\s-.@_]/u  unicorn/regex-shorthand
```

My change is correct, as `-` at the end of a match group means literally `-`. But the optimization suggestion is wrong because it means a range.

I had had these problems with eslint-plugin-unicorn version 14.0.0

### LINKED ISSUE #448: (v14): false-positive "Invalid regular expression: Invalid character class"
Hi! I noticed a bug in the latest release with some regular expressions, as this one:

```typescript
/^test-(?<name>[a-zA-Z-\d]+)$/u
```

It makes ESLint crash (not report an error, actually crash).

fisker:
`regexp-tree` might optimize RegExp to invalid codes, crash ESLint

fixes: #446
fixes #447 
fixes #448

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "eslint-plugin-unicorn",
-	"version": "13.0.0",
+	"version": "14.0.0",
 	"description": "Various awesome ESLint rules",
 	"license": "MIT",
 	"repository": "sindresorhus/eslint-plugin-unicorn",
--- a/rules/prefer-string-slice.js
+++ b/rules/prefer-string-slice.js
@@ -10,6 +10,29 @@ const argumentsVariable = templates.spreadVariable();
 const substrCallTemplate = templates.template`${objectVariable}.substr(${argumentsVariable})`;
 const substringCallTemplate = templates.template`${objectVariable}.substring(${argumentsVariable})`;
 
+const isLiteralNumber = node => node && node.type === 'Literal' && typeof node.value === 'number';
+
+const getNumericValue = node => {
+	if (isLiteralNumber(node)) {
+		return node.value;
+	}
+
+	if (node.type === 'UnaryExpression' && node.operator === '-') {
+		return -getNumericValue(node.argument);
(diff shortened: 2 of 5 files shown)


sindresorhus:
This is a good temporary workaround, but we should find the actual cause for this and get it fixed.

> regexp-tree might optimize RegExp to invalid codes

Can you open an issue on `regexp-tree` about that?

fisker:
turns out this also fixes #448, adding more test

fisker:
@sindresorhus these might related, not confirmed yet

https://github.com/DmitrySoshnikov/regexp-tree/issues/162
https://github.com/BrainMaestro/eslint-plugin-optimize-regex/issues/23

sindresorhus:
Does this also fix https://github.com/sindresorhus/eslint-plugin-unicorn/issues/447 ? Can you add a test for it?

fisker:
> Does this also fix #447 ? Can you add a test for it?

No, this doesn't, different case. If I find anything will do another PR.

fisker:
one second

fisker:
Yes, this fixes #447, I think he didn't use `--fix`, so only suggest the wrong code, but not crash ESLint

fisker:
Add tests, we should do a path release

fisker:
@sindresorhus this `.toRegExp()` might not safe, maybe we should skip the `u` flag?

fisker:
close in favor of #454

PR #452, opened 2019-11-30
Ignore `regexp-tree` parse error on `regex-shorthand` rule

fisker:
If a regex can be parsed by ESLint, but can't parse by `regexp-tree`, we don't know if this regex can be optimized, we should just ignore it


fisker:
close in favor of #454

PR #454, opened 2019-11-30
Update `regex-shorthand`

### LINKED ISSUE #446: (v14): false-positive "Invalid regular expression: Lone quantifier brackets"

### LINKED ISSUE #447: regex-shorthand suggesting invalid regular expression

### LINKED ISSUE #448: (v14): false-positive "Invalid regular expression: Invalid character class"

fisker:
what I've done:

1. simplify logic
2. sort flags, `regexp-tree` do this by default, but missed in previous version
3. ignore regex `regexp-tree` can't parse, can't find a case to test
4. ignore regex with `u` flag that `regexp-tree` can't handle well, avoid crashing ESLint and suggestting wrong code

related PRs:
#452 was tring to ignore parse error
#451 after simplify logic, we caught a crash issue when running integration test, so move the fix into this one

fixes #446
fixes #447
fixes #448

### DIFF
--- a/rules/regex-shorthand.js
+++ b/rules/regex-shorthand.js
@@ -1,6 +1,6 @@
 'use strict';
 const cleanRegexp = require('clean-regexp');
-const {generate, optimize, parse} = require('regexp-tree');
+const {optimize} = require('regexp-tree');
 const getDocumentationUrl = require('./utils/get-documentation-url');
 const quoteString = require('./utils/quote-string');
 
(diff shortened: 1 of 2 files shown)


sindresorhus:
Thanks for fixing all the issues 👌

jdanil:
Thanks @fisker 🙏

PR #468, opened 2019-12-03
Allow disabling of regexp character sorting

### LINKED ISSUE #453: (v14): Make character class sorting in `regex-shorthand` optional or at least documented
Hi,

With the latest 14.0.0, I'm getting the following false positive for `regex-shorthand` (besides those reported).

The rule is expecting sorting within character classes, though this is not advertised as being part of the rule in the docs. E.g., `/#([^)'"]+)/` can be optimized to `/#([^"')]+)/` . I think this should be either advertised as such, or, better yet, made into an option, since it is sometimes clearer to make one's own groupings without sorting, e.g., `[AaQqTt]`. 

Thanks!

voxpelli:
Fixes #453

Improved by https://github.com/DmitrySoshnikov/regexp-tree/pull/200

Does this look like a good addition?

### DIFF
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -33,4 +33,4 @@ const regex = /\d\.\w\-\D/i;
 Type: `boolean`\
 Default: `true`
 
-Disables optimizations that affects the sorting of character classes. Eg. preserves `[AaQqTt]` in its order rather then resorting it to `[AQTaqt]`
+Disables optimizations that affect the sorting of character classes. For example, preserves the order of the characters in `[AaQqTt]` rather than sorting it to `[AQTaqt]`.
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -2,7 +2,6 @@
 
 This rule is fixable.
 
-
 ## Fail
 
 ```js
@@ -14,7 +13,6 @@ const regex = /[^a-zA-Z0-9_]/;
 const regex = /[^a-z0-9_]/i;
 const regex = /[0-9]\.[a-zA-Z0-9_]\-[^0-9]/i;
 ```
-
 
 ## Pass
 
@@ -32,7 +30,7 @@ const regex = /\d\.\w\-\D/i;
 
 ### disableCharacterSorting
 
-Type: `boolean`<br>
+Type: `boolean`\
 Default: `false`
 
 Disables optimizations that affects the sorting of characters.
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -33,4 +33,4 @@ const regex = /\d\.\w\-\D/i;
 Type: `boolean`\
 Default: `true`
 
-Disables optimizations that affects the sorting of characters.
+Disables optimizations that affects the sorting of character classes.
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -28,9 +28,9 @@ const regex = /\d\.\w\-\D/i;
 
 ## Options
 
-### disableCharacterSorting
+### sortCharacterClasses
 
 Type: `boolean`\
-Default: `false`
+Default: `true`
 
 Disables optimizations that affects the sorting of characters.
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -27,3 +27,12 @@ const regex = /\W/;
 const regex = /\W/i;
 const regex = /\d\.\w\-\D/i;
 ```
+
+## Options
+
+### disableCharacterSorting
+
+Type: `boolean`<br>
+Default: `false`
+
+Disables optimizations that affects the sorting of characters.
--- a/docs/rules/regex-shorthand.md
+++ b/docs/rules/regex-shorthand.md
@@ -33,4 +33,4 @@ const regex = /\d\.\w\-\D/i;
 Type: `boolean`\
 Default: `true`
 
-Disables optimizations that affects the sorting of character classes.
+Disables optimizations that affects the sorting of character classes. Eg. preserves `[AaQqTt]` in its order rather then resorting it to `[AQTaqt]`
--- a/rules/regex-shorthand.js
+++ b/rules/regex-shorthand.js
@@ -6,35 +6,13 @@ const quoteString = require('./utils/quote-string');
 
 const message = 'Use regex shorthands to improve readability.';
 
-// TODO: Simplify when/if this PR gets merged: https://github.com/DmitrySoshnikov/regexp-tree/pull/200
-const availableRegexpTreeOptimizations = [
-	'charSurrogatePairToSingleUnicode',
-	'charCodeToSimpleChar',
-	'charCaseInsensitiveLowerCaseTransform',
-	'charClassRemoveDuplicates',
-	'quantifiersMerge',
-	'quantifierRangeToSymbol',
-	'charClassClassrangesToChars',
-	'charClassClassrangesMerge',
-	'charClassToMeta',
-	'charClassToSingleChar',
-	'charEscapeUnescape',
-	'disjunctionRemoveDuplicates',
-	'groupSingleCharsToCharClass',
-	'removeEmptyGroup',
-	'ungroup',
-	'combineRepeatingPatterns'
-];
-
-const generateRegexpTreeWhitelistFromBlacklist = blacklist => availableRegexpTreeOptimizations.filter(optimization => !blacklist.includes(optimization));
-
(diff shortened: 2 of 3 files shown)


docs/rules/regex-shorthand.md
```diff
@@ -27,3 +25,12 @@ const regex = /\W/;
 const regex = /\W/i;
 const regex = /\d\.\w\-\D/i;
 ```
+
+## Options
+
+### disableCharacterSorting
```
sindresorhus:
```suggestion
### sortCharacters
```

And make it true by default.

sindresorhus: Actually, `sortCharactersClasses` may be more correct.

docs/rules/regex-shorthand.md
```diff
@@ -27,3 +25,12 @@ const regex = /\W/;
...
+
+Type: `boolean`\
+Default: `false`
+
+Disables optimizations that affects the sorting of characters.
```
sindresorhus: The description here needs to be clearer about what exactly will be sorted. "characters" is too ambiguous.

rules/regex-shorthand.js
```diff
@@ -6,7 +6,36 @@ const quoteString = require('./utils/quote-string');
 
 const message = 'Use regex shorthands to improve readability.';
 
+const availableRegexpTreeOptimizations = [
```
sindresorhus: Can you add a todo comment about simplifying this when https://github.com/DmitrySoshnikov/regexp-tree/pull/200 is released?

fisker: Why we need that? Why not default to all options and remove disabled items?

fisker:
Did `charClassClassrangesMerge` can actually disable sorting?
If it does, I suggestion use `charClassClassrangesMerge` directly

docs/rules/regex-shorthand.md
voxpelli: @sindresorhus There are two options as I see it:

1. Get into the specifics of regexp-tree and be up front about the disabling of `charClassClassrangesMerge`, which the sorting is a needed side-effect of. This is probably too implementation specific and doesn't convey the purpose clearly.
2. Clearly convey the intended effect by naming it `sortCharacters`/`sortCharacterClasses` (which the description of maybe could be spelled out in a somewhat more specific manner, but I have a hard time pinpointing that, maybe due to me not being the best at Regexp terminology).

I lean towards the second and I get the impression that you do as well?

fisker:
@voxpelli great work, indeed this prevent sorting

rules/regex-shorthand.js
voxpelli: @fisker As far as I see it, we want to convey what rules we do _not_ want, rather than convey what rules we do want. As the important thing in this case is what isn't included, not what is included. Intention becomes more clear if regexp-tree gets handed such a blacklist directly.

Added a TODO

fisker:
@voxpelli would you mind if I refactor the option logic? we don't need that blacklist logic.

voxpelli:
@fisker Because of reasons outlined in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/468#discussion_r353228256 I think the logic around regexp-tree should be kept as an internal implementation of this rule and not be exposed to the outside world.

That way this rule can more easily be adapted to future changes in regexp-tree and/or to future alternative modules.

Therefore I think that all we should do for now is to convey to regexp-tree that we want this very rule to sometimes be disabled, as made possible through https://github.com/DmitrySoshnikov/regexp-tree/pull/200

Reason why I've implemented the blacklist feature myself here is to make this feature be easily upgraded once that PR has been merged and to have the intention in this code to be clear,

I do not like the idea of:

```javascript
{
  charClassClassrangesMerge: false,
  charSurrogatePairToSingleUnicode: true,
}
```

That exposes way too much about the internal implementation and pushes too many decisions onto the user rather than abstract them in a way that makes sense from the purpose of this rule (which has a different target audience than the one of regexp-tree).

rules/regex-shorthand.js
voxpelli: A built in blacklist conveys our intention better + makes any future changes to the set of rules become applied automatically, as it would have been before this PR.

fisker:
@voxpelli Your comment make sense, but I still think let users have all the options is good. 

Also this `charClassClassrangesMerge` not only prevent sorting, also disable ranges from merging as you can see 

```js
/[GgHhIiå.Z:a-f"0-8%A*ä]/

/["%*.0-8:AG-IZa-iäå]/
```

the `a-fghi` is merged to `a-i`, we disable `charClassClassrangesMerge` also disable merging, people may ask why

voxpelli:
@fisker The fact that it disables is a reason why I want to expose the disabling of sorting, as that enables this rule to evolve to separate those additional fixes from the actual sorting, which a raw option of `charClassClassrangesMerge` totally dodges and instead downstreams to `regexp-tree`, where it may not belong as clearly.

Such a shortcoming can be mentioned in the documentation of a `sortCharacterClasses` option and a future version of this module can then fix it, when `regexp-tree` has evolved to such a degree that we can actually fix it.

fisker:
I have a regex [`/Error:?\s+([^\r\n]+)/i`](https://github.com/fisker/fis3-plugins/commit/a27f3eaa8cfa124342e2d4a6ff24795d3d655a6a), yesterday when I update `unicorn`, I got a it fixed to `/error:?\s+([^\n\r]+)/i`, in fact, this my test string is `Error:....`, I put `i` there just in case I may change the message in feature for safty, if I can set [`charCaseInsensitiveLowerCaseTransform`](https://github.com/DmitrySoshnikov/regexp-tree/tree/master/src/optimizer#transforms), I may already change my default, because I prefer not touching it.

I guess not only me, people may also need this. If it happends, we'll need to add another option again. Why not make it easier

voxpelli:
I think we disagree on what makes it easier 🙂

I think exposing of internal implementation details makes the API more complicated while you consider it to make it easier.

I guess we can agree to disagree and have someone else, like @sindresorhus, chime in with his views?

fisker:
//cc @futpib @MrHen may also interested

sindresorhus:
I agree with @voxpelli, we shouldn't expose the internals of `regexp-tree`.

sindresorhus:
> I put i there just in case I may change the message in feature for safty

I would argue this is actually an argument for not exposing all the options. The rule is actually improving your code and you're trying to fight it. By having the `i` in there, even though it's not used, you're actually making your regex harder to read. I think your "just in case" need is better solved by an explanatory code comment.

> I guess not only me, people may also need this. If it happends, we'll need to add another option again.

If there's a valid use-case, we can add a well-documented option for it. Exposing everything just leads to users having to waste too much time reading docs and configuring, and we would have to pin the dependency as every patch release of `regexp-tree` is a potential breaking change for us.

sindresorhus:
@voxpelli The update is out: https://github.com/DmitrySoshnikov/regexp-tree/commits/master Can you update to use the blacklist?

fisker:
@sindresorhus I did the request changes, do we need set it to `false` by default? seems `charClassClassrangesMerge` is not safe #472

voxpelli:
Thanks @fisker! I was out of office for a few days, so unfortunately didn't have time to fix it :)

docs/rules/regex-shorthand.md
sindresorhus: Sorry, missed this. Yes, `2`. Could you also add an example? Not everyone knows what character classes are.

docs/rules/regex-shorthand.md
```diff
@@ -27,3 +25,12 @@ const regex = /\W/;
...
 ```
+
+## Options
+
+### sortCharacterClasses
+
+Type: `boolean`\
+Default: `true`
+
+Disables optimizations that affects the sorting of characters.
```
sindresorhus:
```suggestion
Disables optimizations that affects the sorting of character classes.
```

voxpelli:
@sindresorhus Is this ready to go you think?

sindresorhus:
@voxpelli Did you see https://github.com/sindresorhus/eslint-plugin-unicorn/pull/468#discussion_r360717019?

sindresorhus:
@voxpelli Bump :)

voxpelli:
Done @sindresorhus :) Sorry for me taking so long, I appreciate your bumps to remind me 👍

voxpelli:
Also rebased it 👍

sindresorhus:
Thank you 👍

JounQin:
@sindresorhus Should its default be `false` due to #472, #477, #499?

sindresorhus:
I'm hoping that the author of regexp will fix those extra issues soon. If not, we could do this, yes.

PR #476, opened 2019-12-17
Fix invalid autofix with destructuring assignment in no-for-loops rule

bmish:
Example (original):
```js
for (let i = 0; i < arr.length; i++) {
  const { a, b } = arr[i];
  console.log(a, b);
}
```

Autofix (before this fix):
```js
for (const element of arr) {
  console.log(a, b);
}
```

Autofix (after this fix):
```js
for (const element of arr) {
  const { a, b } = element;
  console.log(a, b);
}
```

When `arr[i]` is used in a destructuring assignment, the entire assignment should not be removed (previous behavior), only `arr[i]` should be replaced with `element` (new behavior).

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -326,18 +326,27 @@ const create = context => {
 					const shouldGenerateIndex = isIndexVariableUsedElsewhereInTheLoopBody(indexVariable, bodyScope, arrayIdentifierName);
 
 					const index = indexIdentifierName;
+					const element = elementIdentifierName || defaultElementName;
 					const array = arrayIdentifierName;
 
-					let element = elementIdentifierName || defaultElementName;
+					let declarationElement = element;
 					let declarationType = 'const';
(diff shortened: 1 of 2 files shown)


fisker:
LGTM, but can we fix this to if `element` is not used?

```
for (const { a, b } of arr) {
  console.log(a, b);
}
```

bmish:
@fisker that's a good idea. However, since that is a much more involved fix, it's probably worth considering as a follow-up improvement.

fisker:
sending a PR to your fork

fisker:
@bmish 

https://github.com/bmish/eslint-plugin-unicorn/pull/1

what do you think? can you merge it?

bmish:
@fisker your improvement looks great, thank you! But I'm not sure how to merge it in my fork due to the conflicts. Perhaps you could just add the commit to this PR?

fisker:
@bmish  conflicts resolved

fisker:
Good job, thank you! @bmish

Issue #481, opened 2019-12-23
v15 is incompatible with Node 8


bmish:
[v15](https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v15.0.0) was just released and is incompatible with Node 8 due to the dependency on semver being upgraded.

```
error semver@7.1.1: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
```
I understand that Node 8 is end-of-life soon, so if this breaking change is intentional, it would help if it was documented in the release notes / CHANGELOG / README.

If the change is not intentional, then semver could be downgraded and a patch released, and the supported Node versions could be documented in the README.

I would personally prefer if Node 8 could be supported for at least a few more months to give other packages using eslint-plugin-unicorn some more time to drop support for it, unless there is an actual need to drop it now.

(also: why didn't the tests catch this breakage? .travis.yml specifies Node 8 as one of the versions to test under.)

sindresorhus:
This was not intentional. I'll make sure it doesn't happen again: https://github.com/sindresorhus/np/issues/480

> (also: why didn't the tests catch this breakage? .travis.yml specifies Node 8 as one of the versions to test under.)

Because npm doesn't enforce the `engine` field. It's really stupid.
https://docs.npmjs.com/misc/config#engine-strict

PR #482, opened 2019-12-23
docs: add minimum Node version requirement to README

bmish:
Related: #481

In a separate PR, the minimum supported eslint version could also be added to this requirements section, if desired.


sindresorhus:
Thanks for the suggestion, but this is already defined in package.json. I don't really want to have it in the readme.

PR #485, opened 2019-12-24
Add `prefer-replace-all` rule

### LINKED ISSUE #433: Rule proposal: `prefer-replace-all`
As the `replaceAll` method is [being added](https://v8.dev/features/string-replaceall) to strings, I think it would be a great opportunity to create a rule which warns on patterns similar to:

```js
str.replace(/literal characters only/g, 'something');
```

Auto-fixing may also be possible:

```js
str.replaceAll('literal characters only', 'something');
```

jmoore914:
Fixes #433

### DIFF
--- a/docs/rules/prefer-replace-all.md
+++ b/docs/rules/prefer-replace-all.md
@@ -1,0 +1,24 @@
+# Prefer `String.prototype.replaceAll` over regex searches with the global flag.
+
+As the replaceAll method is being added to strings, there is no reason to use expensive regex searches when the search string is a string-literal.
+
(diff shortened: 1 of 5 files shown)


jmoore914:
I could significantly improve the regexp to check that there are only literal characters in the search pattern with a negative look behind but that is only supported in node >=8.10
However, in the package.json, eslint is listed as a peer dependency with version >=6.7.1 and that eslint version requires >=8.10. Also, node v8 is going EOL in 5 days so I think it is safe to assume that anybody using a release including this PR would be running at least 8.10, correct?

PR #488, opened 2019-12-26
Add `prefer-replace-all` rule

### LINKED ISSUE #433: Rule proposal: `prefer-replace-all`

jmoore914:
Fixes #433


rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
...
+	const escapedSpecialCharacters = searchPattern.match(/\\[$()*+.?[\\\]^{}]/g);
+	const escapedSpecialCharactersCount = escapedSpecialCharacters ? escapedSpecialCharacters.length : 0;
+	return specialCharactersCount === 2 * escapedSpecialCharactersCount;
+}
+
+function replaceNode(node, fixer) {
+	const stringName = node.callee.object.name;
+	const searchPattern = node.arguments[0].regex.pattern.replace(/\\(.)/g, '$1');
+	const replacePattern = node.arguments[1].value;
+	return fixer.replaceText(node, stringName + '.replaceAll("' + searchPattern + '", "' + replacePattern + '")');
```
fisker: use `quoteString` from `/rules/utils/quote-string.js`

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
...
+}
+
+function checkNode(context, node) {
+	if (checkReplace(node) && checkGlobalFlag(node) && checkLiteralCharactersOnly(node)) {
```
fisker: What do you think name those function with `is` prefix?

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
...
+			node,
+			message: 'Use replaceAll method of string.',
+			fix: fixer => replaceNode(node, fixer)
+		});
+	}
+}
+
+const create = context => {
+	return {
+		CallExpression: node => {
```
fisker: I think we can use `CallExpression[callee.property.name="replace"]` selector here

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+function checkReplace(node) {
+	return node.callee.property.name === 'replace';
+}
+
+function checkGlobalFlag(node) {
+	const searchPattern = node.arguments[0];
+	return Object.prototype.hasOwnProperty.call(searchPattern, 'regex') && searchPattern.regex.flags === 'g';
```
fisker: why not just `searchPattern && searchPattern.regex && searchPattern.regex.flags === 'g'`. 
I understand flags with `i` should not pass, but how about other [flags](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp#Parameters) like `gu`?

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
...
+	const specialCharacters = searchPattern.match(/[$()*+.?[\\\]^{}]/g);
+	const specialCharactersCount = specialCharacters ? specialCharacters.length : 0;
+	const escapedSpecialCharacters = searchPattern.match(/\\[$()*+.?[\\\]^{}]/g);
+	const escapedSpecialCharactersCount = escapedSpecialCharacters ? escapedSpecialCharacters.length : 0;
+	return specialCharactersCount === 2 * escapedSpecialCharactersCount;
+}
+
+function replaceNode(node, fixer) {
+	const stringName = node.callee.object.name;
+	const searchPattern = node.arguments[0].regex.pattern.replace(/\\(.)/g, '$1');
```
fisker: This line is not right, `/\\a/g` is searching for `\a` not `a`, maybe `.replace(/\\\\(.)/g, '\\$1')`?.

Also, funny thing.

I think we can replace it with `.replace(/\\\\/g, '\\')`, then this line will against this rule itself, so maybe `.replaceAll('\\\\', '\\')`?

rules/prefer-replace-all.js
jmoore914: Does the built in replace method have full unicode support? From what I'm reading it seems like it doesn't. And all of the other flags would change the search behavior as well.

rules/prefer-replace-all.js
jmoore914: Hmm. On second thought, that replace shouldn't be necessary at all. However, I do need to add an escape for any double quotes in the original regex expression.

Yeah, that ocurred to me while I was writing it. Very ironic. I decided not to use replaceAll since it doesn't have support in node yet and it would be transpiled back to the expression I have if we ran it through babel.

rules/prefer-replace-all.js
jmoore914: That selector did not work for me, threw an error

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,57 @@
...
+
+function replaceNode(node, fixer) {
+	const stringName = node.callee.object.name;
+	const searchPattern = node.arguments[0].regex.pattern;
+	const replacePattern = node.arguments[1].value;
+	return fixer.replaceText(node, `${stringName}.replaceAll(${quoteString(searchPattern)}, ${quoteString(replacePattern)})`);
+}
+
+function checkNode(context, node) {
+	if (isReplaceMethod(node) && hasGlobalFlag(node) && isLiteralCharactersOnly(node)) {
```
fisker: Can you also add `arguments` length check here?

fisker: Also type check for `arguments[1]`

rules/prefer-replace-all.js
fisker: Works for me, I committed to your fork.

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,53 @@
...
+	const escapedSpecialCharactersCount = escapedSpecialCharacters ? escapedSpecialCharacters.length : 0;
+	return specialCharactersCount === 2 * escapedSpecialCharactersCount;
+}
+
+function replaceNode(node, fixer) {
+	const stringName = node.callee.object.name;
+	const searchPattern = node.arguments[0].regex.pattern;
+	const replacePattern = node.arguments[1].value;
+	return fixer.replaceText(node, `${stringName}.replaceAll(${quoteString(searchPattern)}, ${quoteString(replacePattern)})`);
+}
```
fisker: This part is not right, callee should not touched,  callee might any type not only `Identifier`,  replacement also not safe

```js
const replacement = '0';
const f00 = 'foo'.replace(/o/g, replacement);
```

I suggest `replace` -> `replaceAll`, then `arguments[0]` -> `fixed string`, `arguments[1]` should not touched

rules/prefer-replace-all.js
fisker: Not sure we should check `arguments[1]` has `$0` `$1`

jmoore914: Yep, I can add that.

And what you said below is correct, `arguments[1]` shouldn't be changed at all so there's no need to type check. If it contained `$0` or `$1` the regex would have a match group so it would not pass the check for the search phrase containing only literals.

fisker:
@jmoore914 Can you check https://github.com/sindresorhus/eslint-plugin-unicorn/pull/488/commits/715eb673cae94490b4557930a5aeb3e828c491b4 ? Make sure I didn't miss any test cases

test/prefer-replace-all.js
```diff
@@ -19,56 +19,69 @@ const error = {
...
 			errors: [error]
 		},
+		// Escaped symbols
 		{
-			code: 'foo.replace(/searchPattern/g, \'\\\'escapedQuotes\\\'\')',
-			output: 'foo.replaceAll(\'searchPattern\', \'\\\'escapedQuotes\\\'\')',
+			code: 'foo.replace(/\\./g, bar)',
+			output: 'foo.replaceAll(\'.\', bar)',
 			errors: [error]
 		},
```
fisker: @jmoore914 I think I found a broken case, we still need replace `\`

fisker: Maybe it's not that simple at all, we should test all these cases

```js
/\./g 
/\\./g 
/\\\./g 
```

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
...
+
+function isLiteralCharactersOnly(node) {
+	const searchPattern = node.regex.pattern;
+	const specialCharacters = searchPattern.match(/[$()*+.?[\\\]^{}]/g);
+	const specialCharactersCount = specialCharacters ? specialCharacters.length : 0;
+	const escapedSpecialCharacters = searchPattern.match(/\\[$()*+.?[\\\]^{}]/g);
+	const escapedSpecialCharactersCount = escapedSpecialCharacters ? escapedSpecialCharacters.length : 0;
+	return specialCharactersCount === 2 * escapedSpecialCharactersCount;
+}
```
fisker: I'm sure this can be simplified, can you check it?

rules/prefer-replace-all.js
```diff
@@ -0,0 +1,56 @@
...
+
+			const [search] = arguments_;
+
+			if (!isRegexWithGlobalFlag(search) || !isLiteralCharactersOnly(search)) {
+				return;
+			}
+
+			context.report({
+				node,
+				message: 'Use replaceAll method of string.',
```
fisker:
```suggestion
				message: 'Prefer \`String#replaceAll()\` over \`String#replace()\`.',
```

rules/prefer-replace-all.js
jmoore914: Hmm. One idea would be 
```js
function isLiteralCharactersOnly(node) { 
   const searchPattern = node.regex.pattern;
   const specialCharactersCount = searchPattern.split(/[$()*+.?[\\\]^{}]/).length - 1;
   const escapedSpecialCahractersCount = searchPattern.split(/\\[$()*+.?[\\\]^{}]/).length - 1;
   return specialCharactersCount === 2 * escapedSpecialCharactersCount; 
}
```
It could then turn into:
```js
function isLiteralCharactersOnly(node) { 
   const searchPattern = node.regex.pattern;
   return  searchPattern.split(/[$()*+.?[\\\]^{}]/).length - 1 === 2 *(searchPattern.split(/\\[$()*+.?[\\\]^{}]/).length - 1)
}
```

And if we want, could even turn it into:
```js
function isLiteralCharactersOnly(node) { 
   return  node.regex.pattern.split(/[$()*+.?[\\\]^{}]/).length - 1 === 2 *(node.regex.pattern.split(/\\[$()*+.?[\\\]^{}]/).length - 1)
}
```

I don't know which is most readable though.

fisker: If I understand correctly, this function means check specialCharacters(like `.`, including `\`) length (A) and escaped specialCharacters like `\.` or `\\` length (B), and A should be B * 2 ?

jmoore914: Correct. Every special character is escaped.

fisker: If so, how about 

```js
function isLiteralCharactersOnly(node) { 
   return  /[$()*+.?[\\\]^{}]/.test(pattern.replace(/\\[$()*+.?[\\\]^{}]/g, '')
}
```

equal?

On cellphone, not sure

jmoore914: Very clever. Yeah, that should work. I'll try it out when I'm able.

jmoore914: function isLiteralCharactersOnly(node) { 
   return  !/[$()*+.?[\\\]^{}]/.test(pattern.replace(/\\[$()*+.?[\\\]^{}]/g, '')
}

That worked but when I merged in your other changes, tests started failing

fisker:
@jmoore914 see

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/488#discussion_r362688499

rules/prefer-replace-all.js
fisker: @jmoore914 see

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/488#discussion_r362688499

jmoore914:
> 
> 
> @jmoore914 see
> 
> [#488 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/488#discussion_r362688499)

Newest PR should fix this comment

sindresorhus:
Thanks :)

Issue #295, opened 2019-05-08
A case not handled by `no-for-loop`


chiawendt:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2440%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/295)

```js
/* eslint unicorn/no-for-loop: 2 */

const arr = [];

for (let i = 0, j = arr.length; i < j; i += 1) {
  const element = arr[i];
  console.log(element);
}
```
`no-for-loop` should report on this but it does not.

This slightly unusual pattern is used all over the place in `eslint-plugin-react`.

This issue is a subtask of #250.

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Sponsors (Total: $40.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($40.00)

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

sindresorhus:
@golopot This was not handled by https://github.com/sindresorhus/eslint-plugin-unicorn/pull/297, right?

golopot:
It is not.

IssueHuntBot:
@issuehunt has funded $40.00 to this issue.

---
- Submit pull request via [IssueHunt](https://issuehunt.io/repos/55832243/issues/295) to receive this reward.
- Want to contribute? Chip in to this issue via [IssueHunt](https://issuehunt.io/repos/55832243/issues/295).
- Checkout the [IssueHunt Issue Explorer](https://issuehunt.io/issues) to see more funded issues.
- Need help from developers? [Add your repository](https://issuehunt.io/r/new) on IssueHunt to raise funds.

MrHen:
I took a quick look at this and think this should be a separate rule that auto-fixes this shape into a normal for loop. `no-for-loop` can then convert the normal for loop since it already handles that.

Original:

```
for (let i = 0, j = arr.length; i < j; i += 1) {
  const element = arr[i];
  console.log(element);
}
```

Fixed:

```
[...]

PR #489, opened 2019-12-27
Fix invalid autofix with array destructuring in `no-for-loops` rule

fisker:
Follow up #476 , #476 only fixed object destructuring, not array destructuring

~fixes:​ #295~

~please check https://github.com/sindresorhus/eslint-plugin-unicorn/issues/295#issuecomment-569177568 first when review this PR~

sorry, this didn't fixx #295


fisker:
//cc @bmish

PR #490, opened 2019-12-27
Improve `number-literal-case` rule

fisker:
1. Make it work on `BigInt`, previously crashes on `BigInt`, since `n` is upercased.
2. Lowercase `e` for exponential notation.
3. Add MDN links

fixes: #491

### DIFF
--- a/docs/rules/number-literal-case.md
+++ b/docs/rules/number-literal-case.md
@@ -1,26 +1,42 @@
-# Enforce lowercase identifier and uppercase value for number literals
+# Enforce proper case for numeric literals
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
> Question: should this work on BigInt ? need rename rule?

Yeah, I think it should work on BigInt too. I don't think it's necessary to rename the rule. BigInt is a number too. We can just make it explicit in the description that it works on normal number and BigInt numbers.

fisker:
I'll fix it today

sindresorhus:
Nice work 👌

PR #496, opened 2020-01-03
Add `string-content` rule

### LINKED ISSUE #327: Rule proposal: `string-content`
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/327)

### Good 👍 

```js
const description = 'Add text to user’s comments'
```

### Not good 👎 

```js
const description = 'Add text to user\'s comments'
```

Prior art: https://github.com/jmont/eslint-plugin-smart-quotes

I tried that one but it doesn't have a fixer and it has many false positives. Also strings could be CSS selectors so`[alt=""]` needs to stay as is.

---

_Originally posted by @sindresorhus in https://github.com/sindresorhus/refined-github/pull/2167_:

```js
'use strict';

const create = context => {
	return {
		Literal: node => {
          	const {value} = node;
     
			if (typeof value !== 'string') {
				return;
			}
          
        	const fixedValue = value.replace(/'/, '‘');

			if (fixedValue === value) {
				return;
			}

			context.report({
				node,
				message: 'Use other quote kind',
				fix: fixer => fixer.replaceTextRange([node.range[0] + 1, node.range[1] - 1], fixedValue)
			});
		}
	};
};

module.exports = {
	create,
	meta: {
		type: 'problem',
		fixable: 'code'
	}
};
```

https://astexplorer.net/#/gist/3a67df351a57608ae38a6b55102376d6/c8425e326ddce75bf7242295c8818134ef2d2e16

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

### Submitted pull Requests
- [#496 Add `string-content` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/496)
---

### Tips

---
</details>

fisker:
fixes #327

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#327: Rule proposal: `string-content`](https://issuehunt.io/repos/55832243/issues/327)
---

</details>


sindresorhus:
You're not waiting for feedback from me on anything here, right?

fisker:
I'm not, but I have problem with this, harder than I thought. I'll explain later when I need your help.

fisker:
@sindresorhus Set default to `off`, or disable this rule on our codebase?

sindresorhus:
Yes, it should be off by default. I will enable it in XO with some useful patterns, but I think that's too opinionated for this general plugin.

fisker:
@sindresorhus This PR is ready for review.

Something I need explain first, I set this rule only report one error a time on one string node, for the following reasons.

1. Case: String `ab`, replacement `a -> A, b -> B`

If we check all patterns on the original string, we will get two problem with 2 fixes, so we get fix `ab -> Ab` and `ab -> aB`, but they are conflict, so only `b->B` get replaced.

2. Case: String `a`, replacement `a -> b, b -> c`

If we check `a -> b` first, got result `b`, then check `b -> c` on `b`, got correct result `c`, we report two problems `a -> b, b -> c`, but there is no `b` in the original string.

3. If we consider some replacement set `fix:false`, this get even complicated.

readme.md
```diff
@@ -132,6 +133,7 @@ Configure it in `package.json`.
 - [prefer-trim-start-end](docs/rules/prefer-trim-start-end.md) - Prefer `String#trimStart()` / `String#trimEnd()` over `String#trimLeft()` / `String#trimRight()`. *(fixable)*
 - [prefer-type-error](docs/rules/prefer-type-error.md) - Enforce throwing `TypeError` in type checking conditions. *(fixable)*
 - [prevent-abbreviations](docs/rules/prevent-abbreviations.md) - Prevent abbreviations. *(partly fixable)*
+- [string-content](docs/rules/string-content.md) - ??. *(fixable)*
```
sindresorhus: Missing description

docs/rules/string-content.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Options
+
+Type: `object`
+
+### `patterns`
+
+Type: `object`
+
+You can extend default patterns by passing the `patterns` option.
```
sindresorhus: Linkify `default patterns` to link to the default patterns.

rules/string-content.js
```diff
@@ -0,0 +1,154 @@
...
+const getDocumentationUrl = require('./utils/get-documentation-url');
+const replaceTemplateElement = require('./utils/replace-template-element');
+
+const defaultPatterns = {
+	'\'': '’'
+};
+
+const defaultMessage = 'Prefer `{{suggest}}` over `{{match}}`.';
+
+const escapeTemplateElementRaw = string => string.replace(
```
sindresorhus: This should be moved to the util file.

docs/rules/string-content.md
```diff
@@ -0,0 +1,66 @@
+# Enforce better string context
+
```
sindresorhus: The description should describe some use-cases for this rule.

sindresorhus:
> If we check all patterns on the original string, we will get two problem with 2 fixes

You can give the `report()` method an array of fixes. Does that help?

> An array which includes fixing objects. - https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes

fisker:
> You can give the report() method an array of fixes. Does that help?

I've tried this before,  not useful.

sindresorhus:
Alright. Let's at least document that it only reports one thing per node.

rules/utils/quote-string.js
```diff
@@ -1,9 +1,16 @@
...
 
 @param {string} string - The string to be quoted.
-@returns {string} - The quoted string.
+@param {quote} string - The quote character.
+@returns {string} - The quoted and escaped string.
 */
-module.exports = string => `'${string.replace(/'/g, '\\\'')}'`;
+module.exports = (string, quote = '\'') => {
+	const escaped = string
+		.replace(new RegExp(quote, 'g'), `\\${quote}`)
```
sindresorhus: Are you sure this is enough? Here, I used to use a regex for escaping the string, but it could too complicated and in the end I had to use a proper parser: https://github.com/sindresorhus/to-single-quotes/commit/7f203d0f6a25ab5e74911732b77de02f8dc7eae2

fisker: Wow, I checked the tests on that repo, seems complicated. But no one has report any issue with the old simple encode yet. I guess we can improve this if other case found.

test/prefer-replace-all.js
```diff
@@ -86,7 +86,7 @@ ruleTester.run('prefer-replace-all', rule, {
 		},
 		{
 			code: 'foo.replace(/\\\\\\./g, bar)',
-			output: 'foo.replaceAll(\'\\.\', bar)',
+			output: 'foo.replaceAll(\'\\\\.\', bar)',
```
fisker: The old one is wrong,

```js
foo.replace(/\\\./g, bar)
```

Should fix to

```js
foo.replaceAll('\\.', bar)
```

rules/utils/quote-string.js
fisker: Found broken cases, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/496/commits/b81ea17f7f5cf221dcdf214ab8e789ac18b1f4b5

Similar escape https://github.com/eslint/eslint/blob/9dfc8501fb1956c90dc11e6377b4cb38a6bea65d/lib/rules/quotes.js#L48

sindresorhus:
@fisker I think there's only https://github.com/sindresorhus/eslint-plugin-unicorn/pull/496/files#r386895029 left and this is ready to be merged?

fisker:
@sindresorhus I've add a simple description https://github.com/sindresorhus/eslint-plugin-unicorn/commit/9c0953d0957fe5498e5b072825264079d3844b27 . Not sure what  else you want add here. Maybe you  can  help, I'm really bad at docs .

PR #498, opened 2020-01-17
prefer-node-remove: Only fix when expression is not used

### LINKED ISSUE #497: `prefer-node-remove`: prevent fixer when using return value
`prefer-node-append` currently checks for use of the return value since the replacement of `appendChild`, `append`, does not return a value.

`prefer-node-remove` should do the same, since the old `replaceChild` returned a value (the removed node) while `ChildNode.remove()` does not return a value.

brettz9:
Fixes #497

I also have a minor update and very minor refactoring included with this PR. Let me know if you need them separated out.


rules/prefer-node-remove.js
```diff
@@ -64,11 +77,23 @@ const create = context => {
...
+					parent: grandparent
+				} = (parent || {});
+
+				let fix = fixer => fixer.replaceText(node, `${argumentName}.remove()`);
+
+				if (parent && ignoredParentTypes.includes(parent.type)) {
+					fix = undefined;
+				} else if (grandparent && ignoredGrandparentTypes.includes(grandparent.type)) {
+					fix = undefined;
+				}
```
fisker:
```suggestion
				if (
					(parent && ignoredParentTypes.includes(parent.type)) || 
					(grandparent && ignoredGrandparentTypes.includes(grandparent.type))
				) {
					fix = undefined;
				}
```

brettz9: Thanks! I've applied this, and also to the append rule.

brettz9:
Applied, thanks.

Btw, I'm getting issues when running `npm run lint`, even before my changes. But `npm test` and `npm run integration` are working.

fisker:
@brettz9 If they  are the some, can we add  a function maybe `isReturnValueUsed`(or a better name) in `/rules/utils`?

rules/prefer-node-append.js
```diff
@@ -1,41 +1,18 @@
...
 
 			if (callee.type === 'MemberExpression' && getMethodName(callee) === 'appendChild') {
 				let fix = fixer => fixer.replaceText(callee.property, 'append');
 
-				if (parent && ignoredParentTypes.includes(parent.type)) {
-					fix = undefined;
-				}
-
-				if (grandparent && ignoredGrandparentTypes.includes(grandparent.type)) {
+				if (!isReturnValueUsed(node)) {
```
fisker: Seems not right? If used, should has no fix, right?

brettz9:
actually, on further reflection, I'm thinking `isValueUsed` would be better, since it is not necessarily part of a return. I've added a commit to change, though feel free to revert.

rules/prefer-node-append.js
brettz9: You're right--I inverted it by mistake. Plan to fix now.

brettz9: Fixed.

rules/prefer-node-remove.js
```diff
@@ -64,11 +65,17 @@ const create = context => {
 			)) {
 				const argumentName = getArgumentName(node.arguments);
 
+				let fix = fixer => fixer.replaceText(node, `${argumentName}.remove()`);
+
+				if (isValueUsed(node)) {
+					fix = undefined;
+				}
```
fisker:
```js
				const fix = isValueUsed(node) ? undefined : fixer => fixer.replaceText(node, `${argumentName}.remove()`);
```

fisker: And move it into `if` block, shall we?

brettz9: Ah, yes, all very good. Done.

fisker:
Looks great, thank you very much.

I think `prefer-modern-dom-apis` also missing some check https://github.com/sindresorhus/eslint-plugin-unicorn/pull/362#discussion_r360750630 , If you're interested, maybe you can help, just let you know, no pressure.

I also notice, `prefer-modern-dom-apis` checked `nodeParentType === 'AssignmentExpression'` and we don't do this check here, not sure if is it needed.

brettz9:
Yes, indeed, `nodeParentType === 'AssignmentExpression'` was needed. I've added that and some tests.

PR #503, opened 2020-01-24
prefer-modern-dom-apis: Only fix when expression is not used

brettz9:
Providing a fix related to [this comment](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/498#issuecomment-576365404).

~fix(`prefer-modern-dom-apis`): Ignore other cases where node return value is used~

~- Where parent is an `ArrayExpression`, `IfStatement`, `MemberExpression`,  `Property`, `ReturnStatement`~

~Fixes a test which made a fix which would break because of an inability of the replacement API to return a value.~

~Also provides minor optimization to avoid declaring fixer with value when not needed.
Also applies (and for `prefer-node-*` rules, switches from) checking `ExpressionStatement` in favor of specific parents (`BinaryExpression`, `CallExpression`, `ConditionalExpression`, `LogicalExpression`, `UnaryExpression`), since in theory, the last part of `parentNode.insertBefore(alfa, beta).insertBefore(charlie, delta);` could be checked if the rule were changed to allow such parent expressions rather than requiring a parent member to be a name only (i.e., fixing `parentNode` in `parent.insertBefore()` but ignoring `some.expression().insertBefore()`) since technically such a replacement could be made. But allowing it to be made would allow additional refactoring and potentially reporting long errors, e.g., "Prefer some.very.very.long.expression().before() over some.very.very.long.expression().insertBefore", and I'm not sure it is worth the trouble to support this strange pattern. But I did think it at least more logically clear to check the parents.~

fix(`prefer-modern-dom-apis`): Switch to whitelist for determining where node return values are expected safe not to be used.

For `prefer-modern-dom-apis`, the change will avoid fixers in more cases than just `VariableDeclarator` and `AssignmentExpression` parent types (as previously); will now avoid whenever parent is not an `ExpressionStatement` since many other parents, e.g., `ArrayExpression`, `IfStatement`, `'MemberExpression`, `Property`, `ReturnStatement`, etc. can use a return value.

For `prefer-node-*` rules, rather than blacklisting an albeit more extensive set of parents, will also switch to a whitelist, potentially avoiding fixers on other contexts not previously avoided, e.g., `AwaitExpression`.

The previous avoidance by `prefer-node-*` of fixers with `ExpressionStatement` grandparents is not currently required.

Also:
1. provides minor optimization to avoid declaring fixer with value when not needed.
2. fixes a test which made a fix which would break because of an inability of the replacement API to return a value.
3. refactors to use `isValueNotUsable` over `isValueUsed`

### DIFF
--- a/rules/utils/is-value-not-usable.js
+++ b/rules/utils/is-value-not-usable.js
@@ -3,5 +3,5 @@
 module.exports = function (node) {
 	const {parent} = node;
 
-	return parent && parent.type !== 'ExpressionStatement';
+	return !parent || parent.type === 'ExpressionStatement';
 };
--- a/rules/utils/is-value-not-usable.js
+++ b/rules/utils/is-value-not-usable.js
@@ -1,7 +1,3 @@
 'use strict';
 
-module.exports = function (node) {
-	const {parent} = node;
-
-	return !parent || parent.type === 'ExpressionStatement';
-};
+module.exports = ({parent}) => !parent || parent.type === 'ExpressionStatement';
--- a/rules/prefer-modern-dom-apis.js
+++ b/rules/prefer-modern-dom-apis.js
@@ -1,6 +1,6 @@
 'use strict';
 const getDocumentationUrl = require('./utils/get-documentation-url');
-const isValueUsed = require('./utils/is-value-used');
+const isValueNotUsable = require('./utils/is-value-not-usable');
 
 const getArgumentNameForReplaceChildOrInsertBefore = nodeArguments => {
 	if (nodeArguments.type === 'Identifier') {
@@ -42,15 +42,15 @@ const checkForReplaceChildOrInsertBefore = (context, node) => {
 
 	const preferredSelector = forbiddenIdentifierNames.get(identifierName);
 
-	const fix = isValueUsed(node) ?
+	const fix = isValueNotUsable(node) ?
 		// Report error when the method is part of a variable assignment
 		// but don't offer to autofix `.replaceWith()` and `.before()`
 		// which don't have a return value.
-		undefined :
 		fixer => fixer.replaceText(
 			node,
 			`${oldChildNodeArgument}.${preferredSelector}(${newChildNodeArgument})`
-		);
+		) :
+		undefined;
 
 	return context.report({
 		node,
@@ -103,7 +103,7 @@ const checkForInsertAdjacentTextOrInsertAdjacentElement = (context, node) => {
 		nodeArguments[1]
 	);
 
-	const fix = identifierName === 'insertAdjacentElement' && isValueUsed(node) ?
+	const fix = identifierName === 'insertAdjacentElement' && !isValueNotUsable(node) ?
 		// Report error when the method is part of a variable assignment
 		// but don't offer to autofix `.insertAdjacentElement()`
 		// which doesn't have a return value.
--- a/rules/prefer-modern-dom-apis.js
+++ b/rules/prefer-modern-dom-apis.js
@@ -1,5 +1,6 @@
 'use strict';
 const getDocumentationUrl = require('./utils/get-documentation-url');
+const isValueUsed = require('./utils/is-value-used');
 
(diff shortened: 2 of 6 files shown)


fisker:
@brettz9 

Thanks for fixing this rule.

I think we still miss many cases on that parent type check, eg:

`TemplateLiteral`

```js
`${a.insertBefore(b, c)}` // maybe no one do this way, but still possible
```

`NewExpression`

```js
new Dom(a.insertBefore(b, c))
```

`UnaryExpression`

```js
+a.insertBefore(b, c)
```

Not sure can we do opposite way? only allow certainly types, or can we check all possible node types, and add more.

fisker:
Is this all possible types? https://esprima.readthedocs.io/en/latest/syntax-tree-format.html

brettz9:
Ah, good to know re: other parent types. (The PR is already covering `UnaryExpression`,  btw, but not `TemplateLiteral` or `NewExpression`; I've added tests for these and added them to the blacklisted parents if it is decided to continue that route.)

With ESLint's parser being Espree, I wouldn't be sure that would cover everything, though it does seem to be a [project goal](https://github.com/eslint/espree#esprima-compatibility-going-forward) bring fundamental compatibility with Esprima. However, beyond this, other parsers are being used with eslint, including for TypeScript, so I wouldn't be surprised if there are even more.

As far as inverting the check, the one parent type in use in the tests which isn't caught here is `ExpressionStatement`.  And that would seem to make sense as the one AST I can think of which would be possible and wouldn't have side effects. But I could be missing something. Let me know if you want to go that route.

fisker:
We still not cover all possible cases, like  `await`, `yield` etc. Really hard to cover them all, maybe we can try do `whitelist` check.


/cc @sindresorhus WDYT?

sindresorhus:
> We still not cover all possible cases, like await, yield etc.

What do you mean exactly? `await a.insertBefore(b, c)`?

brettz9:
Yeah, if I may answer, that is correct. As unlikely as such cases may be, there are other types of places in JS where the return value may be significant (`insertBefore` gave a return value, whereas `before` does not, so we don't want to apply a fixer that substitutes one for the other when a return value is being used).

I think we may agree that only allowing fixers when the parent is a `ExpressionStatement` is the better choice, and to my knowledge, there are no other parent types where it would always be safe. The previous approach was to blacklist parent types where the return value was known to be usable (e.g., `LogicalExpression`), but there are quite a few types to cover in JS.

IMO, without going through each type, in either approach, we could be having false negatives or false positives, so I think it would be better to go with the safer approach of only allowing the fixer for `ExpressionStatement`. Besides being the simpler solution, it avoids being potentially over-aggressive in adding a "fix" where it would be harmful--and we can always add to the whitelist if we find cases where it may be needed.

sindresorhus:
>  so I think it would be better to go with the safer approach of only allowing the fixer for ExpressionStatement.

👍

brettz9:
I've pushed a fix that only allows a fixer for `ExpressionStatement` parents.

fisker:
@brettz9 

Should we change `isValueUsed` to negative `isValueNotUsed`, and only fix cases we are sure that value is not used?

brettz9:
I think the result should be the same, but we can invert the naming (though maybe `isValueNotUsable` would be better since we don't know that the value will really be "used", e.g., if a function calling a function with a `ReturnStatement` doesn't use the value returned).

fisker:
Great, let's do this.

brettz9:
I've updated

rules/utils/is-value-not-usable.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = function (node) {
+	const {parent} = node;
+
+	return !parent || parent.type === 'ExpressionStatement';
+};
```
fisker:
```suggestion
module.exports = ({parent}) => !parent || parent.type === 'ExpressionStatement';
```

PR #504, opened 2020-01-25
Switch from Travis CI to GitHub Actions

voxpelli:
Some benefits:

* More granular representation in the UI of which tests fails and which passes
* Very easy to run tests on Windows as well

Example runs can be seen here:

https://github.com/voxpelli/eslint-plugin-unicorn/actions?query=branch%3Atooling%2Fgithub-workflows

Badges like these ones can be added to readme after workflows has been added:

Tests
Other


sindresorhus:
I dunno. GH Actions is promising, but that’s a fairly large amount of config for something the Travis file does in 19 lines.

voxpelli:
@sindresorhus You can do it in fewer files, but I myself has opted for splitting them up in my repositories, to get them as fully separate workflows in the UI and logs.

It would be possible to fully replicate the Travis CI one I believe, but then the bonus of more granular feedback would get lost

voxpelli:
I gave it a second look to see whether I could slim the files down. Now they are pretty much are the simplest bare minimum that I can figure out, where one stills gets granular feedback in the UI.

One could smash the three different tests together even more, but I'm not sure whether it will become more readable or worse + you likely/easily lose some of the feedback in the UI.

One thing though: I do know that `Other` as a label is pretty useless, but it's all my brain can come up with right now 😝

fisker:
@voxpelli I  wrote one few month ago https://github.com/fisker/eslint-plugin-unicorn/blob/github-workflow/.github/workflows/continuous-integration.yml

sindresorhus:
Use `actions/checkout@v2`.

voxpelli:
@sindresorhus Done 👍 

@fisker Nice, my only feedback on that one is that it doesn't give granular feedback on whether it's the linting/integration that fails or something else – just like the Travis tests ;) Any takeaways from when you made that one?

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+
+on: [push]
+
+jobs:
+  test:
+    name: Node ${{ matrix.node_version }} and ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node_version: [8, 10, 12]
```
sindresorhus: Put them on separate lines.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+        os: [ubuntu-latest, windows-latest]
```
sindresorhus: Put them on separate lines.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: Tests
+
+on: [push]
+
+jobs:
+  test:
+    name: Node ${{ matrix.node_version }} and ${{ matrix.os }}
```
sindresorhus:
```suggestion
    name: Node.js ${{ matrix.node_version }} on ${{ matrix.os }}
```

package.json
```diff
@@ -15,7 +15,8 @@
 	},
 	"scripts": {
 		"test": "xo && nyc ava",
-		"lint": "node ./test/lint/lint.js",
+		"test-coverage": "nyc --reporter=lcov ava",
+		"lint": "node ./test/lint/lint.js && xo",
```
sindresorhus: XO is now run twice, both on `test` and `lint`.

sindresorhus:
Alright, let's do this. Can you drop the Travis file and integrate the badges?

package.json
voxpelli: @sindresorhus Yes and no. In the GitHub CI pipe `test-coverage` is run rather than `test`. Partly to get coverage report and partly to move the `xo` part from being run on every Node.js version to being run just once as part of the linting flow.

So, `test.yml` runs `test-coverage` while `other.yml` (suggestion on better name?) runs `lint` and `integration`

I thought it was best to leave the `test` as is, as that's likely the one that gets run manually the most.

voxpelli:
> Alright, let's do this. Can you drop the Travis file and integrate the badges?

Done 👍 

> We should use songle yml for all jobs

I don't see the benefit of mashing them all together in one file.

Now there's at least just two files: One for those tests that don't care about which node.js version they run on and another that tests across all supported node.js versions. So all tests that are similar to one another are in the same file and those that don't really share any similarity, they are kept in separate files.

fisker:
@voxpelli use one file we can use one badge, and they are not that much

voxpelli:
> @voxpelli use one file we can use one badge, and they are not that much

@fisker I don't think having two badges are a bug but a feature. It's to me two very different things whether linting fails or whether tests fails. Tests verifies functionality while linting verifies code quality. Getting good indicators around which one is failing is a benefit on top of Travis CI I think. (Though I do recognize that I haven't contributed a good name for the `Other` badge and that coupling integration and linting in the same file kind of defeats the purpose of the separation, hence why I had three files initially)

@sindresorhus I did a rebase + squash as well as removed Node 8 to match `master` + added Node.js 13 as a separate commit. Any reason why Node.js 13 isn't included right now?

.github/workflows/test.yml
```diff
@@ -0,0 +1,32 @@
...
+      - run: npm install
+      - run: npm run test-coverage
+      - uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.github_token }}
+          parallel: true
+      - uses: coverallsapp/github-action@master
+        with:
+          github-token: ${{ secrets.github_token }}
+          parallel-finished: true
```
fisker: One my experience, this not right, seem need  another job wait for matrix, not sure. I tried in this way before.

So I dont use  parallel https://github.com/fisker/id-placeholder/blob/ae506a0069bb459f6100f2c317f377ba9f8651de/.github/workflows/continuous-integration.yml#L52

Also, this action some times fail for no reason, issue can found on their repo, so I use continue-on-error

.github/workflows/test.yml
```diff
@@ -0,0 +1,32 @@
+name: Tests
+
+on: [push]
```
fisker: pull_request missing ?

.github/workflows/test.yml
voxpelli: The `parallel-finished` is what they document should be used when running multiple tests in parallel: https://github.com/coverallsapp/github-action#inputs

I added the `continue-on-error` on error now 👍 Good feedback!

.github/workflows/test.yml
voxpelli: Added, thanks 👍

fisker:
Like I said, I'm not sure how to make it work, I tried this way before, but seems not right

voxpelli:
Yeah, in the example it does like what I did: https://github.com/coverallsapp/github-action/blob/master/README.md#complete-parallel-job-example 🤔

sindresorhus:
I don't really see the value of two badges either. You have to click the badge regardless to see the details, and neither should be failing.

fisker:
But this example is not on matrix.

We can skip changing this part, just keep them.
If we  find any  problem later, we  can fix  then.

fisker:
One more thing eslint-plugin-ava  already did  this

https://github.com/avajs/eslint-plugin-ava/blob/master/.github/workflows/ci.yml

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  - push
+  - pull_request
+
+jobs:
+  test:
+    name: Node.js ${{ matrix.node_version }} on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
```
fisker: fail-fast: false is needed here

fisker:
@voxpelli found a project use parallel, check the commit history of this file, guess I am right

https://github.com/lo1tuma/eslint-plugin-mocha/blob/e40722d1691f087550a1146b781bcf25631f74c8/.github/workflows/ci.yml

package.json
```diff
@@ -92,6 +92,9 @@
 					"eslint-plugin/require-meta-docs-url": "off"
 				}
 			}
-		]
+		],
+		"rules": {
+			"import/order": "off"
+		}
```
fisker: This is really weird,  see

https://github.com/voxpelli/eslint-plugin-unicorn/runs/437762645?check_suite_focus=true

fisker:
[CI](https://github.com/voxpelli/eslint-plugin-unicorn/actions/runs/37384513)

@voxpelli I did some adjustment.

@sindresorhus You may want take a look.

package.json
fisker: https://github.com/benmosher/eslint-plugin-import/pull/1651

.github/workflows/ci.yml
```diff
@@ -0,0 +1,62 @@
...
+
+  coverage:
+    needs: test
+    runs-on: ubuntu-latest
+    steps:
+        - uses: coverallsapp/github-action@master
+          continue-on-error: true
+          with:
+              github-token: ${{ secrets.github_token }}
+              parallel-finished: true
```
sindresorhus: Why cannot this just be part of the `test` job? I don't see what it does differently from the code in the `test` job either.

package.json
```diff
@@ -14,7 +14,7 @@
 		"node": ">=10"
 	},
 	"scripts": {
-		"test": "xo && nyc ava",
+		"test": "xo && nyc --reporter=lcov ava",
```
sindresorhus: Can you specify this as `nyc` config (in package.json) instead of a CLI argument?

.github/workflows/ci.yml
fisker: This job send a webhook to coveralls, need wait for all test job finished.

See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/504#discussion_r373785339 and https://github.com/sindresorhus/eslint-plugin-unicorn/pull/504#issuecomment-581056148

sindresorhus: Why though? We only need coveralls to run in one of the test job, not all.

fisker: I removed `parallel`, and add a condition. only run on node.js 12 on linux

fisker:
@sindresorhus Do you want a nice title on each task? If not, I think we are good to go

readme.md
```diff
@@ -1,4 +1,4 @@
-# eslint-plugin-unicorn [![Build Status](https://travis-ci.org/sindresorhus/eslint-plugin-unicorn.svg?branch=master)](https://travis-ci.org/sindresorhus/eslint-plugin-unicorn) [![Coverage Status](https://coveralls.io/repos/github/sindresorhus/eslint-plugin-unicorn/badge.svg?branch=master)](https://coveralls.io/github/sindresorhus/eslint-plugin-unicorn?branch=master)
+# eslint-plugin-unicorn ![Build Status](https://github.com/sindresorhus/eslint-plugin-unicorn/workflows/CI/badge.svg) [![Coverage Status](https://coveralls.io/repos/github/sindresorhus/eslint-plugin-unicorn/badge.svg?branch=master)](https://coveralls.io/github/sindresorhus/eslint-plugin-unicorn?branch=master)
```
sindresorhus: Shouldn't the badge be linkified?

sindresorhus:
> Do you want a nice title on each task?

Nah

fisker:
Thanks @voxpelli

fisker:
Looks good

CI: https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/37731558
Coverage: https://coveralls.io/github/sindresorhus/eslint-plugin-unicorn?branch=master

PR #507, opened 2020-02-02
Expand reporting for `prefer-node-remove`

### LINKED ISSUE #506: `prefer-node-remove`: Remove requirement for `removeChild` to be on `parentElement` or `parentNode`
There is a requirement specific to `prefer-node-remove` which expects the parent to be named "parentNode" or "parentElement".

There is no such requirement in `prefer-modern-dom-apis` with `replaceChild` or `insertBefore`, and I think there need be no such requirement.

brettz9:
- `prefer-node-remove`: Expand reporting to cover cases where parent of `removeChild` is other than `parentNode` or `parentElement`.

Fixes #506.


fisker:
Are you willing to do this? 

https://github.com/sindresorhus/eslint-plugin-unicorn/issues/506#issuecomment-581108925

Its Okay if you have no interest.

rules/prefer-node-remove.js
```diff
@@ -59,10 +59,7 @@ const create = context => {
 			const methodName = getMethodName(callee);
 			const callerName = getCallerName(callee);
 
-			if (methodName === 'removeChild' && (
-				callerName === 'parentNode' ||
-				callerName === 'parentElement'
-			)) {
+			if (methodName === 'removeChild') {
 				const argumentName = getArgumentName(node.arguments);
```
fisker: Use selector like https://github.com/sindresorhus/eslint-plugin-unicorn/blob/3e958d0297c8f93cd3e50e22da4f22efc2643fe7/rules/prefer-replace-all.js#L32

brettz9: Done.

rules/prefer-node-remove.js
```diff
@@ -46,31 +36,28 @@ const getArgumentName = arguments_ => {
...
 const create = context => {
 	return {
-		CallExpression(node) {
+		'CallExpression[callee.property.name="removeChild"]'(node) {
 			const {callee} = node;
 
-			if (node.arguments.length === 0 ||
+			if (node.arguments.length !== 1 ||
 				callee.type !== 'MemberExpression' ||
 				callee.computed
```
fisker: This line no need anymore, I think

fisker: I mean `callee.computed`

test/prefer-node-remove.js
```diff
@@ -28,7 +28,8 @@ ruleTester.run('prefer-node-remove', rule, {
 		'foo.parentNode.removeChild(\'bar\')',
 		'foo.parentNode[\'bar\'](foo)',
 		'foo.parentNode[removeChild](foo)',
-		'foo.parentNode.removeChild()'
+		'foo.parentNode.removeChild()',
+		'foo.parentNode.removeChild(bar, baz)'
```
fisker: Add `'foo['removeChild'](bar)', to confirm

rules/prefer-node-remove.js
fisker: Add 'foo['removeChild'](bar)' test, to confirm

fisker:
@brettz9 Let's focus on `prefer-node-remove`, do the selector thing on other rule in another PR, shall we?

rules/prefer-node-remove.js
```diff
@@ -46,34 +36,28 @@ const getArgumentName = arguments_ => {
 
 const create = context => {
 	return {
-		CallExpression(node) {
+		'CallExpression[callee.property.name="removeChild"]'(node) {
 			const {callee} = node;
 
-			if (node.arguments.length === 0 ||
+			if (node.arguments.length !== 1 ||
 				callee.type !== 'MemberExpression' ||
```
fisker: Let's add this to selector too

rules/prefer-node-remove.js
```diff
@@ -46,34 +36,28 @@ const getArgumentName = arguments_ => {
...
-					context.report({
-						node,
-						message: `Prefer \`${argumentName}.remove()\` over \`${callerName}.removeChild(${argumentName})\`.`,
-						fix
-					});
-				}
+			const argumentName = getArgumentName(node.arguments);
+
+			if (argumentName) {
+				const fix = isValueNotUsable(node) ? fixer => fixer.replaceText(node, `${argumentName}.remove()`) : undefined;
```
fisker: I'm not sure, is there a `replaceNode` method? so we can 

1. replace `callee` with `arguments[0]`
2. replace callee.property with `remove`
3. remove `arguments[0]`

If we can do this, we can keep all comments, like

```js
/*comment1*/
foo
/*comment2*/
.removeChild
/*comment3*/
(
/*comment4*/
bar
/*comment5*/
)
```

fisker: I checked the docs. seems not. this is not urgent.

rules/prefer-node-append.js
```diff
@@ -2,14 +2,16 @@
...
-		CallExpression(node) {
+		'CallExpression[callee.property.name="appendChild"]'(node) {
 			const {callee} = node;
 
-			if (callee.type === 'MemberExpression' && getMethodName(callee) === 'appendChild') {
+			if (node.arguments.length !== 1) {
+				return;
+			}
+
+			if (callee.type === 'MemberExpression') {
```
fisker: Wait, is this always true? when `[callee.property.name="appendChild"]`

rules/prefer-node-remove.js
fisker: Wait, is this always true? when [callee.property.name="removeChild"]

fisker: I've send #509 to fix `prefer-node-append`

brettz9: I can look into these changes as I have time, but if there is refactoring into a common utility (e.g., for setting the selector), it would seem we should look at all affected functions together.

fisker:
## @brettz9 already did:

> * `prefer-node-remove`: Expand reporting to cover cases where parent of `removeChild` is other than `parentNode` or `parentElement`.

## I've done:

- replace error message with a general message `` Prefer `childNode.remove()` over `parentNode.removeChild(childNode)`. ``, old one is more specific, but could be `` Prefer `foo.remove()` over `null.removeChild(foo)`. `` in some case
- since we are not checking `parentNode`/`parentElement` anymore, I removed most of them from tests
- simplify tests that can't be fixed.
- exclude `parentNode.removeChild(undefined)` -> `undefined.remove()`

## Need to to (will raise another issue after this merge):

- allow more `childNode`, not only `Identifier` and `ThisExpression`, what I most want support is:

`parentNode.removeChild(some.node)` / `parentNode.removeChild(get.child())`

- investigate what should we do with complicated `callee`

eg:

```js
doSomething1().doSomething2().doSomething3().doSomething4().getParent()
	.removeChild(childNode)
```

we can't just fix to `childNode.remove()`.

/cc @brettz9 @sindresorhus

PR #508, opened 2020-02-02
Enable `prettier` on codebase

fisker:
Prettier sometimes makes codes more readable.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,9 +13,7 @@ module.exports = {
 				ecmaVersion: 2020,
 				sourceType: 'module'
 			},
-			plugins: [
-				'unicorn'
-			],
+			plugins: ['unicorn'],
 			rules: {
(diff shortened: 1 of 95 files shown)


sindresorhus:
I'm not a big fan of Prettier. It does make some code look nicer, but very often it also makes weird choices that makes the code worse. It is just too focused on keeping the line width. I also have a special dislike for it as I waste so much time on PRs where people run Prettier on my code and I have to tell them to undo it.

fisker:
I'm fine with your choice

sindresorhus:
I ran Prettier locally and cherry-picked the changes I agree with: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/8fd4ddac77558b37520886217b5c0fa17af182d2

fisker:
Thanks, looks good.

PR #509, opened 2020-02-02
Refactor `prefer-node-append` to use selector

fisker:
- Simplify logic
- Add `arguments.length` check
- Exclude `SpreadElement`
- Add `TODO` on type check

At last, I add a util function `methodSelector` to do all these checks, planning to add similar checks to `prefer-node-remove` `prefer-replace-all` `prefer-starts-ends-with` `prefer-trim-start-end`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "eslint-plugin-unicorn",
-	"version": "15.0.1",
+	"version": "16.0.0",
 	"description": "Various awesome ESLint rules",
 	"license": "MIT",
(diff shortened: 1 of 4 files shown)


brettz9:
Btw, @fisker , FWIW, inspired in part by your info on selectors, I've created https://github.com/brettz9/eslint-plugin-query for allowing queries against arbitrary selectors. (Though it is usable for expressing as "rules", it may be more useful for queries as per the "Tips" section.)

Just mention in case it may help during development. (Feel free to file an issue to discuss further.)

PR #514, opened 2020-02-07
Add `prefer-ternary` rule

### LINKED ISSUE #486: Rule proposal: `prefer-ternary`
This rule would enforce using ternary expressions to simple if/then/else statements.

# Fail
```js
if(condition){
  a = 1;
}
else{
  a = 2;
}
```
# Pass
```js
a = condition ? 1 : 2
```
```js
if(condition){
  doSomeStuff()
  a = 1;
}
else{
  a = 2;
}
```

This rule could have options for whether it would be enabled for various situations (e.g. variable assignments, return statements, function calls).

I've already written this rule as a normal ESLint plugin so it would just need some minor rewriting to conform to the unicorn linting/documentation

jmoore914:
Fixes #486


docs/rules/prefer-ternary.md
```diff
@@ -0,0 +1,168 @@
+# Prefer ternary expressions over simple `if/then/else` statements
```
fisker: `if-else`?

index.js
```diff
@@ -57,6 +57,7 @@ module.exports = {
 				'unicorn/prefer-spread': 'error',
 				'unicorn/prefer-starts-ends-with': 'error',
 				'unicorn/prefer-string-slice': 'error',
+				'unicorn/prefer-ternary': 'off',
```
fisker: why set to `off`

sindresorhus: It should be on by default.

sindresorhus:
I don't really see the point of all these options to be honest.

test/prefer-ternary.js
```diff
@@ -0,0 +1,286 @@
...
+
+		{
+			code: 'function foo(){if(a){return 1;}else{return 2;}}',
+			output: 'function foo(){return a ? 1 : 2}',
+			errors: [
+				{column: 16, line: 1, type: 'IfStatement'}
+			]
+		},
+		{
+			code: 'function foo(){if(a){return 1;}else{return 2;}}',
```
sindresorhus: This is not very readable. Pretty much all the fixtures needs to be made more readable and should use the same code style as the rest of the project.

docs/rules/prefer-ternary.md
```diff
@@ -0,0 +1,235 @@
...
+### throw
+The throw option determines whether the rule will flag throw statements. It can take a boolean. Default value is true.
+With `{thow: false}` the following would NOT be flagged:
+```js
+if(bar){
+	throw Error(123)
+}
+else{
+	throw Error(456)
+}
```
sindresorhus: This should not be a ternary even with an option.

docs/rules/prefer-ternary.md
```diff
@@ -0,0 +1,235 @@
...
+### call
+The call option determines whether the rule will flag call expressions. It can take a boolean. Default value is false.
+With `{call: true}` the following would be flagged:
+```js
+if(bar){
+	foo()
+}
+else{
+	baz()
+}
```
sindresorhus: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514/files#r380260375. It's a bad practice to put void function calls in a ternary. Only function calls that return a value should be in a ternary.

sindresorhus:
It's simple, if all the conditional branches assign to the same thing, it can be made a ternary, if not, we don't do anything. No options needed.

docs/rules/prefer-ternary.md
fisker: Better example should be 

```js
if(bar){
	throw new Error('foo')
}
else{
	throw new Error('bar')
}
``

File: test/prefer-ternary.js [hunk 6]
```diff
@@ -0,0 +1,286 @@
...
+				} 
+				else{
+					throw Error(456);
+				}`,
+			options: [{throw: true}],
+			output: 'throw a ? Error(123) : Error(456)',
+			errors: [
+				{column: 1, line: 1, type: 'IfStatement'}
+			]
+		},
```
fisker: We need test 3 kinds of throw

```js
throw new Error('foo');
throw getFooError();
throw fooError;
```

docs/rules/prefer-ternary.md
jmoore914: Sorry, I'm not fully understanding. Are the two of you disagreeing that this should be a use case or are you agreeing that it is just a bad example?

jmoore914:
> 
> 
> Basically; If all the conditional branches assign to the same thing, it can be made a ternary, if not, we don't do anything. No options needed.

Yep, that's what my thinking was as well. I would add return as well though or do you disagree? You could say that return would be assigning to the same value which is the function itself.

docs/rules/prefer-ternary.md
sindresorhus: I'm saying that `throw` should never be made into a ternary. @fisker is commenting about the example itself as you're throwing a number, which is not valid, and the code style is off.

sindresorhus:
If all the branches return, then yes, that would be possible too.

fisker:
As I understand, we can check the two branches has same node type.

Same `AssignmentExpression`:

```js
if (foo) {
	bar = 1;
} else {
	bar = 2;
}
```

Same `ReturnStatement`:

```js
if (foo) {
	return 1;
} else {
	return 2;
}
```

Same `ThrowStatement`:

```js
if (foo) {
	return new Error("1");
} else {
	return new Error("2");
}
```

and same `YieldExpression` or same `AwaitExpression`

---

No need care what `1` `2` is. Even

```js
if (foo) {
	return function() {};
} else {
	return a.b.c.d.e.f();
}
```

If you want you can try merge the same part of `1` and `2`, for example:

```js
if (foo) {
	return a(1, 2, 3, 4);
} else {
	return a(1, 2, 3, 5);
}
```

You can simply fix to `return foo ? a(1, 2, 3, 4) : a(1, 2, 3, 5)` or you can fix to `return a(1, 2, 3, foo ? 4 : 5)`, if you want compare this two nodes. I think both acceptable.

fisker:
Another thing, I suggest alway add `()` to the condition, `no-extra-parens` rule can remove it if it's not necessary, this can avoid strange stuff

```js
if (a = foo()) {return A} {return B} 
// return a = foo() ? A : B
// return (a = foo()) ? A : B
```

jmoore914:
@fisker Can you give me an example of an Await statement that would trigger the rule? I am not seeing how it is any different from a normal function call which we do not want included in the rule.

fisker:
```js

if (foo) {
	await promise1;
} else {
	await promise2;
}
```

jmoore914:
I'm not sure whether I would agree that a ternary is preferable in that case. @sindresorhus , do you have any thoughts regarding await?

sindresorhus:
Ideally, it should be:

```js
const something = foo ? promise1 : promise2;
await something;
```

But I think `await (foo ? promise1 : promise2)` is fine too.

fisker:
Please add tests on

```js
if (foo) {
	bar = a ? b : c;
} else {
	bar = 2;
}
```

exclude those assignment/return/yeild that already a ternary

rules/prefer-ternary.js
```diff
@@ -0,0 +1,85 @@
...
+			(consequentType === 'ReturnStatement' ||
+				(consequentType === 'ExpressionStatement' && checkConsequentAndAlternateExpressionStatement(node))));
+	}
+
+	function checkConsequentAndAlternateExpressionStatement(node) {
+		const consequentType = node.consequent.body[0].expression.type;
+		return consequentType === node.alternate.body[0].expression.type &&
+			(consequentType === 'AssignmentExpression' ? compareConsequentAndAlternateAssignments(node) :
+				consequentType === 'YieldExpression');
+	}
```
fisker: This part is not good, you can compare type first, something like

```js
if (isSameType(node)) {
  // Handle `AssignmentExpression`
  // ...
  // Handle `ReturnStatement`
  // ...
  // Handle `ThrowStatement`
  // ...
  // Handle `ExpressionStatement > AwaitExpression`
  // ...
  // ...
}
````

File: rules/prefer-ternary.js [hunk 8]
```diff
@@ -0,0 +1,85 @@
...
+		const ifCondition = sourceCode.getText(node.test);
+		let left = '';
+		let right = '';
+		if (node.consequent.body[0].type === 'ExpressionStatement') {
+			if (node.consequent.body[0].expression.type === 'AssignmentExpression') {
+				prefix = sourceCode.getText(node.consequent.body[0].expression.left) + ' = ';
+				left = sourceCode.getText(node.consequent.body[0].expression.right);
+				right = sourceCode.getText(node.alternate.body[0].expression.right);
+			} else {
+				prefix = 'yield ';
```
fisker: Don't forget `yield*`

fisker:
Braces can also omitted, so we need  test

```js
if (foo)
	bar = 1;
else
	bar = 2;
```

fisker:
@sindresorhus Do you think we should report these two cases?

```js
if (a) {
	return 1;
} else if (b) {
	return 2;
} else {
	return 3;
}
```

```js
if (a) {
	return 1;
} else {
	if (b) {
		return 2;
	} else {
		return 3;
	}
}
```

sindresorhus:
1. Yes, but only if it's not too complicated, and only one else-if.
2. Won't this be caught by `1` on the second fixer pass? First it fixes the inner `if`, and on the second pass, the outer `if`.
3. Same as `2`.

fisker:
Just confirm with you, we  need test those case

jmoore914:
@fisker I added tests for those cases. 

Integration tests are failing for me in the master branch btw

fisker:
@jmoore914 you need fix this https://github.com/facebook/react/blob/7a1691cdff209249b49a4472ba87b542980a5f71/packages/react-dom/src/client/DOMPropertyOperations.js#L183

and `throw` and `yeild*` not handled yet.

jmoore914:
> 
> 
> @jmoore914 you need fix this https://github.com/facebook/react/blob/7a1691cdff209249b49a4472ba87b542980a5f71/packages/react-dom/src/client/DOMPropertyOperations.js#L183
> 
> and `yeild*` not handled yet.

I added support for yield*

I'm confused about how to fix that integration test since the (value: any) isn't valid js

fisker:
@jmoore914 I've fixed integration test for you.

rules/prefer-ternary.js
```diff
@@ -0,0 +1,123 @@
...
+			if (expressionType === 'AssignmentExpression') {
+				prefix = sourceCode.getText(getNodeBody(node.consequent).expression.left) + ' = ';
+				left = getNodeBody(node.consequent).expression.right;
+				right = getNodeBody(node.alternate).expression.right;
+			} else {
+				if (expressionType === 'AwaitExpression') {
+					prefix = 'await ';
+				} else {
+					prefix = getNodeBody(node.consequent).expression.delegate ? 'yield* ' : 'yield ';
+				}
```
fisker: I don't think logic is not right.

We need check `YieldExpression`.

Why it has to be `await` and `yield`? There is other cases.

rules/prefer-ternary.js
```diff
@@ -0,0 +1,123 @@
...
+		const ifCondition = sourceCode.getText(node.test);
+		let left = '';
+		let right = '';
+		if (getNodeBody(node.consequent).type === 'ExpressionStatement') {
+			const expressionType = getNodeBody(node.consequent).expression.type;
+			if (expressionType === 'AssignmentExpression') {
+				prefix = sourceCode.getText(getNodeBody(node.consequent).expression.left) + ' = ';
+				left = getNodeBody(node.consequent).expression.right;
+				right = getNodeBody(node.alternate).expression.right;
+			} else {
```
fisker: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514/files#r392754687

rules/prefer-ternary.js
```diff
@@ -0,0 +1,123 @@
...
+				if (expressionType === 'AwaitExpression') {
+					prefix = 'await ';
+				} else {
+					prefix = getNodeBody(node.consequent).expression.delegate ? 'yield* ' : 'yield ';
+				}
+
+				left = getNodeBody(node.consequent).expression.argument;
+				right = getNodeBody(node.alternate).expression.argument;
+			}
+		} else {
```
fisker: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514/files#r392754687

Need check `ReturnStatement`

rules/prefer-ternary.js
jmoore914: Since this is in the fixFunction, it would not reach that line unless it was await or yield

rules/prefer-ternary.js
jmoore914: Same as above, it wouldn't reach this line unless it was the same type

rules/prefer-ternary.js
jmoore914: Line 43 is the return statement. And as above, it needs to be a return statement because of previous checks

fisker:
@jmoore914 Mind I take this PR over? I found merging two blocks is more complicate than I thought.

fisker:
@sindresorhus

I committed some changes, and some questions:

---

I missed [this comment](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514/files#r380285509), I've already fix

```js
if (test) {
	throw new Error('foo');
} else {
	throw new Error('bar');
}
```

to 

```js
throw test ? new Error('foo') : new Error('bar');
```

Should I remove it?

---

```js
function unicorn() {
  if (test) {
    return;
  } else {
    return 2;
  }
}
```

is not fixable now, because first `ReturnStatement` has no `argument`, should I fix it to

```js
function unicorn() {
  return test ? undefined : 2;
}
```

? `yield` the same.

---

```js
function foo(){
  if (a) {
    return 1;
  } else if (b) {
    return 2;
  } else {
    return 3;
  }
}
```

is fixed to 

```js
function foo(){
  if (a) {
    return 1;
  } else return b ? 2 : 3;
}
```

as discussed here https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514#issuecomment-594328616

But now I feel this case should is less readable, especially if there is many `if/else-if` we only fix the last else, seems strange to me.

Of course,

```js
if (a) {
	return 1;
} else {
	if (b) {
		return 2;
	} else {
		return 3;
	}
}
```

and

```js
if (a) {
	if (b) {
		return 1;
	} else {
		return 2;
	}
} else {
	return 3;
}
```

should be ternary, just the first case, I want to ignore it.


---

```js
function unicorn() {
  if(test){
    foo.bar = a;
  } else{
    foo.bar = b;
  }
}
```

This case should be fixable too, but this will be a little complicated, I suggest improve later.

Means, only assign to same variable cases are reported.


---

@jmoore914 Please please don't hate me.

jmoore914:
> 
> 
> @jmoore914 Mind I take this PR over? I found merging two blocks is more complicate than I thought.

Please do! Things got crazy for me and I haven't had the time to keep up with it

fisker:
Thanks, we are almost done, you can help review, if you got time.

jmoore914:
> 
> 
> @sindresorhus
> 
> I committed some changes, and here are some questions:
> 
> I missed [this comment](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514/files#r380285509), I've already fix
> 
> ```js
> if (test) {
> 	throw new Error('foo');
> } else {
> 	throw new Error('bar');
> }
> ```
> 
> to
> 
> ```js
> throw test ? new Error('foo') : new Error('bar');
> ```
> 
> Should I remove it?
> 

I personally would want that behind an option but I don't feel very strongly about it. Maybe check with@sindresorhus ?

> ```js
> function unicorn() {
>   if (test) {
>     return;
>   } else {
>     return 2;
>   }
> }
> ```
> 
> is not fixable now, because first `ReturnStatement` has no `argument`, should I fix it to
> 
> ```js
> function unicorn() {
>   return test ? undefined : 2;
> }
> ```
> 
> ? `yield` the same.
> 

Hmm, good thought. My feeling is that one branch returns a value and the other does not so the branches are not the same but that's just semantics and I wouldn't disagree with making that fix either.

> ```js
> function foo(){
>   if (a) {
>     return 1;
>   } else if (b) {
>     return 2;
>   } else {
>     return 3;
>   }
> }
> ```
> 
> is fixed to
> 
> ```js
> function foo(){
>   if (a) {
>     return 1;
>   } else return b ? 2 : 3;
> }
> ```
> 
> as discussed here [#514 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514#issuecomment-594328616)
> 
> But now I feel this case is less readable, especially if there is many `if/else-if` we only fix the last else, seems strange to me.
> 
> Of course,
> 
> ```js
> if (a) {
> 	return 1;
> } else {
> 	if (b) {
> 		return 2;
> 	} else {
> 		return 3;
> 	}
> }
> ```
> 
> and
> 
> ```js
> if (a) {
> 	if (b) {
> 		return 1;
> 	} else {
> 		return 2;
> 	}
> } else {
> 	return 3;
> }
> ```
> 
> should be ternary, just the first case, I want to ignore it.
> 

Agreed on all points. Maybe just ignore all nested for simplicity?

> ```js
> function unicorn() {
>   if(test){
>     foo.bar = a;
>   } else{
>     foo.bar = b;
>   }
> }
> ```
> 
> This case should be fixable too, but this will be a little complicated, I suggest improve later.
> 
> Means, only assign to same variable cases are reported.

Makes sense to me.
> 
> @jmoore914 Please please don't hate me.

Absolutely no hate! Much appreciation that you're working on this :)

fisker:
One note `AssignmentExpression.operator` not compared.

fisker:
> Maybe just ignore all nested for simplicity?

The first one is not hard to detect, add one limit to selector will do.

fisker:
@sindresorhus Can you take a look at [this](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514#issuecomment-6245332360) when you got time? Need your opinion to finish this rule.

sindresorhus:
> I missed this comment, I've already fix
> Should I remove it?

I find fixed code pretty ugly and not very readable.

Alternatively, we could fix it to:

```js
const error = test ? new Error('foo') : new Error('bar');
throw error;
```

?

---

> is not fixable now, because first ReturnStatement has no argument, should I fix it to

Yes

> ? yield the same.

:+1:

---

> But now I feel this case is less readable, especially if there is many if/else-if we only fix the last else, seems strange to me.

You're right. We can ignore that case.

---

> This case should be fixable too, but this will be a little complicated, I suggest improve later.

:+1: Open an issue so we don't forget.

test/integration/unicorn/error-name-conflicts.js
```diff
@@ -0,0 +1,12 @@
+function foo() {
+	try {
+	} catch (err) {
+		console.log(err);
+
+		if (test) {
+			throw a;
+		} else {
+			throw b;
+		}
```
fisker, line 10: We have a trouble here, the `catch-error-name` and `prevent-abbreviations` fix `err` to `error`, and this rule is creating `const error = test ? a : b;` cause `Parsing error: Identifier 'error' has already been declared`.

I have no idea how to fix this.

fisker: Actual code  https://github.com/eslint/eslint/blob/b76aef778befb32afe7ad249934b132dc49713d2/lib/shared/config-validator.js#L128

fisker: @futpib 

Glad to see you here, you are the one add `avoidCapture`(great util, BTW), do you have an idea how to fix this problem?

I was thinking use a global `Map` to manage the variables created by our rules, but I'm not sure if it will work. And this may still can't solve the real problem, if there is a new variable created by other rule from another plugin.

fisker:
A note to myself, the if block can be some node's body, which will broken by current fix, need investigate.

Example: 

```js
while (foo) if (test) {throw a} else {throw b}
```

Wrong:

```js
while (foo) const error = test ? a : b;
throw error;
```

Correct:

```js
while (foo) {
  const error = test ? a : b;
  throw error;
}
```

test/integration/unicorn/error-name-conflicts.js
futpib: Even if we make all rules in `eslint-plugin-unicorn` generate non-conflicting names, we can still hit name conflicts across eslint plugins, so I think this is a `eslint` issue.

Having a plugin-global map seems like an ok workaround. (It should probably be a `WeakMap<Scope, Set<string>>` like in `prevent-abbreviations` https://github.com/sindresorhus/eslint-plugin-unicorn/blob/16f6ef33c581782029bf33377f9084a0b2847fc1/rules/prevent-abbreviations.js#L552-L558))

sindresorhus: > we can still hit name conflicts across eslint plugins, so I think this is a eslint issue.

Can you open an issue on https://github.com/eslint/eslint ?

sindresorhus:
Bump

sindresorhus:
@jmoore914 Bump

test/integration/unicorn/error-name-conflicts.js
fisker: https://github.com/eslint/eslint/issues/13706

fisker:
Fixed https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514#discussion_r433762499 with [this suggestion](https://github.com/eslint/eslint/issues/13706#issuecomment-696667545).

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/514#issuecomment-637923335 also fixed.

Ready for review.

fisker:
Result on prettier codebase https://github.com/fisker/prettier/pull/807

PR #519, opened 2020-02-10
Ensure all the rules are compatible with TypeScript-ESLint

JounQin:


JounQin:
It should be an issue integrating with `@typescript/eslint-parser`

rules/no-hex-escape.js
```diff
@@ -10,7 +10,9 @@ function checkEscape(context, node, value) {
 		context.report({
 			node,
 			message: 'Use Unicode escapes instead of hexadecimal escapes.',
-			fix: fixer => fixer.replaceTextRange([node.start, node.end], fixedValue)
+			fix: fixer => typeof node.start === 'undefined' ?
+				fixer.replaceText(node, fixedValue) :
+				fixer.replaceTextRange([node.start, node.end], fixedValue)
```
sindresorhus: Can you drop the conditional? We can just switch completely to `fixer.replaceText`.

sindresorhus:
Would have been nice with an internal rule for us that ensures we don't use `node.start` and `node.end`. Those are deprecated ESLint specific properties and should not be used.

sindresorhus:
There are some other places they're used. Would you be up for converting those too?

```
rules/escape-case.js
55:				escapeNodePosition = [node.start, node.end];

rules/no-hex-escape.js
13:			fix: fixer => fixer.replaceTextRange([node.start, node.end], fixedValue)

rules/no-nested-ternary.js
11:		previousToken.end <= node.start &&

rules/no-nested-ternary.js
13:		nextToken.start >= node.end

rules/no-new-buffer.js
11:				node.start,
```

rules/no-hex-escape.js
JounQin: I don't know why but it throws if I switch completely.

fisker:
```js
 Node {
    type: 'TemplateElement',
    start: 13,
    end: 16,
    range: [ 12, 18 ]
  },
```

seem `TemplateElement` has different start end

fisker:
```suggestion
			fix: fixer => {
				let {range: [start, end], type, tail} = node;

				if (type === 'TemplateElement') {
					start += 1;
					end -= tail ? 1 : 2;
				}

				return fixer.replaceTextRange([start, end], fixedValue);
			}
```

fisker: @JounQin Try this, and add a test `` const foo = `\\xb1${foo}` `` , this covers `node.tail = false`

fisker: Not sure how it works with other parser, seems difference between them

https://github.com/babel/babel/issues/6681

JounQin: It seems much more complicated.

fisker: I think that the way `TemplateElement` works

fisker: see https://github.com/eslint/espree/blob/8a67b2f76f497fc71b11b0a71c3659d3cc0019b9/lib/espree.js#L329

JounQin: @fisker I mean it should have been covered by my original fix, isn't it?

fisker: Not sure, but maybe not . If other parser calculate `range` the eslint ast way, but without `start` this may replacing wrong range, right?

JounQin: OK, changed.

test/no-hex-escape.js
```diff
@@ -187,6 +192,13 @@ ruleTester.run('no-hex-escape', rule, {
 			code: 'const foo = `42\\\\\\x1242\\\\\\x34`',
 			errors: [error],
 			output: 'const foo = `42\\\\\\u001242\\\\\\u0034`'
+		},
+		{
+			// eslint-disable-next-line no-template-curly-in-string
+			code: 'const foo = `\\xb1${foo}`',
```
fisker: Let's test `\\xb1${foo}\\xb1${foo}` make sure this one in middle is also right

JounQin:
@fisker @sindresorhus Please review again.

rules/utils/normalize-range.js
```diff
@@ -0,0 +1,10 @@
+module.exports = node => {
+	let {range: [start, end] = [node.start, node.end], type, tail} = node;
+
+	if (type === 'TemplateElement') {
+		start += 1;
+		end -= tail ? 1 : 2;
+	}
+
+	return [start, end];
+};
```
fisker: I'm not sure we need this, this suppose to only work for case in `no-hex-escape` rule.

The range of `TemplateElement` is not wrong, we just don't want replace `` ` ``, `${` and `}`.

For example we have a `Literal` string, if we want replace string inside quotes, we may `replaceTextRange([start+1, end - 1], 'replacement')`, but if we want replace the whole string, we may do this `replaceText(range, QUOTED_STRING)`, [we do this a lot](https://github.com/sindresorhus/eslint-plugin-unicorn/search?q=quoteString&unscoped_q=quoteString).

fisker: Suggest move these to `no-hex-escape.js`, in other rules we just use `node.range`.

What do you think?

JounQin: Or should I add a second param for this util?

JounQin: > in other rules we just use node.range

`range` is not always available, that's why I add a default value for it: `[node.start, node.end].`

fisker: I don't think we need this. The string case in `no-hex-escape.js`, we are using `node.raw` https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/no-hex-escape.js#L21 . 

So quotes are kept, but if we use `node.value`, we can't say we need `normalize` the range, we just fix the range ourself( or add quotes like we already do in other rules).

fisker: Make sense?

JounQin: Then should `normalize-range` be kept? As @sindresorhus 's [suggestion](#issuecomment-584121177), we should not use `node.start/end` directly, but `node.range` is not available at https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/escape-case.js#L55

JounQin: It would be:

```js
module.exports = node => {
	const {range: [start, end] = [node.start, node.end], type, tail} = node;
	return [start, end];
};
```

fisker: > > in other rules we just use node.range
> 
> `range` is not always available, that's why I add a default value for it: `[node.start, node.end].`

Sorry, I didn't see this until now, can you explain more, or give me a example?

fisker: > It would be:
> 
> ```js
> module.exports = node => {
> 	const {range: [start, end] = [node.start, node.end]} = node;
> 	return [start, end];
> };
> ```

I'm going to take a look at this, thanks.

fisker: I had a quick look ,  this seems a different AST from `regexpp`

https://github.com/mysticatea/regexpp/blob/d1808145b2861738747bb2e742537d87527a6871/src/ast.ts#L60-L71

there is no range, so we should use `.start` and `.end` here, maybe add a comment there. AST in other cases all has `range`, right?

fisker: I suggest remove `normalize-range`, add `TemplateElement` logic to `no-hex-escape.js`, keep using `.start` and `.end` in AST from `regexpp`, and use `.range` directly in  other place.

JounQin: Do you mean keep `normalize-range.js` with:

```js
module.exports = node => {
    // there is no `range` property in AST from `regexpp`
    // reference: https://github.com/mysticatea/regexpp/blob/master/src/ast.ts#L60-L71
	const {range: [start, end] = [node.start, node.end]} = node;
	return [start, end];
};
```

JounQin: I'm aware we would meet other edge cases in the future, keeping `normalize-range` could help.

fisker: I mean remove `normalize-range.js`, 

put this comment in 

```js
// there is no `range` property in AST from `regexpp`
// reference: https://github.com/mysticatea/regexpp/blob/master/src/ast.ts#L60-L71
```
in https://github.com/JounQin/eslint-plugin-unicorn/blob/patch-1/rules/escape-case.js#L56 , and use `.start`/`.end` there.

Other place use `node.range` directly

fisker: > I'm afraid that we would meet other edge cases in the future, keeping `normalize-range` could help.

But if we keep this `normalize-range`, we have to use it whenever we need access `.range`, if there is other case we meet later, we can handle later.

JounQin: @fisker Fine enough.

sindresorhus:
> Would have been nice with an internal rule for us that ensures we don't use node.start and node.end. Those are deprecated ESLint specific properties and should not be used.

Oh yeah => https://eslint.org/blog/2020/02/whats-coming-in-eslint-7.0.0#stricter-ruletester

PR #522, opened 2020-02-11
Rename `regex-shorthand` to `better-regex`

### LINKED ISSUE #469: Rename `regex-shorthand` rule?
`regex-shorthand` do more than just `shorthand` now, should we rename it?
also [`no-unsafe-regex` is buggy](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/153) and already `off` by default, and I looked around, didn't found a better solution, maybe we can remove it, if we found a better way to do it, maybe we can add a option to `regex-shorthand` rule.

fisker:
Fixes #469


docs/rules/better-regex.md
```diff
@@ -0,0 +1,36 @@
+# Enforce the use of regex shorthands to improve readability
```
fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/469#issuecomment-584805859

PR #526, opened 2020-02-12
Improve `no-hex-escape` coverage

fisker:

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
 	},
 	"nyc": {
 		"reporter": [
+			"text",
 			"lcov"
 		]
 	},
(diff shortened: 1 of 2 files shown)


fisker:
There is no coverage reported on PR, I've tried to change the actions config file, but no luck.

sindresorhus:
https://github.com/coverallsapp/github-action/issues/7

fisker:
But i got the coverage result

image


https://github.com/fisker/get-text-nodes/pull/258

With same command

https://github.com/fisker/get-text-nodes/blob/master/.github/workflows/continuous-integration.yml#L52-L57

fisker:
Is there any setting on coveralls.io

PR #528, opened 2020-02-12
Update `escape-case` tests

fisker:
Categorize fixtures.

Improve Literal regex tests.(escaped `\` is not well tested)


sindresorhus:
Nice! I'm always excited about more tests. We can never have enough of those :)

PR #530, opened 2020-02-12
Use placeholder instead of generated error message

fisker:


sindresorhus:
Nice. I didn't know that was possible.

PR #531, opened 2020-02-12
Simplify `escape-case`

fisker:
I'm not sure if I'm missing anything, but seems we don't need parse `regex` at all.


fisker:
@JLHwung You are the one add `regex` parse in #294 , can you take a look?

JLHwung:
@fisker It's neat! Look good to me.

PR #532, opened 2020-02-12
Refactor regex in `escape-case` with lookbehind assertion

fisker:


sindresorhus:
Cool! I would recommend going through our other regexes and see if we can use lookbehind/lookahead anywhere.

fisker:
Hate read regexes, eyes going blind.

PR #534, opened 2020-02-12
Fix `escape-case` missing cases

fisker:
`'\uAaAa'` is missing , should report and fix to `'\uAAAA'`


sindresorhus:

PR #540, opened 2020-02-13
Add `isShadowed` check to `new-for-builtins`

### LINKED ISSUE #122: False positive for the `new-for-builtins` rule

fisker:
Fixes #122

I just steal this function from https://github.com/eslint/eslint/blob/1ee6b6388305a8671c8d4c3cf30c2dbf18a1ff7e/lib/rules/no-alert.js#L48

I've tried to solve this issue before, but failed, this function use range check instead of name check is great idea.

We may need add this check in more rules, I'll check later.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#122: False positive for the `new-for-builtins` rule](https://issuehunt.io/repos/55832243/issues/122)
---

</details>

### DIFF
--- a/test/new-for-builtins.js
+++ b/test/new-for-builtins.js
@@ -8,6 +8,11 @@ const ruleTester = avaRuleTester(test, {
 	},
 	parserOptions: {
 		sourceType: 'module'
+	},
+	// Make sure globals don't effect shadowed check result
+	globals: {
+		Map: 'off',
+		String: 'off'
 	}
 });
 
--- a/test/new-for-builtins.js
+++ b/test/new-for-builtins.js
@@ -114,10 +114,6 @@ ruleTester.run('new-for-builtins', rule, {
 				const foo = new ${object}();
 			}
 		`),
-		...disallowNew.map(object => `
-			const ${object} = function() {};
-			const foo = new ${object}();
-		`),
 		// #122
 		`
 			import { Map } from 'immutable';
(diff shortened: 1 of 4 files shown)


test/new-for-builtins.js
```diff
@@ -18,6 +26,41 @@ const disallowNewError = builtin => ({
...
+	'WeakMap',
+	'Set',
+	'WeakSet',
+	'Promise',
+	'RegExp',
+	'Uint8Array',
+	'Uint16Array',
+	'Uint32Array',
+	'Uint8ClampedArray'
+];
```
sindresorhus: These are just redefinitions of what's in the rule. I think it would be better to move these to util.js and import them both in the rule and here.

test/new-for-builtins.js
```diff
@@ -5,6 +5,14 @@ import rule from '../rules/new-for-builtins';
 const ruleTester = avaRuleTester(test, {
 	env: {
 		es6: true
```
sindresorhus: Let's start using `es2020` for these instead of `es6`.

PR #545, opened 2020-02-13
Clean up `prefer-exponentiation-operator`

fisker:
Should we remove those codes?


sindresorhus:
Yeah, we can just remove it all.

PR #546, opened 2020-02-13
Update `better-regex`

fisker:
- stricter selector
- clean codes
- Categorize fixtures
- More tests


fisker:
image

We got coverage information now, why?

sindresorhus:
Weird. I have no idea why.

PR #552, opened 2020-02-13
Handle currying arrow function in `consistent-function-scoping`

### LINKED ISSUE #391: Is this a `consistent-function-scoping` bug?
I was working on #387 , and found our code 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/92f2da10fd97841e8d674c17f397c39c24bf9fe2/test/integration/test.js#L57-L65

is reported by `consistent-function-scoping`.

This code looks good to me, is this a bug or did I miss something?

/cc @MrHen

fisker:
Fixes #391

### DIFF
--- a/test/consistent-function-scoping.js
+++ b/test/consistent-function-scoping.js
@@ -188,6 +188,14 @@ ruleTester.run('consistent-function-scoping', rule, {
 				}
 				return Bar;
 			};
+		`,
+		// #391
+		outdent`
+			const enrichErrors = (packageName, cliArgs, f) => async (...args) => {
+				try {
+					return await f(...args);
+				} catch (error) {}
+			};
 		`
 	],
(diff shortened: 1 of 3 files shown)


test/consistent-function-scoping.js
```diff
@@ -234,12 +250,6 @@ ruleTester.run('consistent-function-scoping', rule, {
 			`,
 			errors: [arrowError]
 		},
-		{
-			code: outdent`
-				const doFoo = () => bar => bar;
-			`,
-			errors: [arrowError]
-		},
```
fisker: I guess this test case mean to enforce use this style?

```
const movedToOuter = bar => bar;
const doFoo = () => movedToOuter;
```

seems make sense, I'm not doing it right.

PR #553, opened 2020-02-13
Check childScopes references in `consistent-function-scoping`

### LINKED ISSUE #372: False negative? for unicorn/consistent-function-scoping
I like this new rule and I've made change to my code. Here's one WIP snippet where it proved to be problematic though.

```ts
type Data<T> = T extends 'error' ? Error : Record<string, unknown> | unknown[]

type Method = 'info' | 'error'

export function createLogger(name: string) {
    // Two lint errors are on the next line.
    const log = <T extends Method>(method: T) => (data: Data<T>) => {
        try {
            // eslint-disable-next-line no-console
            console[method](JSON.stringify({ name, data }))
        } catch (error) {
            console.error(error)
        }
    }

    return {
        info: log('info'),
        error: log('error'),
    }
}
```

```
6:17  error  Move arrow function to the outer scope  unicorn/consistent-function-scoping
6:50  error  Move arrow function to the outer scope  unicorn/consistent-function-scoping
```

Could this be a bug in `@typescript-eslint`?

@typescript-eslint/eslint-plugin@2.3.0
@typescript-eslint/parser@2.3.0
eslint-plugin-unicorn@11.0.0

### LINKED ISSUE #391: Is this a `consistent-function-scoping` bug?

fisker:
The child scope references is not considered.

Fixes #391 
Fixes #372

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -4,10 +4,9 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const MESSAGE_ID_ARROW = 'ArrowFunctionExpression';
 const MESSAGE_ID_FUNCTION = 'FunctionDeclaration';
 
-function getReferences(scope) {
-	const {references, childScopes} = scope;
-	return references.concat(...childScopes.map(getReferences));
-}
+const getReferences = scope => scope.references.concat(
+	...scope.childScopes.map(scope => getReferences(scope))
+);
 
 function checkReferences(scope, parent, scopeManager) {
 	if (!scope) {
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -6,13 +6,14 @@ const MESSAGE_ID_FUNCTION = 'FunctionDeclaration';
 
 function getReferences(scope) {
 	const {references, childScopes} = scope;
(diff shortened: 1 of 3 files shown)


rules/consistent-function-scoping.js
```diff
@@ -4,12 +4,17 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const MESSAGE_ID_ARROW = 'ArrowFunctionExpression';
 const MESSAGE_ID_FUNCTION = 'FunctionDeclaration';
 
+function getReferences(scope) {
+	const {references, childScopes} = scope;
+	return references.concat(...childScopes.map(getReferences));
```
fisker: Is this code against `no-fn-reference-in-iterator`? Why not caught?

sindresorhus: Yes, it should have been caught. No idea why it wasn't. Can you open an issue for it?

PR #554, opened 2020-02-13
Refactor tests

fisker:


sindresorhus:
This is a good idea. Now we can easily make all rules test against both the ESLint parser, Babel-Eslint, and TypeScript-ESLint.

fisker:
Finally make it work on most rules, some rule can't test with the new logic, ava will throw error on duplicated title, not sure how they are different, will investigate later.

The windows failure should not related.

sindresorhus:
Is this ready?

sindresorhus:
> The windows failure should not related.

What's causing the Windows failures?

fisker:
I don't know why fail on windows, says timeout, but  I don't know how to increase timeout.

sindresorhus:
It's just strange that it fails now and not in master.

fisker:
And it passed on my local windows.

fisker:
I tried again, `npx ava` pass, `yarn test` fails.

fisker:
It seems `npx ava` pass, but `npx nyc ava` fails

sindresorhus:
Maybe just make it `npx nyc ava` => `npx ava` for now until we can figure out the issue?

fisker:
Good to go.

sindresorhus:
Great! 👌

PR #555, opened 2020-02-14
Add `ignore` option to `expiring-todo-comment` rule

### LINKED ISSUE #502: `expiring-todo-comments`: add regex option
I would love to be able to allow certain regexes.

### Use Case

I want to allow TODOs that reference a JIRA or GitHub issue id.

Eg.

```js
// TODO [JIRA-ISSUE-123] fix this
// TODO [#999] fix that
```

Could this be something useful to be included?

Moeriki:
Fixes #502.

### DIFF
--- a/rules/expiring-todo-comments.js
+++ b/rules/expiring-todo-comments.js
@@ -207,10 +207,15 @@ function semverComparisonForOperator(operator) {
 const create = context => {
 	const options = {
 		terms: ['todo', 'fixme', 'xxx'],
+		ignore: [],
 		ignoreDatesOnPullRequests: true,
 		allowWarningComments: true,
 		...context.options[0]
 	};
+
+	const ignoreRegexes = options.ignore.map(
+		pattern => pattern instanceof RegExp ? pattern : new RegExp(pattern, 'u')
+	);
 
 	const sourceCode = context.getSourceCode();
 	const comments = sourceCode.getAllComments();
@@ -250,6 +255,10 @@ const create = context => {
 	const rules = baseRule.create(fakeContext);
 
 	function processComment(comment) {
+		if (ignoreRegexes.some(ignore => ignore.test(comment.value))) {
+			return;
+		}
+
 		const parsed = parseTodoWithArguments(comment.value, options);
 
 		if (!parsed) {
@@ -479,6 +488,10 @@ const schema = [
 				items: {
 					type: 'string'
 				}
+			},
+			ignore: {
+				type: 'array',
+				uniqueItems: true
 			},
 			ignoreDatesOnPullRequests: {
 				type: 'boolean',
--- a/test/expiring-todo-comments.js
+++ b/test/expiring-todo-comments.js
@@ -110,6 +110,22 @@ ruleTester.run('expiring-todo-comments', rule, {
 		{
 			code: '// TODO [but [it will]] [fallback] [[[ to the default ]]] rule [[',
 			errors: []
+		},
+		{
+			code: '// TODO ISSUE-123 fix later',
+			options: [{allowWarningComments: false, ignore: ['ISSUE-\\d+']}]
+		},
+		{
+			code: '// TODO [ISSUE-123] fix later',
+			options: [{allowWarningComments: false, ignore: ['ISSUE-\\d+']}]
+		},
+		{
+			code: '// TODO [1999-01-01, ISSUE-123] fix later',
+			options: [{allowWarningComments: false, ignore: ['ISSUE-\\d+']}]
+		},
+		{
+			code: '// TODO [Issue-123] fix later',
+			options: [{allowWarningComments: false, ignore: [/issue-\d+/i]}]
 		}
 	],
 	invalid: [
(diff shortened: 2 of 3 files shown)


rules/expiring-todo-comments.js
```diff
@@ -207,10 +207,12 @@ function semverComparisonForOperator(operator) {
 const create = context => {
 	const options = {
 		terms: ['todo', 'fixme', 'xxx'],
+		ignore: [],
 		ignoreDatesOnPullRequests: true,
 		allowWarningComments: true,
 		...context.options[0]
 	};
+	const ignoreRegexes = options.ignore.map(ignore => new RegExp(ignore));
```
fisker: Would be nice to allow regex literal here, people may want do `/ignore/i`

fisker: check this https://github.com/sindresorhus/eslint-plugin-unicorn/pull/431

Moeriki: Resolved.

Mimicking the linked PR, I added the `'u'` flag by default, and set array in schema to `uniqueItems: true`.

test/expiring-todo-comments.js
```diff
@@ -110,6 +110,22 @@ ruleTester.run('expiring-todo-comments', rule, {
 		{
 			code: '// TODO [but [it will]] [fallback] [[[ to the default ]]] rule [[',
 			errors: []
+		},
+		{
+			code: '// TODO ISSUE-123 fix later',
```
fisker: I thought those comments requires `:` inside? Can you remove options  make sure  they  are  reported , then add options back.

rules/expiring-todo-comments.js
fisker: Can you  also use  some explanation and  example from  that PR?

test/expiring-todo-comments.js
```diff
@@ -110,6 +110,22 @@ ruleTester.run('expiring-todo-comments', rule, {
...
+		{
+			code: '// TODO [ISSUE-123] fix later',
+			options: [{allowWarningComments: false, ignore: ['ISSUE-\\d+']}]
+		},
+		{
+			code: '// TODO [2020-01-01, ISSUE-123] fix later',
+			options: [{allowWarningComments: false, ignore: ['ISSUE-\\d+']}]
+		},
+		{
+			code: '// TODO [2020-01-01, Issue-123] fix later',
```
fisker: Can  you  use  a more  earlier date ? This  is  not  that  obvious.

test/expiring-todo-comments.js
Moeriki: Not sure if I understand what you mean.

When I remove `ignore: ['ISSUE-\\d+']` the comment is reported as "Unexpected 'todo' comment."

Does that answer it?

test/expiring-todo-comments.js
Moeriki: Will update.

rules/expiring-todo-comments.js
Moeriki: Updated docs to include examples of both string and regex and an small explainer on escaping characters.

rules/expiring-todo-comments.js
```diff
@@ -207,11 +207,16 @@ function semverComparisonForOperator(operator) {
...
 	const options = {
 		terms: ['todo', 'fixme', 'xxx'],
+		ignore: [],
 		ignoreDatesOnPullRequests: true,
 		allowWarningComments: true,
 		...context.options[0]
 	};
 
+	const ignoreRegexes = options.ignore.map(
+		ignore => ignore instanceof RegExp ? ignore : new RegExp(ignore, 'u')
```
fisker:
```suggestion
		pattern => pattern instanceof RegExp ? pattern : new RegExp(pattern, 'u')
```

Moeriki:
I have added two tests. The second one, with two comments, I did notice a potential issue.

When writing the code like this.

```js
			code: `/*
			* TODO fix later
			* TODO ISSUE-123 fix later
			*/`,
```

It would ignore the whole code block. Or maybe it just isn't an error with ESLint `no-warning-comments`?

fisker:
You mean when you use block comment, it get ignored? I think its expected, since thats one comment.

fisker:
Idea just come, out scope this PR, but canwe check this ignore patterns, if its `\d{4}-\d{2}-\d{2}` we can skip the date check, and if its `engine` we can skip engine check.

fisker:
@Moeriki I check the code  again, seems we are not doing right, should just ignore one.

fisker:
@Moeriki ping

Moeriki:
Picking this up again. Sorry for the delay.

I figured out the issue with the block comment. It is not ignoring the whole block, and actually I think the PR is OK to be merged as-is. The problem is an existing bug.

Given `allowWarningComments: false`.

When you have a line comment `// TODO`, Ava will pass `TODO Invalid` to the fake context. And trigger a warning comment.

When you have a block comment on the other hand

```
/*
 * TODO Invalid
 */
```

Then Ava will pass `* TODO Invalid` to the fake context. Because by default `no-warning-comments` is configured with `location: "start"` this **does not** give a warning. That's why it seemed that it was ignoring the whole block.

fisker:
Thanks for the information, does this mean if we check on

```js
			code: `/*
			TODO fix later
			TODO ISSUE-123 fix later
			*/`,
```

and ignore `issue-*` we should still get one warning?

Moeriki:
Yes. I will add this as a test.

Moeriki:
Rebased to master. Squashed everything so far.

fisker:
I  checked the tests again, seems we only tested regex literals, let's add some string patterns. No need squash, just add commits, esaier to review.

fisker:
The integration fails seems unrelated, maybe parser needs update?

Moeriki:
Same test fails on master too.

fisker:
Don't worry, it's unrelated. If you're interest, you can send another update those deps https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/integration/eslint-config-unicorn-tester/package.json , see if it works. If it's not, we'll fix it later.

Moeriki:
There was only one dependency to bump in eslint-config-unicorn-tester.

But I’m afraid it didn’t fix the failed test.

PR #560, opened 2020-02-16
Fix `no-new-buffer` crash on `TypeScript-ESLint`

fisker:
Use `.range[1]` insteadof `.end`, left from #519

### DIFF
--- a/rules/no-new-buffer.js
+++ b/rules/no-new-buffer.js
@@ -1,7 +1,19 @@
 'use strict';
 const getDocumentationUrl = require('./utils/get-documentation-url');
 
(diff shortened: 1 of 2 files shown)


rules/no-new-buffer.js
```diff
@@ -1,15 +1,27 @@
...
+		if (
+			firstArgument.type === 'Literal' &&
+			typeof firstArgument.value === 'number'
+		) {
+			return 'alloc';
+		}
+	}
+
+	return 'from';
+};
```
fisker: Add a `Literal` check.

PR #571, opened 2020-03-02
Linting

brettz9:
- Linting: Apply linting with rc file (unicorn/recommended and eslint:recommended, with a few tweaks), allowing IDEs to display during development
- Linting Add ignore file (to avoid errors when viewing `node_modules` and to enable checks on default-ignored `.eslintrc.js`)
- Linting (package.json): Add recommended `bugs`, `homepage` and `contributors` (used on npmjs.com)
- Refactoring (package.json): Use URLs to facilitate IDE URL opening (e.g., `cmd-O` in Atom when hovering a URL will open it)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,2 @@
+node_modules
+!.eslintrc.js
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,15 @@
+'use strict';
+module.exports = {
+	env: {
+		node: true
+	},
+	extends: ['eslint:recommended', 'plugin:unicorn/recommended'],
+	rules: {
+		semi: ['error'],
+		indent: ['error', 'tab', {
+			'SwitchCase': 1
+		}],
+		// Override eslint:recommended which lacks the option
+		'no-empty': ['error', {allowEmptyCatch: true}]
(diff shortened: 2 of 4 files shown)


fisker:
We have `xo` and `yarn lint` script.

brettz9:
Those must be run manually, no? I'm adding this so that my IDE can discover the settings and report errors while I'm developing on the codebase.

sindresorhus:
Thanks for the PR, but this is not something we want.

sindresorhus:
> Linting: Apply linting with rc file (unicorn/recommended and eslint:recommended, with a few tweaks), allowing IDEs to display during development

There are editor plugins for XO: https://github.com/xojs/xo#editor-plugins

> Linting (package.json): Add recommended bugs, homepage and contributors (used on npmjs.com)

These are inferred from the other fields.

brettz9:
Ok, thanks for the info on the editor plugins. As far as the other items being inferred from other fields, besides the reasons I mentioned (allowing IDEs to easily jump from a link), these are used on the site npmjs.com (which does not do any inferring). So if I go to the npmjs.com page, I have to take extra link jumps to get to the bugs page.

PR #575, opened 2020-03-03
Update `ava` to v3

fisker:


sindresorhus:
I would prefer just waiting until we can soon use `import` natively so we don't have to do all this churn.

fisker:
How about use `{babel: true}` and `@ava/babel`?
We can simply remove it when it's not needed.

sindresorhus:
Sure

PR #578, opened 2020-03-03
`prefer-spread`: Insert semicolon if needed

### LINKED ISSUE #254: `prefer-spread` fixer bug
`prefer-spread` fixer has a bug on no-semi code style

error code like this

```js
const foo = []
Array.from(arr).forEach(doSomething)
```

will fix to code like

```js
const foo = []
[...arr].forEach(doSomething)
```

no semi added before `[` on second line, breaks code

fisker:
I'm sure there is still other case not handled, but should good enough for #254

Fix #254

### DIFF
--- a/rules/prefer-spread.js
+++ b/rules/prefer-spread.js
@@ -20,17 +20,19 @@ const create = context => {
 	const needsSemicolon = node => {
 		const tokenBefore = sourceCode.getTokenBefore(node);
 
-		if (
-			tokenBefore &&
-			tokenBefore.type === 'Punctuator' &&
-			(tokenBefore.value === ']' || tokenBefore.value === ')')
-		) {
-			return true;
-		}
-
 		if (tokenBefore) {
-			const lastBlockNode = sourceCode.getNodeByRangeIndex(tokenBefore.range[0]);
+			const {type, value} = tokenBefore;
+			if (type === 'Punctuator') {
+				if (value === ';') {
+					return false;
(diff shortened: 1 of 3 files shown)


rules/prefer-spread.js
```diff
@@ -14,7 +14,32 @@ const selector = [
 ].join('');
 
 const create = context => {
-	const getSource = node => context.getSourceCode().getText(node);
+	const sourceCode = context.getSourceCode();
+	const getSource = node => sourceCode.getText(node);
+
+	const needsSemicolon = node => {
```
sindresorhus: Should this be in `util`?

fisker: I tried to make a better util function for this, but seems need handle differently in different case.I don't think current implementation can use in other rule, at least not yet.

sindresorhus: Ok. Just something to keep in mind in case we need it in multiple places.

fisker:
Screenshot_2020-03-05-22-20-46-024_com chrome canary

Not important, but  why the last  commit shows by you ? The `ava` one  by me ?  Github  bug?

sindresorhus:
Yes, GitHub bug. I experienced it on another repo earlier too...

sindresorhus:
Can you report it? I’m AFK.

sindresorhus:
https://twitter.com/sindresorhus/status/1235621354528845830

PR #581, opened 2020-03-06
Allow descriptive error names by default in `catch-error-name`

### LINKED ISSUE #192: Allow descriptive error name in catch-error-name
Currently the `catch-error-name` forces to name the `catch` argument `error`, `error2`, `error3` etc... Adding a number at the end of a variable is not really descriptive.

I guess it's ok for the rule to accept `error${index}` as a valid name but the rule should allow and recommend to use a descriptive error name such as `fsError`, `httpError` etc...

lo1tuma:
Fixes #192.

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -34,7 +34,7 @@ const create = context => {
 
 	const options = {
 		name: 'error',
-		caughtErrorsIgnorePattern: '^_$',
+		caughtErrorsIgnorePattern: /^_$|^[A-Za-z0-9]+(e|E)rror$/.source,
 		...context.options[0]
 	};
 
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -34,7 +34,7 @@ const create = context => {
 
 	const options = {
 		name: 'error',
-		caughtErrorsIgnorePattern: /^_$|^[A-Za-z0-9]+(e|E)rror$/.source,
+		caughtErrorsIgnorePattern: /^_$|^[\dA-Za-z]+(e|E)rror$/.source,
 		...context.options[0]
 	};
 
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -101,7 +101,7 @@ You can set the `name` option like this:
 ]
 ```
 
-This option lets you specify a regex pattern for matches to ignore. The default is `^_$`.
+This option lets you specify a regex pattern for matches to ignore. The default allows `_` and descriptive names like `networkError`.
 
 With `^unicorn$`, this would fail:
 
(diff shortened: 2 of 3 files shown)


rules/catch-error-name.js
```diff
@@ -34,7 +34,7 @@ const create = context => {
 
 	const options = {
 		name: 'error',
-		caughtErrorsIgnorePattern: '^_$',
+		caughtErrorsIgnorePattern: '^_$|^error\\w+|^[A-Za-z]\\w*(e|E)rror$',
```
sindresorhus:
```suggestion
		caughtErrorsIgnorePattern: /^_$|^error\w+|^[A-Za-z]\w*(e|E)rror$/.source,
```

sindresorhus: I don't think we should allow `_` in the same.

sindresorhus: This needs to be updated in the docs too.

sindresorhus:
Maybe add a tip to the docs about the fact that even though auto-fix fixes it to `error2`, `error3`, etc, the user is free to rename it to something more descriptive as long as it matches the default pattern.

lo1tuma:
Thanks fo the feedback. I’ve updated the PR and documented the new default setting.

> I don't think we should allow `_` in the same.

Do you mean the case where the name is just `_` or that we allow snake case like `network_error`?

sindresorhus:
Snake case

fisker:
Do we need allow `someError2` ?

rules/catch-error-name.js
```diff
@@ -34,7 +34,7 @@ const create = context => {
 
 	const options = {
 		name: 'error',
-		caughtErrorsIgnorePattern: '^_$',
+		caughtErrorsIgnorePattern: /^_$|^error\w+|^[A-Za-z]\w*(e|E)rror$/.source,
```
sindresorhus: Should we really allow `errorFoo`? Seems like the more correct naming would be `fooError`.

sindresorhus:
> Do we need allow someError2 ?

Let's try without first and see if anyone complains.

rules/catch-error-name.js
sindresorhus: Are you asking "why don't we support regex **literal** directly"?

fisker: Sorry, yes. I mean we can support regex literal and regex string.

sindresorhus: How? The JSON schema does not support regex literal.

fisker: Right,never mind.

sindresorhus: Relevant: https://github.com/eslint/eslint/issues/11318

lo1tuma: Removed support for `errorFoo`.

lo1tuma:
Removed support for snake case.

rules/catch-error-name.js
```diff
@@ -34,7 +34,7 @@ const create = context => {
 
 	const options = {
 		name: 'error',
-		caughtErrorsIgnorePattern: '^_$',
+		caughtErrorsIgnorePattern: /^_$|^[A-Za-z0-9]+(e|E)rror$/.source,
```
fisker:
```suggestion
		caughtErrorsIgnorePattern: /^_$|^[\dA-Za-z]+(e|E)rror$/.source,
```

PR #584, opened 2020-03-08
Add empty `create` function to deprecated rules

fisker:
Not sure how it works, but seems `create` is required. https://github.com/fisker/shared-configs/pull/865/checks?check_run_id=493628377#step:5:14

```text
TypeError: Error while loading rule 'unicorn/regex-shorthand': rule.create is not a function
```

### DIFF
--- a/rules/prefer-exponentiation-operator.js
+++ b/rules/prefer-exponentiation-operator.js
@@ -2,6 +2,7 @@
 const getDocumentationUrl = require('./utils/get-documentation-url');
 
 module.exports = {
+	create() {},
 	meta: {
 		type: 'suggestion',
 		docs: {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Tests are failing

fisker:
@sindresorhus It should return an empty object, I fixed it. 

But another strange thing came up, 

If I use 

```js
module.exports = {
	create: () => ({})
};
```

`eslint-plugin/require-meta-schema` `eslint-plugin/require-meta-docs-description` `eslint-plugin/require-meta-docs-url` are failing.

I notice this before, seems those rules are not working on our other rules file.

This might be a `eslint-plugin-eslint-plugin` bug.

fisker:
@sindresorhus Can you exclude `rules/.DS_Store` file on next release? 

https://unpkg.com/browse/eslint-plugin-unicorn@17.0.0/rules/

sindresorhus:
`npm` should do that automatically, so if it doesn't, it's a bug in `npm`. I remember this being a bug some years ago, so I guess it's come back... I'm using npm v6.14.0.

fisker:
Indeed

https://github.com/npm/cli/issues/440 
https://npm.community/t/831

PR #588, opened 2020-03-11
`consistent-function-scoping`: Ignore within `useEffect` in React Hooks

### LINKED ISSUE #392: `consistent-function-scoping`: Ignore within `useEffect` in React Hooks
In React Hooks, I got:

```js
useEffect(() => {
  async function getItems() {}
  
  getItems()
}, [])
```

And it's definitely better to keep the function in the `useEffect`, to group it together. But this rule recommends moving out the function, of course.

Perhaps perform a check if the function is in `useEffect`, to ignore it?

fisker:
Fixes #392

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -75,7 +75,7 @@ function checkReferences(scope, parent, scopeManager) {
 		);
 }
 
-const isReactHooks = scope =>
+const isReactHook = scope =>
 	scope.block &&
 	scope.block.parent &&
 	scope.block.parent.callee &&
@@ -109,7 +109,7 @@ function checkNode(node, scopeManager) {
 	}
 
 	const parentScope = scopeManager.acquire(parentNode);
-	if (!parentScope || parentScope.type === 'global' || isReactHooks(parentScope)) {
+	if (!parentScope || parentScope.type === 'global' || isReactHook(parentScope)) {
 		return true;
 	}
 
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -60,6 +60,13 @@ function checkReferences(scope, parent, scopeManager) {
 		);
 }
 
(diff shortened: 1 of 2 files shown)


test/consistent-function-scoping.js
```diff
@@ -362,6 +368,25 @@ ruleTester.run('consistent-function-scoping', rule, {
...
+				}, [])
+			`,
+			errors: [
+				{
+					ruleId: 'consistent-function-scoping',
+					messageId: 'FunctionDeclaration',
+					column: 3,
+					line: 3
+				}
+			]
```
fisker: Todo: use function name to test after merge https://github.com/sindresorhus/eslint-plugin-unicorn/pull/589

rules/consistent-function-scoping.js
```diff
@@ -60,6 +60,13 @@ function checkReferences(scope, parent, scopeManager) {
 		);
 }
 
+const isReactHooks = scope =>
```
sindresorhus:
```suggestion
const isReactHook = scope =>
```

It's only a single hook.

rules/consistent-function-scoping.js
```diff
@@ -60,6 +60,13 @@ function checkReferences(scope, parent, scopeManager) {
...
+	scope.block &&
+	scope.block.parent &&
+	scope.block.parent.callee &&
+	scope.block.parent.callee.type === 'Identifier' &&
+	scope.block.parent.callee.name === 'useEffect';
```
sindresorhus: There are more built-in hooks: https://reactjs.org/docs/hooks-reference.html

fisker: Added https://github.com/sindresorhus/eslint-plugin-unicorn/pull/588/commits/4a4f06efd7e67ef779017cbf4b0c8cc8bd3cae66

Feel I don't need remove those hooks doesn't accept function parameter.

PR #589, opened 2020-03-11
`consistent-function-scoping`: Display function name in error message

fisker:
```js
const foo = () => {}
```

Should I use foo as arrow function name?


sindresorhus:
> Should I use foo as arrow function name?

Yes, that's how the `Function#name` algorithm determines the name too.

fisker:
Done.

PR #590, opened 2020-03-11
`consistent-function-scoping`: Check use of `this` and `arguments`

fisker:
`this` is not considered in arrow function before, now fixed.
Also add `arguments` tests.


fisker:
God, more and more complicated when I start to dig, hope this work.

sindresorhus:

PR #603, opened 2020-03-13
`prevent-abbreviations`: Add global variables check

### LINKED ISSUE #445: `prevent-abbreviations` fixer bug on global variable
found this duration #444 , it's in our test

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/a496e96d1e8c4572bffa5c0c4dbb9a8404eb4360/test/prevent-abbreviations.js#L974-L988

but, `property` in first one could be a global variable.

--- FYI: I've checked this code bellow is right

```js
const property = 'outer';
function unicorn(unicorn) {
	const {prop = {}} = unicorn;
	return property;
}
```
fix to

```js
const property = 'outer';
function unicorn(unicorn) {
	const {prop: property_ = {}} = unicorn;
	return property;
}
```

fisker:
Fixes #445

### DIFF
--- a/rules/utils/avoid-capture.js
+++ b/rules/utils/avoid-capture.js
@@ -41,9 +41,9 @@ function unicorn() {
 	};
 }
 */
-const isProbablyGlobal = (name, scopes) => scopes.some(scope =>
+const isUnresolvedName = (name, scopes) => scopes.some(scope =>
 	scope.references.some(reference => reference.identifier && reference.identifier.name === name && !reference.resolved) ||
-	isProbablyGlobal(name, scope.childScopes)
+	isUnresolvedName(name, scope.childScopes)
 );
 
 const isSafeName = (name, scopes, ecmaVersion, isStrict) => {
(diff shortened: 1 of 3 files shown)


rules/utils/avoid-capture.js
```diff
@@ -28,10 +28,17 @@ const nameCollidesWithArgumentsSpecial = (name, scopes, isStrict) => {
 	return isStrict || scopes.some(scopeHasArgumentsSpecial);
 };
 
+const maybeGlobal = (name, scopes) => scopes.some(
```
sindresorhus:
```suggestion
const isProbablyGlobal = (name, scopes) => scopes.some(
```

?

And maybe a doc comment that explains what maybe/probably implies. Like when can it not be correct.

fisker:
DO NOT merge, find another case, not right

```js
				function unicorn() {
					const prop = 1;
					return function () {
						return property;
					};
				}
```

to

```js
				function unicorn() {
					const property = 1; // Should be `const property_ = 1;`
					return function () {
						return property;
					};
				}
```

PR #604, opened 2020-03-13
Add `prefer-set-has` rule

### LINKED ISSUE #495: Rule proposal: `prefer-set-has`
Enforce use `Set` for existence check over `Array#includes`.

# Fail

```js
const array = ['foo', 'bar'];

if (array.includes('foo')) {}
```

# Pass

```js
const set = new Set(['foo', 'bar']);

if (set.has('foo')) {}
```

This rule might should not fail on this

```js
if (['foo', 'bar'].includes('foo')) {}
```

Maybe only check variables and check references. Also perhaps the rule name `prefer-set-has` ?

real world case: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d638e5477d7240ea777dcaf249d0121a5a1910eb/rules/prefer-node-append.js#L6-L17

fisker:
Fixes #495

### DIFF
--- a/test/prefer-set-has.js
+++ b/test/prefer-set-has.js
@@ -197,6 +197,6 @@ ruleTester.run(ruleId, rule, {
 				...createError('foo'),
 				...createError('bar')
 			]
-		},
+		}
 	]
 });
--- a/test/prefer-set-has.js
+++ b/test/prefer-set-has.js
@@ -1,6 +1,6 @@
 import test from 'ava';
 import avaRuleTester from 'eslint-ava-rule-tester';
-import outdent from 'outdent';
+import {outdent} from 'outdent';
 import rule from '../rules/prefer-set-has';
 
 const ruleId = 'prefer-set-has';
--- a/test/prefer-set-has.js
+++ b/test/prefer-set-has.js
@@ -7,7 +7,8 @@ const ruleId = 'prefer-set-has';
 
 const ruleTester = avaRuleTester(test, {
 	parserOptions: {
-		ecmaVersion: 2020
+		ecmaVersion: 2020,
+		sourceType: 'module'
 	}
 });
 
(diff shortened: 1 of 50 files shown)


test/prefer-set-has.js
```diff
@@ -0,0 +1,208 @@
...
+		outdent`
+			function unicorn() {
+				const foo = [1, 2, 3];
+			}
+			const exists = foo.includes(1);
+		`,
+
+		outdent`
+			export const foo = [1, 2, 3];
+			const exists = foo.includes(1);
```
fisker: Is there any other way can expose a variable need  to be  handled?

sindresorhus: Sorry, I don't understand the question.

fisker: We need check variable only called `.includes` and make sure it can't modify from other file. I think `export const foo` is the only way to make foo exposed without accessing `foo`?

sindresorhus: Maybe this?

```js
module.exports = [1, 2, 3];
const exists = module.exports.includes(1);
```

---

Can you add tests for these cases too? Just to ensure they're covered.

```js
const foo = [1, 2, 3];
export {foo};
const exists = foo.includes(1);
```

```js
const foo = [1, 2, 3];
export default foo;
const exists = foo.includes(1);
```

```js
const foo = [1, 2, 3];
export {foo as bar};
const exists = foo.includes(1);
```

```js
const foo = [1, 2, 3];
module.exports = foo;
const exists = foo.includes(1);
```

```js
const foo = [1, 2, 3];
exports = foo;
const exists = foo.includes(1);
```

```js
const foo = [1, 2, 3];
module.exports.foo = foo;
const exists = foo.includes(1);
```

fisker:
```js
module.exports = [1, 2, 3];
const exists = module.exports.includes(1);
```

This won't reported, because only `VariableDeclaration` is allowed. Other tests added https://github.com/sindresorhus/eslint-plugin-unicorn/pull/604/commits/3282cccdc83f6bbae6fb5e2a192312f591333e5b .

PR #559, opened 2020-02-16
Add more projects to the integration test

fisker:
Add some new projects to tests
Add `vue-eslint-parser` tests
Remove `eslint-plugin-unicorn`, `yarn lint` already cover that.

Found 3 bugs during this, two fixed #560 #558 , one open #561

PR #606, opened 2020-03-16
`prevent-abbreviations`: Fix crash on `TypeParameter` with default value

fisker:


test/integration/test.js
```diff
@@ -112,6 +112,7 @@ const projects = [
 	},
 	'https://github.com/chakra-ui/chakra-ui',
 	'https://github.com/ReactTraining/react-router',
+	'https://github.com/facebook/relay',
```
fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/559#issuecomment-586711964

Issue #438, opened 2019-11-16
Make `prevent-abbreviations` rule work better with TypeScript ESLint


sindresorhus:
Currently, the rule's auto-fixer removes all types. Would be useful to make it preserve types when https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/parser is the parser.

PR #608, opened 2020-03-16
Keep `typeAnnotation` when replacing `identifier`

### LINKED ISSUE #347: Rules improvements with the help of `@typescript-eslint/parser`
I don't have experience with writing ESLint rules but I am wondering if `@typescript-eslint/parser` can help improve some of the rules. For example, `prefer-event-key` can extend its reach by detecting that the object type is KeyboardEvent and make the check.

```ts
function onKeyDown(e: KeyboardEvent) {
  if (e.keyCode) {
    // ...
  }
}
```

fisker:
Based on #606 , not sure if other rules need this to fix ~#438~ , I'll extract this to utils and check other rules later

Fixes #347

### DIFF
--- a/docs/rules/prefer-string-slice.md
+++ b/docs/rules/prefer-string-slice.md
@@ -4,17 +4,15 @@
 
 This rule is partly fixable.
 
-
 ## Fail
 
 ```js
-foo.substr(1, 2);
-foo.substring(1, 3);
+foo.substr(start, length);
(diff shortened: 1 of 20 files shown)


fisker:
@sindresorhus I've handled `typeAnnotation` in `prevent-abbreviations`, and Extract a useful function `renameIdentifier` to utils. 

I will check identifier renaming in other rules in separate PRs later.

PR #613, opened 2020-03-17
`catch-error-name`: Keep `typeAnnotation` when replacing `identifier`

### LINKED ISSUE #347: Rules improvements with the help of `@typescript-eslint/parser`

### LINKED ISSUE #438: Make `prevent-abbreviations` rule work better with TypeScript ESLint

fisker:
- [x] catch-error-name
- [x] prevent-abbreviation #608

Close #438

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -2,6 +2,7 @@
 const astUtils = require('eslint-ast-utils');
 const avoidCapture = require('./utils/avoid-capture');
 const getDocumentationUrl = require('./utils/get-documentation-url');
+const renameIdentifier = require('./utils/rename-identifier');
 
 // Matches `someObj.then([FunctionExpression | ArrowFunctionExpression])`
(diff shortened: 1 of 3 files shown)


fisker:
I checked all rules, turns out we only rename `identifier` in `catch-error-name` and 
 `prevent-abbreviation`.

Incoming rule `prefer-set-has` #604 may also need this fix.

All done.

sindresorhus:
Nice work 👏

PR #622, opened 2020-03-20
Add `prefer-number-properties` rule

### LINKED ISSUE #594: Rule proposal: Prefer `Number` methods
### Fail

```js
parseInt(…);
parseFloat(…);
isNaN(…);
isFinite(…);
```

### Pass

```js
Number.parseInt(…);
Number.parseFloat(…);
Number.isNaN(…);
Number.isFinite(…);
```

This rule cannot be auto-fixable as the globals methods have slightly different behavior. However, it **should** use the ESLint suggestions API.

Should we also enforce `Number.NaN` over `NaN`? They equivalent. I'm leaning towards "yes", for consistency, and also less use of floating globals.

fisker:
Fixes #594


docs/rules/prefer-number-properties.md
```diff
@@ -0,0 +1,60 @@
+# Prefer `Number.{parseInt,parseFloat,isNaN,isFinite}()` and `Number.NaN` over global ones.
```
sindresorhus:
```suggestion
# Prefer `Number` static properties over global ones.
```

?

We don't really need to mention all the methods in the title.

docs/rules/prefer-number-properties.md
```diff
@@ -0,0 +1,60 @@
...
+- [`Number.parseFloat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseFloat) over [`parseFloat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) *(fixable)*
+- [`Number.isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) over [`isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN)
+- [`Number.isFinite()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite) over [`isFinite()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isFinite)
+- [`Number.NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/NaN) over [`NaN`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) *(fixable)*
+
+This rule is partly fixable.
```
sindresorhus: Maybe explain why the other are not fixable? Because they have slightly different behavior.

rules/prefer-number-properties.js
```diff
@@ -0,0 +1,121 @@
...
+
+			if (node.parent && node.parent.type === 'VariableDeclarator' && node.parent.id === node) {
+				return;
+			}
+
+			if (node.parent && node.parent.type === 'TSDeclareFunction' && node.parent.id === node) {
+				return;
+			}
+
+			// TODO: not sure how this works
```
sindresorhus: I would take a look at TypeScript-ESLint.

https://github.com/typescript-eslint/typescript-eslint/blob/afa7900285ad6d31d2acaff0107fbd19bad4ff06/packages/eslint-plugin-internal/src/rules/prefer-ast-types-enum.ts#L45

sindresorhus (changes requested):

sindresorhus:
Perfect 👌

KyleHorkley:
Do you think we could add the requirement for `parseFloat` as well? `parseInt` is included here but not `parseFloat`. I know there's the `prefer-number-coercion` rule, but if that's disabled I think it still make sense to include here.

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3307

PR #624, opened 2020-03-23
Fix integration tests

fisker:


fisker:
Can't figure a way to ignore files, I'm going to disable tests on `typescript` project.

PR #632, opened 2020-03-26
Add `prefer-default-parameters` rule

### LINKED ISSUE #29: Prefer default parameters
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/29)

Applies only to `es6` mode:

Bad:

``` js
function (foo) {
  foo = foo || 'bar';
  // ...
}
```

Good:

``` js
function (foo = 'bar') {
  // ...
}
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $60.00)

### Submitted pull Requests
- [#632 Add `prefer-default-parameters` rule](https://issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/632)
---

### Tips

---
</details>

medusalix:
Some notes: I've tried to keep the rule simple by only reporting `literal` assignments (see [this](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/29#issuecomment-502794939) comment). Combining parameters with e.g. a return value of another function or variable might be desired in some cases and should in my opinion not be reported. I've had to work around arrow functions with only one parameter and no parentheses (`foo => {}`) to make sure the default parameter gets inserted correctly by including the missing parentheses.

Fixes #29

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#29: Prefer default parameters](https://issuehunt.io/repos/55832243/issues/29)
---

</details>

### DIFF
--- a/rules/prefer-default-parameters.js
+++ b/rules/prefer-default-parameters.js
@@ -24,6 +24,68 @@ const isDefaultExpression = (left, right) =>
 	right.left.type === 'Identifier' &&
 	right.right.type === 'Literal';
 
+const containsCallExpression = (source, node) => {
+	if (!node) {
+		return false;
+	}
+
+	if (node.type === 'CallExpression') {
+		return true;
+	}
+
+	for (const key of source.visitorKeys[node.type]) {
+		const value = node[key];
+
+		if (Array.isArray(value)) {
+			for (const element of value) {
+				if (containsCallExpression(source, element)) {
+					return true;
+				}
+			}
+		} else if (containsCallExpression(source, value)) {
+			return true;
+		}
+	}
+
+	return false;
+};
+
+const hasSideEffects = (source, function_, node) => {
+	for (const element of function_.body.body) {
+		if (element === node) {
+			break;
+		}
+
+		// Function call before default-assignment
+		if (containsCallExpression(source, element)) {
+			return true;
+		}
+	}
+
+	return false;
+};
+
+const hasExtraReferences = (assignment, references, left) => {
+	// Parameter is referenced prior to default-assignment
+	if (assignment && references[0].identifier !== left) {
+		return true;
+	}
+
+	// Old parameter is still referenced somewhere else
+	if (!assignment && references.length > 1) {
+		return true;
+	}
+
+	return false;
+};
+
+const isLastParameter = (parameters, parameter) => {
+	const lastParameter = parameters[parameters.length - 1];
+
+	// See 'default-param-last' rule
+	return parameter && parameter === lastParameter;
+};
+
 const needsParentheses = (source, function_) => {
 	if (function_.type !== 'ArrowFunctionExpression' || function_.params.length > 1) {
 		return false;
@@ -83,26 +145,18 @@ const create = context => {
 			return;
 		}
 
-		const variable = findVariable(context.getScope(), secondId);
-
-		// Parameter is referenced prior to default-assignment
-		if (assignment && variable.references[0].identifier !== left) {
-			return;
-		}
-
-		// Old parameter is still referenced somewhere else
-		if (!assignment && variable.references.length > 1) {
-			return;
-		}
-
-		const parameter = currentFunction.params.find(parameter =>
+		const {references} = findVariable(context.getScope(), secondId);
+		const {params} = currentFunction;
+		const parameter = params.find(parameter =>
 			parameter.type === 'Identifier' &&
 			parameter.name === secondId
 		);
-		const lastParameter = currentFunction.params[currentFunction.params.length - 1];
 
-		// See 'default-param-last' rule
-		if (!parameter || parameter !== lastParameter) {
+		if (
+			hasSideEffects(source, currentFunction, node) ||
+			hasExtraReferences(assignment, references, left) ||
+			!isLastParameter(params, parameter)
+		) {
 			return;
 		}
 
--- a/rules/prefer-default-parameters.js
+++ b/rules/prefer-default-parameters.js
@@ -78,16 +78,20 @@ const create = context => {
 			right: {raw: literal}
 		} = right;
 
-		// Check if literal is assigned to the same identifier
+		// Parameter is reassigned to a different identifier
 		if (assignment && firstId !== secondId) {
 			return;
 		}
 
 		const variable = findVariable(context.getScope(), secondId);
-		const maxReferences = assignment ? 2 : 1;
 
-		// Check if identifier is referenced elsewhere
-		if (variable.references.length > maxReferences) {
+		// Parameter is referenced prior to default-assignment
+		if (assignment && variable.references[0].identifier !== left) {
+			return;
+		}
+
+		// Old parameter is still referenced somewhere else
+		if (!assignment && variable.references.length > 1) {
 			return;
 		}
 
--- a/rules/prefer-default-parameters.js
+++ b/rules/prefer-default-parameters.js
@@ -88,17 +88,8 @@ const create = context => {
 		);
 		const lastParameter = currentFunction.params[currentFunction.params.length - 1];
 
-		if (!parameter) {
-			return;
-		}
-
 		// See 'default-param-last' rule
-		if (parameter !== lastParameter) {
-			context.report({
-				node,
-				messageId: MESSAGE_ID
-			});
-
+		if (!parameter || parameter !== lastParameter) {
 			return;
 		}
 
--- a/rules/prefer-default-parameters.js
+++ b/rules/prefer-default-parameters.js
@@ -2,6 +2,7 @@
 const getDocumentationUrl = require('./utils/get-documentation-url');
 
 const MESSAGE_ID = 'preferDefaultParameters';
+const MESSAGE_ID_SUGGEST = 'preferDefaultParametersSuggest';
 
 const assignmentSelector = [
 	'ExpressionStatement',
@@ -61,9 +62,11 @@ const fixDefaultExpression = (fixer, source, node) => {
 
 const create = context => {
 	const source = context.getSourceCode();
-	let currentFunction;
+	const functionStack = [];
 
 	const checkExpression = (node, left, right, assignment) => {
+		const currentFunction = functionStack[functionStack.length - 1];
+
 		if (!currentFunction || !isDefaultExpression(left, right)) {
 			return;
 		}
@@ -95,25 +98,28 @@ const create = context => {
 		context.report({
 			node,
 			messageId: MESSAGE_ID,
-			fix: fixer => [
-				fixer.replaceText(parameter, replacement),
-				fixDefaultExpression(fixer, source, node)
-			]
+			suggest: [{
+				messageId: MESSAGE_ID_SUGGEST,
+				fix: fixer => [
+					fixer.replaceText(parameter, replacement),
+					fixDefaultExpression(fixer, source, node)
+				]
+			}]
 		});
 	};
 
 	return {
 		FunctionDeclaration: node => {
-			currentFunction = node;
+			functionStack.push(node);
 		},
 		ArrowFunctionExpression: node => {
-			currentFunction = node;
+			functionStack.push(node);
 		},
 		'FunctionDeclaration:exit': () => {
-			currentFunction = undefined;
+			functionStack.pop();
 		},
 		'ArrowFunctionExpression:exit': () => {
-			currentFunction = undefined;
+			functionStack.pop();
 		},
 		[assignmentSelector]: node => {
 			const {left, right} = node.expression;
@@ -137,7 +143,8 @@ module.exports = {
 		},
 		fixable: 'code',
 		messages: {
-			[MESSAGE_ID]: 'Prefer default parameters over reassignment.'
+			[MESSAGE_ID]: 'Prefer default parameters over reassignment.',
+			[MESSAGE_ID_SUGGEST]: 'Replace reassignment with default parameter.'
 		}
 	}
 };
--- a/rules/prefer-default-parameters.js
+++ b/rules/prefer-default-parameters.js
@@ -120,16 +120,10 @@ const create = context => {
 	};
 
 	return {
-		FunctionDeclaration: node => {
+		':function': node => {
 			functionStack.push(node);
 		},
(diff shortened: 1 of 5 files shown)


rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,142 @@
...
+
+	return {
+		FunctionDeclaration: node => {
+			currentFunction = node;
+		},
+		ArrowFunctionExpression: node => {
+			currentFunction = node;
+		},
+		'FunctionDeclaration:exit': () => {
+			currentFunction = undefined;
```
fisker, line 112: I think this is not right,  if assign is after a  inner function, should still in the outer function.

rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,142 @@
...
+const declarationSelector = [
+	'VariableDeclaration',
+	'[declarations.0.type="VariableDeclarator"]'
+].join('');
+
+const isDefaultExpression = (left, right) =>
+	left &&
+	right &&
+	left.type === 'Identifier' &&
+	right.type === 'LogicalExpression' &&
```
fisker, line 20: you  forgot  check  operator

rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,142 @@
...
+			left: {name: secondId},
+			right: {raw: literal}
+		} = right;
+
+		// Check if literal is assigned to the same identifier
+		if (assignment && firstId !== secondId) {
+			return;
+		}
+
+		const parameter = currentFunction.params.find(parameter =>
```
fisker, line 81: I think we only need check the laster parameter.

test/prefer-default-parameters.js
```diff
@@ -0,0 +1,156 @@
...
+		},
+		{
+			code: 'function abc(foo) { foo = foo || true; }',
+			output: 'function abc(foo = true) { }',
+			errors: [error]
+		},
+		{
+			code: 'function abc(foo, bar) { foo = foo || \'bar\'; baz(); }',
+			output: 'function abc(foo = \'bar\', bar) { baz(); }',
+			errors: [error]
```
fisker, line 72: This is why I said, we shoul only check last parameter, not good code.

test/prefer-default-parameters.js
```diff
@@ -0,0 +1,156 @@
...
+		},
+		{
+			code: 'function abc(foo) { const bar = foo || \'bar\'; }',
+			output: 'function abc(bar = \'bar\') { }',
+			errors: [error]
+		},
+		{
+			code: 'function abc(foo) { let bar = foo || \'bar\'; }',
+			output: 'function abc(bar = \'bar\') { }',
```
fisker, line 81: Not sure about this `function abc(foo = \'bar\') { const bar = foo;}`, this should easier, but you already did better. Any chance this would break something?

fisker:
One more thing, since the fix  I not  safe , we should use suggestion api, but don't worry, we can easily convert it.  [example](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/prefer-number-properties.js]

test/prefer-default-parameters.js
medusalix: Why do you think only the last parameter should be checked?

medusalix:
The fix should be safe because I only handle the cases where a literal is used as a default parameter. See [this](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/29#issuecomment-502794939) comment.

medusalix:
> Can you update code with code in function body to multiple lines?

Do you mean the test cases?

test/prefer-default-parameters.js
medusalix: Yeah, if some code relies on the presence of the `foo` parameter it might berak. I'm going to use the suggestions API for this.

fisker:
```js
'function abc() { let foo = 123; const bar = foo || \'bar\' }',
```

to

```js
`
function abc() {
  let foo = 123;
  const bar = foo || 'bar' 
}
`,
```

test/prefer-default-parameters.js
fisker: We can check if `foo` is used. If not used, in the way you already did, otherwise fix to the code I wrote.

WDYT?

test/prefer-default-parameters.js
fisker: See https://eslint.org/docs/rules/default-param-last

rules/prefer-default-parameters.js
medusalix: Fixed using a function stack in the latest commit.

medusalix:
> I think we only need check the last parameter.

We still need to *check* all of them to see if it's actually one of the parameters, but I chose to skip the fixing if it's not the last parameter.

rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,161 @@
...
+		},
+		ArrowFunctionExpression: node => {
+			functionStack.push(node);
+		},
+		'FunctionDeclaration:exit': () => {
+			functionStack.pop();
+		},
+		'ArrowFunctionExpression:exit': () => {
+			functionStack.pop();
+		},
```
fisker, line 134: I think `:function` and `::function:exit` should work.

fisker:
> We still need to check all of them to see if it's actually one of the parameters, but I chose to skip the fixing if it's not the last parameter.

I think we can simple this logic, only report one problem each time on one function. but we need consider following cases

`function foo(a, b, c=1){}` check `b` not `c`, 
`function foo(a, b, ...c){}` also check `b`.

fisker:
Another edge case, since the assignment can be used as value.

```js
function foo(a){
 doSomething(
   a = a || 1
)

doAnotherThing(a)
}
```

or

```js
function foo(a){
 const b = a = a || 1
}
```

So, we should choose leave `a` there or ignore those case.

My suggestion, alway leave `a` there unless parent is `ExpressionStatement`.

fisker:
@medusalix What do  you think the suggestion above?

medusalix:
Sorry for the late reply, I'm pretty busy right now, I'll take a look at your suggestions when I get the time.

medusalix:
> Another edge case, since the assignment can be used as value.

Yeah, that's already checked by the selector: `ExpressionStatement[expression.type="AssignmentExpression"]`. I've included both of your tests.

test/prefer-default-parameters.js
```diff
@@ -0,0 +1,346 @@
...
+				}
+			`]
+		}),
+		invalidTestCase({
+			code: outdent`
+				function abc(foo, bar) {
+					foo = foo || 'bar';
+					baz();
+				}
+			`
```
fisker, line 158: I thought we agree only check the last parameter?

medusalix: Yes, sorry, I assumed we would still report them. I fixed it in the last commit.

fisker:
1. Please fix tests I added.
2. `??` operator is supported by eslint now, please check it too.

rules/prefer-default-parameters.js
```diff
@@ -115,6 +121,14 @@ const create = context => {
 			functionStack.push(node);
 		},
 		':function:exit': () => {
+			for (const [id, report] of reportedIdentifiers) {
```
fisker, line 124: @medusalix 

We use `variable` to analysis the identifiers https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/prefer-set-has.js#L143.

medusalix: Ah, even better. No need to reinvent the wheel I guess.

fisker: I'm not sure if we can check this,

```js
			function abc(foo) {
				baz();
				foo = foo || 'bar';
				function baz() {
					console.log(foo);
				}
			}
```

In this case, `foo` is used before the assignment, but how to know?

medusalix: Yeah, I don't think we can check whether certain functions have side-effects. But we use the suggestions API, so users still have the option to ignore the fixes.

medusalix:
I moved one of your test cases because it can be fixed. References are now checked using `findVariable`.

fisker:
> I moved one of your test cases because it can actually be fixed. References are now checked using `findVariable`.

Good, you checked the `foo`

rules/prefer-default-parameters.js
fisker: I don't know, though users can ignore the fix, but it will still reported.

This seems the tricky part.

fisker: `eslint-utils` has a function called [`hasSideEffect`](https://eslint-utils.mysticatea.dev/api/ast-utils.html#hassideeffect), maybe we can check all nodes before the assignment, ignore those cases hasSideEffect?

medusalix: That works, but it also ignores this (and 5 other test cases):

```js
function abc(bar) {
	foo();
	bar = bar || 123;
}
```

It's only a heuristic, so we get lots of valid test cases that would otherwise be considered invalid.

fisker: @sindresorhus What do you think?

sindresorhus: I'm fine with ignoring such cases. That code should use default parameters anyway.

rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,155 @@
...
+			right: {raw: literal}
+		} = right;
+
+		// Check if literal is assigned to the same identifier
+		if (assignment && firstId !== secondId) {
+			return;
+		}
+
+		const variable = findVariable(context.getScope(), secondId);
+		const maxReferences = assignment ? 2 : 1;
```
fisker, line 87: I feel this not good, but I don't know what to do

```js
			function abc(foo = 123) {
				foo = foo || 'bar';
				console.log(foo)
			}
```

fisker:
I think the problem we can't resolve here is

1.

```js
function abc(foo) {
	foo = foo || 'bar';
	console.log(foo);
}
```

VS


```js
function abc(foo) {
	console.log(foo);
	foo = foo || 'bar';
}
```

2.

```js
function abc(foo) {
	useFoo();
	foo = foo || 'bar';

	function useFoo() {
		console.log(foo);
	}
}
```

VS

```js
function abc(foo) {
	notUseFoo();
	foo = foo || 'bar';

	function notUseFoo() {
	}
}
```

sindresorhus:
Bump

medusalix:
I think I've found a pretty good solution for the first point @fisker outlined above:

1. If the parameter is being reassigned, make sure that there are no references to the parameter before the default-assignment:

```js
foo = foo || 123; // Reference 1 and 2
console.log(foo); // Reference 3
// First reference is the default-assignment; invalid, can be fixed.
```

```js
console.log(foo); // Reference 1
foo = foo || 123; // Reference 2 and 3
// First reference is the console.log statement; valid.
```

2. If the parameter is default-assigned to a new variable, make sure that there are no additional references to the old parameter *at all*:

```js
const bar = foo || 'bar'; // Reference 1
console.log(foo, bar);    // Reference 2
// More than one reference; valid.
```

```js
const bar = foo || 'bar'; // Reference 1
console.log(bar);
// Only one reference; invalid, can be fixed.
```

medusalix:
I've done some tests with the `hasSideEffect` function. I believe implementing side-effect checking defeats the purpose of this rule. Consider the following example:

```js
function abc(foo) {
	let bar = true;
	bar = false;

	foo = foo || 123;
	console.log(foo);
}
```

This example would be ignored by the rule because `hasSideEffect` returns `true` for the `bar = false` statement, even though it doesn't have any effects on the `foo` parameter *at all*.

@fisker I think ignoring cases where there is any `CallExpression` before the default-assignment would be a better solution. What do you think?

fisker:
I'm running out of ideas. :smile:

rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,159 @@
...
+				fix: fixer => [
+					fixer.replaceText(parameter, replacement),
+					fixDefaultExpression(fixer, source, node)
+				]
+			}]
+		});
+	};
+
+	return {
+		':function': node => {
```
fisker, line 127: Do you think it should work on `ObjectMethod`? 

```js
foo = {
  method(foo) {
    foo = foo || 123;
  }
};
```

and `ClassMethod`

```js
class Foo {
  method(foo) {
    foo = foo || 123;
  }
}
```

test/prefer-default-parameters.js
```diff
@@ -0,0 +1,403 @@
...
+						bar = bar || 'foo';
+					}
+					function ghi(bay = 'bar') {
+					}
+					foo = foo || 'bar';
+				}
+			`]
+		})
+	]
+});
```
fisker, line 403: `FunctionExpression` seems not tested

```js
foo = function (foo){
  foo = foo || 123;
}
```

fisker (changes requested):

rules/prefer-default-parameters.js
medusalix: Yes, I think it should. I added some tests.

test/prefer-default-parameters.js
medusalix: I added some tests.

medusalix:
@fisker I've implemented the `CallExpression` check I described in my last comment and added lots of tests for it.

rules/prefer-default-parameters.js
```diff
@@ -0,0 +1,213 @@
...
+			}
+		} else if (containsCallExpression(source, value)) {
+			return true;
+		}
+	}
+
+	return false;
+};
+
+const hasSideEffects = (source, function_, node) => {
```
sindresorhus, line 53: https://eslint-utils.mysticatea.dev/api/ast-utils.html#hassideeffect ?

sindresorhus (changes requested):

sindresorhus (approved):

sindresorhus:
// @fisker

fisker (approved):
I forgot the review progress, I read all comments from beginning again, seems good enough, we can still improve it if any issue found.

PR #633, opened 2020-03-26
Add `no-object-as-default-parameter` rule

### LINKED ISSUE #208: Rule proposal: No default parameter options

medusalix:
Simply reports any default-assignments of objects to function parameters as described in [this](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/208#issuecomment-576647974) comment.

Fixes #208

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#208: Rule proposal: No default parameter options](https://issuehunt.io/repos/55832243/issues/208)
---

</details>


fisker:
How about `no-object-as-default`

fisker:
On second though, will people think this rule forbid this following syntax?

```js
export {
    object as default,
// ^
    foo as bar
};
```

medusalix:
No, I don't think users will mix that up. The description is pretty clear about the rule's purpose.

sindresorhus:
I would name the rule `no-object-as-default-parameter` to make it clearer.

docs/rules/no-object-as-default.md
```diff
@@ -0,0 +1,29 @@
+# Disallow the use of objects as default parameters
+
+Default parameters should not be passed to a function through an object. The `foo = {a: false}` parameter works fine if only used with one option. As soon as additional options are added, you risk replacing the whole `foo = {a: false, b: true}` object when passing only one option: `{a: true}`. For this reason, object destructuring should be used instead.
```
sindresorhus, line 3:
```suggestion
Default parameters should not be passed to a function through an object literal. The `foo = {a: false}` parameter works fine if only used with one option. As soon as additional options are added, you risk replacing the whole `foo = {a: false, b: true}` object when passing only one option: `{a: true}`. For this reason, object destructuring should be used instead.
```

rules/no-object-as-default.js
```diff
@@ -0,0 +1,36 @@
...
+
+module.exports = {
+	create,
+	meta: {
+		type: 'problem',
+		docs: {
+			url: getDocumentationUrl(__filename)
+		},
+		messages: {
+			[MESSAGE_ID]: 'Do not assign default object to parameter `{{parameter}}`.'
```
sindresorhus, line 33:
```suggestion
			[MESSAGE_ID]: 'Do not use an object literal as default for parameter `{{parameter}}`.'
```

?

rules/no-object-as-default.js
```diff
@@ -0,0 +1,36 @@
...
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+const MESSAGE_ID = 'noObjectAsDefault';
+
+const objectParameterSelector = [
+	'AssignmentPattern',
+	'[left.type="Identifier"]',
+	'[right.type="ObjectExpression"]',
+	'[right.properties.length>0]'
+].join('');
```
fisker, line 11: If I remember correctly, the `AssignmentPattern` not only can use as parameter, maybe assignment default value? Anyway, we can use a stricter selector.

fisker: Confirmed `const {foo={bar:1}} = baz;`

medusalix:
> Please add tests about class/object methods.

Could you please elaborate?

rules/no-object-as-default-parameter.js
```diff
@@ -4,6 +4,8 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const MESSAGE_ID = 'noObjectAsDefaultParameter';
 
 const objectParameterSelector = [
+	':matches(FunctionDeclaration, ArrowFunctionExpression)',
```
fisker, line 7: I think `:function` should work.

rules/no-object-as-default-parameter.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+const MESSAGE_ID = 'noObjectAsDefaultParameter';
+
+const objectParameterSelector = [
+	':matches(FunctionDeclaration, ArrowFunctionExpression)',
+	'>',
+	'AssignmentPattern',
```
fisker, line 9: Though another child(function.id) can't be `AssignmentPattern`, I still prefer use `AssignmentPattern.params`.

fisker:
Please test


```js
class A {
  method(foo = {bar:1}) {}
}
```

```js
const A = class {
  method(foo = {bar:1}) {}
}
```

```js
object = {
  method(foo = {bar:1}) {}
}
```

rules/no-object-as-default-parameter.js
fisker: If all your tests passed, you must missed the `FunctionExpression`, pleaseadd one.

medusalix:
> const foo = ({bar: baz = {a = false, b = 123}}) => {};

Is this valid syntax? It outputs `Shorthand property assignments are valid only in destructuring patterns`.

rules/no-object-as-default-parameter.js
medusalix: Yep, I've missed that one. Fixed.

fisker:
`const adc = () => (foo = {bar: 1});`  This one looks similar to default parameter, but it's not , let's add it to test too.

fisker:
> > const foo = ({bar: baz = {a = false, b = 123}}) => {};
> > const foo = ({foo = {a = false, b = 123}}) => {};
> 
> Is this syntactically correct? It gives me `Shorthand property assignments are valid only in destructuring patterns`.

Okay , if  it's not  valid, I  mean  to  test  deep destructuring.


> > Maybe also constructor and getter setter
> 
> I've added tests for the constructor and setter, it doesn't seem like getters allow parameters.

Okay.

fisker:
Looks  good , I'll check  again when ai  back to my  laptop.

medusalix:
Thank you. I really appreciate your constructive feedback, it's always very helpful 👍.

rules/no-object-as-default-parameter.js
```diff
@@ -0,0 +1,36 @@
...
+	'[left.type="Identifier"]',
+	'[right.type="ObjectExpression"]',
+	'[right.properties.length>0]'
+].join('');
+
+const create = context => {
+	return {
+		[objectParameterSelector]: node => {
+			context.report({
+				node,
```
fisker, line 17: @medusalix  Do you think we should report on parameter name? `node.left`

medusalix: Yes, makes sense.

sindresorhus:
Great! Looks good to me too 👍

PR #634, opened 2020-03-27
Simplify `no-console-spaces`

fisker:


rules/utils/replace-string-raw.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = (fixer, node, replacement) =>
```
sindresorhus: Would be good to add a short comment when you add utilities what they do and what problem they solve.

fisker: Done https://github.com/sindresorhus/eslint-plugin-unicorn/pull/634/commits/399abaf746cba7229cc3ff561d38d092043e4a8a

PR #636, opened 2020-03-27
Add `no-null` rule

### LINKED ISSUE #612: Rule proposal: `no-null`
https://github.com/sindresorhus/meta/issues/7

# Fail

```js
const foo = null;
```

# Pass

```js
let foo;
```

Exclude `Object.create`

```js
let foo = Object.create(null);
```

- There is already a plugin do this, https://github.com/nene/eslint-plugin-no-null/blob/master/index.js
- Rule proposal: no-null in `ESLint` project https://github.com/eslint/eslint/issues/6701
- TypeScript Coding guidelines https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

himynameisdave:
Adds a `no-null` rule to disallow use of `null` in favor of `undefined`.

Resolves #612

### DIFF
--- a/rules/no-null.js
+++ b/rules/no-null.js
@@ -30,6 +30,7 @@ const create = context => ({
 
 		const fix = fixer => fixer.replaceText(node, 'undefined');
 
+		/* istanbul ignore next */
 		const {parent = {}} = node;
 		if (isLooseEqual(parent)) {
 			problem.fix = fix;
--- a/rules/no-null.js
+++ b/rules/no-null.js
@@ -1,0 +1,23 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+const create = (context) => ({
+  Literal: (node) => {
+    if (node.raw === 'null') {
+      context.report({
+        node: node,
+        message: 'Use undefined instead of null'
+      });
+    }
+  },
+});
+
+module.exports = {
+  meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocumentationUrl(__filename)
+		},
+	},
+  create,
+};
--- a/rules/no-null.js
+++ b/rules/no-null.js
@@ -1,23 +1,29 @@
 'use strict';
 const getDocumentationUrl = require('./utils/get-documentation-url');
 
-const create = (context) => ({
-  Literal: (node) => {
-    if (node.raw === 'null') {
-      context.report({
-        node: node,
-        message: 'Use undefined instead of null'
-      });
-    }
-  },
+// The number of chars in "Object.create("
+const objectCreateCharCount = 14;
+const objectCreate = 'Object.create(null';
+// Check if the 14 chars preceding `null` are "Object.create("
+const isObjectCreate = (context, node) => context.getSource(node, objectCreateCharCount) === objectCreate;
+
+const create = context => ({
+	Literal: node => {
+		if (node.raw === 'null' && !isObjectCreate(context, node)) {
+			context.report({
+				node,
+				message: 'Use undefined instead of null'
+			});
+		}
+	}
 });
 
 module.exports = {
-  meta: {
+	meta: {
 		type: 'suggestion',
 		docs: {
 			url: getDocumentationUrl(__filename)
-		},
+		}
 	},
-  create,
+	create
 };
--- a/rules/no-null.js
+++ b/rules/no-null.js
@@ -38,14 +38,14 @@ const create = context => ({
 			problem.suggest = [
 				{
 					messageId: SUGGESTION_REMOVE_MESSAGE_ID,
-					fix: fixer => fixer.replaceText(node, '')
+					fix: fixer => fixer.remove(node)
 				}
 			];
 		} else if (parent.type === 'VariableDeclarator' && parent.init === node && parent.parent.kind !== 'const') {
 			problem.suggest = [
 				{
 					messageId: SUGGESTION_REMOVE_MESSAGE_ID,
-					fix: fixer => fixer.replaceTextRange([parent.id.range[1], node.range[1]], '')
+					fix: fixer => fixer.removeRange([parent.id.range[1], node.range[1]])
 				}
 			];
 		} else {
--- a/rules/no-null.js
+++ b/rules/no-null.js
@@ -2,6 +2,10 @@
 const getDocumentationUrl = require('./utils/get-documentation-url');
 const methodSelector = require('./utils/method-selector');
 
+const ERROR_MESSAGE_ID = 'error';
+const SUGGESTION_REPLACE_MESSAGE_ID = 'replace';
+const SUGGESTION_REMOVE_MESSAGE_ID = 'remove';
+
 const objectCreateSelector = methodSelector({
 	object: 'Object',
 	name: 'create',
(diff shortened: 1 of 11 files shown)


rules/no-null.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+// The number of chars in "Object.create("
+const objectCreateCharCount = 14;
+const objectCreate = 'Object.create(null';
+// Check if the 14 chars preceding `null` are "Object.create("
+const isObjectCreate = (context, node) => context.getSource(node, objectCreateCharCount) === objectCreate;
```
fisker: This is bad idea 

```js
Object.create( null );
```

```js
Object.create(
null
);
```

```js
Object
  .create(null);
```

Should use selector or check parent.

fisker:
```js
const methodSelector = require("./utils/method-selector");

const objectCreateSelector = methodSelector({
	object: 'Object',
	name: 'create',
	length: 1
});

const selector = [
	`:not(${objectCreateSelector})`,
	'>',
	'Literal',
	'[raw="null"]',
].join('');
```

I think this selector will work.

rules/no-null.js
himynameisdave: True, it felt a bit hacky. What is the ESLint plugin API for getting a parent node?

himynameisdave:
So I guess the next thing to do in this PR is replace all the use of `null` in this plugin codebase:

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/636/checks?check_run_id=542224252

fisker:
@himynameisdave Please add `"null"` and similar tests like https://github.com/sindresorhus/eslint-plugin-unicorn/blob/5de72d87b1219ff8f09580388032b4dd52092dc7/test/prefer-spread.js#L29-L46 to tests.

fisker:
This rule definitely can't auto fix for most cases, but should we fix `foo == null` to `foo == undefined` (this is very common, though I never use it)?

And should we supply a suggestion fix? If so, I'd like to fix `return null;` to `return;`, not `return undefined;`

himynameisdave:
@fisker note that auto-fixing this case:

```diff
-let foo = null;
+let foo = undefined;
```

...will break [this core rule](https://eslint.org/docs/rules/no-undef-init) for some folks. I always find it annoying when an ESlint auto-fix causes other errors.

fisker:
@himynameisdave 
I'm  talking about suggestions api,  https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions we use it for unsafe fixing.

sindresorhus:
Agreed

fisker:
@himynameisdave Can you check `Allow edits from maintainers.`(on right side of this page, at the bottom), I'll push some codes.

fisker:
I've add all fix mentioned in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/636#issuecomment-605946940

fisker:
Do we need option to ignore `null` compare?

sindresorhus:
> Do we need option to ignore null compare?

I don't think we need that.

fisker:
I have a feel, people going to hate it because of compare.  Some value has to compare with `null` https://github.com/sindresorhus/eslint-plugin-unicorn/blob/0c31a0b5d08d29e529a36205c7b8beaa454c1229/rules/no-unreadable-array-destructuring.js#L6

We fobind use `null` the only reason to use `null` is check values from somewhere else. I suggest we add this option, if you don't like it we can set default to `false`.

sindresorhus:
Sorry, I thought you were talking about only `foo == null`. Yes, an option for what you described above does make sense.

fisker:
Default to `true` and only check `===` / `!==`.

If people follow this rule strictly, only needs is strict compare value from other module with null. We don't want handle `array.includes` or `set.has` or `Object.is` , not worth to do this.

fisker:
@himynameisdave Can you do this?

himynameisdave:
I should have some time this weekend to do this 👍

fisker:
I can help, if you don't have time.

himynameisdave:
@fisker yeah that'd be great, or even if you could just point me to the right place to get started.

I understand that we'd basically need to parse out a new option like:

```js
const checkEquality = context.options[0] || { checkEquality: true };
```

...but I'm not sure if this should all be inside the existing selector that we already have, and which test cases we need to add to account for this.

fisker:
I guess a function like `isStrictEqual` is easier.

rules/no-null.js
```diff
@@ -0,0 +1,78 @@
...
+		type: 'suggestion',
+		docs: {
+			url: getDocumentationUrl(__filename)
+		},
+		messages: {
+			[ERROR_MESSAGE_ID]: 'Use `undefined` instead of `null`.',
+			[SUGGESTION_REPLACE_MESSAGE_ID]: 'Replace `null` with `undefined`.',
+			[SUGGESTION_REMOVE_MESSAGE_ID]: 'Remove `null`.'
+		},
+		fixable: true
```
fisker, line 76:
```suggestion
		fixable: "code"
```

fisker:
I updated the suggestions for `ReturnStatement` and `VariableDeclarator`.

When I try to apply suggestion to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/aa00647236b1bb851f2a5c44f8651c56a8b49c9b/rules/no-unused-properties.js#L142 with `return;`, the core rule `array-callback-return` complained, so I decide add both `remove` and `replace` suggestions, let the user decide which one to apply.

fisker:
Another thing came up, about `JSON.stringify(value, null, 2)` the [`replacer` parameter ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Parameters)is suggested to use `null` on MDN.

I always use `null` instead of `undefined`, though `JSON.stringify({foo: 'bar'}, undefined, 2)` can also work.

@sindresorhus WDYT?  You used [a lot in your repos](https://github.com/search?q=user%3Asindresorhus+JSON.stringify+null&type=Code) too.

sindresorhus:
I want to enforce use of undefined for JSON.stringify. The current usage of null is just because I haven’t gotten around to changing it yet. This rule will make that easier to do.

rules/no-unused-properties.js
```diff
@@ -139,7 +139,7 @@ const create = context => {
 							return {identifier: parent};
 						}
 
-						return null;
+						return undefined;
```
sindresorhus, line 142: @fisker Seems like we have an idea for another rule here. `no-return-undefined`. Since `return` and `return` is the same thing.

sindresorhus:
Nice work 👍

rules/no-unused-properties.js
fisker: See this https://github.com/sindresorhus/eslint-plugin-unicorn/pull/636#issuecomment-611326069

sindresorhus: I'm not talking about this rule. I'm talking in general about `return undefined` being used (anywhere).

fisker: `no-useless-undefined`

- `return undefined;`
- `let a = undefined;`
- `call(undefined)`
- more?

sindresorhus:
```js
const {foo = undefined} = {};

function ({foo = undefined} = {}) {}

function (foo = undefined) {}
```

PR #639, opened 2020-03-27
`prefer-number-properties`: Test suggestions output

fisker:
Didn't know this can be tested.

https://eslint.org/docs/developer-guide/nodejs-api#testing-suggestions

### DIFF
--- a/test/prefer-number-properties.js
+++ b/test/prefer-number-properties.js
@@ -37,7 +37,7 @@ const typescriptRuleTester = avaRuleTester(test, {
 	parser: require.resolve('@typescript-eslint/parser')
 });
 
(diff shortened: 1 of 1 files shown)


fisker:
I wonder is there a way to apply suggestions fix via CLI, so we can test in integration-test.

sindresorhus:
> I wonder is there a way to apply suggestions fix via CLI, so we can test in integration-test.

Maybe worth opening an ESLint issue about that?

fisker:
Let me check how vscode-eslint works first.

fisker:
Opened an issue, https://github.com/eslint/eslint/issues/13103

PR #640, opened 2020-03-27
Rewrite `prefer-node-remove`

### LINKED ISSUE #548: `prefer-node-remove`: Do not remove `callee` from `.removeChild()`
Ref: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/507#issuecomment-585156112

`parentNode.removeChild()` call can have complicated `callee`

Eg: 

```js
function doSomeThingWithChild(node) {
	// do your thing!
	return node;
}

doSomeThingWithChild(childNode).parentNode.removeChild(childNode);
```

Currently, we fix this to `childNode.remove()`, which is not safe. 

Maybe we should not fix in this rule, or use [suggestion](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).

Another related issue: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/549

### LINKED ISSUE #549: `prefer-node-remove`: Allow more types of `childNode`
Ref: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/507#issuecomment-585156112

Currently, this rule only check that `childNode` is `ThisExpression` or `Identifier`, should allow more AST Types, like:

```js
parentNode.removeChild(some.node);
                       ^^^^^^^^^ `MemberExpression`
parentNode.removeChild(getChildNode());
                       ^^^^^^^^^^^^^^ `CallExpression`
```

More cases: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/fe711c851b75d663c04bfa1c2e07649cb7ba58c0/test/prefer-node-remove.js#L61-L65

Note: some case may hard to fix.

Eg:

```js
parentNode.removeChild(await getChild());

// ->
await getChild().remove(); // Not right

// Should be parenthesed
(await getChild()).remove();

// Even might need semicolon
const foo = []
;(await getChild()).remove();

```

Another related issue: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/548

fisker:
Fixes #549
Fixes #548

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
 		"clean-regexp": "^1.0.0",
 		"eslint-ast-utils": "^1.1.0",
 		"eslint-template-visitor": "^1.1.0",
+		"eslint-utils": "^2.0.0",
 		"import-modules": "^2.0.0",
 		"lodash": "^4.17.15",
 		"read-pkg-up": "^7.0.1",
--- a/rules/prefer-node-remove.js
+++ b/rules/prefer-node-remove.js
@@ -1,9 +1,10 @@
 'use strict';
-const {isParenthesized} = require('eslint-utils');
+const {isParenthesized, hasSideEffect} = require('eslint-utils');
 const getDocumentationUrl = require('./utils/get-documentation-url');
 const methodSelector = require('./utils/method-selector');
 const {notDomNode} = require('./utils/not-dom-node');
 const needsSemicolon = require('./utils/needs-semicolon');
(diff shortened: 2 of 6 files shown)


fisker:
I've done:

- Report on any type of `childNode`
- Remove auto fix, add suggestions if `parentNode` has side effect or Value of `CallExpression` is used


@medusalix You're the one implement this rule first, would be great if you can help review this PR.
@brettz9 You may interest to help check all these changes.

sindresorhus:
Looks good to me 👌

PR #643, opened 2020-03-30
`throw-new-error`: Minor refactor

fisker:
- Add `Identifier` check to avoid `customError` test on `undefined`
- Report on `error` instead of `ThrowStatement`

Before:
image

After:
image

- 100% coverage

### DIFF
--- a/rules/throw-new-error.js
+++ b/rules/throw-new-error.js
@@ -1,18 +1,23 @@
 'use strict';
 const getDocumentationUrl = require('./utils/get-documentation-url');
 
+const selector = [
+	'ThrowStatement',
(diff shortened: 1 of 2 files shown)


sindresorhus:
I like how you make many of the rules be more specific about where the error is. I wonder if there's any way we could test it, to ensure what part of the output will be marked as the error.

fisker:
Seems we can test `line` and `column`, but not range, some rule did test them

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/08274f9603092070083642f2a6be25bc16d72c23/test/prefer-spread.js#L56-L57

PR #647, opened 2020-03-31
`catch-error-name`: Simplify logic

fisker:


rules/catch-error-name.js
```diff
@@ -31,7 +32,7 @@ const create = context => {
 
 	const options = {
 		name: 'error',
-		caughtErrorsIgnorePattern: /^_$|^[\dA-Za-z]+(e|E)rror$/.source,
+		caughtErrorsIgnorePattern: /^[\dA-Za-z]+[e|E]rror$/.source,
```
fisker: L47 already ignored `_`

fisker: See https://github.com/sindresorhus/eslint-plugin-unicorn/issues/648 , will fix it later.

Issue #648, opened 2020-03-31
`catch-error-name` bugs


fisker:
[Example in docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/catch-error-name.md) says 

> `_` is allowed when the error is not used

[caughtErrorsIgnorePattern](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/catch-error-name.md#caughterrorsignorepattern) options says

> The default allows `_` and descriptive names like `networkError`.

----

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/b1ae9822926cca47d3621f8f7685f9c9b78e16b8/test/catch-error-name.js#L214

Currently we report it, without auto-fix. This should not be a problem.

PR #649, opened 2020-03-31
`catch-error-name`: Ignore destructuring

fisker:
Fix part of #648


fisker:
@sindresorhus Can you have a quick look, confirm this change?

sindresorhus:
👍

PR #650, opened 2020-03-31
`catch-error-name`: Only allow `_` when it's not used

### LINKED ISSUE #648: `catch-error-name` bugs

fisker:
Fixes #648

### DIFF
--- a/docs/rules/catch-error-name.md
+++ b/docs/rules/catch-error-name.md
@@ -4,7 +4,9 @@ Applies to both `try/catch` clauses and `promise.catch(…)` handlers.
 
 The desired name is configurable, but defaults to `error`.
 
-This rule is fixable unless the reported code was destructuring an error.
+The error name `_` is ignored, if the error is not used.
+
+This rule is fixable.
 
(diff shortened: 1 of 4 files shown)


fisker:
@sindresorhus Another change need confirm, previously we ignore `_`, now we ignore `_` only if it's not used.

sindresorhus:
Yup. This is correct.

PR #662, opened 2020-04-03
`prevent-abbreviations`: Allow ignore words contains number

### LINKED ISSUE #567: prevent-abbreviations: whitelist has no effect on file name
### Environment

* OS Version: Windows 10
* Node.js Version: `13.7.0`
* ESLint: `6.8.0`  
* `eslint-plugin-unicorn`: `16.1.1`

### Actual behavior

```
eslint-plugin-unicorn-repro/some.spec.e2e.js
1:1  error  Please rename the filename `./some.spec.e2e.js`.
Suggested names are: `some.spec.error2error.js`, `some.spec.error2event.js`, `some.spec.event2error.js`, ... (1 more omitted).
A more descriptive name will do too
unicorn/prevent-abbreviations

```

### Expected behavior

There are no errors, because `e2e` is a valid word.

Also this works for the variable name, but not for file name.

### Steps to reproduce

1. `npm i eslint eslint-plugin-unicorn`
2. Create `.eslintrc.json` file
3. Create `some.spec.e2e.js` file with content: `var e2e = 1;`
4. `npx eslint .`

### Config

```json
{
    "plugins": ["unicorn"],
    "rules": {
            "error",
            {
                "whitelist": {
                    "e2e": true
                }
            }
        ]
    }
}
```

fisker:
Only replace whitelist word contains number, we can add more if we find other cases.

Fixes #567

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -304,6 +304,14 @@ const getNameReplacements = (name, options, limit = 3) => {
 		};
 	}
 
+	const whitelistWordPlaceholders = [];
+	for (const whitelistWord of whitelist.keys()) {
+		if (/\d/.test(whitelistWord) && name.includes(whitelistWord)) {
+			const index = whitelistWordPlaceholders.push(whitelistWord);
+			name = name.split(whitelistWord).join(`\0placeholder_${index}_placeholder\0`)
+		}
+	}
+
 	// Split words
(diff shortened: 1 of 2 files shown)


sindresorhus:
@fisker Would be great if you could comment on why you close a PR when you do. It's not clear to me why this was closed.

fisker:
Sorry, I commented in the  original issue https://github.com/sindresorhus/eslint-plugin-unicorn/issues/567#issuecomment-608330093 . closing it  because I found that this solution  is not a good idea.

PR #666, opened 2020-04-07
Rewrite `no-fn-reference-in-iterator`

### LINKED ISSUE #418: `no-fn-reference-in-iterator` autofix might break code
Origin code
```js
const fn = (x, y) => x + y;

[1, 2, 3].map(fn);
```
After autofix
```js
const fn = (x, y) => x + y;

[1, 2, 3].map(x => fn(x));
```

fisker:
Improvements:

- Remove all auto-fix, there is no way to know how many arguments `iterator` accept, so we can't fix it.
- Add `flatMap` method
- Ignore `lodash` and `underscore`
- Support more types of `iterator` 

Regressions:

~- `CallExpression` was supported, I removed it for now, we need handle more types, like `foo.map(lib.fn)`(MemberExpression)~ Update: fixed.

Fixes #418

### DIFF
--- a/test/no-fn-reference-in-iterator.js
+++ b/test/no-fn-reference-in-iterator.js
@@ -157,6 +157,21 @@ ruleTester.run('no-fn-reference-in-iterator', rule, {
 			})
 		),
 
+		// Actual messages
+		{
+			code: 'foo.map(fn)',
+			errors: [
+				{
+					message: 'Do not pass function `fn` directly to `map()`.',
+					suggestions: [
+						{desc: 'Replace function `fn` with `(element) => fn(element)`.'},
+						{desc: 'Replace function `fn` with `(element, index) => fn(element, index)`.'},
+						{desc: 'Replace function `fn` with `(element, index, array) => fn(element, index, array)`.'}
+					]
+				}
+			]
+		},
+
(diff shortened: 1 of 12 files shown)


test/no-fn-reference-in-iterator.js
```diff
@@ -1,114 +1,212 @@
...
-			errors,
-			output: 'foo.every(x => fn(x))'
+			code: 'foo.reduce(fn)',
+			errors: [
+				{
+					message: 'Do not pass function `fn` directly to `reduce()`.',
+					suggestions: [
+						{desc: 'Replace function `fn` with `(accumulator, element) => fn(accumulator, element)`.'},
+						{desc: 'Replace function `fn` with `(accumulator, element, index) => fn(accumulator, element, index)`.'},
+						{desc: 'Replace function `fn` with `(accumulator, element, index, array) => fn(accumulator, element, index, array)`.'}
```
fisker, line 182: Is this message too much?

fisker: Simplified a little bit ``Replace function `fn` with `… => fn(accumulator, element, index, array)`.``

sindresorhus:
Very nice work!

PR #671, opened 2020-04-09
Add `prefer-optional-catch-binding` rule

### LINKED ISSUE #670: Rule proposal: prefer-optional-catch-binding
There is  already a  proposal in  eslint  project, but not  accepted there, I'd like to have it here .

Orignal proposal: https://github.com/eslint/eslint/issues/10434

Autofix for no-unused-vars "caughtErrors: all" is  also rejected  https://github.com/eslint/eslint/issues/11640

fisker:
Fixes #670

Note: I'm not check destructuring, so the following code will pass

```js
try {} catch({message}) {}
```

This can get really complicated,

Case 1:

```js
try {} catch({nonExistsProperty = thisWillExecute()}) {}
```

In this case `nonExistsProperty` is not used, removing `thisWillExecute()` is not safe.

Case 2:

```js
try {} catch({a, b = a}) {}
```

I'm not sure we can track both `a` and `b`.

Note2: white spaces not removed.


rules/prefer-optional-catch-binding.js
```diff
@@ -0,0 +1,61 @@
...
+				messageId: ERROR_MESSAGE_ID,
+				data: {name},
+				fix: fixer => {
+					const tokenBefore = context.getTokenBefore(node);
+					const tokenAfter = context.getTokenAfter(node);
+
+					/* istanbul ignore next */
+					if (!isOpeningParenToken(tokenBefore) || !isClosingParenToken(tokenAfter)) {
+						throw new Error('Unexpected token.');
+					}
```
fisker, line 36: I don't want fixer remove the wrong token, not sure is there a case can make this throws.

sindresorhus: How is the stack trace in ESLint if you throw? It's important that the user can see where it failed, so we can fix.

If the stack trace is not useful, maybe do a `context.report` here instead?

rules/prefer-optional-catch-binding.js
```diff
@@ -0,0 +1,61 @@
...
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocumentationUrl(__filename)
+		},
+		fixable: 'code',
+		messages: {
+			[ERROR_MESSAGE_ID]: 'Remove unused catch binding paramter `{{name}}`.'
```
sindresorhus, line 58:
```suggestion
			[ERROR_MESSAGE_ID]: 'Remove unused catch binding parameter `{{name}}`.'
```

Do we really need to use the word `parameter`? I think just "catch binding" is fine.

rules/prefer-optional-catch-binding.js
sindresorhus: How is the stack trace in ESLint if you throw? It's important that the user can see where it failed, so we can fix.

If the stack trace is not useful, maybe do a `context.report` here instead?

fisker: Eslint will crash, with this

```text
\eslint-plugin-unicorn\node_modules\eslint\lib\linter\linter.js:1181
            throw err;
            ^

Error: Unexpected token.
Occurred while linting \eslint-plugin-unicorn\1.js:1
    at Object.fix (\eslint-plugin-unicorn\rules\prefer-optional-catch-binding.js:35:13)
    at normalizeFixes (\eslint-plugin-unicorn\node_modules\eslint\lib\linter\report-translator.js:178:28)
    at \eslint-plugin-unicorn\node_modules\eslint\lib\linter\report-translator.js:343:49
    at Object.report (\eslint-plugin-unicorn\node_modules\eslint\lib\linter\linter.js:905:41)
    at CatchClause>Identifier.param (\eslint-plugin-unicorn\rules\prefer-optional-catch-binding.js:25:12)
    at \eslint-plugin-unicorn\node_modules\eslint\lib\linter\safe-emitter.js:45:58
    at Object.emit (\eslint-plugin-unicorn\node_modules\eslint\lib\linter\safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (\eslint-plugin-unicorn\node_modules\eslint\lib\linter\node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (\eslint-plugin-unicorn\node_modules\eslint\lib\linter\node-event-generator.js:283:22) {
  currentNode: Node {
    type: 'Identifier',
    start: 13,
    end: 14,
    loc: SourceLocation { start: [Position], end: [Position] },
    range: [ 13, 14 ],
    name: 'e',
    parent: Node {
      type: 'CatchClause',
      start: 7,
      end: 17,
      loc: [SourceLocation],
      range: [Array],
      param: [Circular],
      body: [Node],
      parent: [Node]
    }
  }
}
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

like `SyntaxError` wrong fixing 

```text
E:\project\@forks\eslint-plugin-unicorn\1.js
  1:13  error  Parsing error: Unexpected token )

```

But more detail.

fisker:
@sindresorhus This is ready, have to land in v20 now :(

sindresorhus:
Sorry, I misunderstood. I thought you intended to fix https://github.com/sindresorhus/eslint-plugin-unicorn/pull/671/files#diff-52502d7490c2a15f41e0def7598c29a4R109 in this PR and I was waiting for that.

fisker:
Right. I forgot that, now we have time, I'll try to fix it.

fisker:
```js
try {
} catch(error) {
	var error = 1;
}
```

This one seems hard to fix, the core rule `no-unused-vars` has [lots of codes](https://github.com/eslint/eslint/blob/master/lib/rules/no-unused-vars.js) to check, really too complicated.  And it's only about `var` because you can't 

```js
try {
} catch(error) {
	const error = 1;
}
// VM170:3 Uncaught SyntaxError: Identifier 'error' has already been declared
```

OR

```js
try {
} catch(error) {
	function error(){}
}
// VM191:3 Uncaught SyntaxError: Identifier 'error' has already been declared
```

I'm going to give up.

sindresorhus:
> And it's only about var because you can't

Just leave it. I really don't care about `var`.

In general, just ignore any legacy syntax. We should not waste time on that.

fisker:
I alway try to to better, if it's not hard . :)

fisker:
Integration  test failure due to https://github.com/then/is-promise/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc , they publish a `module:type` package with wrong `main` field.

fisker:
Are we good to merge?

PR #676, opened 2020-04-10
Clarify variable names in `prefer-set-has` rule doc

bmish:


fisker:
`item` is not a good  name, MDN  use [`valueToFind `](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)

bmish:
Good point! Updated.

fisker:
Thanks

fisker:
Thanks

PR #679, opened 2020-04-10
Simplify `prefer-set-has`

fisker:
- Simplify `prefer-set-has` with `eslint-utils` module
- Add `lodash` example mentioned in #641


sindresorhus:
Nice simplification!

PR #684, opened 2020-04-14
Cache integration test fixtures

fisker:


.github/workflows/ci.yml
```diff
@@ -48,4 +48,13 @@ jobs:
         with:
           node-version: 12
       - run: npm install
+      - uses: actions/cache@v1
+        # Prevent error on re-run, ref https://github.com/actions/cache/issues/107#issuecomment-556037278
+        if: github.event_name == 'push' || github.event_name == 'pull_request'
+        with:
+          path: test/integration/fixtures
+          key: v1-build-cache-${{ github.ref }}-${{ hashFiles('package.json') }}-${{ hashFiles('test/integration/*') }}-
```
fisker, line 56: The hashes not really necessary, but the cache can't be deleted/updated, we need a way to update the cache. https://github.com/actions/cache/issues/2

fisker:
Cache these repos seems not helping, some project actually takes 15 mins to run eslint.

PR #691, opened 2020-04-22
Tweak integration tests

fisker:
- ~Set `CI` environment variable on CI, [`ci-info` stable version hasn't support github actions](https://github.com/watson/ci-info/pull/42)~
- Keep downloaded repos, some projects are huge, we don't want download again and again on local tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,6 @@ jobs:
 
   integration:
     runs-on: ubuntu-latest
-    env:
-      CI: 1
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


sindresorhus:
FYI: https://github.blog/changelog/2020-04-15-github-actions-sets-the-ci-environment-variable-to-true/

fisker:
Add argument support, `yarn integration np ora` will only test `np` and `ora`

sindresorhus:
> Add argument support, npm run integration np ora will only test np and ora

Can you add a `readme.md` in the integration directory that just documents this. Just a single sentence like the above, so this info doesn't get lost to time.

PR #706, opened 2020-05-02
`prefer-set-has`: Ignore arrays only checking existence once

### LINKED ISSUE #696: `prefer-set-has` is little bit annoying
When the `.includes()` check not in a `for` , `while` or used by another function seems not really necessary to use a `Set`.

fisker:
Fixes #696

### DIFF
--- a/rules/prefer-set-has.js
+++ b/rules/prefer-set-has.js
@@ -110,6 +110,7 @@ const isIncludesCall = node => {
 const multipleCallNodeTypes = new Set([
 	'ForOfStatement',
 	'ForStatement',
+	'ForInStatement',
 	'WhileStatement',
 	'DoWhileStatement',
 	'FunctionDeclaration',
--- a/rules/prefer-set-has.js
+++ b/rules/prefer-set-has.js
@@ -107,7 +107,7 @@ const isIncludesCall = node => {
 	);
 };
 
-const multipleCallNodeTypes = [
+const multipleCallNodeTypes = new Set([
 	'ForOfStatement',
 	'ForStatement',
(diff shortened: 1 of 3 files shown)


rules/prefer-set-has.js
```diff
@@ -115,7 +144,14 @@ const create = context => {
...
 				identifiers.length === 0 ||
-				identifiers.some(node => !isIncludesCall(node))
+				identifiers.some(identifier => !isIncludesCall(identifier))
+			) {
+				return;
+			}
+
+			if (
+				identifiers.length === 1 &&
+				identifiers.every(identifier => !isMultipleCall(identifier, node))
```
fisker, line 154: I use `.every()` on purpose, we might add option to ignore `2` :)

rules/prefer-set-has.js
```diff
@@ -107,6 +107,35 @@ const isIncludesCall = node => {
...
+	'ForOfStatement',
+	'ForStatement',
+	'WhileStatement',
+	'DoWhileStatement',
+	'FunctionDeclaration',
+	'FunctionExpression',
+	'ArrowFunctionExpression',
+	'ObjectMethod',
+	'ClassMethod'
+]);
```
fisker, line 120: These types are all I got in mind could call `.includes()` multiple times. Maybe more from other parser, need feedback.

fisker: Forgot `ForInStatement`, added in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/706/commits/aaa3638f7fdc48d4d283df6bc6595160032f45eb.

PR #707, opened 2020-05-02
`consistent-function-scoping`: Ignore IIFE

### LINKED ISSUE #685: consistent-function-scoping: option to exclude IIFE
For code intended to run in a shared global scope (e.g. client-side libraries), it is not uncommon to use an [immediately invoked function expression (IIFE)](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) to wrap parts of the code to avoid adding functions to the global scope.  This currently runs afoul of the `unicorn/consistent-function-scoping` rule:

```js
(function() {
  function onFocusChange() {
    // ...
  }
  window.addEventListener('blur', onFocusChange);
  window.addEventListener('focus', onFocusChange);
}());
```

Disabling `unicorn/consistent-function-scoping` for each function in the IIFE can be tedious, and disabling it for the entire IIFE would miss potential issues in nested functions.  Is there any chance you might consider an option to not warn about moving functions out of an IIFE?

Thanks,
Kevin

fisker:
Fixes #685

### DIFF
--- a/docs/rules/consistent-function-scoping.md
+++ b/docs/rules/consistent-function-scoping.md
@@ -71,3 +71,23 @@ function doFoo(FooComponent) {
 	return Bar;
 };
 ```
+
+[Immediately invoked function expression (IIFE)](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) are ignored:
+
+```js
+(function () {
+	function doFoo(bar) {
+		return bar;
+	}
+})();
+```
+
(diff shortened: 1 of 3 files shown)


sindresorhus:
I would add a sentence to the docs that the rule ignores IIFEs.

kevinoid:
Looks great!  Thanks very much @fisker and @sindresorhus!

PR #709, opened 2020-05-03
`prefer-string-slice`: Handle negative length in `.substr()`

### LINKED ISSUE #700: prefer-string-slice makes a breaking change for negative lengths
Consider the following autofix:

`"Sample".substr(0, "Sample".lastIndexOf('/'))` returns empty string. 
Due to ECMA-262 "If length is a negative number, it is treated as 0."

Unicorn changes it to:
`"Sample".slice(0, "Sample".lastIndexOf('/'))` returns "Sampl".

I'm not sure if autofix shouldn't be breaking. 
But i think you need to be aware of this behavior.

UPD: I see that in other places it correctly added Math.Max for `substring` but `substr` seems broken.

fisker:
- Rename `slice` to `sliceArguments`
- Handle negative length in `.substr()`
- Improve coverage
- Remove dead code

Fixes #700

### DIFF
--- a/rules/prefer-string-slice.js
+++ b/rules/prefer-string-slice.js
@@ -63,20 +63,20 @@ const create = context => {
 			const firstArgument = argumentNodes[0] ? sourceCode.getText(argumentNodes[0]) : undefined;
 			const secondArgument = argumentNodes[1] ? sourceCode.getText(argumentNodes[1]) : undefined;
 
-			let slice;
+			let sliceArguments;
 
 			if (argumentNodes.length === 0) {
-				slice = [];
+				sliceArguments = [];
 			} else if (argumentNodes.length === 1) {
-				slice = [firstArgument];
+				sliceArguments = [firstArgument];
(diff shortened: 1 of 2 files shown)


rules/prefer-string-slice.js
```diff
@@ -111,27 +118,27 @@ const create = context => {
...
+					sliceArguments = [Math.max(0, firstNumber)];
 				} else if (isLengthProperty(argumentNodes[0])) {
-					slice = [firstArgument];
+					sliceArguments = [firstArgument];
 				} else {
-					slice = [`Math.max(0, ${firstArgument})`];
+					sliceArguments = [`Math.max(0, ${firstArgument})`];
 				}
 			} else if (argumentNodes.length === 2) {
-				const secondNumber = argumentNodes[1] ? getNumericValue(argumentNodes[1]) : undefined;
```
fisker, line 127: If `argumentNodes.length === 2` , `argumentNodes[1]` has to be a node.

PR #718, opened 2020-05-04
Add `no-useless-undefined` rule

### LINKED ISSUE #683: Rule proposal: `no-useless-undefined`
# Fail

```js
function foo() {
	return undefind;
}
```

```js
let a = undefined;
```

```js
call(undefined);
```

```js
call(foo, bar, undefined);
```

```js
const {foo = undefined} = {};
```

```js
function foo({bar = undefined} = {}) {}
```

```js
function foo(bar = undefined) {}
```

# Pass

```js
function foo() {
	return;
}
```

```js
let a;
```

```js
call();
```

```js
call(foo, bar);
call(foo, bar,);
```

```js
const {foo} = {};
```

```js
function foo({bar} = {}) {}
```

```js
function foo(bar) {}
```

Discussion https://github.com/sindresorhus/eslint-plugin-unicorn/pull/636#discussion_r407150865

Not sure about this, 

```js
const noop = () => undefined;

// prefer this?
const noop = () => {};
```

fisker:
Fixes #683

### DIFF
--- a/test/no-useless-undefined.js
+++ b/test/no-useless-undefined.js
@@ -116,6 +116,6 @@ ruleTester.run('better-regex', rule, {
 			code: 'function foo([bar = undefined] = []) {}',
 			output: 'function foo([bar] = []) {}',
 			errors
-		},
+		}
 	]
 });
--- a/test/no-useless-undefined.js
+++ b/test/no-useless-undefined.js
@@ -29,7 +29,9 @@ ruleTester.run('no-useless-undefined', rule, {
 		'foo(undefined, bar);',
 		'const {foo} = {};',
 		'function foo({bar} = {}) {}',
-		'function foo(bar) {}'
+		'function foo(bar) {}',
+		// I guess nobody use this, but `yield* undefined;` is valid code, and `yield*;` is not
+		'function* foo() {yield* undefined;}'
 	],
 	invalid: [
 		{
--- a/test/no-useless-undefined.js
+++ b/test/no-useless-undefined.js
@@ -1,4 +1,5 @@
 import test from 'ava';
+import {outdent} from 'outdent';
 import avaRuleTester from 'eslint-ava-rule-tester';
 import rule from '../rules/no-useless-undefined';
 
@@ -78,7 +79,7 @@ ruleTester.run('no-useless-undefined', rule, {
 		},
 		{
 			code: 'foo(undefined, undefined);',
-			output: 'foo(undefined);',
+			output: 'foo();',
 			errors
 		},
 		{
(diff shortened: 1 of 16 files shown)


fisker:
Ready for review, hope I get everything covered, really lots of cases.

I'll fix lint issue when it's approved.

test/no-useless-undefined.js
```diff
@@ -0,0 +1,202 @@
...
+		'foo(bar,);',
+		'foo(undefined, bar);',
+		'const {foo} = {};',
+		'function foo({bar} = {}) {}',
+		'function foo(bar) {}'
+	],
+	invalid: [
+		{
+			code: 'function foo() {return undefined;}',
+			output: 'function foo() {return ;}',
```
sindresorhus, line 37: I assume the space between `return` and `;` is intentional? And that the intention is that other style rules would fix it.

fisker: I simply remove the `undefined` node for `return` and `yield`  cases, so the space is not taking care, do you want me to remove it?

sindresorhus: Yes

fisker: I tested `xo`, [`semi-spacing`](https://eslint.org/docs/rules/semi-spacing) can take care both `return ;` and `yield ;`, but I'm going to remove the space here.

fisker: Done in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/718/commits/6f4b355a1c276d7bc7007168c9653d2329ece00d

fisker:
To make this rule work together with [`array-callback-return`](https://eslint.org/docs/rules/array-callback-return#top).

1) Disable `array-callback-return` in `recommended config`
2) Set `array-callback-return` to `[
				"error",
				{
					"allowImplicit": true
				}
			]`, but I don't think we can do this in `recommended config`, maybe in `xo` ?

[`allowImplicit`](https://eslint.org/docs/rules/array-callback-return#options)

I suggest do `2` in `xo`, WDYT?

test/no-useless-undefined.js
```diff
@@ -0,0 +1,219 @@
...
+			output: outdent`
+				foo(
+					undefined,
+					bar,
+				)
+			`,
+			errors: [
+				{
+					messageId,
+					// The 2rd `undefined`
```
yakov116, line 168: 2nd ? 3rd?

fisker: 2nd, thanks

sindresorhus:
> Set array-callback-return to [ "error", { "allowImplicit": true } ], but I don't think we can do this in recommended config, maybe in xo ?

That option is `true` by default, so XO already has that.

What we do need to do is to mention in the docs that that option should not be `false`.

sindresorhus:
https://github.com/xojs/eslint-config-xo/commit/f3c5dd2816c3dbb98acad74cf137631dc8818297 Will be part of the next minor XO release.

PR #724, opened 2020-05-11
Fixed issue requiring eslint sub-rule with VSCode-eslint

Sembiance:
So eslint-plugin-unicorn causes an error when using the VSCode eslint extension:
`Error: Cannot find module 'eslint/lib/rules/no-warning-comments'`

It does work correctly when just running `eslint` from the console.

I think it's because the VSCode eslint extension loads the rules/*.js files individually and thus the baseRule isn't able to be included the way it's attempting to be included.

The fix is pretty simple, to include it directly, based on the rules file location.

This fixes the issue for VSCode eslint and also still works fine from console too.

### DIFF
--- a/rules/expiring-todo-comments.js
+++ b/rules/expiring-todo-comments.js
@@ -2,7 +2,8 @@
 const readPkgUp = require('read-pkg-up');
 const semver = require('semver');
 const ci = require('ci-info');
-const baseRule = require('eslint/lib/rules/no-warning-comments');
+const path = require('path');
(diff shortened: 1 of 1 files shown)


sindresorhus:
Tests are failing.

Sembiance:
Yah, I noticed that. Maybe my fix isn't the proper way to fix it for a general release. Since unicorn works fine via eslint command line, I'm just gonna close this pull request and it'll be up to the VSCode eslint folks to fix their side.

PR #735, opened 2020-05-17
Add `prefer-array-find` rule

### LINKED ISSUE #730: Rule proposal: `prefer-array-find`
Prefer `.find()` over `.filter()[0]`, since `find` breaks the loop as soon as it finds a match.

## Fail
```js
const item = array.filter(x => x === '🦄')[0];
```

## Pass
```js
const item = array.find(x => x === '🦄');
```

fisker:
Fixes #730


test/prefer-find.js
```diff
@@ -0,0 +1,113 @@
...
+					;
+			`,
+			output: outdent`
+				const item = array
+					// comment 1
+					.find(
+						// comment 2
+						x => x === '🦄'
+					)
+					// comment 4
```
yakov116, line 107: 1,2........ 4 😄

fisker: Because we are not 3 year baby :smile:

docs/rules/prefer-find.md
```diff
@@ -0,0 +1,21 @@
+# Prefer `.find(…)` over first element of `.filter(…)` result
```
sindresorhus, line 1:
```suggestion
# Prefer `.find(…)` over the first element from `.filter(…)`
```

docs/rules/prefer-find.md
```diff
@@ -0,0 +1,21 @@
...
+
+[`Array#find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) breaks the loop as soon as it finds a match.
+
+This rule is fixable.
+
+## Fail
+
+```js
+const item = array.filter(x => x === '🦄')[0];
+```
```
sindresorhus, line 11: How about

```js
const [item] = array.filter(x => x === '🦄');
```

and

```js
const result = array.filter(x => x === '🦄');
const item = result[0];
```

?

rules/prefer-find.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+const methodSelector = require('./utils/method-selector');
+
+const MESSAGE_ID_ZERO_INDEX = 'prefer-find-over-filter-zero-index';
```
sindresorhus, line 5:
```suggestion
const MESSAGE_ID_ZERO_INDEX = 'prefer-array-find-over-filter-zero-index';
```

sindresorhus:
The rule should be named `prefer-array-find`.

fisker:
I thought about this rule name too, I agree it should be `prefer-array-find`, and I even want rename `no-reduce` to `no-array-reduce`, then I realize that rule already published in v20.

docs/rules/prefer-find.md
fisker: The first one:

```js
const [item = foo] = array.filter(x => x === '🦄');
```

-> 

```js
const item = array.find(x => x === '🦄') || foo;
```

?

Not safe, find result could be `0 | false | '' | null`

The second one seems more complicated.

fisker:
```js
function foo([bar] = array.filter(baz)) {}
```

This can't fix right?

fisker:
I'm going to add suggestion to

```js
const [foo = bar] = array.filter(fn);
```

Suggestion 1:

```js
const foo = array.find(fn) || bar;
```

Suggestion 2:

```js
const foo = array.find(fn) ?? bar;
```


God, I thought this is simple, so I started it.

fisker:
Following cases not handled.

```js
const foo = array.filter(fn);
const bar = foo.shift();
```

I didn't handle this, because `.shift()` has side effect, we should know it's only called once, and not in a `function/while/for`.

```js
const foo = array.filter(fn);
while (foo.shift()) {}
```

---

```js
const foo = array.filter(fn);
const [bar] = foo;
```

I didn't handle this, because we need handle/exclude this

```js
const foo = array.filter(fn);
const [bar = baz] = foo;
```

I use suggestion for `const [foo = baz] = array.filter(fn);`, but we can't use suggestion for this. because we are reportting problem on `.filter`, suggest to use `??` or `||` will be weird, we should decide ignore this case or do something different.

---

```js
let bar;
const foo = array.filter(fn);
[bar] = foo;
```

Same as second one.
```js
let bar;
const foo = array.filter(fn);
[bar = baz] = foo;
```

docs/rules/prefer-find.md
sindresorhus: > Maybe just

👍

sindresorhus:
> function foo([bar] = array.filter(baz)) {}
This can't fix, right?

No point in handling that. I have never seen such code in practice.

fisker:
@oscarmarcelo Can you  help review this?

docs/rules/prefer-array-find.md
```diff
@@ -0,0 +1,33 @@
...
+This rule is fixable unless default values are used in declaration or assignment.
+
+## Fail
+
+```js
+const item = array.filter(x => x === '🦄')[0];
+```
+
+```js
+const item = array.filter(x => x === '🦄').shift();
```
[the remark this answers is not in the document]
fisker: No... there is no `.findRight()`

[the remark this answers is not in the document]
fisker:
```js
const item = [1, 2, 3].filter(x => x < 3).shift();
// -> 1
```

```js
const item = [1, 2, 3].find(x => x < 3);
// -> 1
```

```js
const item = [1, 2, 3].filter(x => x < 3).pop();
// -> 2
```

sindresorhus:
Nice work as always, @fisker 👍

PR #745, opened 2020-05-24
Use singular array element variable name in autofix for `no-for-loop` rule

### LINKED ISSUE #743: `no-for-loop` rule could use smarter element variable name in autofix
If the loop is iterating an array with a plural name (such as `plugins`), the generated element variable in the autofix could use the singular version of the array name (`plugin`), instead of just the generic name `element`.

Original code:

```js
for (let i = 0; i < plugins.length; i++) {
  console.log(plugins[i]);
}
```

After autofix:

```js
for (const element of plugins) {
  console.log(element);
}
```

Desired autofix:

```js
for (const plugin of plugins) {
  console.log(plugin);
}
```

bmish:
If the loop is iterating an array with a plural name (such as `plugins`), the generated element variable in the autofix could use the singular version of the array name (`plugin`), instead of just the generic name `element`.

Fixes #743. CC: @mongoose700

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -270,16 +270,10 @@ const getChildScopesRecursive = scope => [
 
 const getSingularName = originalName => {
 	const singularName = singular(originalName);
-	if (
-		!singularName ||
-		singularName === originalName
-	) {
-		// Bail if we can't produce a singular name that differs from the original name.
-		return;
+	if (singularName !== originalName) {
+		return singularName;
 	}
-
-	return singularName;
-};
+}
 
 const create = context => {
 	const sourceCode = context.getSourceCode();
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -3,6 +3,7 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const isLiteralValue = require('./utils/is-literal-value');
 const {flatten} = require('lodash');
 const avoidCapture = require('./utils/avoid-capture');
+const {singular} = require('pluralize');
 
 const defaultElementName = 'element';
 const isLiteralZero = node => isLiteralValue(node, 0);
@@ -267,6 +268,19 @@ const getChildScopesRecursive = scope => [
 	...flatten(scope.childScopes.map(scope => getChildScopesRecursive(scope)))
 ];
 
+const getSingularName = originalName => {
+	const singularName = singular(originalName);
+	if (
+		!singularName ||
+		singularName === originalName
+	) {
+		// Bail if we can't produce a singular name that differs from the original name.
+		return;
+	}
+
+	return singularName;
+};
+
 const create = context => {
 	const sourceCode = context.getSourceCode();
 	const {scopeManager} = sourceCode;
@@ -342,7 +356,7 @@ const create = context => {
 
 					const index = indexIdentifierName;
 					const element = elementIdentifierName ||
-						avoidCapture(defaultElementName, getChildScopesRecursive(bodyScope), context.parserOptions.ecmaVersion);
+						avoidCapture(getSingularName(arrayIdentifierName) || defaultElementName, getChildScopesRecursive(bodyScope), context.parserOptions.ecmaVersion);
 					const array = arrayIdentifierName;
 
 					let declarationElement = element;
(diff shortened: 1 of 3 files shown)


fisker:
https://github.com/facebook/react/blob/3ca1904b37ad1f527ff5e31b51373caea67478c5/packages/react-devtools-shared/src/__tests__/storeStressSync-test.js#L117

bmish:
@fisker just updated to avoid using reserved JavaScript keywords.

fisker:
```js
for (const case_ of cases) {}
```

```js
for (const element of cases) {}
```

Which one do you prefer?

bmish:
@fisker no strong preference, either works for me.

test/no-for-loop.js
```diff
@@ -482,6 +482,46 @@ ruleTester.run('no-for-loop', rule, {
...
+		`),
+		// Singularization (capital letters, multiple words):
+		testCase(outdent`
+			for (let i = 0; i < LARGE_CITIES.length; i++) {
+				console.log(LARGE_CITIES[i]);
+			}
+		`, outdent`
+			for (const LARGE_CITY of LARGE_CITIES) {
+				console.log(LARGE_CITY);
+			}
```
snowteamer, line 524: Does the singularization also work for camel-case identifiers, e.g. `largeCities` ?

bmish: Yep!

```
Welcome to Node.js v12.16.2.
> const pluralize = require('pluralize')
undefined
> pluralize.singular('largeCities')
'largeCity'
```

fisker:
I prefer

```js
for (const case_ of cases) {}
```

over

```js
for (const element of cases) {}
```

And

```js
for (const function_ of functions) {}
```

over

```js
for (const element of functions) {}
```


So, we need fix #747 first, then we can use `avoidCapture` to change `case` to `case_`

sindresorhus:
I also prefer `case_`.

test/no-for-loop.js
snowteamer: That's great ! It would be nice to add an example and a test for this case.

snowteamer:
I kinda dislike both of them, since I think one should try not using identifiers like `cases` or `functions` in the first place.

That being said, I also prefer the `case_` form, because I think the trailing underscore may serve as a nice visual reminder to the user that she should think about renaming the variable.

test/no-for-loop.js
bmish: Added a camelCase test case.

bmish:
Rebased to take advantage of `avoidCapture` for producing the `case_` variable name.

fisker:
```js
invalid: [
	...[
	  ['plugin', 'plugins'],
	  ['person', 'people'],
	  ['element', 'list'],
	].map(([elementName, arrayName]) => 
	  testCase(
	    `for(const i = 0; i < ${arrayName}.length; i++) {console.log(${arrayName}[i])}`,
	    `for(const ${elementName} of ${arrayName}) {console.log(${elementName})}`,
	  )
	),
]
```

Can you rewrite tests to this? So we can test more cases easier.

fisker:
I ask you again, don't squash!

bmish:
@fisker good idea, updated test cases. Sorry, squashing is a habit, I will try to avoid it.

fisker:
Add some word ends with `s`? Even if it was wrong. 
Also add one test that uses `i`.

fisker:
Add a word that has more than one plural, `girlsAndBoys`?

rules/no-for-loop.js
```diff
@@ -267,6 +268,19 @@ const getChildScopesRecursive = scope => [
...
+	if (
+		!singularName ||
+		singularName === originalName
+	) {
+		// Bail if we can't produce a singular name that differs from the original name.
+		return;
+	}
+
+	return singularName;
+};
```
fisker, line 282:
```suggestion
const getSingularName = originalName => {
	const singularName = singular(originalName);
	if (singularName !== originalName) {
		return singularName;
	}
}
```

Will this work?

bmish: Yep that will work.

sindresorhus:
Nice work :)

fisker:
@bmish Do you interested in improving `no-fn-reference-in-iterator` rule? We use `element` there too. No pressure.

bmish:
@fisker I'll see if I have a chance sometime. Can you file a ticket for now?

fisker:
@bmish #754

PR #748, opened 2020-05-25
`no-for-loop`: Use generator function instead of return array

fisker:
Is this more readable?


sindresorhus:
Yup. More readable.

PR #749, opened 2020-05-25
Avoid naming collision with default array element variable in autofix for `no-for-loop` rule

### LINKED ISSUE #747: `no-for-loop` fix to invalid code
We didn't use [`avoidCapture`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/utils/avoid-capture.js) in this rule.

```js
			for (let i = 0; i < arr.length; i += 1) {
				console.log(arr[i]);
				const element = foo();
				console.log(element);
			}
```

Fixed to

```js
			for (const element of arr) {
				console.log(element);
				const element = foo();
				console.log(element);
			}
```

bmish:
Fixes #747. Will also help prepare us for #745.

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -2,6 +2,7 @@
 const getDocumentationUrl = require('./utils/get-documentation-url');
 const isLiteralValue = require('./utils/is-literal-value');
 const {flatten} = require('lodash');
+const avoidCapture = require('./utils/avoid-capture');
 
 const defaultElementName = 'element';
 const isLiteralZero = node => isLiteralValue(node, 0);
@@ -261,6 +262,13 @@ const getReferencesInChildScopes = (scope, name) => {
 	];
 };
 
+const getChildScopesRecursive = scope => {
+	return [
+		scope,
+		...flatten(scope.childScopes.map(s => getChildScopesRecursive(s)))
+	];
+};
+
 const create = context => {
 	const sourceCode = context.getSourceCode();
 	const {scopeManager} = sourceCode;
@@ -335,7 +343,7 @@ const create = context => {
 					const shouldGenerateIndex = isIndexVariableUsedElsewhereInTheLoopBody(indexVariable, bodyScope, arrayIdentifierName);
 
 					const index = indexIdentifierName;
-					const element = elementIdentifierName || defaultElementName;
+					const element = elementIdentifierName || avoidCapture(defaultElementName, getChildScopesRecursive(bodyScope), context.parserOptions.ecmaVersion);
 					const array = arrayIdentifierName;
 
 					let declarationElement = element;
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -264,7 +264,7 @@ const getReferencesInChildScopes = (scope, name) => {
 
 const getChildScopesRecursive = scope => [
 	scope,
-	...flatten(scope.childScopes.map(s => getChildScopesRecursive(s)))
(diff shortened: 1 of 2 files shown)


fisker:
Please add more cases like https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/catch-error-name.js or
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/test/prevent-abbreviations.js

bmish:
@fisker I added one more test case.

test/no-for-loop.js
```diff
@@ -482,6 +482,121 @@ ruleTester.run('no-for-loop', rule, {
...
+					console.log(arr[i], element);
+				}
+			}
+		`, outdent`
+			for (const element_ of arr) {
+				function element__(element) {
+					console.log(element_, element);
+				}
+			}
+		`),
```
fisker, line 528: This one need fix, we are not checking the childscope.

bmish: Updated to check child scopes.

bmish:
@fisker thanks a lot for adding additional test cases. I fixed the failure.

rules/no-for-loop.js
```diff
@@ -335,7 +343,7 @@ const create = context => {
 					const shouldGenerateIndex = isIndexVariableUsedElsewhereInTheLoopBody(indexVariable, bodyScope, arrayIdentifierName);
 
 					const index = indexIdentifierName;
-					const element = elementIdentifierName || defaultElementName;
+					const element = elementIdentifierName || avoidCapture(defaultElementName, getChildScopesRecursive(bodyScope), context.parserOptions.ecmaVersion);
```
fisker, line 346: I think this is not right. 

```js
			for (let i = 0; i < arr.length; i += 1) {
				console.log(arr[1]);
				function foo(element) {
					console.log(element);
				}
			}
```

It should allow use variable `element`.

this `foo` function scope should not included, because it didn't use `arr[1]`, this is why `avoidCapture` didn't recursive automaticlly.

We need collect copes from references.

---
BTW: don't squash commits, this makes review harder.

bmish: A few questions for you.

1. By `arr[1]`, do you mean `arr[i]`? The rule doesn't report a violation with `arr[1]`.
2. In your example, I see that we technically can use `element` as our new array variable name, but wouldn't that be unnecessarily confusing? The [no-shadow](https://eslint.org/docs/rules/no-shadow) rule suggests this is bad practice.

fisker: 1. Yes, `arr[1]`, sorry.
2. Yes `no-shadow` suggest no variable shadowing, it's default to `function`, do you use `all` option? My example is about function scope, but there are other scopes.

This example

```js
for (let i = 0; i < errors.length; i++) {
  console.log(errors[i]);

  try {
    doSomethingThrowError();
  } catch(error) {}
}
```

After #745 we will fix it to 

```js
for (const error_ of errors) {
  console.log(error_);

  try {
    doSomethingThrowError();
  } catch(error) {}
}
```

I don't want use `error_`, I can use `error`.

fisker: Anyway, this is not very important, let's keep it this way now.

I'm going to check this PR again.

fisker:
Thanks for fixing this.

rules/no-for-loop.js
bmish: Thanks for the explanation. I will follow-up with your recommendation if I have a chance.

PR #756, opened 2020-05-29
Ignore more types in `no-fn-reference-in-iterator` and `no-reduce` rule

### LINKED ISSUE #755: unicorn/no-fn-reference-in-iterator - wrong context
The lint is claiming a MongoDB query (`.find`)

```js
// Example
// error  Do not pass function `collection` directly to `.find(…)`  unicorn/no-fn-reference-in-iterator
const results = await collection
  .find({
    $and: [cursorQuery, params.query]
  }, {
    projection: params.projection
  })
  .sort($sort)
  .limit(params.limit + 1)
  .toArray()
```

fisker:
- Ignore more types that can't be a `function` on `no-fn-reference-in-iterator` and `no-reduce` rule.

Including some unsafe node types, like

```
a.find(a = fn);
a.find(await fn);
a.find(fn());
a.find(fn1 || fn2);
a.find(new Foo());
a.find(new Function());
a.find(tag`unicorn`);
a.find(this);
```

They could be function, but not very common. Some might need discuss. like `fn()` and `fn1 || fn2`.

- Fix some logic error in `no-reduce` rule.

Fixes #755

### DIFF
--- a/rules/no-fn-reference-in-iterator.js
+++ b/rules/no-fn-reference-in-iterator.js
@@ -2,6 +2,7 @@
 const {isParenthesized} = require('eslint-utils');
 const getDocumentationUrl = require('./utils/get-documentation-url');
 const methodSelector = require('./utils/method-selector');
+const {notFunctionSelector} = require('./utils/not-function');
 
 const ERROR_WITH_NAME_MESSAGE_ID = 'error-with-name';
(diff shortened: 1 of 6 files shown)


fisker:
Do not merge, adding more.

PR #757, opened 2020-06-01
`no-null`: allow ignore `==` and `!=`

### LINKED ISSUE #703: Can not ignore unicorn/no-null in some case
.eslintrc.js: rules `'unicorn/no-null': 0,`

code (Note the line of code marked `FIXME`): 

```jsx
// import 'core-js/stable';
// import 'regenerator-runtime/runtime';
import uniq from 'lodash/uniq';
import React from 'react';
import ReactDom from 'react-dom';
import { createStore, combineReducers, applyMiddleware, compose } from 'redux';
import { Provider } from 'react-redux';
import { createBrowserHistory } from 'history';
import { ConnectedRouter, connectRouter, routerMiddleware } from 'connected-react-router';
import createSagaMiddleware from 'redux-saga';

import modelManager, { getReducer, getSaga, SHOW, HIDE } from '@/models';
import initRequest from '@/utils/request';
import App from '@/routes/App';

const history = createBrowserHistory();

// Build the middleware for intercepting and dispatching navigation actions
const reduxRouterMiddleware = routerMiddleware(history);
const reduxSagaMiddleware = createSagaMiddleware();

// window.__REDUX_DEVTOOLS_EXTENSION__
// const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({
//   //
// }) || compose
const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
const connectRouterReducer = connectRouter(history);

const allModelKeys = modelManager.keys();

const injectedModels = [];
const loadModels = keys => {
  injectedModels.push(
    ...uniq(keys)
      .filter(key => !injectedModels.some(m => m.namespace === key) && allModelKeys.includes(key))
      .map(key => {
        const model = modelManager.load(key);
        if (model == null) { // **FIXME target line 1** 
          throw new Error(`[${key}]: Model load fails`);
        }
        return model.default || model;
      })
      .filter(Boolean)
  );
};

// redux 初始化默认加载的model
const initialModels = ['global', 'login', 'user', 'venue', 'datatable', 'extfield'];

loadModels(initialModels);

const createNewReducers = () => {
  const reducers = {};
  injectedModels.forEach(model => {
    reducers[model.namespace] = getReducer(model);
  });
  return combineReducers({
    ...reducers,
    loading: (state, action) => {
      switch (action.type) {
        case SHOW:
        case HIDE:
          return {
            ...state,
            effects: {
              ...state.effects,
              [action.payload.actionType]: action.type === SHOW,
            },
          };
        default:
          return state || {};
      }
    },
    router: connectRouterReducer,
  });
};

const runSaga = (keys = [], hasLoadKeys = []) => {
  injectedModels
    .filter(model => keys.includes(model.namespace) && !hasLoadKeys.includes(model.namespace))
    .forEach(model => {
      reduxSagaMiddleware.run(getSaga(model));
    });
};

const store = createStore(
  createNewReducers(),
  // preloadedState
  {
    loading: {
      effects: {},
    },
  },
  composeEnhancers(
    applyMiddleware(
      reduxRouterMiddleware,
      // promiseMiddleware
      // eslint-disable-next-line unicorn/consistent-function-scoping
      () => next => action => {
        const { type } = action;
        if (typeof type === 'string') {
          const [namespace, effectName] = type.split('/');
          // model.effects 的一定都是 effect
          if (
            injectedModels
              .filter(model => model.namespace === namespace)
              .some(model => model.effects && model.effects[effectName])
          ) {
            return new Promise((resolve, reject) => {
              next({
                ...action,
                _resolve: resolve,
                _reject: reject,
              });
            });
          }
        }
        return Promise.resolve(next(action));
      },
      reduxSagaMiddleware
    )
  )
);

initRequest(store.dispatch, store);

runSaga(initialModels, []);

const appManager = {
  injectModel: model => {
    if (model == null || model.length === 0) { // **FIXME target line 2**
      return;
    }
    const hasLoadKeys = injectedModels.map(m => m.namespace);
    loadModels(model);
    store.replaceReducer(createNewReducers());
    runSaga(model, hasLoadKeys);
  },
};

ReactDom.render(
  <Provider store={store}>
    <ConnectedRouter history={history}>
      <App app={appManager} />
    </ConnectedRouter>
  </Provider>,
  // eslint-disable-next-line unicorn/prefer-query-selector
  document.getElementById('root')
);

```

It report 

image

If I fix any one error, the other will no longer report the error.

fisker:
Fixes #703

### DIFF
--- a/docs/rules/no-null.md
+++ b/docs/rules/no-null.md
@@ -44,6 +44,20 @@ Strict equality(`===`) and strict inequality(`!==`) is ignored by default.
 if (foo === null) {}
 ```
 
+### checkLooseEquality
+
+Type: `boolean`\
+Default: `true`
+
+Set to `false` to ignore loose equality(`==`) and equality inequality(`!=`).
+
+#### Pass
+
+```js
+// eslint unicorn/no-null: ["error", {"checkLooseEquality": false}]
+if (foo == null) {}
+```
+
 ## Why
 
 - [“Intent to stop using `null` in my JS code”](https://github.com/sindresorhus/meta/issues/7).
--- a/docs/rules/no-null.md
+++ b/docs/rules/no-null.md
@@ -49,7 +49,7 @@ if (foo === null) {}
 Type: `boolean`\
 Default: `true`
 
-Set to `false` to ignore loose equality(`==`) and equality inequality(`!=`).
+Set to `false` to ignore loose equality(`==`) and loose inequality(`!=`).
 
 #### Pass
 
--- a/rules/no-null.js
+++ b/rules/no-null.js
@@ -23,7 +23,8 @@ const isLooseEqual = node => node.type === 'BinaryExpression' && ['==', '!='].in
 const isStrictEqual = node => node.type === 'BinaryExpression' && ['===', '!=='].includes(node.operator);
 
 const create = context => {
-	const {checkStrictEquality} = {
+	const {checkStrictEquality, checkLooseEquality} = {
+		checkLooseEquality: true,
 		checkStrictEquality: false,
 		...context.options[0]
 	};
@@ -38,7 +39,10 @@ const create = context => {
 			/* istanbul ignore next */
 			const {parent = {}} = node;
 
-			if (!checkStrictEquality && isStrictEqual(parent)) {
+			if (
+				(!checkStrictEquality && isStrictEqual(parent)) ||
+				(!checkLooseEquality && isLooseEqual(parent))
+			) {
 				return;
 			}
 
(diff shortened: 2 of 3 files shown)


sindresorhus:
I disagree with this. The whole point of the rule is to disallow `null`. Seems like `eqeqeq` will get an option for `undefined` soon: https://github.com/eslint/eslint/pull/13303 I think we should just wait for that and then add to the docs that users should set that option if they use `eqeqeq`.

fisker:
I don't think `eqeqeq` is going to have that option, https://github.com/eslint/eslint/pull/13303 is closed, at least not "soon". 

We can remove it if `undefined` option added in `eqeqeq`.

sindresorhus:
Adding "yet another option" is the lazy way out. I'd rather just add a note about the incompatibility with `eqeqeq` and encourage users to get ESLint to fix this. If we add this option, users will use it and ESLint will never get fixed...

PR #764, opened 2020-06-02
`prevent-abbreviations`: Fix optional parameter

### LINKED ISSUE #763: prevent-abbreviations removes optional declaration of TypeScript parameter
Repro:

```ts
const foo = (extraParams?: string) => {}
foo()
```
autofixes to
```ts
const foo = (extraParameters: string) => {}
foo()
```

which no longer compiles.

fisker:
Fixes #763

### DIFF
--- a/rules/utils/rename-identifier.js
+++ b/rules/utils/rename-identifier.js
@@ -28,7 +28,7 @@ function renameIdentifier(identifier, name, fixer, sourceCode) {
 
 	// `typeAnnotation`
 	if (identifier.typeAnnotation) {
-		return fixer.replaceText(identifier, `${name}${sourceCode.getText(identifier.typeAnnotation)}`);
+		return fixer.replaceText(identifier, `${name}${identifier.optional ? '?' : ''}${sourceCode.getText(identifier.typeAnnotation)}`);
 	}
 
 	return fixer.replaceText(identifier, name);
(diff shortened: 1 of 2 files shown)


rules/utils/rename-identifier.js
```diff
@@ -28,7 +28,7 @@ function renameIdentifier(identifier, name, fixer, sourceCode) {
 
 	// `typeAnnotation`
 	if (identifier.typeAnnotation) {
-		return fixer.replaceText(identifier, `${name}${sourceCode.getText(identifier.typeAnnotation)}`);
+		return fixer.replaceText(identifier, `${name}${identifier.optional ? '?' : ''}${sourceCode.getText(identifier.typeAnnotation)}`);
```
sindresorhus, line 31: This is kinda fragile. I'm surprised ESLint doesn't have a fixer utility to simply rename a `node`. If we could just set the `identifier.name` property instead, we wouldn't have to specially handle `identifier.optional` and `identifier.typeAnnotation`.

What do you think about proposing such utility to ESLint?

---

There might also be other rules of ours where we need to handle `identifier.optional`

futpib: There might be a way to modify the node, render it and then use `teplaceText`. The only tricky part is to choose the right renderer for JS/TS. This begs for a utility module

fisker: @futpib What do you mean `render`? Get the original text?

PR #772, opened 2020-06-17
`consistent-function-scoping`: Fix wrong detect of IIFE

### LINKED ISSUE #770: consistent-function-scoping: regression: ignored in function argument
The `consistent-function-scoping` rule no longer warns about functions defined inside a function being passed as an argument.  For example:

```js
process.nextTick(() => {
  function returnsZero() {
    return true;
  }
  process.exitCode = returnsZero();
});
```

In `eslint-plugin-unicorn@19.0.1` this would cause:

    example.js:2:3: Move function `returnsZero` to the outer scope. [Error/unicorn/consistent-function-scoping]

In `eslint-plugin-unicorn@20.0.0` and later (up to 7781c3f, current `master`) it does not cause an error.

I bisected the regression to 1f4413d from #707.  Perhaps the IIFE identification is a bit too aggressive?  (I wouldn't consider any of the functions in the above example to be immediately-invoked, but perhaps I'm missing something?)

Thanks again,
Kevin

fisker:
Fixes #770

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,7 +51,8 @@ module.exports = {
 				'unicorn/prefer-array-find': 'error',
 				'unicorn/prefer-dataset': 'error',
 				'unicorn/prefer-event-key': 'error',
-				'unicorn/prefer-flat-map': 'error',
+				// TODO: Enable this by default when targeting Node.js 12.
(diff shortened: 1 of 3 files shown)


sindresorhus:
Can you fix the conflict?

PR #779, opened 2020-06-22
Refactor rules to use `messageId`

fisker:
- [x] better-regex
- [x] catch-error-name
- [x] consistent-function-scoping
- [x] custom-error-definition
- [x] error-message
- [x] escape-case
- [x] expiring-todo-comments
- [x] explicit-length-check
- [x] filename-case
- [x] import-index
- [x] import-style
- [x] new-for-builtins
- [x] no-abusive-eslint-disable
- [x] no-array-instanceof
- [x] no-console-spaces
- [x] no-fn-reference-in-iterator
- [x] no-for-loop
- [x] no-hex-escape
- [x] no-keyword-prefix
- [x] no-nested-ternary
- [x] no-new-buffer
- [x] no-null
- [x] no-object-as-default-parameter
- [x] no-process-exit
- [x] no-reduce
- [x] no-unreadable-array-destructuring
- [x] no-unsafe-regex
- [x] no-unused-properties
- [x] no-useless-undefined
- [x] no-zero-fractions
- [x] number-literal-case
- [x] prefer-add-event-listener
- [x] prefer-array-find
- [x] prefer-dataset
- [x] prefer-event-key
- [x] prefer-exponentiation-operator
- [x] prefer-flat-map
- [x] prefer-includes
- [x] prefer-modern-dom-apis
- [x] prefer-negative-index
- [x] prefer-node-append
- [x] prefer-node-remove
- [x] prefer-number-properties
- [x] prefer-optional-catch-binding
- [x] prefer-query-selector
- [x] prefer-reflect-apply
- [x] prefer-replace-all
- [x] prefer-set-has
- [x] prefer-spread
- [x] prefer-starts-ends-with
- [x] prefer-string-slice
- [x] prefer-text-content
- [x] prefer-trim-start-end
- [x] prefer-type-error
- [ ] ~prevent-abbreviations~ This message is too complicated
- [x] regex-shorthand
- [x] string-content
- [x] throw-new-error


sindresorhus:
Wow. This must have been a huge effort. Nice work!

Issue #782, opened 2020-06-24
Unsafe replacement with .startsWith() when doing xo --fix


timlesallen:
`/^NN|VB/.test(partOfSpeech`) →  `partOfSpeech.startsWith('NN|VB')`

This actually broke my code. :)

I hope I'm filing this in the right place.  Or should it rather be under `xo`?

fisker:
This is the right place.

fisker:
Would you like to send a PR fix this? This is easy one, just add `|` to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/459408580b953cfd9940d4d2a0df0f397f537739/rules/prefer-starts-ends-with.js#L14

timlesallen:
@fisker Ok will do.

timlesallen:
@fisker There you go :) https://github.com/sindresorhus/eslint-plugin-unicorn/pull/783

PR #783, opened 2020-06-24
`prefer-starts-ends-with`: Fix pipe in regex

timlesallen:
Fix the case where `/^A|B/.test(s)` → `s.startsWith('A|B')`.

Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/782

### DIFF
--- a/test/prefer-starts-ends-with.js
+++ b/test/prefer-starts-ends-with.js
@@ -27,7 +27,8 @@ const validRegex = [
 	/^foo/i,
 	/^foo/m,
 	/^foo/im,
-	/^A|B/
+	/^A|B/,
+	/A|B$/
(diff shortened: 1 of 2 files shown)


test/prefer-starts-ends-with.js
```diff
@@ -26,7 +26,8 @@ const validRegex = [
 	/\^foo/,
 	/^foo/i,
 	/^foo/m,
-	/^foo/im
+	/^foo/im,
+	/^A|B/
```
fisker, line 30:
```suggestion
	/^A|B/,
	/A|B$/
```

fisker:
Thank you, I'll merge after CI pass.

PR #786, opened 2020-06-29
`consistent-function-scoping`: Fix wrong report when using JSX

fisker:
~I'm not sure if this regression of #551, but I think the old one also not right, the function position should not effect the result.~

~In this PR, I simply ignore all functions if JSX is used.~

~Maybe we can improve this rule later.~

It should be regression of #551

### DIFF
--- a/test/consistent-function-scoping.js
+++ b/test/consistent-function-scoping.js
@@ -608,7 +608,7 @@ ruleTester.run('consistent-function-scoping', rule, {
 			`,
 			errors: [createError('function \'baz\'')]
 		},
-		// Jsx
+		// JSX
 		{
 			// Function `c` is not used, but we can't tell
 			code: outdent`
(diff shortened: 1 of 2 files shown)


fisker:
Incident https://www.githubstatus.com/, please don't merge.

test/consistent-function-scoping.js
```diff
@@ -593,6 +607,23 @@ ruleTester.run('consistent-function-scoping', rule, {
 				)
 			`,
 			errors: [createError('function \'baz\'')]
+		},
+		// Jsx
```
sindresorhus, line 611:
```suggestion
		// JSX
```

fisker:
The  logic  seems still  not  right , I'll check  again

sindresorhus:
Is this PR ready to be merged?

fisker:
No, it's not  fixing anything.  Sorry!


closing now

PR #789, opened 2020-07-03
Add `import-style` rule

### LINKED ISSUE #211: Rule proposal: Don't import `util` Node.js core module directly

futpib:
fixes #211

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#211: Rule proposal: Don't import `util` Node.js core module directly](https://issuehunt.io/repos/55832243/issues/211)
---
</details>


fisker:
Are we going to allow check `./util'?

futpib:
I would prefer not to. If we are to support relative paths properly, we are to invite all of complexities of `eslint-plugin-import` (pluggable resolvers). I'd rather leave it as undefined behaviour (which it currently is), advise against it in the docs or throw if relative path is found in options.

manovotny:
I stumbled across this while looking for a way to enforce the use of `const` / `require` instead of accidental `import` for apps / packages that haven't opted into the [experimental support in newer node versions](https://nodejs.org/api/esm.html#esm_enabling) yet. 

Do y'all think this would be something appropriate to add to this `import-style` rule or something completely different? I also didn't see anything else out there in native eslint, but please forgive me if I overlooked something.

rules/import-style.js
```diff
@@ -0,0 +1,320 @@
...
+			messageId: MESSAGE_ID,
+			data
+		});
+	};
+
+	return templates.visitor({
+		ImportDeclaration(node) {
+			if (!checkImport) {
+				return;
+			}
```
fisker, line 180: Can we add `listener`, and don't add `ImportDeclaration` at all if `checkImport === false`

docs/rules/import-style.md
```diff
@@ -0,0 +1,87 @@
+# Enforce specific import styles per module
+
+Sometimes a module contains unrelated functions like `util`, thus it is a good practice to enforce destructuring or named imports here, other times in modules like `path` it is good to do default import as they have similar functions, all likely to be utilised.
+
+This rule defines 4 import types:
+* `unassigned` - `import 'foo'` or `require('foo')`
```
fisker, line 6: Isn't this called `empty import`?

rules/import-style.js
```diff
@@ -0,0 +1,320 @@
...
+
+			const allowedImportStyles = styles.get(moduleName);
+			const actualImportStyles = getActualAssignmentTargetImportStyles(assignmentTargetNode);
+
+			report(node, moduleName, actualImportStyles, allowedImportStyles);
+		},
+
+		'ExpressionStatement > CallExpression[callee.name=\'require\']'(node) {
+			if (!checkRequire) {
```
fisker, line 222: `ExpressionStatement > CallExpression[callee.name="require"][arguments.length=1][arguments.0.type="Literal"]`

This work?

docs/rules/import-style.md
futpib: The name is inspired by [no-unassigned-import](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unassigned-import.md) rule

rules/import-style.js
```diff
@@ -0,0 +1,316 @@
...
+				const allowedImportStyles = styles.get(moduleName);
+				const actualImportStyles = getActualAssignmentTargetImportStyles(assignmentTargetNode);
+
+				report(node, moduleName, actualImportStyles, allowedImportStyles);
+			}
+		};
+	}
+
+	if (checkRequire) {
+		visitor = {
```
fisker, line 223: Why not `vister.xxx = `? I think it can  reduce indentation level.

rules/import-style.js
```diff
@@ -0,0 +1,316 @@
...
+
+				const allowedImportStyles = styles.get(moduleName);
+				const actualImportStyles = getActualAssignmentTargetImportStyles(assignmentTargetNode);
+
+				report(node, moduleName, actualImportStyles, allowedImportStyles, true);
+			}
+		};
+	}
+
+	return templates.visitor(visitor);
```
fisker, line 251: I haven't see  `templates.visitor`, should it skip if `vistor` is  empty?

rules/import-style.js
```diff
@@ -0,0 +1,316 @@
...
+			}
+		};
+	}
+
+	if (checkRequire) {
+		visitor = {
+			...visitor,
+
+			'ExpressionStatement > CallExpression[callee.name=\'require\'][arguments.length=1]'(node) {
+				const moduleName = getStringIfConstant(node.arguments[0], context.getScope());
```
fisker, line 227: Did  this  mean , we  support

```
const foo  = "chalk"
const bar = require(foo)
```

?

futpib: I suppose we do. I used it only because it reads better than direct AST analysis.

rules/import-style.js
futpib: I don't think performance gain here is worth the code complication.

futpib:
@fisker I added `export ... from ...` support but chose to leave it disabled by default (behind `checkExportFrom` option).

The original motivation for this rule (as far as I understand) was to force the named import of unrelated functions and default or namespace import of related functions. `export ... from ...` allows you to turn unrelated-functions-module (like `util`) into a related-functions-module (like `chalk`) like this:
```js
export {promisify, callbackify} from 'util';
```
I think the above code should not be reported. @sindresorhus, hope this decision is fine by you.

rules/import-style.js
fisker: Add a test? Maybe also `const bar = require('cha' + 'lk')`.

rules/import-style.js
```diff
@@ -0,0 +1,368 @@
...
+			return;
+		}
+
+		const data = {
+			allowedStyles: joinOr([...allowedImportStyles.keys()]),
+			moduleName
+		};
+
+		context.report({
+			node,
```
fisker, line 195: I think report on Import `.source` and `require` call or first argument is better.

Can we also add [some visualize test](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/788/files) to see the range?

test/import-style.js
```diff
@@ -0,0 +1,550 @@
...
+			errors: [{
+				messageId: 'importStyle',
+				data: {
+					allowedStyles: 'named, namespace or default',
+					moduleName: 'no-unassigned'
+				}
+			}]
+		}
+	].map(test => addDefaultOptions(test))
+});
```
fisker, line 550: Let's test 

```js
const {} = require('foo')
```

and

```js
import {} from 'foo'
```

too, they are valid code, just make sure we don't crash on them.

rules/import-style.js
futpib: Why report the module name when it's the part that's not supposed to be fixed at all?

```diff
-const util = require('util');
+const {inspect} = require('util');
```

fisker: You are right, report on specifiers should be more reasonable.

But I guess we can ignore this for now.

test/import-style.js
```diff
@@ -0,0 +1,603 @@
...
+		'const x = require(\'namespace\')',
+		'import * as x from \'namespace\'',
+		outdent`
+			async () => {
+				const x = await import('namespace');
+			}
+		`,
+		'export * from \'namespace\'',
+
+		'const {x} = require(\'named\')',
```
fisker, line 105: Do we support check `const x = require('named').x`?

futpib: I added the test as `valid`. I hope people use the `prefer-destructuring` rule.

rules/import-style.js
futpib: Sorry, I'm not convinced indentation level is bad enough to minimize it here (or bad at all). 

Will fix if another reviewer votes for it too.

futpib:
I addressed the remaining review comments.

sindresorhus:
Great work on this, @futpib 👍🏻

PR #793, opened 2020-07-07
Add ev-to-event default replacement to prevent-abbreviations rule

mvargeson:
This adds a new default replacement to the `prevent-abbreviations` rule for `ev` -> `event`.

I'm not sure if there is a lot of process around adding something like this so feel free to close this or ask for changes that I may have missed.

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -94,9 +94,9 @@ const defaultReplacements = {
 	err: {
 		error: true
 	},
-  ev: {
-    event: true
-  },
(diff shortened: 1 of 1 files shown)


sindresorhus:
Have you actually seen this abbreviation being used in real code?

mvargeson:
I have.

mvargeson:
A quick browsing around https://github.com/search?l=JavaScript&q=ev&type=Code shows quite a few instances of this abbreviation being used for what it's worth.

PR #809, opened 2020-08-15
Allow `Children` in `no-fn-reference-in-iterator`

remcohaszing:
This allows the use of `Children.map(children, fn)` in
`no-fn-reference-in-iterator`. This handles the following situation.

```js
import { Children } from 'react';

export default ({ children ]) => (
  Children.map(children, (child) => child)
);
```

The casing for `React.Children` in the test has been changed to better reflect
reality.

Refs #121

### DIFF
--- a/test/no-fn-reference-in-iterator.js
+++ b/test/no-fn-reference-in-iterator.js
@@ -88,7 +88,8 @@ ruleTester.run('no-fn-reference-in-iterator', rule, {
 		'_.map(fn)',
 		'Async.map(list, fn)',
 		'async.map(list, fn)',
-		'React.children.forEach(children, fn)',
+		'React.Children.forEach(children, fn)',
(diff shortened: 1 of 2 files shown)


test/no-fn-reference-in-iterator.js
```diff
@@ -88,7 +88,8 @@ ruleTester.run('no-fn-reference-in-iterator', rule, {
 		'_.map(fn)',
 		'Async.map(list, fn)',
 		'async.map(list, fn)',
-		'React.children.forEach(children, fn)',
+		'React.Children.forEach(children, fn)',
+		'Children.forEach(children, fn)',
```
sindresorhus, line 92:
```suggestion
		'Children.forEach(children, fn)', // `import {Children} from 'react';`
```

PR #814, opened 2020-09-01
`no-array-callback-reference`: Ignore `AwaitExpression` except `reduce` and `reduceRight`

### LINKED ISSUE #813: unicorn/no-fn-reference-in-iterator mistaking .find method?
I'm using a third-party API that contains a method called find. It is not find as in the collection find, but rather their own find. It expects a scalar value (id) and then does its thing.

XO is throwing the following error:

```
✖   82:41  Do not pass function directly to .find(…).  unicorn/no-fn-reference-in-iterator
```

Given that the parameter is scalar, how could this happen? Sample code:

```javascript
const clientId = 20
const client = await oidc.Client.find(clientId)
```

If this is still a valid application of the rule, what would the correct code look like? The [rule doc](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-fn-reference-in-iterator.md) explains the code could be wrapped in an arrow function, which does pass the text. But, is this right?

```javascript
const clientId = 20
const client = await oidc.Client.find(() => clientId)
```

fisker:
Fixes #813


fisker:
I set this as draft for now, there are new cases coming up, see https://github.com/sindresorhus/eslint-plugin-unicorn/issues/813#issuecomment-684139507

sindresorhus:
I think the discussion in #813 has been resolved, so this could be moved forward.

RobertoMachorro:
> I think the discussion in #813 has been resolved, so this could be moved forward.

@sindresorhus does this mean this draft will be merged? Or is the solution something else? Thanks!

sindresorhus:
> @sindresorhus does this mean this draft will be merged? Or is the solution something else? Thanks!

It means that this PR needs to be updated based on the resolution in #813.

PR #815, opened 2020-09-01
`number-literal-case`: Fix false positive on `@typescript-eslint/parser`

### LINKED ISSUE #812: number-literal-case breaks with bigints
There isn't a good way to write hex big int literals with this rule enabled.

`0x7Fn` is marked as an error presumably because of the lower case `n`.

fisker:
Fixes #812

- Run tests on `espree`, `babel-eslint`, `@typescript-eslint/parser`
- Add some `0n` tests


hayes:
thanks for fixing this so quickly!

fisker:
This also fix `bigint` with "numeric separator"

PR #821, opened 2020-09-01
Switch `babel-eslint` to `@babel/eslint-parser`

fisker:


fisker:
It seems too early to do this.

PR #826, opened 2020-09-15
`filename-case`: Ignore more index files

### LINKED ISSUE #824: Ignore index.ts on "filename-case" by default
As the document said, `index.js` is ignored by the `filename-case` rule, how about `index.ts` (**.TS** instead of **.JS**)?

fisker:
Fixes #824


papb:
I think `.jsx` should have been added too, shouldn't it?

papb:
Friendly bump @fisker

fisker:
Sorry, missed this message. PR welcome.

PR #830, opened 2020-09-24
Track JSX presence per-function, fixing some false negatives

willheslam:
Removing some false negatives of movable functions that were ignored because they were defined after some JSX

This is a follow up to https://github.com/sindresorhus/eslint-plugin-unicorn/commit/8a999c028f19337e8241f706b2bb1c4e8975efac

I noticed this example has a false negative:
```javascript
function Foo() {
  const Bar = <div />
  function doBaz() {
    return 42
  }
  return <div>{ doBaz() }</div>
}
```
where `doBaz` comes after `Bar`, so it's treated as unmovable because the rule is using a singular boolean flag.
If `const Bar = <div />` is moved to after `doBaz`, it correctly identifies it as movable because the flag hasn't been set yet.
The scope has to reach module-level again before the flag is turned off, meaning any safely-movable functions succeeding some JSX, or succeeding another *scope* with some JSX in are ignored.

This PR changes the rule to track whether JSX has been identified more finely, so it no longer ignores safe functions like these.

I tried this change out on a large-ish React codebase and whilst it picked up a previous false negative that could be safely moved, it didn't newly identify any false positives, so that's a good sign above and beyond the test examples.

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -153,17 +153,16 @@ const create = context => {
 	const {scopeManager} = sourceCode;
 
 	const functions = [];
-	let hasJsx = false;
 
 	return {
-		'ArrowFunctionExpression, FunctionDeclaration': node => functions.push(node),
+		'ArrowFunctionExpression, FunctionDeclaration': () => functions.push(false),
 		JSXElement: () => {
 			// Turn off this rule if we see a JSX element because scope
 			// references does not include JSXElement nodes.
-			hasJsx = true;
+			functions[functions.length - 1] = true;
 		},
 		':matches(ArrowFunctionExpression, FunctionDeclaration):exit': node => {
-			if (!hasJsx && !checkNode(node, scopeManager)) {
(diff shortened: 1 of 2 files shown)


fisker:
I think this is what I'm trying to do in  https://github.com/sindresorhus/eslint-plugin-unicorn/pull/786, but I got myself confused...

Can you add tests from https://github.com/sindresorhus/eslint-plugin-unicorn/pull/786 here?

willheslam:
Hi @fisker, I'm happy to add extra test cases to this PR to make sure we're covering every angle. :)
Looking briefly at your PR, some of those tests look like they overlap with the ones already added here - as you're familiar with that PR, do you mind pointing out to me which ones you think should be added?

I've covered:
1. Conservatively ignoring functions with JSX in that otherwise could be safely moved
2. JSX in the same scope, preceding a function that could be safely moved
3. JSX in a nested scope, preceding a function that could be safely moved

fisker (approved):

Issue #819, opened 2020-09-01
Rule proposal: Collapsible if-statements


sindresorhus:
The rule should identify nested if-statements that can be combined into one.

Would be useful if the rule could be auto-fixable.

>
> Instead of:
> 
> ```ts
> if (foo) {
>     if (bar) {
>         // ...
>     }
> }
> ```
> 
> Prefer:
> 
> ```ts
> if (foo && bar) {
>     // ...
> }
> ```

From: https://github.com/typescript-eslint/typescript-eslint/issues/279

sindresorhus:
Accepted. PR welcome.

fisker:
Rule name? `collapse-if`?

fregante:
This can only work if it’s not followed by `else`

fisker:
I think [this comment](https://github.com/typescript-eslint/typescript-eslint/issues/279#issuecomment-463791071) make sense, let's name it as `no-lonely-if` too, hope it will merge to core someday.

fregante:
Hasn’t eslint stopped adding rules to core?

fisker:
No, but @sindresorhus made this proposal here. I guess he want have this rule in `unicorn` first?

fisker:
Anyway, the rule is ready https://github.com/sindresorhus/eslint-plugin-unicorn/pull/936

fregante:
According to their [rule suggestion issue template](https://github.com/eslint/eslint/issues/new?assignees=&labels=triage%2C+rule%2C+feature&template=NEW_RULE.md&title=):

> New rules must be related to ECMAScript features added within the last 12 months

But I suppose this rule could still be accepted as a _modification_ of the existing one.

sindresorhus:
> No, but @sindresorhus made this proposal here. I guess he want have this rule in unicorn first?

Yeah, I have given up bothering. The process there is way too slow and they reject almost everything anyway.

PR #833, opened 2020-09-26
Add `numeric-separators-style` rule

### LINKED ISSUE #828: Rule proposal: Enforce numeric separators style
Numeric separators are a new JavaScript feature that has [reached stage 4](https://github.com/tc39/proposal-numeric-separator) and allow numbers to be written like this:
```js
1_000_000_000
101_475_938.38 
0.000_001
1e10_000

// It also works with binary, hexadecimal, BigInts, Octals...
0b1010_0001_1000_0101
0xA0_B0_C0
9_223_372_036_854_775_807n
0o1234_5670
```

It would be great to have a rule to enforce its usage.
We could, for example, enforce grouping hexadecimal by 2, binary and octal by 4, numbers and BigInts by 3... I think that this preferred group length should be configurable.

It would also be great to have some more options such as:
- Only apply this rule if there is more than a certain number of digits. Example, if we disable the rule for less than 5 digits, then `1_024` will fail and `1024` will pass, but if we put it to 0, then `1_024` will pass and `1024` will fail
- The "preferred group length" could also be used to determine if a group is not of a valid length. It will check all non-leading groups, so in the case of `1_2_34_567`, it will be `2`, `34`, `567`, and see if it matches the preferred group length, here 3. If we fix this, we get `1_234_567`, which satisfy the "preferred group length" rule

With those two things in mind, maybe a configuration like this would be appropriate:
```js
// General:
{
    type: [minThreshold, preferedGroupLength],
}

// Example:
{
  hexadecimal: [0, 2],
  binary: [0, 4],
  octal: [0, 4],
  bigint: [5, 3],
  number: [5, 3],
}
```

~~I'm willing to try to make a PR for that, although this would be my first time creating an ESLint rule...~~

EDIT: Sorry for all the edits but I keep having new/better ideas, and I struggle to formulate them 😅

EDIT 2: I've created the rule and it works pretty well. [Here is the doc](https://gist.github.com/noftaly/23207a7c4dd04fdb0bedcdda3cfb3b0f) if you want to see the final result. I'll make a PR only if you guys are in favor of this. There are ~60 tests, with all scenarios I could imagine. The integration test was successful too.

noftaly:
A new rule to enforce the [numeric separators](https://github.com/tc39/proposal-numeric-separator)' style. See #828 for more information.

Concerning the code, I decided to group digits together by splitting them into an array of chunks and then joining them with an underscore. Although this was the simplest way I found, it required pre-processing and post-processing the value, to remove/add prefixes/suffixes (such as `0x`, or `n` for big ints), but also add offset to some numbers because the `chunk()` function adds the rest of the numbers at the end (i.e. create an array with strings of length 3, 3, 1), whereas in some cases you want the rest to be at the beginning (1, 3, 3, such as in this number `1_000_000`).
The code might also be longer than what you would expect for this rule, but I stumbled upon multiple challenges, such as the large variety of Literal that support the numeric separator notation, but also decimal numbers and the exponential notation.

I had one question: what to do with decimals? Currently, the extra group must be at the end, such as `3.000_000_1`, but it might look weird in some cases (pretty rare), where you would want something like this `3.1_000_000`... I don't think it should be changed but feel free to tell me and I'll change (or make it configurable).

You can look at the documentation for this rule, [here](https://gist.github.com/noftaly/23207a7c4dd04fdb0bedcdda3cfb3b0f).

I'll be glad to fix anything you'd like me to change 😄

Fixes #828


rules/numeric-separators-style.js
```diff
@@ -0,0 +1,133 @@
...
+
+	const base = numberGroups.map(part => part.join('_')).join('.');
+
+	const power = powerGroups.length > 0 ? 'e' + powerGroups.join('_') : '';
+	return addLiteralNotation(notation, base + power);
+}
+
+const create = context => ({
+	Literal: node => {
+		if (['number', 'bigint'].includes(typeof node.value)) {
```
fisker, line 105: value of bigint can be null, need check node.bigint

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,133 @@
...
+		if (powerPart.length < minimumThreshold) {
+			powerGroups.push(powerPart);
+		} else {
+			powerGroups.push(...getChunks(powerPart, preferedGroupLength, false));
+		}
+	}
+
+	const base = numberGroups.map(part => part.join('_')).join('.');
+
+	const power = powerGroups.length > 0 ? 'e' + powerGroups.join('_') : '';
```
fisker, line 99: Exponential notation maybe negative.

test/numeric-separators-style.js
```diff
@@ -0,0 +1,308 @@
...
+			output: 'var foo = 3.6e12_000'
+		},
+		{
+			code: 'var foo = -1200000e5',
+			options: [{number: {minimumThreshold: 5, preferedGroupLength: 3}}],
+			errors: [error],
+			output: 'var foo = -1_200_000e5'
+		}
+	]
+});
```
fisker, line 308: Legacy `0777777` need test.

rules/numeric-separators-style.js
noftaly: It already takes it into account.
If you meant the number before the `e` can be negative, then this test passes `-123_456e100`, and this `-123456e100` is correctly fixed
If you meant the power part, after the `e`, can be negative, then I haven't done any test cases but it works too. I'll add some, though.

rules/utils/get-number-literal-type.js
```diff
@@ -0,0 +1,33 @@
...
+ * @returns {undefined|"bigint"|"hexadecimal"|"binary"|"octal"|"number"}
+ */
+function getNumberLiteralType(node) {
+	if (!node?.type === 'Literal' || !['number', 'bigint'].includes(typeof node.value)) {
+		return;
+	}
+
+	if (node.bigint) {
+		return 'bigint';
+	}
```
fisker, line 14: bigint  can  be  binary too, we  should just strip `n` and treat rest part as  number

rules/numeric-separators-style.js
fisker: I mean  10000e-10000

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Options
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are 5 options, `hexadecimal`, `binary`, `octals`, `bigint`, `number`, which all
+refer to their corresponging type. Each of them has to be associated with an array
+containing 2 values:
+- The first one is the minimum threshold. You can't use numeric separators if the number has
```
fisker, line 39: How about `minimumDigits`

noftaly: It is indeed more explicit and easier to understand. Should I also update for the configuration, or only in the description, in the documentation?

fisker: Ah, didn't notice it's array.

rules/utils/get-number-literal-type.js
noftaly: What do you mean?

rules/utils/get-number-literal-type.js
```diff
@@ -0,0 +1,33 @@
+'use strict';
+/**
+ * Return the type of the number in the node
+ * @param {Node} node - The node to check. Its type must be "Literal", and its value must be a number.
+ * @returns {undefined|"bigint"|"hexadecimal"|"binary"|"octal"|"number"}
+ */
+function getNumberLiteralType(node) {
+	if (!node?.type === 'Literal' || !['number', 'bigint'].includes(typeof node.value)) {
```
fisker, line 8: Did you mean `node?.type !== 'Literal'`? But why `?.` Node.type  always exits.

fisker (approved):

rules/utils/get-number-literal-type.js
noftaly: I did mean `node?.type !== 'Literal'`, you're correct.
This is a function and everything can be passed here, thus node might not be a Node object. Should we consider that `node` is always an instance of `Node`? Should I check if `node instanceof Node`? Because it's true that currently it is not consistent at all.
Anyway the optional chaining syntax should be removed, because it is the cause of all those failing CI (which I'm currently fixing)

test/numeric-separators-style.js
noftaly: How should it work then? Is the leading zero considered as a "regular" digit, and should be correctly grouped, or is it considered as a "prefix" and ignores grouping?
Maybe it can even be fixed to `0o`, as `0` is now deprecated? Although this might be a breaking change, users of eslint and eslint-plugin-unicorn should have at least node 10, and afaik, `0o` is supported in node 10... (need verification though)

rules/utils/get-number-literal-type.js
fisker: 0b1010n

rules/numeric-separators-style.js
noftaly: Then i mark this as resolve because the test `-100_000e-10_000` passes correctly 🙂

rules/utils/get-number-literal-type.js
noftaly: Yes but what do you mean by strip the `n` and treat the rest as a number?
Do you mean that if there is already a prefix, such as `0b` then we just completely remove the `n`?

rules/utils/get-number-literal-type.js
fisker: If you don't pass something that's not a Node, I don't think you need add this test. If someone need check  it, he can add it when he use this.  Normally we do `!node ||`

docs/rules/numeric-separators-style.md
noftaly: Oh you're right, i forgot to change the doc after sindresorhus suggested to make it an object. I'll change that.

rules/utils/get-number-literal-type.js
fisker: I mean you should just strip `n` and return the "type" as `binary`

We don't need option for `bigint`, `0b1010n` should follow `binary` options.

test/numeric-separators-style.js
fisker: Just make sure  we don't break it, like this  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/f141d08829085cb314bc78a4bdf59e2c739dedb8/test/number-literal-case.js#L25

fisker: Just make sure  we don't break it, like this  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/f141d08829085cb314bc78a4bdf59e2c739dedb8/test/number-literal-case.js#L25

rules/numeric-separators-style.js
fisker: No, you didn't ignore `-` in  this part 

`powerPart.length < minimumThreshold`

noftaly: Oh nice catch. Will fix.

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,133 @@
...
+	return addLiteralNotation(notation, base + power);
+}
+
+const create = context => ({
+	Literal: node => {
+		if (['number', 'bigint'].includes(typeof node.value) || node.bigint) {
```
fisker, line 105:
```suggesttion
		if (typeof node.value === 'number' || node.bigint) {
```

rules/utils/get-number-literal-type.js
fisker: > strip n 

I mean, like this

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/834/files#diff-d6dd6615153d8c4ed594a894e169f055R22

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,133 @@
...
+	chunks = chunkString(string, size);
+	chunks[0] = chunks[0].replace(offset, '');
+	return chunks;
+}
+
+function parseNumber(string) {
+	const groups = string.split(/(?=\.|e)/);
+	const wholePart = groups.find(grp => !grp.startsWith('.') && !grp.startsWith('e'));
+	const decimalPart = groups.find(grp => grp.startsWith('.'));
+	const powerPart = groups.find(grp => grp.startsWith('e'));
```
noftaly, line 63: Also support exponential notation with an uppercase `E`

rules/utils/get-number-literal-type.js
noftaly: Oh i just finished that and I did not do it like this... Check my last commit (will be pushed in a couple of minutes) and leave a review if you want me to do it like this 😄

rules/numeric-separators-style.js
noftaly: [Fixed](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/833/commits/a4c69d3c2bcd0849686e883d9e971007f5986fc0)

rules/numeric-separators-style.js
noftaly: And support correctly lowercase hexadecimal (with lowercase `e`)

noftaly:
Everything I could though of is here, and all reviews were resolved.
There is only the codecov/patch test not passing, because of the util function I created. I put it into `utils` for future usage, so I added some validations checks at the beginning to ensure the value is of the correct type, and of course, in my case, it will always be. Should I add checks for the `getNumberLiteralType` function to satisfy the code coverage check, or should we just ignore it?

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+## Options
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are 5 options, `hexadecimal`, `binary`, `octals`, `bigint`, `number`, which all
+refer to their corresponging type. Each of them has to be associated with an object
+containing 2 fields:
+- `minimumThreshold`: It has to be a number. It corresponds to the threshold of digits in
```
fisker, line 39: Rename to `minimumDigits`

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,70 @@
...
+const foo = 0o10_4421;
+const foo = 1_294_287_712n;
+```
+
+
+## Options
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are 5 options, `hexadecimal`, `binary`, `octals`, `bigint`, `number`, which all
```
fisker, line 36: Remove `bigint`

rules/utils/get-number-literal-type.js
```diff
@@ -0,0 +1,33 @@
...
+ * Return the type of the number in the node
+ * @param {Node} node - The node to check. Its type must be "Literal", and its value must be a number.
+ * @returns {undefined|"bigint"|"hexadecimal"|"binary"|"octal"|"number"}
+ */
+function getNumberLiteralType(node) {
+	if (!node || node.type !== 'Literal' || (typeof node.value !== 'number' && !node.bigint)) {
+		return;
+	}
+
+	if (node.raw.startsWith('0x')) {
```
fisker, line 12: `0XdEED_BEEF`

fisker (changes requested):

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,151 @@
...
+			numberGroups[0] = getChunks(wholePart, preferedGroupLength, false);
+		}
+	}
+
+	if (decimalPart) {
+		numberGroups[1] = getChunks(decimalPart, preferedGroupLength, true);
+	}
+
+	if (powerPart) {
+		if (powerPart.replace('-', '').length < minimumThreshold) {
```
fisker, line 105: `1e+10`

rules/utils/get-number-literal-type.js
noftaly: Oh, I did not know that prefixes could be uppercase too... That raises another issue: if someone uses uppercase, then the autofix will fix to lowercase... Is this an issue or is it ok?

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,151 @@
...
+});
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			url: getDocumentationUrl(__filename)
+		},
+		fixable: 'code'
```
fisker, line 149: `schema` required. Example: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/rules/better-regex.js#L102

rules/utils/get-number-literal-type.js
fisker: I prefer handle uppercase in this rule, people may disable `number-literal-case`, especially with `eslint-config-prettier` it's disabled by default https://github.com/prettier/eslint-config-prettier/blob/5d479e943313d078129ad3308bf4d16f69eeda53/unicorn.js#L6

fisker:
`0xe` failed , source https://github.com/mozilla/pdf.js/blob/master/src/core/jbig2.js#L2004

log https://github.com/sindresorhus/eslint-plugin-unicorn/pull/833/checks?check_run_id=1172323645#step:5:352

fisker:
I suggest, we do this logic

```
Literal 
  ->  if (bigint) -> remove `n` from `raw` as `raw`
  ->  if (number) -> use `raw`
raw
  -> check `prefix`
      -> if (hasPrefix) format(dataPart)
      -> if (noPrefix) 
           -> split `e` -> format(numberPart) format(powerPart)
format(numberPart)
  -> split `.` -> format(decimalPart) format(fractionalPart)
```

noftaly:
Nice suggestion! I was also telling myself that it was a lot of code for not much, and it really looked like code that just fixed specific cases rather than being scalable and comprehensive. I'm working on that 🙂

rules/numeric-separators-style.js
noftaly: Thanks! I'm not really familiar with schemas tho. Can you tell me if it is valid? Default values apparently doesn't work as the integration test is failing because of an incorrect (inexistent) configuration.

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,134 @@
...
+function isLegacyOctal(node) {
+	return getNumberLiteralType(node) === 'number' && node.raw.startsWith('0');
+}
+
+function hasPrefix(raw) {
+	const rawLower = raw.toLowerCase();
+	return rawLower.startsWith('0x') || rawLower.startsWith('0b') || rawLower.startsWith('0o');
+}
+
+const chunkString = (string, size) => chunk(string, size).map(elt => elt.join(''));
```
sindresorhus, line 15: Use a more descriptive name than `elt`.

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,73 @@
...
+const foo = 0b1000_1111;
+const foo = 0o10_4421;
+const foo = 1_294_287_712n;
+```
+
+## Options
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are 4 options, `hexadecimal`, `binary`, `octals`, `number`, which all refer to their corresponding type. Each of them has to be associated with an object
```
sindresorhus, line 33: Typo: `octals`

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,73 @@
...
+containing 2 fields:
+- `minimumDigits`: It has to be a number. It corresponds to the threshold of digits in a number, where you shouldn't use numeric separator. For example, if you put 5 as the minimum threshold, then `1024` will pass because it has 4 digits, and `1_024` will fail. However `1_000_000` will still pass and `1000000` will still fail.
```
sindresorhus, line 35:
```suggestion
- `minimumDigits`: It has to be a number. It corresponds to the threshold of digits in a number, where you shouldn't use a numeric separator. For example, if you put 5 as the minimum threshold, then `1024` will pass because it has 4 digits, and `1_024` will fail. However `1_000_000` will still pass and `1000000` will still fail.
```

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,73 @@
...
+- `preferedGroupLength`: It has to be a number. It is the size of a group, if we don't count the size of the first one (which can be of any size as long as it is equal to or less than the number specified here).
```
sindresorhus, line 36:
```suggestion
- `preferedGroupLength`: It has to be a number. It is the size of a group if we don't count the size of the first one (which can be of any size as long as it is equal to or less than the number specified here).
```

sindresorhus (changes requested):

sindresorhus:
I think the options docs text could use more work.

noftaly:
Thanks for the review. I applied requested changes and improved the documentation to make it easier to read and look more like other Unicorn documentations.

Concerning the `codecov/patch` test: it is failing because of the util function I created. I put it into `utils` for future usage, so I added some validations checks at the beginning to ensure the value is of the correct type. Of course in my case, it will always be valid, hence the first check is never called. Should I add checks for the getNumberLiteralType function to satisfy the code coverage check, or should we just ignore it?

fisker:
If you don't mind, I want do some refactor.

noftaly:
Sure! Go ahead!

fisker:
I've done refactoring, WDYT?

noftaly:
I was not expecting that much refactoring, but yes, perfect! You reduced the number of lines which is a good thing, and it looks much cleaner with regexes.
Lint and Integration tests are not passing though.

fisker:
Lint and Integration tests failed because default options not applied, can you fix that?

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,86 @@
...
+containing 2 fields:
+
+**`minimumDigits`**
+
+Type: `number`
+The minimum of digits in a number, where you shouldn't use numeric separator.
+
+Example: With 5 as the minimum digits, `1024` will pass because it has 4 digits, and `1_024` will fail.
+
+**`preferedGroupLength`**
```
fisker, line 43: This is misspelled , should be `preferredGroupLength`. Maybe we should use `groupLength`, simplier.

noftaly:
> Lint and Integration tests failed because default options not applied, can you fix that?

Well I can try but as I told you previously, this is my first time making a rule so I'm not familiar with all that stuff and I don't really know how to do it...

fisker:
You can use this example 	https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e502f429abc0378cc6e4506a9659e3476a8daf58/rules/expiring-todo-comments.js#L214, once it's done, the options with default  value need remove

noftaly:
Is this what you meant? If no, sorry but I did not really understand 😅...

fisker:
I'm so sorry... I mean in tests,

> the options with default value in tests need remove

noftaly:
Oh sure. But are the changes I made for the default value ok?

noftaly:
Concerning the lint test, failing: it does not fail locally, and the proposition it makes is almost exactly the same as the original and does not optimise the RegEx... Should I apply it anyway or can we ignore it ?
Proposition (`unicorn/better-regex`):
```diff
- /^(?<number>.*?)(?:(?<mark>e)(?<sign>[-+]?)?(?<power>\d+))?$/i
+ /^(?<number>.*?)(?:(?<mark>e)(?<sign>[+-]?)?(?<power>\d+))?$/i
```

fisker:
Apply

noftaly:
Applied. All tests are passing now, except `codecov/patch`. What do we do about it?

fisker:
> Oh sure. But are the changes I made for the default value ok?

I think it's not, you can try pass options `{binary: {minimumDigits:0}}` to see if it work as expected, unfortunately I'm not familiar with this part too, need look for the docs.

fisker:
> Applied. All tests are passing now, except `codecov/patch`. What do we do about it?

Cover the missing line by add test `const foo = "1234567n"` or some other `Literal` node.

noftaly:
> Cover the missing line by add test const foo = "1234567n" or some other Literal node.

Thanks.

> I think it's not, you can try pass options {binary: {minimumDigits:0}} to see if it work as expected, unfortunately I'm not familiar with this part too, need look for the docs.

It does not work and after checking the doc, I can't find any way of adding default values. I suggest merging the two objects together (with lodash/merge), like this, the user can only specify some fields in the documentation and it will work.
I don't know if this is how you would do it normally tho.

fisker:
> I suggest merging the two objects (user + defaults) together (with lodash/merge)

Yes, try it.

noftaly:
I've done it.
All test pass, so unless you wan't to add or refactor some things, then this PR is ready for the final review!

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,142 @@
...
+	return formatNumber(number, formatOption) + mark + sign + addSeparator(power, options['']);
+}
+
+const create = context => {
+	const defaultOptions = {
+		hexadecimal: {minimumDigits: 0, groupLength: 2},
+		binary: {minimumDigits: 0, groupLength: 4},
+		octal: {minimumDigits: 0, groupLength: 4},
+		number: {minimumDigits: 5, groupLength: 3}
+	};
```
fisker, line 73: Move defaultOptions to outer scope.

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,142 @@
...
+	const rawOptions = merge(defaultOptions, context.options[0]);
```
fisker, line 74: `_.merge` and `_.defaultsDeep` which is better in this case?  Not sure about difference between them.

test/numeric-separators-style.js
```diff
@@ -0,0 +1,332 @@
...
+			code: 'var foo = -1200000e5',
+			errors: [error],
+			output: 'var foo = -1_200_000e5'
+		},
+		{
+			code: 'var foo = 3.65432E12000',
+			errors: [error],
+			output: 'var foo = 3.654_32E12_000'
+		}
+	]
```
fisker, line 331: Uppercase prefix not tested, options not tested.

fisker (changes requested):

rules/numeric-separators-style.js
noftaly: Looking at lodash's code, merge might be slightly faster, because `_.defaultsDeep` is base on `_.mergeWith`, which is based on the exact same function as `_.merge`.

fisker:
It will be great if you can add some comments to the tests, like this https://github.com/sindresorhus/eslint-plugin-unicorn/blob/18bb3a3207e0ae3d8199b7ef82494c5bedcbdae3/test/number-literal-case.js#L39

rules/numeric-separators-style.js
fisker: We  need deep merge, right? Does `_.merge` do that?

fisker: And make sure it don't modify defaultOptions

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,142 @@
...
+			} else {
+				return;
+			}
+
+			const fixed = format(number.replace(/_/g, ''), options) + suffix;
+
+			if (raw !== fixed) {
+				context.report({
+					node,
+					message: 'Invalid group length in numeric value.',
```
fisker, line 104: Please use `messageId`, I don't want do this #779 again.

rules/numeric-separators-style.js
noftaly: Well.. As weird as it seems, I can't move it out of its scope... Then it is undefined in `create()`. Is it really a problem ?

rules/numeric-separators-style.js
```diff
@@ -0,0 +1,142 @@
...
+	return parts.join('_');
+}
+
+function addSeparatorFromLeft(value, options) {
+	return addSeparator(value, options, true);
+}
+
+function formatNumber(value, options) {
+	const parts = value.split('.');
+	const [decimal, fractional] = parts;
```
fisker, line 34: `decimal` seems should name as `integer`, sorry for my bad english.

noftaly: No problem ^^

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,86 @@
...
+const foo = 0b1000_1111;
+const foo = 0o10_4421;
+const foo = 1_294_287_712n;
+```
+
+## Options
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are 4 options, `hexadecimal`, `binary`, `octal`, `number`, which all refer to their corresponding type. Each of them has to be associated with an object
```
fisker, line 33: Add MDN links for  `hexadecimal` https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/number-literal-case.md

rules/numeric-separators-style.js
fisker: Seen  this?  Is it the reason for  https://github.com/sindresorhus/eslint-plugin-unicorn/pull/833#discussion_r495986350?

docs/rules/numeric-separators-style.md
noftaly: I added links for all types, as they did in the documentation you linked.

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,86 @@
...
+### Examples
+
+```js
+// eslint unicorn/numeric-separators-style: ["error", {number: {minimumDigits: 0, groupLength: 3}}]
+const foo = 100; // Pass
+const foo = 1_000; // Pass
+const foo = 1_000_000; // Pass
+const foo = 0.000_0001; // Fail
+
+// eslint unicorn/numeric-separators-style: ["error", {octal: {minimumDigits: 0, groupLength3}}]
```
fisker, line 58:
```suggestion
// eslint unicorn/numeric-separators-style: ["error", {octal: {minimumDigits: 0, groupLength: 3}}]
```

rules/numeric-separators-style.js
noftaly: Oh I missed your messages! Sorry. It might, let me try, I'll get back to you in a minute

noftaly: Yes, `_.merge` does deep merge.
`_.defaultsDeep` seems to have a little problem, if there are multiple properties, and won't modify them all:
If i have `context.options[0] { number: { minimumDigits: 6 } }`, it will give me `rawOptions { number: { minimumDigits: 6 } }`, which is not what we want.
I have 3 failure in the tests with `_.defaultsDeep`.

fisker:
Looks good already, I'll take a deeper look later, thank you!

noftaly:
No problem! Thanks for your help!

fisker:
Want do more rules? This one https://github.com/sindresorhus/eslint-plugin-unicorn/issues/819 seems simple. :smile:

noftaly:
Why not! I'll take a look and see if I have the time for it. Thanks for the proposition !

fisker:
Added some commits

fisker (approved):
The coding part looks good, but I think the docs still need some improvement.

Like

- Explain how `minimumDigits` works, for example `123.456e+789`, we check integer fractional power part separately, and `0x` `n` `.` not count, etc
- Explain fractional part is formatted from left

I'm really bad at docs, so I'm not going to review the docs changes.

noftaly:
> Added some commits, let me know if you find something wrong.

Seems good, but I find the schema part less readable...
I won't fix this as I don't have a good solution in mind other than reverting this change.

I'll work on the documentation 🙂

fisker:
> but I find the schema part less readable

It's less readable maybe because I use `fromPairs(Object.entries)`,  I first wrote

```
properties: {
	hexadecimal: formatOptionsSchema('hexadecimal'),
	// ...
}
```

Then I think I can use `defaultOptions` and DRY it, so it's in current way.


But it's stricter, previously we allow any number, with `groupLength: 0` we'll even run into a infinity loop.

noftaly:
Yeah it's definitely better, that's for sure.
But if the readability can be improve it wouldn't hurt. Otherwise no problem, it is still understandable.

Is the documentation OK, now ?

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,114 @@
...
+
+The minimum of digits in a number where you shouldn't use numeric separator.
+
+Example: With `5` as the minimum digits, `1024` will pass because it has 4 digits, and `1_024` will fail.
+
+**`groupLength`**
+
+Type: `number`
+
+The size a group should be.
```
sindresorhus, line 47: Not clear to a user what a "group" is.

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,114 @@
...
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are four group types, [`hexadecimal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Hexadecimal), [`binary`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Binary), [`octal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Octal) and [`number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type). Each of them has to be associated with an object containing the following options:
+
+**`minimumDigits`**
+
+Type: `number`
+
+The minimum of digits in a number where you shouldn't use numeric separator.
```
sindresorhus, line 39:
```suggestion
The minimum amount of digits in a number where you shouldn't use a numeric separator.
```

I also don't think the sentence makes it clear enough what the option does.

docs/rules/numeric-separators-style.md
```diff
@@ -0,0 +1,114 @@
...
+const foo = 0b1000_1111;
+const foo = 0o10_4421;
+const foo = 1_294_287_712n;
+```
+
+## Options
+
+If you want a custom group size for a specific number type, you can specify it here.
+
+There are four group types, [`hexadecimal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Hexadecimal), [`binary`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Binary), [`octal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Octal) and [`number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type). Each of them has to be associated with an object containing the following options:
```
sindresorhus, line 33:
```suggestion
There are four number types; [`hexadecimal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Hexadecimal), [`binary`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Binary), [`octal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Octal) and [`number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type). Each of them has to be associated with an object containing the following options:
```

sindresorhus: Also, this sentence makes it sound like you have to specify all number types.

sindresorhus (changes requested):

sindresorhus:
The code looks good, but the docs could be made clearer.

noftaly:
I tried to make a clearer, and add a description at the top.
Unfortunately, I'm not English and this is the best I can do in terms of clarity...

fisker:
Need update snapshots

noftaly:
> Need update snapshots

I'm sorry but I have no idea what snapshots are and how to update them 😬

fisker:
`npx ava test/numeric-separators-style.js -u`

fisker:
@sindresorhus  Do you think this is bug in ava?

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/833/checks?check_run_id=1181693091#step:5:7359

Shouldn't it `+`? That's a new line added.

noftaly:
> `npx ava test/numeric-separators-style.js -u`

Thanks but that did not updated anything 😬no file where modified...

fisker:
Maybe you didn't pull? I'll commit.

noftaly:
> Maybe you didn't pull? I'll commit.

Oh yes... My bad

noftaly:
Hey, just want to let maintainers know that all reviews were addressed and there are none left, and all tests passed:
this PR is now ready to be merged (or re-review) 😄

sindresorhus (approved):

fisker:
Can you rebase and update the snapshot again?  Apologize for the frequency update in `visualizeRuleTester`.

noftaly:
Uhhh I might have done something wrong 😂 It was my first rebase ever and i had merge conflict...
It looks like there are also the changes of `prefer-ternary`... How can I resolve that ?
*really sorry 😅*

fisker:
Fixed it.

fisker:
Good job, thanks.

noftaly:
Thanks for your help!

sindresorhus:
Does anyone remember why we didn't enable this one by default?

fisker:
Because it's not supported in Node.js 10?  I didn't find support info on https://node.green

PR #835, opened 2020-09-27
`consistent-function-scoping`: Check anonymous functions

fisker:
```js
				const doFoo = function() {
					const doBar = function(bar) {
						return bar;
					};
				};
```

Function `doBar` should be able to move to outer scope.


test/consistent-function-scoping.js
```diff
@@ -356,6 +390,47 @@ ruleTester.run('consistent-function-scoping', rule, {
...
+		},
+		{
+			code: outdent`
+				const doFoo = function() {
+					const doBar = function(bar) {
+						return bar;
+					};
+				};
+			`,
+			errors: [createError('function')]
```
fisker, line 411: This function should has name `doBar`, https://github.com/mysticatea/eslint-utils/pull/14

sindresorhus: Is this PR blocked by that?

fisker: No, we can improve message later.

sindresorhus:

PR #839, opened 2020-09-29
`visualizeRuleTester`: Support multiple errors

fisker:
Next, I'll try to support `options`, and snapshot `output`. Maybe we can end up run tests without the `RuleTester`.

Result with multiple errors https://github.com/sindresorhus/eslint-plugin-unicorn/pull/839/files#diff-272d8dc0108ac6532205aff54da6651eR1


sindresorhus:

PR #849, opened 2020-10-03
fix: Prefer-flat-map rule: allow .map(...).flat(Infinity)

William-Schutte:
Small fix that no longer show error when using Infinity in `.map(...).flat(Infinity)`

### DIFF
--- a/rules/prefer-flat-map.js
+++ b/rules/prefer-flat-map.js
@@ -141,8 +141,10 @@ const create = context => ({
 
 		if (
 			node.arguments.length === 1 &&
-			node.arguments[0].type === 'Literal' &&
-			node.arguments[0].value !== 1
+			((node.arguments[0].type === 'Literal' &&
+			node.arguments[0].value !== 1) ||
+			(node.arguments[0].type === 'Identifier' &&
+			node.arguments[0].value === 'Infinity'))
 		) {
 			return;
 		}


rules/prefer-flat-map.js
```diff
@@ -141,8 +141,10 @@ const create = context => ({
 
 		if (
 			node.arguments.length === 1 &&
-			node.arguments[0].type === 'Literal' &&
-			node.arguments[0].value !== 1
+			((node.arguments[0].type === 'Literal' &&
+			node.arguments[0].value !== 1) ||
+			(node.arguments[0].type === 'Identifier' &&
+			node.arguments[0].value === 'Infinity'))
```
noftaly, line 147: An Identifier has a `name`, not a value 😄
```suggestion
			node.arguments[0].name === 'Infinity'))
```

sindresorhus:
This needs a test. And also reference the issue this fixes using the `fixes` syntax.

fisker:
I think this should only report on  `.flat()` and `.flat(1)`.

Should ignore `.flat(foo)` `.flat(foo.bar)` and many other Nodes

rules/prefer-flat-map.js
William-Schutte: Oh, duh. Thank you!

fisker:
Close in favor of #859

PR #850, opened 2020-10-04
`explicit-length-check`: Use `'non-zero': 'greater-than'` by default

### LINKED ISSUE #702: Pick a default `non-zero` option for `explicit-length-check`
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/f1ced468120686263a45a9993d2f02343f70c6f3/docs/rules/explicit-length-check.md#non-zero-comparisons

I suggest picking a default value for the `non-zero` option.

## not-equal

Pros:

1. `!== 0` is the inverse of `=== 0`, like "is it not empty?" and "is it empty?"
2. `!== 0` does not perform type coercion (considering that the general consensus is that implicit type coercion is bad)

Cons:

1. If `array.length` is `undefined`, `array.length !== 0` would evaluate to `true`, which is incorrect if we are asking "is it not empty?"

## greater-than

Pros:

1. If `array.length` is `undefined`, `array.length > 0` evaluates to `false`, which is "more correct" than the `!== 0` case
2. (subjective) A positive question in a condition is more readable than a negative question, like `is it populated?` for `> 0` vs `is it not empty?` for `!== 0`

Cons:

1. Performs type coercion (this actually might be good in this case, see pro # 1)

## greater-than-or-equal

Pros:

1. same as for greater-than
2. same as for greater-than

Cons:

1. same as for greater-than
2. (unconfirmed) More mentally demanding because it actually has two conditions in one expression (`greater than 1 OR equal to 1`)
3. (unconfirmed) Used less often than `!== 0` or `> 0`

I didn't perform a thorough analysis when I started using `!== 0` in my code, and now I think that actually `> 0` is a better option.

EvgenyOrekhov:
Note that this is a breaking change since the rule wasn't enforcing any non-zero comparison style by default before, but now it enforces the `greater-than` style.

Fixes #702

### DIFF
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -1,6 +1,6 @@
 # Enforce explicitly comparing the `length` property of a value
 
-Enforce explicitly checking the length of a value array in an `if` condition, rather than checking the truthiness of the length.
+Enforce explicitly checking the length of a value array in an `if` condition, rather than checking the truthiness of the length, and enforce comparison style.
 
 This rule is partly fixable.
 
@@ -10,6 +10,7 @@ This rule is partly fixable.
 if (string.length) {}
 if (array.length) {}
 if (!array.length) {}
+if (array.length !== 0) {}
 ```
 
 ### Pass
@@ -17,14 +18,13 @@ if (!array.length) {}
 ```js
 if (string.length > 0) {}
 if (array.length > 0) {}
-if (array.length !== 0) {}
 if (array.length === 0) {}
 ```
 
 
 ## Zero comparisons
 
-Enforce comparison with `!== 0` when checking for zero length.
+Enforce comparison with `=== 0` when checking for zero length.
 
 ### Fail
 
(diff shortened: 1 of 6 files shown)


docs/rules/explicit-length-check.md
```diff
@@ -35,13 +35,13 @@ if (string.length < 1) {}
 ### Pass
 
 ```js
-if (array.length !== 0) {}
+if (array.length === 0) {}
```
EvgenyOrekhov, line 38: I believe these two were typos because this section talks about **zero** comparisons. So I fixed them.

docs/rules/explicit-length-check.md
```diff
@@ -10,21 +10,21 @@ This rule is partly fixable.
...
 if (array.length > 0) {}
-if (array.length !== 0) {}
 if (array.length === 0) {}
 ```
 
 
 ## Zero comparisons
 
-Enforce comparison with `!== 0` when checking for zero length.
+Enforce comparison with `=== 0` when checking for zero length.
```
EvgenyOrekhov, line 27: I believe this was a typo because this section talks about zero comparisons. So I fixed it.

docs/rules/explicit-length-check.md
EvgenyOrekhov, line 38: I believe this was a typo because this section talks about zero comparisons. So I fixed it.

fisker (approved):
Looks good.

PR #851, opened 2020-10-04
Add `prefer-math-trunc` rule

### LINKED ISSUE #827: Rule proposal: `no-bitwise-trunc`
Prefer the [`Math.trunc`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) method for truncating numbers instead of the `| 0` hack which doesn't even work in some cases (such as `3000000000.1 | 0`, which gives `-1294967296`).

## Fail
```js
const bar = foo | 0;
```

## Pass
```js
const bar = Math.trunc(foo);
```

[Relevant SO question](https://stackoverflow.com/questions/7487977/using-bitwise-or-0-to-floor-a-number)

noftaly:
Added `prefer-math-trunc` rule, which forbids the use of `| 0` over `Math.trunc()` to truncate numbers.
One thing i'm not really sure about is the error message. It says not to use `| 0` to truncate numbers, but maybe it should say no to use `bitwise OR with 0`? Idk, let me know what you think of.

Fixes #827


noftaly:
I have a question: What to do with variables (or member expressions)? The fix can break one's code because `undefined | 0` gives `0`, but `Math.trunc(undefined)` gives `NaN`...
What should we do?

rules/no-bitwise-trunc.js
```diff
@@ -0,0 +1,54 @@
...
+	'[right.value=0]'
+].join('');
+
+const assignementExpressionSelector = [
+	'AssignmentExpression',
+	'[operator="|="]',
+	'[left.type=/(Literal|Identifier)/]',
+	'[right.type="Literal"]',
+	'[right.value=0]'
+].join('');
```
fisker, line 23: No need care about left, use `SourceCode.getText` to get the whole left

fisker (changes requested):

rules/no-bitwise-trunc.js
```diff
@@ -0,0 +1,54 @@
...
+const messages = {
+	[MESSAGE_ID]: 'Use `Math.trunc` instead of `| 0`.'
+};
+
+const binaryExpressionSelector = [
+	'BinaryExpression',
+	'[operator="|"]',
+	'[left.type=/(Literal|Identifier)/]',
+	'[right.type="Literal"]',
+	'[right.value=0]'
```
fisker, line 14: I'm not sure. But I think `| "0"` will get selected, we can use `[left.raw=0]`, just ignore `| 0.0`.

noftaly: Actually `| "0"` behave the exact same as `0` (it will truncate the number, and it has the same ""bug""). I think it should be OK if it is also auto-fixed. This will also allow `0.0` to be taken into account.

fisker: will `null` get selected? I'm not sure.

noftaly: image
Idk i'm lost, it works with *everything*

docs/rules/no-bitwise-trunc.md
```diff
@@ -0,0 +1,23 @@
+# Enforce the use of `Math.trunc` instead of `| 0`
+
+Enforces a convention of using [`Math.trunc()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) instead of `| 0` (bitwise OR with 0) for clarity and more reliable results.
+
+The `| 0` hack helps truncate numbers but is not clear and does not work in some cases.
```
sindresorhus, line 5: Can you linkify `some cases` to more info about that?

sindresorhus (changes requested):

sindresorhus:
The rule name could be improved though:

- `no-bitwise-number-truncation`
- `prefer-math-trunc`

rules/no-bitwise-trunc.js
fisker: I prefer only check `| 0`

papb:
I like `prefer-math-trunc`  :+1:

docs/rules/no-bitwise-trunc.md
noftaly: Sure! Done in the next commit.

noftaly:
I renamed the rule and adressed all reviews :)
The CI seems to be failing, but because of another rule, not prefer-math-trunc. `npm run test` is successful on my laptop, which is up to date with this PR...

test/prefer-math-trunc.js
```diff
@@ -0,0 +1,113 @@
...
+});
+
+const visualizeTester = visualizeRuleTester(test, {
+	parserOptions: {
+		ecmaVersion: 2021
+	}
+});
+
+visualizeTester.run('prefer-math-trunc', rule, [
+	'const foo = 10.01 | 0;'
```
fisker, line 112: Please test

```js
foo |= 0;
```

too

rules/prefer-math-trunc.js
```diff
@@ -0,0 +1,54 @@
...
+	'AssignmentExpression',
+	'[operator="|="]',
+	'[right.type="Literal"]',
+	'[right.raw=0]'
+].join('');
+
+const create = context => {
+	return {
+		[binaryExpressionSelector]: node => {
+			const lhs = context.getSourceCode().getText(node.left);
```
fisker, line 26: Move `context.getSourceCode()` to outer.

fisker (changes requested):
Please add some cases that have comments inside or around. Also some parenthesized cases.

test/prefer-math-trunc.js
```diff
@@ -35,6 +39,16 @@ ruleTester.run('prefer-math-trunc', rule, {
...
 		},
+		{
+			code: 'const foo = (1 + 2 / 3.4) | 0;',
+			errors: [error],
+			output: 'const foo = Math.trunc(1 + 2 / 3.4);'
+		},
+		{
+			code: 'const foo = bar((1.4 | 0) + 2);',
+			errors: [error],
+			output: 'const foo = bar((Math.trunc(1.4)) + 2);'
```
noftaly, line 50: Do you know how I can make it so it does not require the parenthesis around `Math.trunc(1.4)` @fisker ?

fisker: I think we can keep it.

test/prefer-math-trunc.js
```diff
@@ -0,0 +1,153 @@
...
+		},
+		{
+			code: outdent`
+				const foo = 3.4; // comment 1
+				foo |= 0; // comment 2
+			`,
+			errors: [error],
+			output: outdent`
+				const foo = 3.4; // comment 1
+				foo = Math.trunc(foo); // comment 2
```
fisker, line 135: Let's add this one`/* comment 1 */ foo /* comment  2 */ .bar /* comment 3 */ |= /* comment 4 */ 0 /* comment 5*/;`

noftaly: Should it just delete all the comments?

fisker (approved):

sindresorhus:
Nice work, @noftaly 👍🏻

PR #855, opened 2020-10-09
Support `~~x` in `prefer-math-trunc` rule

noftaly:
I just discovered a new way to truncate a number with bitwise operators: `~~3.9`. It has the same downsides as the `| 0` way, so I added it to the rule.
I also discovered another way to do it but it seems a bit too "extreme" (and rare) according to me so I did not include it in the PR: `3.9 & 3.9` (or `foo &= foo`).


test/snapshots/prefer-math-trunc.js.md
```diff
@@ -22,6 +22,22 @@ Generated by [AVA](https://avajs.dev).
...
+    `␊
+    Input:␊
+      1 | const foo = ~~10.01;␊
+    ␊
+    Output:␊
+      1 | const foo = Math.trunc(10.01);␊
+    ␊
+    Error 1/1:␊
+    > 1 | const foo = ~~10.01;␊
+        |              ^^^^^^ Use `Math.trunc` instead of `~~`.␊
```
fisker, line 36: Better report on first `UnaryExpression`

fisker (changes requested):

fisker:
Read the parenthesized tests make me think, will this fail?


```js
foo = (0, 1) | 0;
```

An edge case, but I wounder what the result will it be.

noftaly:
It (correctly?) reported `(0, 1.2) | 0`, to use Math.trunc, but it fixed it with `Math.trunc(0, 1.2)` which is not what we want afaik. It should be `Math.trunc((0, 1.2))` right ?

fisker:
Ah, just as I thought, this also means many other rules not handle similar cases correctly, you can fix it if you want.

> It should be Math.trunc((0, 1.2))

noftaly:
Ok. Should I just do a "naive" if to check if it is a SequenceExpression, or is there a better way to do it with ESLint? I watched at the AST on astexplorer, and parenthesis are never captured.

fisker:
Check `SequenceExpression` and add `()`

fisker (approved):
Good job

fisker:
Sorry, can you test `~~~1.2`?

noftaly:
> Sorry, can you test ~~~1.2, ~(~~1.2), ~~(~1.2)? Are the the same? IDK.

The 3 are the same because the AST doesn't care about parenthesis, so in those 3 cases, the nodes will be only
`UnaryExpression "~" > UnaryExpression "~" > UnaryExpression "~"`.

But still, it does not currently pass because it outputs 2 errors instead of 1:
```js
AssertionError {
    actual: 2,
    code: 'ERR_ASSERTION',
    expected: 1,
    generatedMessage: false,
    operator: 'strictEqual',
    message: `Should have 1 error but had 2: [␊
      {␊
        ruleId: 'prefer-math-trunc',␊
        severity: 1,␊
        message: 'Use `Math.trunc` instead of `~~`.',␊
        line: 1,␊
        column: 13,␊
        nodeType: 'UnaryExpression',␊
        messageId: 'bitwiseNo',␊
        endLine: 1,␊
        endColumn: 22,␊
        fix: { range: [Array], text: 'Math.trunc(~1.44)' }␊
      },␊
      {␊
        ruleId: 'prefer-math-trunc',␊
        severity: 1,␊
        message: 'Use `Math.trunc` instead of `~~`.',␊
        line: 1,␊
        column: 14,␊
        nodeType: 'UnaryExpression',␊
        messageId: 'bitwiseNo',␊
        endLine: 1,␊
        endColumn: 22,␊
        fix: { range: [Array], text: 'Math.trunc(1.44)' }␊
      }␊
    ]`,
  }
```

noftaly:
Well sorry but it's been 45 minutes and I really can't figure out why they are called twice... This might be an ESLint behavior I don't know about or just me being too tired 😂 Any help would be appreciated 🙂

fisker:
Sorry, I was busy just now. It's easy to fix, but I mean are these result the same?

fisker:
If you want select the second `~` use `:not(UnaryExpression[operator="~"])` selector, or check the parent in js, if you want select the third `~`, use js to check the child.

fisker:
All these cases calculate from the third one first? So we should fix to `~Math.trunc(1.2)`, not `Math.trunc(~1.2)`, right? So we should select the second `~`.

fisker:
Similar selector https://github.com/sindresorhus/eslint-plugin-unicorn/blob/616033db7cc2ee0544d9ff94b2868b9c1defe1fa/rules/no-nested-ternary.js#L12

noftaly:
> If you want select the second ~ use :not(UnaryExpression[operator="~"]) > UnaryExpression[operator="~"] > UnaryExpression[operator="~"] selector, or check the parent in js, if you want select the third ~, use js to check the child.

With this, all the tests pass, except `~(~~1.1)` which does not output `~Math.trunc(1.1)` but `Math.trunc(~1.1)`.
As I told, you the AST does not take parenthesis into account making this change much harder. I'll try with the inner-most `~ even though I don't really know how to do it.

fisker (approved):

docs/rules/prefer-math-trunc.md
```diff
@@ -1,23 +1,46 @@
-# Enforce the use of `Math.trunc` instead of `| 0`
+# Enforce the use of `Math.trunc` instead of bitwise operators
 
-Enforces a convention of using [`Math.trunc()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) instead of `| 0` (bitwise OR with 0) for clarity and more reliable results.
+Enforces a convention of using [`Math.trunc()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc) instead of `| 0` ([`bitwise OR`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR) with 0) or `~~x` (two [`bitwise NOT`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_NOT)) for clarity and more reliable results.
 
-The `| 0` hack helps truncate numbers but is not clear and does not work in [some cases](https://stackoverflow.com/a/34706108/11687747).
+These hacks helps truncate numbers but they are not clear and do not work in [some cases](https://stackoverflow.com/a/34706108/11687747).
```
yakov116, line 5:
```suggestion
These hacks help truncate numbers, but they are not clear and do not work in [some cases](https://stackoverflow.com/a/34706108/11687747).
```

noftaly:
Thanks for your help!
Just one thing, now `~~(~1.42)` gives `~Math.trunc(1.42)` and `~(~~1.41)` gives `~(Math.trunc(1.41))`. Is it good? If yes I'll push some more tests

docs/rules/prefer-math-trunc.md
noftaly: Ups sorry I did not see I could commit your suggestion directly 😄 well it is now pushed. Thanks!

fisker:
I've tested https://github.com/sindresorhus/eslint-plugin-unicorn/pull/855/files#diff-71932de485bb3e23da1b2c58e9ce54f1dab22a23ddb196a2dc96b567e5559659R259

noftaly:
Yes but I'll add them also in the "invalid" section, for the autofix

fisker:
Output in snapshot https://github.com/sindresorhus/eslint-plugin-unicorn/pull/855/files#diff-9b02400258b2a0180cdb2d846de70166907382e011c7fd8f7b0c0ecb91230af2R117

noftaly:
So you don't want to add them in `invalid` as well?

fisker:
No, there are the same. Don't you think `visualizeTester` is better?

noftaly:
Yes it's true 😄 well you know better, we'll leave it like that !

fisker:
The `visualizeTester` is designed for easier test on `invalid` tests. Now it can replace most cases without `avaRuleTester`.

fisker:
For such a rule like this one, we don't even need `avaRuleTester` for invalid cases.

fisker:
Do you think we should support 

```js
1.23 >> 0;

1.23 << 0;

1.23 & 0xFFFFFFFFFFFF;

1.23 ^ 0;
```

fisker:
I'm going to merge this one first.

fisker:
Next time, try use other branch (not `master`) to send PR,

noftaly:
> Next time, try use other branch (not master) to send PR,

Sure, will do!

> Do you think we should support

Wow, that's a lot of other ways to do it! I think why not, as I told you I also discovered one: `1.1 & 1.1` or `foo &= foo`. I'll try to add them 🙂

fisker:
I havn't seen people use `1.1 & 1.1` & `foo &= foo`, maybe not worth to speed time for them.

PR #861, opened 2020-10-14
Support more bitwise operations to `prefer-math-trunc`

noftaly:
Added support for
- `x >> 0` ([`Signed Right Shift`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Right_shift) with 0)
- `x << 0` ([`Left Shift`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Left_shift) with 0)
- `x ^ 0` ([`bitwise XOR Shift`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR) with 0)

Thanks @fisker for these ideas 🙂


fisker:
```js
10000.01 & 0xF;
10000.01 & 0xFFFF;
```

This is not safe, I'm not sure if we want check `&`. Anyone have better idea?

rules/prefer-math-trunc.js
```diff
@@ -3,27 +3,33 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
...
+const MESSAGE_ID_BITWISE_XOR = 'bitwiseXor';
+const MESSAGE_ID_BITWISE_AND = 'bitwiseAnd';
 const messages = {
 	[MESSAGE_ID_BITWISE_OR]: 'Use `Math.trunc` instead of `| 0`.',
-	[MESSAGE_ID_BITWISE_NOT]: 'Use `Math.trunc` instead of `~~`.'
+	[MESSAGE_ID_BITWISE_NOT]: 'Use `Math.trunc` instead of `~~`.',
+	[MESSAGE_ID_BITWISE_SIGNED_RIGHT_SHIFT]: 'Use `Math.trunc` instead of `>> 0`.',
+	[MESSAGE_ID_BITWISE_LEFT_SHIFT]: 'Use `Math.trunc` instead of `<< 0`.',
+	[MESSAGE_ID_BITWISE_XOR]: 'Use `Math.trunc` instead of `^ 0`.',
+	[MESSAGE_ID_BITWISE_AND]: 'Use `Math.trunc` instead of `& 0xF`.'
```
fisker, line 16: No need use so many messageIds,

```js
[MESSAGE_ID_BITWISE]: 'Use Use `Math.trunc` instead of `{{operator}} {{value}}`'
```

```js
context.report({
	messageId: MESSAGE_ID_BITWISE,
	data: {
		operator: '&',
		value: '0xFFFFF'
	}
});
```

rules/prefer-math-trunc.js
```diff
@@ -33,6 +39,23 @@ const createBitwiseNotSelector = (level, isNegative) => {
 	return isNegative ? `:not(${selector})` : selector;
 };
 
+// All binary expressions
+const binarySelectors = [
```
fisker, line 43: Listen to `BinaryExpression` and `AssignmentExpression` selector, check `operator` and `right` inside,

```js

const operators = new Map([
	['|', 0],
	// ...
]);

'BinaryExpression[right.type="Literal"]'(node) {
	const {operator, right} = node;
	if (operators.has(node.operator) && 
		typeof right.value === 'number' && 
		right.value === operators.get(node.operator)
	) {
		// ...
	}
}
```

papb:
Good catch @fisker, I think someone should carefully check the ecmascript spec to check that each of these methods are really guaranteed to correspond to trunc

fisker (changes requested):

test/prefer-math-trunc.js
```diff
@@ -12,15 +12,26 @@ const ruleTester = avaRuleTester(test, {
...
+const MESSAGE_ID_BITWISE_LEFT_SHIFT = 'bitwiseLeftShift';
+const MESSAGE_ID_BITWISE_XOR = 'bitwiseXor';
+const MESSAGE_ID_BITWISE_AND = 'bitwiseAnd';
 
 const errorsBitwiseOr = [{messageId: MESSAGE_ID_BITWISE_OR}];
 const errorsBitwiseNot = [{messageId: MESSAGE_ID_BITWISE_NOT}];
+const errorsBitwiseSignedRightShift = [{messageId: MESSAGE_ID_BITWISE_SIGNED_RIGHT_SHIFT}];
+const errorsBitwiseLeftShift = [{messageId: MESSAGE_ID_BITWISE_LEFT_SHIFT}];
+const errorsBitwiseXor = [{messageId: MESSAGE_ID_BITWISE_XOR}];
+const errorsBitwiseAnd = [{messageId: MESSAGE_ID_BITWISE_AND}];
```
fisker, line 25: The `invalid` cases for `ruleTester` can test with `errors: 1`, but make sure every message is tested in `visualizeTester`

fisker:
@papb Do you think we should just ignore `&`?

papb:
> @papb Do you think we should just ignore `&`?

Yes. And also make sure via looking at the ecmascript spec that the other ideas (`x >> 0`, `x << 0` and `x ^ 0`) indeed behave like Math.trunc in every reasonable situation.

rules/prefer-math-trunc.js
noftaly: Ohh I didn't know that! It will be way better

rules/prefer-math-trunc.js
noftaly: I'll do it but why do you want that? To have a more readable code? Is it faster?

fisker: It's more flexible, 

```js
const operators = new Map([
	['|', 0],
	// ...
	['SOME_NEW_OPERATOR', A_NEW_VALUE]
]);
```

fisker: Image if we already doing this, we only need add 

```js
	['<<', 0],
	['>>', 0],
```

All done.

test/prefer-math-trunc.js
noftaly: I'm not sure I understood, please tell me if my last 2 commits are not OK 😄

noftaly:
I've made this very quickly to show most of the cases:
image
We can see that the only differences are when the left-hand side is `undefined`, because `Math.trunc` gives `NaN`, but the bitwise operators give `0`

docs/rules/prefer-math-trunc.md
```diff
@@ -44,3 +67,19 @@ foo = Math.trunc(3.3);
...
+```js
+const foo = 0 << 3.3;
+```
+
+```js
+const foo = 0 ^ 3.3;
+```
+
+```js
+const foo = 3.3 & 0x0FA9;
```
fisker, line 84: Please remove `&` from docs

rules/prefer-math-trunc.js
```diff
@@ -41,26 +30,41 @@ const bitwiseNotUnaryExpressionSelector = [
...
 	return {
-		[bitwiseOrBinaryExpressionSelector]: node => {
-			context.report({
-				node,
-				messageId: MESSAGE_ID_BITWISE_OR,
-				fix: fixer => fixer.replaceText(node, mathTruncFunctionCall(node.left))
-			});
+		'BinaryExpression[right.type="Literal"]': node => {
+			const {operator, right} = node;
+			if (binaryOperators.has(operator) && right.value === 0) {
```
fisker, line 43: We are now using  `value === 0`, we need add some tests to cover other `0`

```js
foo | .0;
foo | 0.;
foo | 0.00_00;
foo | 0b0;
foo | 0x0000_0000_0000;
```

Please add

```js
foo | (0.4 + .6);
```

as well

fisker (changes requested):

rules/prefer-math-trunc.js
noftaly: Uh yes, just wondering, how can I make the math expressions work?? With an eval? Making the computations directly in the code? Or does ESLint have a util function to do that?

fisker: No, we don't need make it work.

fisker: In [this rule](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/859/files#diff-a1cd7e3eacfd287d9522bdaa755f05111c27b9de74197392c5ceb4b6b68a2762R40) there're several cases equals to `1`, but we don't want it work.

rules/prefer-math-trunc.js
```diff
@@ -41,26 +30,41 @@ const bitwiseNotUnaryExpressionSelector = [
...
 		},
-		[bitwiseOrAssignmentExpressionSelector]: node => {
-			context.report({
-				node,
-				messageId: MESSAGE_ID_BITWISE_OR,
-				fix: fixer => fixer.replaceText(node, `${sourceCode.getText(node.left)} = ${mathTruncFunctionCall(node.left)}`)
-			});
+		'AssignmentExpression[right.type="Literal"]': node => {
+			const {operator, right, left} = node;
+			if (assignmentOperators.has(operator) && right.value === 0) {
```
fisker, line 57:
```suggestion
			if (binaryOperators.has(operator.slice(0, -1)) && right.value === 0) {
```

So we don't have to add an operator in two places.

test/prefer-math-trunc.js
fisker: Your commit is fine, but since we have `visualizeTester`, we can test messages in that, and in `ruleTester` you can use

```js
{
	code: 'foo',
	ouput: 'bar',
	errors: 3
}
```

To make it easier.

rules/prefer-math-trunc.js
noftaly: Oh you want me to add them to the `valid` case?

fisker: Yes

fisker: Add `foo & 0xFFFFFFFFFFFF` too, we do't want other people think about `&` again.

test/prefer-math-trunc.js
noftaly: I'm sorry but I'm not really familiar with the `visualizeTest`... How can I test the message inside? And why did you put `errors: 3` in the ruleTester? (I don't understand the `3`)

fisker: number means numbers of errors

fisker: > How can I test the message inside?

The snapshot is an assertion.

test/prefer-math-trunc.js
```diff
@@ -59,6 +65,32 @@ ruleTester.run('prefer-math-trunc', rule, {
...
+			errors: errorsBitwise,
+			output: 'const foo = Math.trunc(1.4);'
+		},
+		{
+			code: 'const foo = 1.4 | 0b0;',
+			errors: errorsBitwise,
+			output: 'const foo = Math.trunc(1.4);'
+		},
+		{
+			code: 'const foo = 1.4 | 0x0000_0000_0000;',
```
fisker, line 85: These cases need go to `visualizeTester`, because the message will be `` `| 0x0000_0000_0000 ` ``, different from `` `| 0 ` ``, and it's not tested.

fisker: You know what, let's move all invalid tests to `visualizeTester`.

fisker: I'll commit.

noftaly: Ok I'll let you do it

fisker:
Do you think we should merge `BinaryExpression` and `AssignmentExpression` ? https://github.com/sindresorhus/eslint-plugin-unicorn/pull/861/commits/0be93e24b2b8c12b7add4b972f4c175418492d47

fisker (approved):

noftaly:
Yes why not! I thought of it multiple times because they have very similar code...
Maybe we could make the fix a variable to make it more readable, or do it like this:
```js
isAssignment ?
    sourceCode.getText(left) + ' = ' + mathTruncFunctionCall(left) : // Or a template litteral, as you wish
    mathTruncFunctionCall(left);
```

PR #866, opened 2020-10-16
`no-console-spaces`: Improve report location

fisker:
Directly point to the `SPACE`.


rules/no-console-spaces.js
```diff
@@ -23,60 +23,60 @@ const selector = methodSelector({
 });
 
 // Find exactly one leading space, allow exactly one space
-const hasLeadingSpace = value => value.length > 1 && value.charAt(0) === ' ' && value.charAt(1) !== ' ';
+const hasLeadingSpace = value => /^ [^ ]/.test(value);
```
sindresorhus, line 26: I generally try to avoid regex when it's not necessary. In this case, the regex makes the code both slower and less readable.

sindresorhus:
Linting is failing.

PR #869, opened 2020-10-17
Fixed example `prefer-math-trunc`'s `README.md`

hong4rc:
Fixed position of number:

```js
const foo = 0 >> 3.3; // 0
const foo = 3.3 >> 0; // 3
```

```js
const foo = 0 << 3.3; // 0
const foo = 3.3 << 0; // 3
```


fisker:
Sorry, the old one is correct.

hong4rc:
> Sorry, the old one is correct.

I think you should merge this PR or fix it by yourself

PR #872, opened 2020-10-19
`prefer-math-trunc`: Update examples

fisker:
`const foo = 1.23 | 0;` is less common than `const foo = bar | 0;`

//cc @noftaly


noftaly (approved):

sindresorhus:
Is it clear enough to the user that `bar` is a number? Maybe make it explicit.

fisker:
Nobody will write this code 

`const foo = 37.4 | 0;`

maybe 

```js
const foo = 37.4;
console.log(foo | 0);
```

?

sindresorhus:
Yes, the latter.

PR #874, opened 2020-10-20
`prefer-number-properties`: Add `Infinity` support

### LINKED ISSUE #863: Add `Infinity` support to `prefer-number-properties` rule
Fail:

```js
const foo = Infinity;
const bar = -Infinity;
```

Pass

```js
const foo = Number.POSITIVE_INFINITY;
const bar = Number.NEGATIVE_INFINITY;
```

I [mentioned this in the original proposal](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/594#issuecomment-601560552), but [@sindresorhus prefer use `Infinity`](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/594#issuecomment-601564678).

Can we add this support behind options?

fisker:
Fixes #863


docs/rules/prefer-number-properties.md
```diff
@@ -58,3 +68,46 @@ const foo = Number.isFinite(10);
...
+
+```js
+const isNegativeZero = value => value === 0 && 1 / value === Number.NEGATIVE_INFINITY;
+```
+
+# Options
+
+Type: `object`
+
+### Infinity
```
sindresorhus, line 84:
```suggestion
### infinity
```

The option name should use normal property casing.

fisker: I think in this case  better use the name of the global object?

Maybe we'll need add `NaN` too

sindresorhus: That will just be annoying for users. `{Infinity: true}` will cause lint errors for users using JS configs. I don't think it's worth it just to match the exact casing of the global name.

fisker: Renamed to `checkInfinity`

PR #877, opened 2020-10-20
`no-useless-undefined`: Add `checkArguments` option

### LINKED ISSUE #760: no-useless-undefined - should not be fixable
As it stands, the rule corrects `console.log(undefined)` to `console.log()`. These have different runtime behaviour, so it isn't safe to autofix. The fixer should be downgraded to a suggestion.

### LINKED ISSUE #801: Make `no-useless-undefined` work with TypeScript
My original code:
```ts
  const [currentWorkspace, setCurrentWorkspace] = useState<IWorkspace>();

  const onAddWorkspaceClick = () => {
    setCurrentWorkspace(undefined);
  };
```

After I enable `no-useless-undefined`, it forces me to remove `undefined`. Then TypeScript complains:

image

any suggestions?

fisker:
Fixes #760
Fixes #801


futpib:
Consider renaming `ignoreArguments` to `checkArguments` (and inverting it's meaning accordingly). #881

docs/rules/no-useless-undefined.md
```diff
@@ -82,6 +82,31 @@ function foo({bar}) {
 foo();
 ```
 
+## Options
+
+Type: `object`
+
+### ignoreArguments
+
+Type: `boolean`<br>
```
sindresorhus, line 91:
```suggestion
Type: `boolean`\
```

fisker:
Done.

docs/rules/no-useless-undefined.md
```diff
@@ -82,6 +82,31 @@ function foo({bar}) {
...
+## Options
+
+Type: `object`
+
+### checkArguments
+
+Type: `boolean`\
+Default: `true`
+
+Pass `checkArguments: false` to disable check on function call.
```
sindresorhus, line 94: The description should describe what this option does when `true`, not `false`.
```suggestion
Pass `checkArguments: false` to disable check on function call arguments.
```

PR #878, opened 2020-10-20
`no-for-loop`: Respect declaration kind

### LINKED ISSUE #742: `no-for-loop` produces invalid autofix when mutation is present
If the current element gets mutated in the loop body, we can't use `const` for it, and probably want to use `let` instead.

Original code:

```js
for (let i = 0; i < plugins.length; i++) {
  let plugin = plugins[i];
  plugin = calculateSomeNewValue();
  // ...
}
```

After autofix (invalid, fails the [no-const-assign](https://eslint.org/docs/rules/no-const-assign) lint rule):

```js
for (const plugin of plugins) {
  plugin = calculateSomeNewValue();
  // ...
}
```

Here is one autofix that would be valid:

```js
for (let plugin of plugins) {
  plugin = calculateSomeNewValue();
  // ...
}
```

fisker:
Fixes #742


test/no-for-loop.js
```diff
@@ -725,10 +725,72 @@ ruleTesterEs5.run('no-for-loop', rule, {
...
+				for (let i = 0; i < positions.length; i++) {
+					let last: vscode.Position | vscode.Range = positions[i];
+					let selectionRange = allProviderRanges[i];
+				}
+			`,
+			output: outdent`
+				for (let [i, last]: [number, vscode.Position | vscode.Range] of positions.entries()) {
+					let selectionRange = allProviderRanges[i];
+				}
+			`,
```
fisker, line 743: The old code fix this to 

```js
				for (let [i, last: vscode.Position | vscode.Range] of positions.entries()) {
					let selectionRange = allProviderRanges[i];
				}
```

This cause parser error.

I don't use typescript, is this good?

sindresorhus: Yes, that looks correct.

PR #883, opened 2020-10-21
`no-for-loop`: Improve output when using the TypeScript parser

fisker:
Space between `id` and `typeAnnotation` was not removed .


sindresorhus:
You're being super productive 🙌🏻

PR #886, opened 2020-10-21
`prefer-number-properties`: Support `computed` key of `ClassProperty` `MethodDefinition` `Property`

fisker:
Not common use, but possible.


sindresorhus:
I cannot imagine anyone doing this, but I'm not against making the rules super sturdy either. 👍🏻

sindresorhus:
Travis is failing though.

PR #887, opened 2020-10-21
Add `prefer-array-some` rule

### LINKED ISSUE #736: Rule proposal: `prefer-array-some`
# Fail

```js
if (array.find(element => element === '🦄')) {
}
```

```js
const foo = array.find(element => element === '🦄') ? bar : baz;
```

# Pass

```js
if (array.some(element => element === '🦄')) {
}
```

```js
const foo = array.some(element => element === '🦄') ? bar : baz;
```

Just experienced one  few days ago, and kind of follow up for https://github.com/sindresorhus/eslint-plugin-unicorn/issues/730

fisker:
Fixes #736


docs/rules/prefer-array-some.md
```diff
@@ -0,0 +1,41 @@
+# Prefer `.some(…)` over `.find(…)`.
+
+Prefer use [`Array#some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) over [`Array#find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) when testing array.
```
sindresorhus: "when testing array" is not clear enough about what it does.

sindresorhus (changes requested):

sindresorhus:
Is it possible to add autofixing?

docs/rules/prefer-array-some.md
fisker: But  we don't know what `.find` is doing.

fisker:
> Is it possible to add autofixing?

It's possible, but not safe.

fisker:
Note to my self, forgot negative test


```js
if (!foo.find(fn)){}
```

docs/rules/prefer-array-some.md
sindresorhus: Something like:

> when ensuring at least one element in the array passes a given check.

fisker:
After a second thought, this rule, should not related to `if` nor  `ternary`, it should be any expression returns a boolean.


```js
const foo = !array.find(element => element === foo);
```

Should be reported too.


The same thing is missing in `explicit-length-check` rule.

And it seems hard to implement.

Maybe I'll use some logic from `explicit-length-check` , so we can check `if (!foo.find(fn)){}` `if (foo.find(fn)  || bar){}` etc

fisker:
Ready for review.

sindresorhus:
Nice! I really like how simple the rule logic is because it's just using a reusable utility. 👍🏻

PR #891, opened 2020-10-21
`prevent-abbreviations`: Add `ignore` option

### LINKED ISSUE #567: prevent-abbreviations: whitelist has no effect on file name

### LINKED ISSUE #752: prevent-abbreviations: How to avoid "e" matching "e2e"?
`prevent-abbreviations` has a replacement defined for `e` to `error` or `event`. Unfortunately, something like `e2e` also matches this replacement, e.g. a file like `e2e-test-utils.js`. The suggested replacements are then `error2error-test-utils.js` or `event2event-test-utils.js`, which don't make sense and would likely confuse people.

What I would like to do is define `e2e` as a replacement key, so that I can set it to `false` to simply allow `e2e` or define a replacement `end-to-end`. But right now, even if that is defined, it still matches the `e` replacement.
Could the logic be changed so that if the `e2e` matches, other replacements are not tried anymore? Maybe in the order they are defined, or ordered descending by their length (i.e. their specificity)

fisker:
Fixes #567
Fixes #752

Maybe we can remove `whitelist` option later.

### DIFF
--- a/docs/rules/prevent-abbreviations.md
+++ b/docs/rules/prevent-abbreviations.md
@@ -8,7 +8,6 @@ You can find the default replacements [here](https://github.com/sindresorhus/esl
 
 This rule is fixable only for variable names with exactly one replacement defined.
 
-
 ## Fail
 
 ```js
@@ -22,7 +21,6 @@ const e = document.createEvent('Event');
 ```js
 class Btn {}
 ```
-
 
 ## Pass
 
@@ -59,7 +57,6 @@ const levels = {
 // Property is not checked by default
 this.evt = 'click';
 ```
-
 
 ## Options
 
@@ -105,7 +102,7 @@ The example below:
 
 ### extendDefaultReplacements
 
-Type: `boolean`<br>
+Type: `boolean`\
 Default: `true`
 
 Pass `"extendDefaultReplacements": false` to override the default `replacements` completely.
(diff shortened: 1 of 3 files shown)


futpib:
I think we could solve those issues by fixing `whitelist` and without introducing regexps.

We could make whitelist match any subsequence of the split identifier. For example (following up this https://github.com/sindresorhus/eslint-plugin-unicorn/issues/567#issuecomment-591007369):
* For `fooE2eBar` all of these would be checked against whitelist:
* * `foo`
* * `e`
* * `2`
* * `bar`
* * `barE`
* * `e2`
* * `2e`
* * `eBar`
* * `fooE2`
* * `e2e`
* * `2eBar`
* * `fooE2e`
* * `e2eBar`
And if any of them is in the whitelist that subsequence is ignored.

But in my opinion it should not ignore the whole identifier. For example, if the user had `e2e` in their `whitelist`, `e2eErr` should still be fixed to `e2eError`.

fisker:
I don't think there is a good solution for these cases, #567 has been there for a long time. 

Instead of blocking people use this rule, I believe give an option to let them use on there own risk is the right thing to do.

sindresorhus:
@futpib Can you open a new issue for your above comment? I think we should improve this rule, but I also think it's valuable to let users easily ignore things like this. There might be other cases we just cannot fix and this option let's user ignore instead of having to disable the whole rule.

PR #893, opened 2020-10-27
Update comments' accuracy in the default config

noftaly:
[NodeJS 14 just transitioned to Long Term Support!](https://github.com/nodejs/node/releases/tag/v14.15.0)

Hence, I've updated the default rules accordingly (`numeric-separators-style`: `off` -> `error`), and I've clarified the LTS version where the `String#replaceAll` method will be shipped.

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,8 @@ module.exports = {
 				'unicorn/no-useless-undefined': 'error',
 				'unicorn/no-zero-fractions': 'error',
 				'unicorn/number-literal-case': 'error',
-				'unicorn/numeric-separators-style': 'error',
+				// TODO: Enable this by default when NodeJS 10 is deprecated.
+				'unicorn/numeric-separators-style': 'off',
 				'unicorn/prefer-add-event-listener': 'error',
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -49,8 +49,7 @@ module.exports = {
 				'unicorn/no-useless-undefined': 'error',
 				'unicorn/no-zero-fractions': 'error',
 				'unicorn/number-literal-case': 'error',
-				// TODO: Enable this by default when it's shipping in a Node.js LTS version.
-				'unicorn/numeric-separators-style': 'off',
```
sindresorhus, line 53: The comment was wrong. We need to wait until April next year when Node.js 10 is obsolete. https://github.com/nodejs/Release

sindresorhus (changes requested):

index.js
noftaly: Oh ok 🙂

sindresorhus:
Why did you close? We can still update the comments.

noftaly:
Ups sorry, did not thought of that 😅

PR #899, opened 2020-11-05
Report more specific error in `import-style` when no style is allowed

FloEdelmann:
Given a configuration like this:

```js
  'unicorn/import-style': ['error', {
    styles: {
      fs: {
        unassigned: false,
        default: false,
        namespace: false,
        named: false,
      },
      'fs/promises': {
        named: true,
      },
    },
  }],
```

With this configuration, importing `fs` is effectively forbidden (in favor of `fs/promises` in this case). However, the reported error message is misleading:

```
Use unassigned, default, namespace or named import for module `fs`.
```

This PR adds a special error message for this case:

```
Importing module `fs` is not allowed.
```

### DIFF
--- a/rules/import-style.js
+++ b/rules/import-style.js
@@ -3,9 +3,11 @@ const {defaultsDeep} = require('lodash');
 const {getStringIfConstant} = require('eslint-utils');
 const eslintTemplateVisitor = require('eslint-template-visitor');
 
-const MESSAGE_ID = 'importStyle';
+const MESSAGE_ID_IMPORT_STYLE = 'importStyle';
+const MESSAGE_ID_IMPORT_STYLE_NONE = 'importStyleNone';
 const messages = {
-	[MESSAGE_ID]: 'Use {{allowedStyles}} import for module `{{moduleName}}`.'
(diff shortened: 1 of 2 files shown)


futpib:
You should probably use `no-restricted-imports` or `no-restricted-modules` for that.

FloEdelmann:
> You should probably use `no-restricted-imports` or `no-restricted-modules` for that.

Indeed, I'll use that. Nevertheless, when specifiying any `importStyle: false`, the reported message will include that `importStyle` even though it's not allowed, so this PR to fix the message wording is still relevant.

futpib:
That is indeed a wrong message for the configuration. I submitted an alternative fix for it.

Another issue here is that configurations like below in my opinion presents a misuse of this rule.
```js
      fs: {
        unassigned: false,
        default: false,
        namespace: false,
        named: false,
      },
```
I think it should be simply disallowed and `no-restricted-imports` or `no-restricted-modules` rulse should be recommended to the user when such configuration is detected.

FloEdelmann:
That's fine for me, too.

If at one point, an error is thrown for such configurations, please also make sure to suggest using `no-restricted-imports` or `no-restricted-modules` instead.

PR #907, opened 2020-11-24
`new-for-builtins`: Remove auto-fix for `new String`, `new Boolean()`, and `new Number()`

### LINKED ISSUE #901: <new-for-builtins> : potential side effects auto fix
This is more of a question than actual bug report.

`new-for-builtins` autofixes this code:

```js
const str = new String('test')
```

to this:
```js
const str = String('test')
```

The problem is that 2 values are not identical.
For example:

```js
String('test') === String('test') // true
new String('test') === new String('test') // false
```

Is this intended behavior?

The rule was first implemented in #105

fisker:
`new Symbol()` and `new BigInt()`  still auto-fix, they can't call with `new` anyway.

Fixes #901

### DIFF
--- a/rules/new-for-builtins.js
+++ b/rules/new-for-builtins.js
@@ -31,15 +31,20 @@ const create = context => {
 			const {name} = callee;
 
 			if (disallowNew.has(name) && !isShadowed(context.getScope(), callee)) {
-				context.report({
+				const problem = {
 					node,
 					messageId: 'disallow',
-					data: {name},
-					fix: fixer => fixer.removeRange([
+					data: {name}
+				}
+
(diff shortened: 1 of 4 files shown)


rules/new-for-builtins.js
```diff
@@ -31,15 +31,20 @@ const create = context => {
...
 
 			if (disallowNew.has(name) && !isShadowed(context.getScope(), callee)) {
-				context.report({
+				const problem = {
 					node,
 					messageId: 'disallow',
-					data: {name},
-					fix: fixer => fixer.removeRange([
+					data: {name}
+				}
```
sindresorhus, line 38:
```suggestion
				};
```

PR #910, opened 2020-11-30
Fix `false` import styles seen in reported problem messages

futpib:
Fixes #899 (Alternative fix)

### DIFF
--- a/rules/import-style.js
+++ b/rules/import-style.js
@@ -172,7 +172,7 @@ const create = context => {
 	styles = new Map(
 		Object.entries(styles).map(
 			([moduleName, styles]) =>
-				[moduleName, new Map(Object.entries(styles))]
+				[moduleName, new Set(Object.entries(styles).filter(([_, isAllowed]) => isAllowed).map(([moduleName]) => moduleName))]
 		)
 	);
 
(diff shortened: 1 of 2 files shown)


sindresorhus (approved):

rules/import-style.js
```diff
@@ -172,7 +172,7 @@ const create = context => {
 	styles = new Map(
 		Object.entries(styles).map(
 			([moduleName, styles]) =>
-				[moduleName, new Map(Object.entries(styles))]
+				[moduleName, new Set(Object.entries(styles).filter(([_, isAllowed]) => isAllowed).map(([moduleName]) => moduleName))]
```
fisker, line 175:
```suggestion
				[moduleName, new Set(Object.entries(styles).filter(([, isAllowed]) => isAllowed).map(([moduleName]) => moduleName))]
```

fisker (approved):
LGTM, I think we can remove the unused key?

rules/import-style.js
fisker: The variable name seems not right too?

```suggestion
				[moduleName, new Set(Object.entries(styles).filter(([, isAllowed]) => isAllowed).map(([importStyle]) => importStyle))]
```

PR #916, opened 2020-12-05
`numeric-separators-style`: Add `checkOnlyIfSeparator` option

### LINKED ISSUE #868: numeric-separators-style: option to accept numbers without grouping
I'm a big fan of the new [`numeric-separators-style`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/numeric-separators-style.md) rule from #833.  Great idea!

I'd like to enable the `numeric-separators-style` rule to catch incorrect groupings without requiring that all numbers use separators.  For example:
```js
const foo = 1234;  // Pass: No separators
const bar = 1_234; // Pass: Correct grouping
const baz = 12_34; // Fail: Incorrect grouping
```
I couldn't find a way to achieve this.  Configuring `minimum` between 0 and 4, `foo` and `baz` fail.  With `minimum` 5 or above, `bar` and `baz` fail.

**Is this an feature that you might consider?  If so, how to configure?**

An idea:  Additional options for each number type:

| Name | Description |
| --- | --- |
| `requireLeft` | Require separators to the left of the decimal point for numbers with at least `minimum` digits to the left.  (Current behavior: `true`) |
| `requireRight` | Require separators to the right of the decimal point for numbers with at least `minimum` digits to the right.  (Current behavior: `true`) |
| `forbidLeft` | Forbid separators to the left of the decimal point for numbers with less than `minimum` digits to the left.  (Current behavior: `true`) |
| `forbidRight` | Forbid separators to the right of the decimal point for numbers with less than `minimum` digits to the right.  (Current behavior: `true`) |

The desired example behavior could then be accomplished with `requireLeft: false` (to never require grouping) or something like `minimum: 7, forbidLeft: false` to require grouping 7-digits or more, but allow grouping fewer.

noftaly:
Added an option for each literal type, named `checkOnlyIfSeparator`.
As explained in the docs, it:
Check if the group sizes are valid **only** if there are groups, separated with an `_`.

Example:
```js
const foo = 100000; // Pass, because there are no groups, so no check was performed.
const bar = 1_000_000; // Pass, because the group sizes are correct.
const baz = 10_0_000_0; // Fail, because there are groups, but they are incorrect.
```

This option is disabled by default.

Closes #868


fisker:
I'm not sure if we should add  this option, but you didn't test `12_34.5678e1234.5678`, should we ignore all, or just part?

noftaly:
Considering it has a separator, I think the entire number should be checked, to be more consistent.

> I'm not sure if we should add this option

No problem, I did not know if I should wait for a "help-wanted" label, so I just did it anyway 😄 Don't hesitate to close the PR if you don't think it should be added.

sindresorhus:
While I wouldn't personally use this option, I can see how it would be useful if you want to enforce "correctness", but not style.

docs/rules/numeric-separators-style.md
```diff
@@ -33,6 +33,19 @@ If you want a custom group size for a specific number type, you can specify it h
 
 There are four number types; [`hexadecimal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Hexadecimal), [`binary`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Binary), [`octal`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Octal) and [`number`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type). Each of them can be associated with an object containing the following options:
 
+**`checkOnlyIfSeparator`**
+
+Type: `boolean`
```
sindresorhus, line 38:
```suggestion
Type: `boolean`\
Default: `false`
```

docs/rules/numeric-separators-style.md
```diff
@@ -33,6 +33,19 @@ If you want a custom group size for a specific number type, you can specify it h
...
+
+Check if the group sizes are valid **only** if there are groups separated with an `_`.
+
+Example:
```
sindresorhus, line 42:
```suggestion
Example:

```

sindresorhus (changes requested):

sindresorhus:
> Considering it has a separator, I think the entire number should be checked, to be more consistent.

👍

sindresorhus:
@fisker Are you ok with merging this?

fisker:
Yes, I'm fine with this, but I think the option can have a better name, like `onlyIfSeparatorExists` or `onlyIfSeparatorPresents`?

And do we want a top level one to easier set for all?

Also, need more tests on different options, and seperator in different position of `1234.5678e1234`

sindresorhus:
> onlyIfSeparatorExists

👍

sindresorhus:
Alternatively: `onlyIfContainsSeparator`

sindresorhus:
> And do we want a top level one to easier set for all?

👍

noftaly:
> And do we want a top level one to easier set for all?

Just to be sure, do you want only *one* at the top level, or one for each types *and* a "master" setting at the top level that overrides other settings if set?

fisker:
Both, if top one is set, all default to it.

noftaly:
> Both, if top one is set, all default to it.

Doesn't it add too many complexity and redundancy. I thought of doing that too but I think it might confuse the user and clutter the options...

fisker:
```js
{
  onlyIfContainsSeparator: true,
  binary: {
    onlyIfContainsSeparator: false
  }
}
```

Pretty clear to me, think this way you can set `onlyIfContainsSeparator` at  top level, but can be override on each.

noftaly:
Okay, I'll do that then 👍🏼

noftaly:
Just curious about the actions, why does the lint errors appear in the build job and not in the lint job? (cf [my last failing commit](https://github.com/sindresorhus/eslint-plugin-unicorn/runs/1722998311)

fisker:
`lint` is  not really  `lint`, maybewe need rename it, it's `run rules on codebase`, 😄

fisker:
I changed the logic a little bit, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/916/commits/ebf5aeee6f8e3054e0a0bd6681bdb76c853a8afd, I think this is more reasonable.

PR #920, opened 2020-12-06
Add `prefer-array-index-of` rule

### LINKED ISSUE #919: Rule proposal: `prefer-array-index-of`
All built-ins have [`Array#indexOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) in addition to [`Array#findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex). Prefer `Array#indexOf()` over use of `Array#findIndex()` passing a function that tests equality as parameter.

This rule is fixable.

## Fail

```js
[].findIndex(x => x === 'foo');
values.findIndex(x => 'foo' === x);
values.findIndex(x => {return x === 'foo';});
['foobar'].findIndex(x => {return 'foo' === x;});
values.findIndex(function (x) {return x === 'foo';});
['foobar'].findIndex(function (x) {return 'foo' === x;});
```

## Pass

```js
[].findIndex(i => i === list[i]);
[].findIndex(x => x === 'foo' && isValid);
[].findIndex(x => y === 'foo');
[].findIndex(x => y.x === 'foo');
[].findIndex(x => x == 'foo');
['foobar'].findIndex(function (x) {return 'foo' == x;});
[].findIndex(x => x !== 'foo');
values.findIndex(function (x) {return x !== 'foo';});
```

Fabrice-TIERCELIN:
Add `prefer-indexof` rule

I have successfully run `npm test`. I have successfully manually tested the lint and the fix on a project. I have run `npm run integration` but my node is too old.

Fixes #919


rules/prefer-indexof.js
```diff
@@ -0,0 +1,141 @@
...
+};
+
+const templates = eslintTemplateVisitor();
+
+const objectVariable = templates.variable();
+const argumentsVariable = templates.spreadVariable();
+
+const findIndexCallTemplate = templates.template`${objectVariable}.findIndex(${argumentsVariable})`;
+
+const getSearchedConstant = node => {
```
fisker, line 17: This function can be simplified to use an esquery selector.

```js
// Matches `foo.findIndex(bar => bar === baz)`
ArrowFunctionExpression > BinaryExpression.body[operator="==="]

// Matches `foo.findIndex(bar => {return bar === baz})`
// Matches `foo.findIndex(function (bar) {return bar === baz})`
:matches(ArrowFunctionExpression, FunctionExpression) > BlockStatement[length=0].body > ReturnStatement.body > BinaryExpression.argument[operator="==="]
```

Can you try it?

rules/prefer-array-index-of.js
```diff
@@ -0,0 +1,168 @@
...
+	};
+
+	return {
+		[selector](node) {
+			const [callback] = node.arguments;
+			const binaryExpression = callback.body.type === "BinaryExpression" ?
+				callback.body :
+				callback.body.body[0].argument;
+			const [element] = callback.params;
+			const {left, right} = binaryExpression;
```
fisker, line 128: @Fabrice-TIERCELIN I've selected all the nodes you need, do you think you can continue work on this? You only need check `left` or `right` is the same `Identifier` with `element`.

Fabrice-TIERCELIN:
May [`node.callee.object`](https://github.com/Fabrice-TIERCELIN/eslint-plugin-unicorn/blob/2afc4ecd6dc80c746ae1923c3f74f507f41fe8df/rules/prefer-array-index-of.js#L100) be `undefined`?

fisker:
No, it can't

Fabrice-TIERCELIN:
Have you got common code to check _extended_ literals like:

- `-8`
- `1 + 1`
- `'foo' + 'bar'`
- `10 * (5 - 9)`
- ...

?

fisker:
No, you are not doing right, you should not care about the searching value, you only need care about `left/right` is the same as `element`

fisker:
Meaning `foo.findIndex(bar => bar === a.b.c)` should be fixed to `foo.indexOf(a.b.c)` too, WDYT?

Fabrice-TIERCELIN:
I think I should only check it's an expression that doesn't change, that is to say not that for example:

`foo.findIndex(bar => bar === a.getValueAndUpdateIt())`

In this case, check that it's a literal is just a first objective that can be improved in the future.

fisker:
Yes, I considered that, you can use `hasSideEffect` to make sure it don't has SideEffect. So we can fix it, otherwise, I think we can use ESLint suggestion 

And I suggest still report it , because I can't image anyone really do something in findIndex call

rules/prefer-array-index-of.js
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			const problem = {
+				node,
+				messageId: MESSAGE_ID_FINDINDEX
+			};
+
+			const searchedConstantText = sourceCode.getText(constant);
+			const objectText = getNodeText(node.callee.object);
+			problem.fix = fixer => fixer.replaceText(node, `${objectText}.indexOf(${searchedConstantText})`);
```
fisker, line 101: Let's fix in this way

1. replace `node.callee.property` with `"indexOf"`
2. replace `node.arguments[0]` with `searchedConstantText`

fisker: And please use [`isParenthesized`](https://eslint-utils.mysticatea.dev/api/ast-utils.html#isparenthesized) to check parenthesize.

Fabrice-TIERCELIN:
The parenthesis should be checked around what?

fisker:
It's needed for 2 places, but not common case.

`foo.findIndex(element => element === (a, b))`

This one need add `()`

`foo.findIndex((element => element === (a, b)))`

This one already has `()` around argument[0], don't need.

But I think it's okay to ignore.

rules/prefer-array-index-of.js
```diff
@@ -0,0 +1,122 @@
...
+			if (
+				left.type === 'Identifier' &&
+				!left.callee &&
+				parameter.name === left.name
+			) {
+				item = left;
+				passiveExpression = right;
+			} else if (
+				right.type === 'Identifier' &&
+				!right.callee &&
```
fisker, line 67: `Identifier` don't have `callee` property

rules/prefer-array-index-of.js
```diff
@@ -75,12 +75,8 @@ const create = context => {
 
 			if (
 				parameter.type !== 'Identifier' ||
```
fisker, line 77: parameter always `Identifier`, L22 and L31

fisker: And `passiveExpression` always truly.

rules/prefer-array-index-of.js
Fabrice-TIERCELIN: OK

rules/prefer-array-index-of.js
Fabrice-TIERCELIN: OK, useless check.

fisker:
We need test each lines in the selector, example 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/43b3324988d56530f76107d2d0e01d73eed6da88/test/no-fn-reference-in-iterator.js#L68-L81

Fabrice-TIERCELIN:
Do you know why `passiveExpressionScope` is empty?

fisker:
What do you want to achieve

Fabrice-TIERCELIN:
I want to ignore such cases:

`['foo', 'bar'].findIndex(x => x === 'foo' + x);`

So I want to searche references of `x` in the expression to be sure the expression won't change.

fisker:
Let me think a little bit.

fisker:
Does `scopeManager.acquire(callback)` / scopeManager.acquire(callback.body) work?

Fabrice-TIERCELIN:
It only works for `scopeManager.acquire(callback)`.

fisker:
Is it the right function scope?

Fabrice-TIERCELIN:
I have no idea. I saw this code in `no-for-loop`.

Fabrice-TIERCELIN:
About parenthesis with this case:

`foo.findIndex((element => element === a))`

...it consists in parsing the code, whatever there is parenthesis or not. So we should adapt the selector to match the code whatever there is parenthesis or not. That's it?

fisker:
Don't need care about it, I mean if you want keep `()` around `right`, you can omit it in this case, similar case

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/898fcb48d06a0740818773ccf2d586af9241e548/rules/prefer-starts-ends-with.js#L80

fisker:
Why [this](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/920/commits/0bdcaabd9894a85c4fbca61f2908aed0ef4db8b9) get passed? Because they `hasSideEffect`?

fisker:
I pushed many changes, please take a look.

Fabrice-TIERCELIN:
OK so parenthesis are automatically handled by the selector.
I don't understand why you have removed all the invalid?...
I suppose you would like to rename `MESSAGE_ID_FINDINDEX` in the tests.

fisker:
> so parenthesis are automatically handled by the selector.

No, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/920/commits/1af4e8b654568074b59e51dcfd041921f809a755 do the job.

> I don't understand why you have removed all the invalid?...

It's much easier to write tests with `test.visualize()`, we can see the report location and message. https://github.com/sindresorhus/eslint-plugin-unicorn/pull/920/commits/e5405073be1d6a3cfdaf739e1c573a650f695dd7 https://github.com/sindresorhus/eslint-plugin-unicorn/pull/920/commits/6d3f2f476dcc527af77fe08b8c680a1f1187c7f5

> I suppose you would like to rename MESSAGE_ID_FINDINDEX in the tests.

Let's remove it, errors already tested by `test.visualize()`, we can simply use `errors: 1(errors count)` in `ts` tests.

Fabrice-TIERCELIN:
OK, I have checked it. It's good for me.

fisker:
You didn't consider this https://github.com/sindresorhus/eslint-plugin-unicorn/pull/920/commits/0bdcaabd9894a85c4fbca61f2908aed0ef4db8b9, but it just get passed, there should be something wrong.

docs/rules/prefer-array-index-of.md
```diff
@@ -0,0 +1,31 @@
+# Prefer `Array#indexOf()` over `Array#findIndex()` when searching the index of an item
+
+All built-ins have [`Array#indexOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) in addition to [`Array#findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex). Prefer `Array#indexOf()` over use of `Array#findIndex()` passing a function that tests equality as parameter
```
sindresorhus, line 3: All built-ins do not have an `.indexOf` method. I think this text needs more work.

docs/rules/prefer-array-index-of.md
```diff
@@ -0,0 +1,31 @@
+# Prefer `Array#indexOf()` over `Array#findIndex()` when searching the index of an item
```
sindresorhus, line 1: This should be the same as the readme description.

rules/prefer-array-index-of.js
```diff
@@ -0,0 +1,118 @@
+'use strict';
+const {hasSideEffect, isParenthesized} = require('eslint-utils');
+const getDocumentationUrl = require('./utils/get-documentation-url');
+const methodSelector = require('./utils/method-selector');
+
+const MESSAGE_ID_FIND_INDEX = 'findIndex';
+const messages = {
+	[MESSAGE_ID_FIND_INDEX]: 'Use `.indexOf()`, rather than `.findIndex()`, when searching the index of an item.'
```
sindresorhus, line 8:
```suggestion
	[MESSAGE_ID_FIND_INDEX]: 'Use `.indexOf()` instead of `.findIndex()` when looking for the index of an item.'
```

sindresorhus (changes requested):

rules/prefer-array-index-of.js
Fabrice-TIERCELIN: I will do it as you want it. By the past, I used to use _instead of_, but I realized that it is ambiguous. When you say "x instead of y", you don't know what is good and what is bad. You can say: _"You suck at using `.findIndex()` instead of `.indexOf()`."_. _Rather than_ is not ambiguous. That's why I pretty like _prefer over_ too.

docs/rules/prefer-array-index-of.md
Fabrice-TIERCELIN: Is it better now?

rules/prefer-array-index-of.js
Fabrice-TIERCELIN: It seems that `prefer-array-index-of.js.snap` needs to be updated but I don't know how.

Fabrice-TIERCELIN:
You say that `foo.findIndex(x => x === "foo" + x)` is not handled as a valid case but I wanna say it's handled [here](https://github.com/Fabrice-TIERCELIN/eslint-plugin-unicorn/blob/748693a7fcc7320ba8d734f89c3b2df2c778bbb6/rules/prefer-array-index-of.js#L85). Isn't it?

And the following valid cases are valid because they use named functions and I think we don't handle named functions.

WDYT?

fisker:
I'm not talking about `x`, I mean function name, `arguments`, `this`(I think that's all possible things related to the callback function.)  I mean they should be valid, but we don't have any code considered about them, the tests should fail.

Fabrice-TIERCELIN:
I think that a named function is parsed as `FunctionDeclaration`, not `FunctionExpression` and that's why it is ignored. OK, let's check the `id` to be sure.

Here, we have two choices:
1. Forbid the named functions,
2. Allow the named functions and add all the checks needed for this.

I think that we should ignore the named functions as it goes too far.

fisker:
> I think that a named function is parsed as FunctionDeclaration, not FunctionExpression

No, it's not.

Check `[].findIndex(function a() {})` on https://astexplorer.net/

fisker:
@Fabrice-TIERCELIN

Pushed some changes, 

1. Possible variables in the `callback` function: the parameter, the function name, `this`, `arguments`(All the possible things I can think), the all checked in other side of `===` expression.
Any use of these variables will be ignored. 
2. I've add suggestions for cases that "hasSideEffect", meaning, `foo.findIndex(x => x === some.call())` will be reported, but since it's not safe to fix, it will not auto-fixed. Mentioned before https://github.com/sindresorhus/eslint-plugin-unicorn/pull/920#issuecomment-741604447, but you 
 didn't implement.

docs/rules/prefer-array-index-of.md
fisker: This is not done.

Fabrice-TIERCELIN:
Yes, it is parsed as `FunctionExpression`. I thought we were using [esquery](https://github.com/estools/esquery) 😞

Thanx for implementing the suggestion. I don't know if I have badly understood your comment or wanted not to understand it 😄

docs/rules/prefer-array-index-of.md
Fabrice-TIERCELIN: I don't understand. I would say [this](https://github.com/Fabrice-TIERCELIN/eslint-plugin-unicorn/blob/prefer-indexof-over-findindex/docs/rules/prefer-array-index-of.md#prefer-arrayindexof-over-arrayfindindex-when-searching-the-index-of-an-item) is the _readme description_. Or you want to say this line should be the same as the description below, as this is a title. Or you wanna say it should be the same as the [message id](https://github.com/Fabrice-TIERCELIN/eslint-plugin-unicorn/blob/prefer-indexof-over-findindex/rules/prefer-array-index-of.js#L10).

I thought you wanted to say that the link to mozilla were not formatted the same way as below.

fisker:
> I don't know if I have badly understood your comment or wanted not to understand it

Sorry for my bad English, I'm not a native speaker.

docs/rules/prefer-array-index-of.md
```diff
@@ -0,0 +1,29 @@
+# Prefer [`Array#indexOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) over [`Array#findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) when looking for the index of an item
```
fisker, line 1: We normally add links in details. and this should be the same as the on in readme.md

docs/rules/prefer-array-index-of.md
```diff
@@ -0,0 +1,29 @@
...
+['foobar'].findIndex(x => {return 'foo' === x;});
+values.findIndex(function (x) {return x === 'foo';});
+['foobar'].findIndex(function (x) {return 'foo' === x;});
+```
+
+## Pass
+
+```js
+[].findIndex(i => i === list[i]);
+[].findIndex(x => x === 'foo' && isValid);
```
fisker, line 22: This one is not valid cases anymore, please fix it.

Also, please put these in separate markdown code blocks. example https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/error-message.md#fail

docs/rules/prefer-array-index-of.md
Fabrice-TIERCELIN: Is it OK now?

docs/rules/prefer-array-index-of.md
fisker: In readme.md it's

> - [prefer-array-index-of](docs/rules/prefer-array-index-of.md) - Prefer `.includes()` over `.indexOf()` when checking for existence or non-existence. *(fixable)*

docs/rules/prefer-array-index-of.md
```diff
@@ -0,0 +1,65 @@
...
+['foobar'].findIndex(function (x) {return 'foo' == x;});
+```
+
+```js
+[].findIndex(x => x !== 'foo');
+```
+
+```js
+values.findIndex(function (x) {return x !== 'foo';});
+```
```
fisker, line 65: Since we don't care about the `callee.object`, let's use `const index = foo.findIndex` in these examples,

please add this to "pass" section


```js
const index = foo.indexOf('foo');
```

fisker (changes requested):

fisker (approved):
LGTM, thank you!

sindresorhus:

docs/rules/prefer-array-index-of.md
```diff
@@ -1,6 +1,6 @@
...
 
-[`Array#findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) is made for more complex criteria. If you are just looking for the index where a given object is present, then the code can be reduced. It concerns any search with a literal, a variable or any expression that hasn't any side effect. However, if the expression you are looking for has side effect, relies on element related to the function (the name of the function, its arguments...), the case is still valid.
+[`Array#findIndex()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) is intended for more complex needs. If you are just looking for the index where the given item is present, then the code can be simplified. This applies to any search with a literal, a variable, or any expression that doesn't have any side effects. However, if the expression you are looking for has side effects, relies on element related to the function (the name of the function, its arguments, etc.), the case is still valid.
```
fisker, line 3: >  However, if the expression you are looking for has side effects, relies on element related to the function (the name of the function, its arguments, etc.), the case is still valid.

This part seems not right, the searching value has "side effects"  not valid.

```js
const index = foo.findIndex(x => x === call());
```

This `call` has "side effects", but it will still reported without "auto-fix".

fisker (changes requested):

docs/rules/prefer-array-index-of.md
Fabrice-TIERCELIN: I have added _**explicit** side effects_.

Fabrice-TIERCELIN: In [`getBinaryExpressionSelector`](https://github.com/Fabrice-TIERCELIN/eslint-plugin-unicorn/blob/845d4727aea41e335ac227dbbdd4160de0a4f595/rules/prefer-array-index-of.js#L17), I want to add a criteria that avoids the existence of `call()`, something like:
```
:not(:has(*[expression.callee.name="call"][expression.type="CallExpression"]))
```

But I don't know how to tell to search only in the binary expression.

fisker (approved):

Issue #912, opened 2020-12-03
prevent-abbreviations: Fix objects must not be overlapped in a report.


AriPerkkio:
Hello, I'm testing stability of well known community ESLint plugins with `eslint-remote-tester`. This ESLint plugin seems to contain a rule which causes linter to crash. ESLint rules should not crash in any condition since this makes all valid linting problems disappear. If this is a false flag please let me know.

This issue was spotted by automated CI run: https://github.com/AriPerkkio/eslint-remote-tester/runs/1487452403

### Crashing rule:

`prevent-abbreviations`

If name of the `VariableExpression` and name of the `TSTypeReference` are identical `unicorn/prevent-abbreviations` crashes.

### Minimal repro:

This seems to be valid typescript:
https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApQPYAdkG8BQyRywAJgFzIgCuAtgEbQDc+AvvvghiAM5hqYsldNmQBePCQrIAjMlYt8EAB5YMUfqQjxqAG34isTIA

```ts
interface Prop {
    id: number;
}

const Prop: Prop = { id: 1 };

export default Prop;
```

This causes ESlint to dispaly an error:
```
$ ./node_modules/.bin/eslint lib

Oops! Something went wrong! :(

ESLint: 7.12.1

AssertionError [ERR_ASSERTION]: Fix objects must not be overlapped in a report.
    at mergeFixes (/<root-path-removed>/node_modules/eslint/lib/linter/report-translator.js:152:9)
    at normalizeFixes (/<root-path-removed>/node_modules/eslint/lib/linter/report-translator.js:182:16)
    at /<root-path-removed>/node_modules/eslint/lib/linter/report-translator.js:347:49
    at Object.report (/<root-path-removed>/node_modules/eslint/lib/linter/linter.js:920:41)
    at checkVariable (/<root-path-removed>/node_modules/eslint-plugin-unicorn/rules/prevent-abbreviations.js:653:11)
    at checkPossiblyWeirdClassVariable (/<root-path-removed>/node_modules/eslint-plugin-unicorn/rules/prevent-abbreviations.js:567:10)
    at /<root-path-removed>/node_modules/eslint-plugin-unicorn/rules/prevent-abbreviations.js:657:39
    at Array.forEach (<anonymous>)
[...]

PR #923, opened 2020-12-07
Add weekly scheduled smoke tests

AriPerkkio:
Adds weekly run smoke tests. See https://github.com/sindresorhus/eslint-plugin-unicorn/issues/912#issuecomment-739621530. Note that this PR does not change the existing integration tests. 

**Why:**

Unit tests cover only cases which are expected. The AST produced by Javascript and Typescript can cause very unexpected conditions. These often lead into null pointers and can crash the linter.

ESLint rules should never crash. A crashing rule causes linter to lose all valid linting problems.

**How:**

Example run: https://github.com/AriPerkkio/eslint-plugin-unicorn/actions/runs/405055179
Contains 1832 repositories. A single run takes ~2 hours on Github Actions CI.

Job is scheduled to be run on Sunday 00:00. There is also an option for running this job manually. 

All comments are welcome!


.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 14
+    - run: yarn install
+    - run: yarn link
+    - run: yarn link eslint-plugin-unicorn
+    - run: yarn smoke
+      env:
+        CI: true
```
fisker, line 21: GitHub Actions set this automaticly

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,21 @@
...
+  workflow_dispatch:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: 14
```
fisker, line 15:
```suggestion
```

Let's remove this, use the latest LTS

fisker (changes requested):

.github/workflows/smoke-test.yml
AriPerkkio: What's the syntax for `LTS`? By removing this the CI will default to some old version.

```
Expected version ">=12.11". Got "10.23.0"
```
https://github.com/AriPerkkio/eslint-plugin-unicorn/runs/1509597287?check_suite_focus=true

package.json
```diff
@@ -96,7 +98,10 @@
 			"plugin:eslint-plugin/all"
 		],
 		"ignores": [
-			"test/integration/{fixtures,unicorn}/**"
+			"test/integration/{fixtures,unicorn}/**",
+			"test/smoke/**",
```
fisker, line 102: Why are we ignoring this file?

.github/workflows/smoke-test.yml
AriPerkkio: This seems to work now with `actions/setup-node@v2-beta`.

fisker: Do we need care about Node.js version?

If not let's use whatever this action decided, so we don't need update it in future.

AriPerkkio: Minimum requirement is `12.11` which introduced some new features to `worker_threads`, required by `eslint-remote-tester`. 

The v14 was originally copied from integration step:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d669707d18609b9b4672c7aa768076d01d671948/.github/workflows/main.yml#L50-L54

fisker: Okay, add it back.

package.json
AriPerkkio: Ignoring removed. Applied required styling to contents of `test/smoke/**`.

test/smoke/repositories.json
```diff
@@ -0,0 +1,1834 @@
+[
+    "AriPerkkio/js-framework-playground",
```
fisker, line 2: I think @sindresorhus prefer to use `Tab`

.github/workflows/smoke-test.yml
```diff
@@ -0,0 +1,19 @@
...
+    - run: yarn install
+    - run: yarn link
+    - run: yarn link eslint-plugin-unicorn
```
fisker, line 18: Will `npm link` work? Other CI jobs uses `npm`

AriPerkkio: I've found that `yarn link` is extremely faster than `npm link`, the difference is significant. In the workflow we can use `npm install`, but for creating the package link I would recommend to use yarn.

fisker: I use `yarn` too, but in this repo seems all commands using `npm`.

AriPerkkio: Now using `npm`. I tested one CI run and it seems to work fine.

fisker:
I wounder if it can open an issue or something to notify maintainers when problem found.

AriPerkkio:
> I wounder if it can open an issue or something to notify maintainers when problem found.

There is an option available in `eslint-remote-tester.config.js` which can be used to implement features like this. https://github.com/AriPerkkio/eslint-remote-tester#eslint-remote-testerconfigjs

```js
    /**
     * Optional callback invoked once scan is complete.
     *
     * @param {{
     *     repository: string,
     *     repositoryOwner: string,
     *     rule: string,
     *     message: string,
     *     path: string,
     *     link: string,
     *     extension: string,
     *     source: string,
     *     error: (string|undefined),
     * }[]} results Results of the scan, if any
     * @returns {Promise<void>|void}
     */
    onComplete: async function onComplete(results) {
        // Extend the process with custom features, e.g. send results to email, create issues to Github...
    },
```

fisker (approved):
Looks good, can you fix conflicts?

AriPerkkio:
Rebased from upstream master. Should be good to go.

Thanks for the review @fisker.

.gitignore
```diff
@@ -4,3 +4,5 @@ yarn.lock
 coverage
 package-lock.json
 /test/integration/fixtures
+.cache-eslint-remote-tester
+eslint-remote-tester-results
```
fisker, line 8:
```suggestion
eslint-remote-tester-results

```

sindresorhus (approved):

fisker:
Thanks!

PR #925, opened 2020-12-07
Revert "`error-message`: Check `AggregateError` too (#921)"

fisker:
Sorry, #921 is a mistake, [`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError/AggregateError) accept `message` as second argument, not first.


sindresorhus:
I missed that too..

PR #933, opened 2020-12-10
Add  `empty-brace-spaces` rule

### LINKED ISSUE #900: Rule proposal: `empty-brace-space`
This rule would enforce spaces between braces in empty anonymous functions and try/catch braces. It is important because it is a syntax style I [need to follow](https://github.com/sindresorhus/internal-ip/pull/36#discussion_r527539515) but isn't automatically caught by xo already.

## Fail

```js
try { } catch { }

() => { }
```

## Pass

```js
try {} catch {}

() => {}
```

fisker:
I renamed `empty-brace-space` to `empty-brace-spaces`, consistent with `no-console-spaces` rule name.

Fixes #900


rules/empty-brace-spaces.js
```diff
@@ -0,0 +1,54 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+const MESSAGE_ID = 'empty-brace-spaces';
+const messages = {
+	[MESSAGE_ID]: 'Do not add spaces between braces.'
+};
+
+const selector = `:matches(${
```
Richienb, line 9: Isn't it easier to just do this?

```js
const selector = ":matches(BlockStatement[body.length=0], ClassBody[body.length=0], ObjectExpression[properties.length=0])"
```

test/empty-brace-spaces.js
```diff
@@ -0,0 +1,89 @@
...
+	'try {} catch(foo){/* */}',
+	'try {} catch(foo){} finally {/* */}',
+	'do {/* */} while (foo)',
+	'while (foo){/* */}',
+	'foo = () => {/* */}',
+	'foo = function (){/* */}',
+	'function foo(){/* */}',
+	'foo = {/* */}',
+	'class Foo {bar() {/* */}}',
+	'foo = class {bar() {/* */}}'
```
Richienb, line 27: We also need to test for `() => {}`.

rules/empty-brace-spaces.js
```diff
@@ -0,0 +1,54 @@
...
+		'ClassBody[body.length=0]',
+		'ObjectExpression[properties.length=0]'
+	].join(', ')
+})`;
+
+const create = context => {
+	const sourceCode = context.getSourceCode();
+	return {
+		[selector](node) {
+			let [start, end] = node.range;
```
sindresorhus, line 21: We should not use the unofficial properties `start` and `end`.

rules/empty-brace-spaces.js
```diff
@@ -0,0 +1,54 @@
...
+					end: sourceCode.getLocFromIndex(end)
+				},
+				messageId: MESSAGE_ID,
+				fix: fixer => fixer.replaceTextRange([start, end], '')
+			});
+		}
+	};
+};
+
+const schema = [];
```
sindresorhus, line 41: I don't see the point of adding empty schema.

sindresorhus (changes requested):

rules/empty-brace-spaces.js
fisker: Yes, same. But I prefer current one, it's more readable add diff friendly, if we want add more.

```diff
const selector = `:matches(${
	[
		'BlockStatement[body.length=0]',
		'ClassBody[body.length=0]',
+		'ImportNamespaceSpecifier[assertions.length=0]',
		'ObjectExpression[properties.length=0]'
	].join(', ')
})`;
```

VS

```diff
- const selector = ":matches(BlockStatement[body.length=0], ClassBody[body.length=0], ObjectExpression[properties.length=0])"
+ const selector = ":matches(BlockStatement[body.length=0], ClassBody[body.length=0], ImportNamespaceSpecifier[assertions.length=0], ObjectExpression[properties.length=0])"

rules/empty-brace-spaces.js
fisker: That's destructuring `.range`  :smile:

sindresorhus: Oh... I missed that.

test/empty-brace-spaces.js
fisker: It's already tested https://github.com/sindresorhus/eslint-plugin-unicorn/pull/933/files#diff-049935872bb22d7098f0c8cbe3b5319cf79af03b3769b1387f3d3fa20cb8a546R22

rules/empty-brace-spaces.js
fisker: It's in template, forgot to remove.

BTW: Do you know `eslint-plugin-eslint-plugin` not working on our rules? And actually this is [enforced by it](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/require-meta-schema.md).

sindresorhus: Not sure why it didn't catch this. It has caught issues in the past though.

PR #935, opened 2020-12-11
Add `prefer-date-now` rule

### LINKED ISSUE #909: Rule proposal: `prefer-date-now`
Prefer `Date.now()` over `new Date().getTime()`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

Because `Date.now()` is shorter and nicer.

## Fail

```js
new Date().getTime();
```

## Pass

```js
Date.now();
```

fisker:
Fixes #909


sindresorhus:
Nice work as always 👍

PR #936, opened 2020-12-14
Add `no-lonely-if` rule

### LINKED ISSUE #819: Rule proposal: Collapsible if-statements

fisker:
Fixes #819


sindresorhus:
Looks good 👍🏻

PR #938, opened 2020-12-18
Rewrite `explicit-length-check` rule

### LINKED ISSUE #794: `explicit-length-check`: Should be able to fix these cases
```js
if (array.length) {}
```

```js
if (!array.length) {}
```

```js
if (!!array.length) {}
```

```js
if (foo || array.length) {}
```

```js
if (foo && array.length) {}
```

fisker:
#### the old logic:
 report problems one case by one case

#### the new logic, 
  1. If node matches wanted style, done.
  2. If node is checking non-zero/zero length, directly fix to wanted style
  in this way we don't need add fix for any new case, only need add check logic.

---
With the new logic, the following changes are easy to be done:

- Added auto-fix for all cases
- Check `==` and `!=` operator too
- Forbid logic like `if (!(foo.length === 0)) {}`, we can know it's checking non-zero, it should be `if (foo.length > 0) {}`
- Added `yoda` style check

Other changes,

- Simplify messages, just tell what should use.

For case `if (foo.length) {}`

```diff
- `length` property should be compared to a value.
+ Use `.length === 0` when checking length is zero.
```

For case `if (!foo.length) {}`

```diff
- `length` property should be compared to a value.
+ Use `.length > 0` when checking length is not zero.
```

Fixes #794


sindresorhus:
Very nice!

PR #939, opened 2020-12-18
Fix link in `new-for-builtins` rule documentation

FloEdelmann:


yakov116:
Shouldn't it be `they return` not returns?

fisker (approved):
Thanks.

PR #941, opened 2020-12-20
`explicit-length-check`: Check more cases

fisker:
This rule should work in any cases return a `boolean`, add missing

- `while (foo.length){}`
- `do {} while (foo.length);`
- `if (; foo.length;){}`

I'll send more if I figure something else, this is actually preparing for https://github.com/sindresorhus/eslint-plugin-unicorn/pull/887

---

Coming changes in my mind:

- `!foo.length` should check in any place, I'll fix in another PR.
- LogicExpression not in `if` should check too.


fisker:
Merging this, so I can continue work on it, let me know if you found something wrong

PR #946, opened 2020-12-21
`prefer-date-now`: Elaborate on rule rationale

jdanil:
Was trying to think of more reasons to justify enforcing this other than style.


sindresorhus (approved):

fisker:
Thanks!

PR #948, opened 2020-12-21
Rename `no-fn-reference-in-iterator` to `no-array-callback-reference`

### LINKED ISSUE #945: Rename `no-fn-reference-in-iterator`
`fn` is [abbreviation listed in `prevent-abbreviations`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/83a64538bc1662c4bac9822263d8ec008c6dfee5/rules/prevent-abbreviations.js#L109-L111), we should avoid use it ourself

Suggestions:

- `no-array-iterator-reference`
- `no-array-callback-reference` [MDN use `callback` term in all their docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#Parameters), and [the spec uses `callbackfn`](https://www.ecma-international.org/ecma-262/11.0/index.html#sec-array.prototype.map)

fisker:
Fixes #945


fisker:
Close in favor of #959

PR #951, opened 2020-12-22
`explicit-length-check`: Check `Boolean(foo.length)`

fisker:


rules/explicit-length-check.js
```diff
@@ -62,9 +72,15 @@ const lengthSelector = [
 
 function getBooleanAncestor(node) {
 	let isNegative = false;
-	while (isLogicNotArgument(node)) {
-		isNegative = !isNegative;
-		node = node.parent;
+	while (true) {
```
fisker, line 75: @sindresorhus How do we write this? `eslint-disable` ?

sindresorhus: Yes, `eslint-disable`. Because in most cases when you want to use `while (true)`, there's usually a safer alternative.

PR #952, opened 2020-12-22
`explicit-length-check`: Check unsafe `LogicalExpression`s

fisker:


fisker:
I think this will be the last PR for this rule. Some changes will help #887

fisker:
Possible improvements in future:

1. Remove `>= 1` option, highly doubt someone prefer this.
2. Make it work for `Set#size`

sindresorhus:
> Possible improvements in future:

Can you move this to an issue so we don't forget?

sindresorhus:

PR #959, opened 2020-12-23
Rename rules

### LINKED ISSUE #945: Rename `no-fn-reference-in-iterator`

fisker:
- `no-fn-reference-in-iterator` -> `no-array-callback-reference`
- `no-array-instanceof` -> `no-instanceof-array`
- `no-reduce` -> `no-array-reduce`
- `prefer-dataset` -> `prefer-dom-node-dataset`
- `prefer-flat-map` -> `prefer-array-flat-map`
- `prefer-replace-all` -> `prefer-string-replace-all`
- `prefer-starts-ends-with` -> `prefer-string-starts-ends-with`
- `prefer-text-content` -> `prefer-dom-node-text-content`
- `prefer-trim-start-end` -> `prefer-string-trim-start-end`
- `prefer-event-key` -> `prefer-keyboard-event-key`
- `prefer-node-append` -> `prefer-dom-node-append`
- `prefer-node-remove` -> `prefer-dom-node-remove`

Fixes #945


fisker:
`dataset` is provide by [`HTMLOrForeignElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset), is it okay to name `prefer-node-dataset`?

sindresorhus:
> dataset is provide by HTMLOrForeignElement, is it okay to name prefer-node-dataset?

Sure, `HTMLElement` inherits from `Node`.

> prefer-event-key to prefer-keyboard-event-key, KeyboardEvent.key?

👍

fisker:
As we are renaming batch of rule names, can you go over all rule names, check if any other rule need rename, so we can do it in one PR.

sindresorhus:
Maybe `prefer-negative-index` => `prefer-negative-index-in-slice` / `prefer-slice-negative-index`, but not sure it's worth renaming just for that.

fisker:
It also work for `.splice`, I think we should keep it.

sindresorhus:
Oh right. I missed the `p`. Then I think your list is good.

fisker:
Thanks for updating the documentation!

futpib:
> - `prefer-dataset` -> `prefer-node-dataset`
> - `prefer-text-content` -> `prefer-node-text-content`

Wouldn't `dom-node` be less ambiguous as `node` can mean Node.js?

- `prefer-dataset` -> `prefer-dom-node-dataset`
- `prefer-text-content` -> `prefer-dom-node-text-content`

fisker:
I use `node` because we already have `prefer-node-append` and `prefer-node-remove`.

I don't have strong opinion, but if we are doing this, these two need rename too.

To be honest, I really don't know what `prefer-text-content` for, when I first saw this rule.

sindresorhus:
I agree, `dom-node` is better.

fisker:
`prefer-query-selector` -> `prefer-dom-query-selector` too?

sindresorhus:
> prefer-query-selector -> prefer-dom-query-selector too?

Nah. There's no confusion there.

PR #961, opened 2020-12-24
Use `@babel/eslint-parser` in tests

fisker:
I failed to make it work on integration test, only switched tests to use new parser, the old `babel-eslint` parser is renamed to `test.babelLegacy()`, it still needed for some test.


sindresorhus:
@fisker Is this something you'd like to get into the next release? I was planning to release what's in master, but I can wait if this is something you want to include.

fisker:
As I can see `@babel/eslint-parser` not using in many repos, and nobody has filed any issue on this, no need rush on this one.

PR #968, opened 2020-12-29
Fix deprecated rules

fisker:
I was stupid, forgot to return rules in `createDeprecatedRules` function.


sindresorhus:
What are the implications of this? Is it a breaking change now?

fisker:
I think this fix v25, if people have deprecated rulesin config or `eslint-disable`d there were errors(can't find rule), this fix that.

PR #970, opened 2020-12-30
Add `prefer-regexp-test` rule

### LINKED ISSUE #732: Rule proposal: `prefer-regexp-test`
Prefer `RegExp.prototype.test` over `String.prototype.match` for truthiness checks of (inline) regexes.

- [Performance](https://stackoverflow.com/a/10940138/271577)
- More semantically clear at a glance how the result is expected to be used. Admittedly, newcomers may think "match" sounds more like it would be a truthy check than "test", but at least for those familiar with the language details, it is more confined and clear as to intent.

## Fail

```js
if (str.match(/hasTheMagic/)) {
}
```

```js
const func = str.match(/hasTheMagic/) ? process(str) : null;
```

## Pass

```js
if ((/hasTheMagic/).test(str)) {
}
```

```js
const func = (/hasTheMagic/).test(str) ? process(str) : null;
```

```js
const matches = str.match(/hasTheMagic/);
```

(The following passes given that changing to `test` could be problematic since it could contain a global regex and such expressions passed to `test` (as with `exec`) behave non-deterministically due to `lastIndex` incrementing)
```js
if (str.match(hasTheMagicRegex)) {
}
```

Apologies this is not a "funded" request with PR planned--just if you felt it fit.

fisker:
Fixes #732


rules/prefer-regexp-test.js
```diff
@@ -0,0 +1,101 @@
...
+	name: 'match',
+	length: 1
+});
+
+const create = context => {
+	const sourceCode = context.getSourceCode();
+
+	return {
+		[regExpExecCallSelector](node) {
+			if (isBooleanNode(node)) {
```
sindresorhus, line 29: I would do an early-return here too.

rules/prefer-regexp-test.js
```diff
@@ -0,0 +1,101 @@
...
+							regexpNode.type === 'Identifier' ||
+							regexpNode.type === 'MemberExpression' ||
+							regexpNode.type === 'CallExpression' ||
+							(regexpNode.type === 'NewExpression' && regexpText.endsWith(')'))
+						)
+					) {
+						regexpText = `(${regexpText})`;
+					}
+
+					// All nodes pass `isBooleanNode` can't have ASI problem
```
sindresorhus, line 82:
```suggestion
					// The nodes that pass `isBooleanNode` cannot have an ASI problem.
```

sindresorhus (changes requested):

sindresorhus:
This is not waiting for https://github.com/sindresorhus/eslint-plugin-unicorn/pull/976, right?

fisker:
> This is not waiting for #976, right?

Current implementation not relay on #976, but better swith to the utility I'm going to add in #976 before merge.

sindresorhus:
> but better swith to that utility I'm going to add in #976 before merge.

👍

sindresorhus:
Tests are failing.

fisker:
Fixed.

PR #976, opened 2020-12-31
`prefer-string-starts-ends-with`: Fix missing parentheses for some cases

fisker:
Noticed this bug during #970

```js
/^a/.test(foo + bar);
```

is fixed to 

```js
foo + bar.startsWith('a');
```

Should be

```js
(foo + bar).startsWith('a');
```

I didn't add to many cases in `needAddParenthesesToMemberExpressionObject`, for now, only some nodes can be `string`. I believe there are other rules need this, we'll add more when we need.


rules/utils/need-add-parentheses-to-member-expression-object.js
```diff
@@ -0,0 +1,49 @@
...
+
+			return false;
+		}
+
+		default:
+			return true;
+	}
+}
+
+module.exports = needAddParenthesesToMemberExpressionObject;
```
sindresorhus, line 49:
```suggestion
module.exports = shouldAddParenthesesToMemberExpressionObject;
```

Same with the filename.

rules/utils/need-add-parentheses-to-member-expression-object.js
```diff
@@ -0,0 +1,49 @@
...
+	switch (node.type) {
+		// This is not a full list. Some other nodes like `FunctionDeclaration` don't need parentheses,
+		// but it's not possible to be in the place we are checking at this point.
+		case 'Identifier':
+		case 'MemberExpression':
+		case 'CallExpression':
+		case 'ChainExpression':
+		case 'TemplateLiteral':
+			return false;
+
```
sindresorhus, line 22:
```suggestion
```

rules/utils/need-add-parentheses-to-member-expression-object.js
```diff
@@ -0,0 +1,49 @@
...
+					return true;
+				}
+			}
+
+			return false;
+		}
+
+		case 'Literal': {
+			/* istanbul ignore next */
+			if (typeof node.value === 'number' && /^\d+$/.test(node.raw)) {
```
sindresorhus, line 37: Won't this fail on `1.1` or `1e+2` and more? If this is intentional, it would be good with a code comment.

fisker: Yes, `1.1.toString();` an  `2e+2.toString();` are  valid.

fisker: Found 
a  better on in eslint core, https://github.com/eslint/eslint/blob/cc4871369645c3409dc56ded7a555af8a9f63d51/lib/rules/no-extra-parens.js#L1011, I didn't consider numberic seperators.

[`no-extra-parens` rule use the same pattern in initial commit.](https://github.com/eslint/eslint/commit/89a163062cdcd1cf073530c58ed8ff411cf81d11#diff-df54692ec82456195837b4983ce1f0b16b3fef0b7f8956fc2d07fcba0e4059dfR187) 😄

PR #977, opened 2021-01-01
Remove `unicorn/import-index` from recommended preset

sindresorhus:
It's not compatible with ESM which requires a filename. In April, Node.js packages can target Node.js 12 which supports ESM, so I think it makes sense to remove this from the recommended preset.

### DIFF
--- a/index.js
+++ b/index.js
@@ -50,7 +50,7 @@ module.exports = {
 				'unicorn/expiring-todo-comments': 'error',
 				'unicorn/explicit-length-check': 'error',
 				'unicorn/filename-case': 'error',
-				'unicorn/import-index': 'error',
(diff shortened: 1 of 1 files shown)


fisker:
How about just set default value of [`ignoreImports`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/import-index.md#ignoreimports) to `true`?
I'm not if it work for `export ... from`, but we can ignore it too if it's not, cjs module will still exists for a while, this rule can be useful.

fisker:
Sorry, on a second thought, there is nothing wrong to use `require("./index")`, so I'm okay with this.

fisker (approved):

PR #986, opened 2021-01-02
Don't consider `??` operator LogicExpression as a `boolean` value node

fisker:


sindresorhus:
Relevant issue: https://github.com/estree/estree/issues/203

PR #998, opened 2021-01-07
Exclude folders like `test` from npm package

bmish:
A lot of extra files are being included in the npm package unnecessarily.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+.github/
+scripts/
+test/


sindresorhus:
We are already doing that: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e2f94fe0c8ac003af2c9a30bfbf3973e17447f25/package.json#L23-L26

bmish:
Thanks. I was mistaken when I checked this.

PR #1008, opened 2021-01-12
`visualizeRuleTester`: Support test `suggestions`

fisker:


test/utils/visualize-rule-tester.js
```diff
@@ -1,5 +1,5 @@
 'use strict';
-const {Linter} = require('eslint');
+const {Linter, SourceCodeFixer} = require('eslint/lib/linter');
```
fisker, line 2: There is no public api to apply suggestion fix, is it okay to use this internal one?

sindresorhus: Yes, since it's just the tests. But you should open an issue on ESLint about exposing this.

PR #1010, opened 2021-01-12
Add auto-fix for `no-unreadable-array-destructuring`

### LINKED ISSUE #746: Auto fix for `no-unreadable-array-destructuring`
[no-unreadable-array-destructuring](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-unreadable-array-destructuring.md)

Maybe not too much cases, but some cases can add autofix or suggestion.

Like

```js
const [,, foo] = bar;
```
To
```js
const foo = bar[2];
```

fisker:
Only if single element(or rest elements) in VariableDeclarator is fixable.

Fixes #746


sindresorhus:
Are there some unsafe cases we could add suggestions for?

fisker:
yes, but not real world case....

```js
const [first,second,,,,,,] = foo;
```

TO

```js
const [first,second,] = foo;
```

:smile:

sindresorhus:
Yeah, no point in supporting unrealistic cases.

PR #1011, opened 2021-01-12
100% code coverage

### LINKED ISSUE #1003: 100% code coverage
Not important, but would be nice.

We can use `/* istanbul ignore next: Some comment */` type comments where it doesn't make sense to test a code path.

fisker:
Fixes #1003, 100% code coverage

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -46,11 +46,13 @@ function checkReferences(scope, parent, scopeManager) {
 		const identifierScope = scopeManager.acquire(identifier);
 
 		// If we have a scope, the earlier checks should have worked so ignore them here
+		/* istanbul ignore next: Hard to test */
 		if (identifierScope) {
 			return false;
 		}
 
 		const identifierParentScope = scopeManager.acquire(identifier.parent);
+		/* istanbul ignore next: Hard to test */
 		if (!identifierParentScope) {
 			return false;
 		}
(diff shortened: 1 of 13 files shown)


rules/prefer-query-selector.js
```diff
@@ -106,9 +106,6 @@ const create = context => {
 		[selector](node) {
 			const method = node.callee.property.name;
 			const preferedSelector = forbiddenIdentifierNames.get(method);
```
sindresorhus, line 108: Unrelated, but I noticed a typo here.
```suggestion
			const preferredSelector = forbiddenIdentifierNames.get(method);
```

rules/custom-error-definition.js
```diff
@@ -161,14 +161,6 @@ const customErrorDefinition = (context, node) => {
 };
 
 const customErrorExport = (context, node) => {
-	if (!node.left.object || node.left.object.name !== 'exports') {
-		return;
-	}
-
-	if (!node.left.property) {
-		return;
-	}
```
fisker, line 170: Moved this part to selector

rules/prefer-query-selector.js
```diff
@@ -88,39 +88,36 @@ const hasValue = node => {
...
 
 const create = context => {
 	return {
 		[selector](node) {
 			const method = node.callee.property.name;
-			const preferedSelector = forbiddenIdentifierNames.get(method);
-			if (!preferedSelector) {
-				return;
-			}
+			const preferredSelector = forbiddenIdentifierNames.get(method);
```
fisker, line 108: Selector already list the methods, so we always have `preferredSelector`

rules/consistent-function-scoping.js
```diff
@@ -119,9 +121,6 @@ function checkNode(node, scopeManager) {
 	}
 
 	let parentNode = node.parent;
-	if (!parentNode) {
```
fisker, line 122: Function parent can't be null

sindresorhus:
Nice! 🎉

PR #1013, opened 2021-01-12
`prefer-default-parameters`: Fix non-iterable `visitorKeys`

### LINKED ISSUE #966: TypeError: source.visitorKeys[node.type] is not iterable
eslint-plugin-unicorn@25.0.0

`prefer-default-parameters.js:36:38`

by code:

```js
const length = quantity || 0;
```

medusalix:
ESLint's `visitorKeys` apparently doesn't have an entry for `ExperimentalRestProperty` nodes, which leads to a `TypeError` when iterating over the keys in `containsCallExpression`. ~I couldn't find a way to trigger this behavior in a test case.~

Fixes #966


fisker:
It's babel ast , use  babel parser to test it.

test/prefer-default-parameters.js
```diff
@@ -1,13 +1,5 @@
-import test from 'ava';
```
sindresorhus, line 1: > 100644 → 100755

Don't change the mode of the file. See the UI above here.

rules/prefer-default-parameters.js
```diff
@@ -33,7 +33,14 @@ const containsCallExpression = (source, node) => {
 		return true;
 	}
```
sindresorhus, line 35: Can you add a code comment for the part that is specifically for the Babel AST.

test/prefer-default-parameters.js
medusalix: Sorry.

fisker:
`eslint-visitor-keys` seems have `ExperimentalRestProperty`, https://github.com/eslint/eslint-visitor-keys/blob/2d7be11e4d13ac702c9fe3c529cadbd75b370146/lib/visitor-keys.json#L90

fisker:
Located problem `babel-eslint` use `require("@babel/types").VISITOR_KEYS`, and `ExperimentalRestProperty` was removed.

fisker:
Do you think we can fallback to `Object.keys(node)` as `visitorKeys`? But should exclude some keys like `parent`.

fisker:
@medusalix Can you check this line, https://github.com/sindresorhus/eslint-plugin-unicorn/blob/b7cbd99dc4376484fff8599d90ce8b253d097708/rules/consistent-destructuring.js#L95 should have the same problem.

medusalix:
> `eslint-visitor-keys` seems have `ExperimentalRestProperty`, [eslint/eslint-visitor-keys@`2d7be11`/lib/visitor-keys.json#L90](https://github.com/eslint/eslint-visitor-keys/blob/2d7be11e4d13ac702c9fe3c529cadbd75b370146/lib/visitor-keys.json#L90)

@fisker That's great. I've added `getKeys` from `eslint-visitor-keys` as a fallback instead of returning `true` by default.

fisker (approved):

fisker:
I changed to use `eslintVisitorKeys.KEYS`, `getKeys` will return some objects not a `Node`, if we use it, we'll need check `node.type` to make sure it's a ESTree node.

But with `eslintVisitorKeys.KEYS`, I think `ExperimentalRestProperty` is the only thing we got from there, maybe we can just hard code visitorKeys for `ExperimentalRestProperty`?

medusalix:
> But with `eslintVisitorKeys.KEYS`, I think `ExperimentalRestProperty` is the only thing we got from there, maybe we can just hard code visitorKeys for `ExperimentalRestProperty`?

There are unfortunately more cases that have to be considered: `ChainExpression`, `ExperimentalSpreadProperty` and `ImportExpression` are also not included in ESLint's visitor keys.
I've added tests for `ChainExpression` and `ImportExpression`. The `ExperimentalSpreadProperty` was apparently already changed to `SpreadElement`.

PR #1014, opened 2021-01-12
`prefer-math-trunc`: Use suggestion instead of fix for `x | 0`

### LINKED ISSUE #911: Incorrect autofix for prefer-math-trunc rule
eslint-plugin-unicorn@23.0.0

Before:

```js
// Very basic hash implementation that mimics Java's hash function, used for generating unique ids for aggregate items
// Found here: https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery
export function generateHash(input) {
  let hash = 0;
  for (let i = 0; i < input.length; i++) {
    const char = input.charCodeAt(i);
    hash = (hash << 5) - hash + char;
    hash |= 0; // Convert to 32bit integer
  }
  return hash;
}
```

Example result (correct):

```
generateHash('abc123;')
-1207861333
```

After autofix by [unicorn/prefer-math-trunc](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v23.0.0/docs/rules/prefer-math-trunc.md):

```js
// Very basic hash implementation that mimics Java's hash function, used for generating unique ids for aggregate items
// Found here: https://stackoverflow.com/questions/7616461/generate-a-hash-from-string-in-javascript-jquery
export function generateHash(input) {
  let hash = 0;
  for (let i = 0; i < input.length; i++) {
    const char = input.charCodeAt(i);
    hash = (hash << 5) - hash + char;
    hash = Math.trunc(hash); // Convert to 32bit integer
  }
  return hash;
}
```

Example result (after autofix, incorrect):

```
generateHash('abc123;')
3087105963
```

Looks related to #862.

noftaly:
Use a suggestion rather than an autofix for the `| 0` case, to prevent incorrectly transforming Int32's casts.

I've only done it for `| 0` but I can do it for all the cases too (`~~x`, `x >> 0`, `x << 0`, `x ^ 0`).

*~~I've not updated the test because I have no idea how to do so with the "visual tester". Does it support suggestions? If no, do I have to move all tests containing `| 0` to the invalid array?
I've already done something like this, taking inspiration from the `prefer-node-remove` rule ; but I don't know if I should really move all of them, if this is the right thing to do, and mostly, how to deal with the "multiple-in-one" tests, such as `const foo = ~~(bar| 0);`~~*
**With the new suggestion support to the visuazizetester, this is no longer applicable**

Fixes #911


fisker:
Suggestion support for `visualizeTester` was  added few hours ago #1008, please rebase and update snapshots.

noftaly:
> Suggestion support for `visualizeTester` was added few hours ago #1008, please rebase and update snapshots.

Yes I'm doing this right now, although I've never done it before, so it might take a little moment 😄

fisker (approved):

sindresorhus:
@noftaly Thanks for the PR 👍🏻

Issue #937, opened 2020-12-15
Rule proposal: `no-continuous-array-push`


fisker:
Forbid calling `push()` multiple times on the same array.

## Fail

```js
foo.push(1);
foo.push(2);
```

## Pass

```js
foo.push(1, 2);
```

```js
foo.push(1);
length = foo.push(2);
foo.push(3);
```

----
I think it should not work for `.unshift()`, since

```js
foo.unshift(1);
foo.unshift(2);
```

Not equal to
```js
foo.unshift(1, 2);
```

sindresorhus:
> Not equal to

It is if you reverse the argument order.

fisker:
> It is if you reverse the argument order.

Right, I was thinking

```js
foo.unshift(...[1,2]);
foo.unshift(...[3,4]);
```

Not equal to

```js
foo.unshift(...[3,4], ...[1,2]);
```

I was wrong

fisker:
But

``js
foo.unshift(a());
foo.unshift(b());
```

``js
foo.unshift(b(), a());
```

Might not right

sindresorhus:
> foo.unshift(b(), a());

Do you mean if `b()` mutates some global state? Yeah, we can skip function calls.

fisker:
Push may also not safe

```js
foo.push(1);
foo.push(a());
```

If a use `foo`

sindresorhus:
We should still report in these "unsafe" cases as they're very unlikely to happen in real code, but we should use a suggestion instead of auto-fix.

sindresorhus:
This is now accepted.

fisker:
Name?

fisker:
How about let's forbid `.pop` too :smile:

Fail

```js
foo.pop();
foo.pop();
```

Pass

```js
foo.length -= 2;
```

futpib:
I prefer one push call per push as you can easily count pushes at a glance if written this way.

fisker:
Which one you prefer

```js
foo.push(...a);
foo.push(...b);
```

vs

```js
foo.push(...a, ...b);
```

sindresorhus:
@futpib 

```js
foo.push(
    1,
    2
);
```

?

silverwind:
Also similar `concat` usage could warn, thought it's likely even rarer in real code:

```js
foo.concat(bar).concat(baz);
```
```js
let x = foo.concat(bar);
x = x.concat(baz);
```

[...]

PR #1015, opened 2021-01-13
Add `no-array-push-push` rule

fisker:
Part of #937, only work for `.push()`


rules/no-array-push-push.js
```diff
@@ -0,0 +1,124 @@
...
+			const index = value.indexOf(node);
+
+			if (index !== -1) {
+				return value[index - 1];
+			}
+		}
+	}
+
+	/* istanbul ignore next */
+	throw new Error('Can not find the first `Array#push()` call.');
```
sindresorhus, line 49:
```suggestion
	throw new Error('Cannot find the first `Array#push()` call.');
```

sindresorhus: Should it really throw? If it is supposed to catch a case that should not happen, I think we should have some text that tells the user to open an issue.

docs/rules/no-array-push-push.md
```diff
@@ -0,0 +1,29 @@
+# Enforce combine multiple `Array#push()` into one call
```
sindresorhus, line 1:
```suggestion
# Enforce combining multiple `Array#push()` into one call
```

docs/rules/no-array-push-push.md
```diff
@@ -0,0 +1,29 @@
...
+
+[`Array#push()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push) accepts multiple arguments, multiple calls should combine into one.
```
sindresorhus, line 3:
```suggestion
[`Array#push()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/push) accepts multiple arguments. Multiple calls should be combined into one.
```

sindresorhus (changes requested):

sindresorhus:
Not sure about the rule name. How about `no-multiple-array-push`?

fisker:
I like `no-array-push-push`, like core rule [`no-plusplus`](https://eslint.org/docs/rules/no-plusplus).

rules/no-array-push-push.js
fisker: This function is a little bit hacky, I'm not sure if it can cover all cases, I can add a issue link.

sindresorhus:
Ok. Sure. I don't feel strongly about it.

fisker:
#1019

PR #1017, opened 2021-01-14
Add `no-array-for-each` rule

### LINKED ISSUE #955: Rule proposal: `no-array-for-each`
`for...of` is more readable.

## Fail

```js
foo.forEach(x => callback(x));
```

## Pass

```js
for (const x of foo) {
	callback(x);
}
```

fisker:
Fixes #955

~Not done yet, still two `TODO`s need fix.~


rules/no-array-for-each.js
```diff
@@ -0,0 +1,284 @@
...
+	// Check `callback.params`
+	const parameters = callback.params;
+	if (
+		!(parameters.length === 1 || parameters.length === 2) ||
+		parameters.some(parameter => parameter.type !== 'Identifier')
+	) {
+		return false;
+	}
+
+	// TODO: check parameters conflicts
```
fisker, line 212: Not really sure how to check this,

Should not fix



```js
foo.forEach(foo => bar());
````

->

```js
for (const foo of foo) bar();
````



```js
a[foo].forEach(foo => bar());
````

->

```js
for (const foo of a[foo]) bar();
````

-----

Should fix




```js
a((foo) => foo).forEach(foo => bar());
````
->
```js
for (const foo of a((foo) => foo)) bar();
````

Comment (fisker) [reply rules/no-array-for-each.js]: Figured a way to detect this, but not very efficient.

Comment (fisker) [reply rules/no-array-for-each.js]: A similar logic maybe possible to auto fix destructuring parameters, but I'm going to try in a seperate PR.

**Comment by fisker** [comment]
There are some typescript parser fix errors, I'll look into it later.

**Comment by sindresorhus** [comment]
This is really well done (as always). And a super useful rule. 🙌🏻

File: test/snapshots/no-array-for-each.js.md [hunk 2]
```diff
@@ -0,0 +1,1176 @@
...
+        |                 ^^^^^^^ Do not use `Array#forEach(…)`.␊
+    `
+
+## Invalid #30
+      1 | a((foo, index) => foo + index).forEach((foo, index) => bar());
+
+> Output
+
+    `␊
+      1 | for (const [foo, index] of a((foo, index) => foo + index).entries())  bar();␊
```
fisker, line 400: I was stupid, forgot to swap parameters.

fregante:
You're killin' it, @fisker! 🥳

isghe:
just add to my rules: `"unicorn/no-array-for-each": 0`

PR #1018, opened 2021-01-14
Add `no-this-assignment` rule

### LINKED ISSUE #997: Rule proposal: `no-this-assignment`
Forbid assign `this` to another variable, I don't see any reason to do this.

## Fail

```js
const that = this;
function foo() {
	return that.foo;
}
foo();
```

```js
const that = this;
class Foo {
	bar() {
		console.log(that.foo)
	}
}
new Foo().bar();
```

## Pass

```js
const foo = () => this.foo;
foo();
```

```js
function foo(someObject) {
	return someObject.foo;
}
foo(this);
```

```js
class Foo {
	constructor(someObject) {
		this.someObject = someObject
	}
	bar() {
		console.log(this.someObject.foo)
	}
}

new Foo(this).bar()
```

I guess we should allow this case

```js
const foo = {someObject: this};
```

Because we don't know how this `foo` is used.

fisker:
Fixes #997


docs/rules/no-this-assignment.md
```diff
@@ -0,0 +1,47 @@
+# Disallow assign `this` to variable
+
```
sindresorhus, line 2: Needs a sentence on "why".

docs/rules/no-this-assignment.md
```diff
@@ -0,0 +1,47 @@
+# Disallow assign `this` to variable
```
sindresorhus, line 1:
```suggestion
# Disallow assigning `this` to a variable
```

docs/rules/no-this-assignment.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Fail
+
+```js
+const foo = this;
+setTimeout(function () {
+	foo.bar();
+}, 1000);
+
```
sindresorhus, line 10:
```suggestion
const foo = this;

setTimeout(function () {
	foo.bar();
}, 1000);
```

sindresorhus (changes requested):

PR #1021, opened 2021-01-15
Update `prefer-optional-catch-binding` fixer

fisker:
Update to use utility added in #1020, previously we remove spaces after catch parameter, now remove spaces before it.


fisker:
Hold this due to possible bug in `eslint-utils`

sindresorhus:
There has not been any activity on https://github.com/mysticatea/eslint-utils/issues/15 for a year. Maybe we should just fork it?

Or maybe @mysticatea would be willing to give us commit access?

fisker:
Let's close this, it's not fixing anything, not important.

PR #1025, opened 2021-01-15
`prefer-ternary`: Add `only-single-line` option

### LINKED ISSUE #889: prefer-ternary: option to ignore when multiple lines
There should be an option for the rule prefer-ternary to ignore when there are multiple lines.
I just updated my config, and it reported me that this was wrong:

```js
if (process.platform === 'win32') {
    await exec(oneLine`
        A long line,
        which is on multiple lines
        etc etc
    `, { cwd: paths.project });
} else {
     await exec(oneLine`
          Same, but
          not on windows,
          etcetc
      `, { cwd: paths.project });
}
```

I think ternary expressions are really great for short lines, like in the examples on the docs, but for long lines like this, it should be skipped.

fisker:
Fixes #889 

//cc @noftaly @szmarczak


noftaly:
Nice 🙂 Thanks !

I think it is great except the documentation is not really clear and could be explained better 😄

fisker:
I'm really bad at docs, much appreciate if  you can help.

docs/rules/prefer-ternary.md
```diff
@@ -120,3 +120,26 @@ if (test) {
...
+
+## Options
+
+Type: `string`\
+Default: `always`
+
+- `always` (default)
+	- Always check `IfStatement`
+- `only-single-line`
+	- Only check if `test`, `consequent`, and `alternate` of `IfStatement` are single line
```
noftaly, line 132:
```suggestion
- `always` (default)
	- Always report when using an `IfStatement` where a ternary expression can be used.
- `only-single-line`
	- Only check if the content of the `IfStatement` is less than one line long.
```
Maybe we loose a bit in precision (I did not tell about `test`, `consequent` and `alternate`) but I think we gain in clarity as it uses broader terms.
I'm not bilingual though so there still might be a better way to formulate it 😄

fisker: The content seems not clear, maybe   ``the content of `if` and `else` ``?

noftaly: You're right, do you want me to edit my suggestion or do will you do a commit ?

fisker: You  can  edit, or  maybe

`` the content of the condition,  `if` and `else` block ``.?

noftaly: what about ``Only check if the content of the `if` and/or `else` block is less than one line long.``?

docs/rules/prefer-ternary.md
```diff
@@ -120,3 +120,26 @@ if (test) {
 	baz = 2;
 }
 ```
+
+## Options
+
+Type: `string`\
+Default: `always`
```
sindresorhus, line 127:
```suggestion
Default: `'always'`
```

PR #1027, opened 2021-01-17
`better-regex`: Disable  broken `combineRepeatingPatterns` transform

### LINKED ISSUE #994: Broken autofix with `better-regex` and multiple optional spaces
The following throws a ` /\s?\s?/ can be optimized to /\s?{2}/       unicorn/better-regex`

```js
const regex = /\s?\s?/;
```

which auto-fixes to:

```js
const regex = /\s?{2}/;
```

This isn't valid though, and throws a parsing error with `Nothing to repeat`. `\s{0,2}` would be a safer auto-fix here I think, allowing for 0-2 of the previous token. The `{0,2}` fix is done as expected with `/a?a?/`, so it seems `\s` is being handled differently.

Further investigation shows that this looks to be an issue with the `regexp-tree` module:
```js
const {optimize} = require('regexp-tree');
const result = optimize(/\s?\s?/);
/*
_ast: {
    type: 'RegExp',
    body: { type: 'Alternative', expressions: [Array] },
    flags: ''
  },
  _source: null,
  _string: '/\\s?{2}/',
  _regexp: null,
  _extra: null
*/
```
Opened an issue on that repo: https://github.com/DmitrySoshnikov/regexp-tree/issues/216

Cherry:
This is a temporary fix for #994 until this can be resolved upstream. I considered adding an option to the rule to enable/disable this transform, but since this is likely to just be a temporary fix, didn't think it was worth it.

I added a quick test case as from the original issue, but let me know if there's anything else you'd like to see here.

Closes #994

### DIFF
--- a/rules/better-regex.js
+++ b/rules/better-regex.js
@@ -12,7 +12,10 @@ const messages = {
 const create = context => {
 	const {sortCharacterClasses} = context.options[0] || {};
 
-	const ignoreList = ['combineRepeatingPatterns'];
+	const ignoreList = [
+		// #994
+		'combineRepeatingPatterns'
+	];
 
 	if (sortCharacterClasses === false) {
 		ignoreList.push('charClassClassrangesMerge');
--- a/rules/better-regex.js
+++ b/rules/better-regex.js
@@ -12,7 +12,7 @@ const messages = {
 const create = context => {
 	const {sortCharacterClasses} = context.options[0] || {};
 
-	const ignoreList = [];
+	const ignoreList = ['combineRepeatingPatterns'];
 
 	if (sortCharacterClasses === false) {
 		ignoreList.push('charClassClassrangesMerge');
--- a/test/better-regex.js
+++ b/test/better-regex.js
@@ -70,7 +70,10 @@ test({
 		'new RegExp()',
 
 		// #472
-		'/[ ;-]/g'
+		'/[ ;-]/g',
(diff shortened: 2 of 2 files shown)


sindresorhus (approved):

fisker:
Disabling this will also disable some useful case like `/aaaa/`, do you think we can traverse the optimized ast to detect such a case, then run `optimize` again with disabling `combineRepeatingPatterns`?

fisker:
For example 

```js
> require('regexp-tree').optimize(/\s?\s?/)._ast.body.expressions[0]
{
  type: 'Repetition',
  expression: {
    type: 'Repetition',
    expression: {
      type: 'Char',
      value: '\\s',
      kind: 'meta',
      symbol: undefined,
      codePoint: NaN
    },
    quantifier: { type: 'Quantifier', kind: '?', greedy: true }
  },
  quantifier: { type: 'Quantifier', kind: 'Range', from: 2, to: 2, greedy: true }
}
```

We can check `Repetition`'s `quantifier` and `expression`, or just simple regex like `\?\[\d+` to test the optimized string.

Cherry:
Closing this as the upstream issue was fixed in https://github.com/DmitrySoshnikov/regexp-tree/issues/216#issuecomment-762119610. I'll open a new PR shortly to simply bump the `regexp-tree` dependency.

PR #1031, opened 2021-01-18
`better-regexp`: bump `regexp-tree` dependency

### LINKED ISSUE #994: Broken autofix with `better-regex` and multiple optional spaces

Cherry:
Fixes #994


fisker (commented):
Can you add test from that issue?

fisker (approved):
Thanks!

Cherry:
Added 2 test cases, one for `/s?/s?` which shouldn't be optimised (as per https://github.com/DmitrySoshnikov/regexp-tree/issues/216#issuecomment-762073297). And then another for `/s?/s?/s?` which should.

test/better-regex.js
```diff
@@ -70,7 +70,10 @@ test({
 		'new RegExp()',
 
 		// #472
-		'/[ ;-]/g'
+		'/[ ;-]/g',
+
+		// 994
```
fisker, line 75:
```suggestion
// #994
```

fisker (approved):

fisker:
Can we  also  test  `/\s{0,2}/` :smile:

Cherry:
Added. `/\s?\s?/` and `/\s{0,2}/` are functionally equivalent and both do not get optimised - the transform comes into effect once you hit 3 optional repeating characters now from what I can tell, and as per [this test](https://github.com/DmitrySoshnikov/regexp-tree/blob/0421d57324defe30ff15a5dbb5af5c08f751a795/src/optimizer/__tests__/optimizer-integration-test.js#L119). I included a link to the comment about this.

fisker:
Good job.

PR #1036, opened 2021-01-19
Refactor: replace `isSameNode` with `hasSameRange`

fisker:
`Node`s has same range didn't mean they are the same node,

example

```
a
```

Both `Program`, `ExpressionStatement`, `Identifier` have same range `[0, 1]`.

Add `hasSameRange` to use it detect shorthands


sindresorhus:
Can you reopen the issue in the main Babel repo? Per https://github.com/babel/babel-eslint/issues/809#issuecomment-668300104

fisker:
I'll confirm this bug still exits first, but it doesn't matter, because `babel-eslint` will have this bug forever...

sindresorhus:
> but it doesn't matter, because babel-eslint will have this bug forever...

Sure, but at some point (sooner than later), we'll remove support for `babel-eslint` and require users to use `@babel/eslint-parser`.

fisker:
I've confirmed, this is fixed in `@babel/eslint-parser`

PR #1037, opened 2021-01-19
`no-new-buffer`: Use suggestion for unknown arguments

fisker:
1. Improve arguments detection with `getStaticValue`
2. Improve fix function
3. Use `suggestion` API when arguments are unknown


rules/no-new-buffer.js
```diff
@@ -1,40 +1,91 @@
...
-	return 'from';
 };
 
+function fix(node, sourceCode, method) {
+	return function * (fixer) {
+		const [start] = node.range;
+		let end = start + 3; // `3` = length of `new`
+		const textAfter = sourceCode.text.slice(end);
+		const [leadingSpaces] = textAfter.match(/^\s*/);
+		end += leadingSpaces.length;
```
sindresorhus, line 51: This is the exact same as https://github.com/sindresorhus/eslint-plugin-unicorn/blob/ca2f54f609ee46bf68503accfddb5b4ca31755d8/rules/new-for-builtins.js#L55-L59 Maybe it could be reused?

fisker: I'll do a clean up later, because I don't want add a function to do `removeNewAndSpaces`, I want add `removeLeadingSpaceAfter(index)`, I think there are other places can use it.

fisker: I searched codebase, seems only these two places, I'll do it now.

PR #1038, opened 2021-01-19
`no-array-for-each`: Support auto fix destructuring parameters

fisker:


fisker:
Close in favor of #1228

PR #1042, opened 2021-01-20
`prefer-spread`: Fix more `.concat` cases

### LINKED ISSUE #1041: `prefer-spread`: Support auto fix for more cases
#1029 only fix `.concat()` with one argument, should be able to fix more cases, ref https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1029#discussion_r559988242

fisker:
1. Support fix empty `.concat()` call.
2. If all fixable, auto fix multiple arguments, remove `.concat(...)`. 
3. If some arguments at beginning are fixable, auto-fix code with them and keep `.concat(...)` with rest arguments.
4. If FIRST argument is not fixable, suggestion will fix FIRST argument and fixable arguments after first arguments, this can reduce one step when apply suggestions.
5. `.concat(...a)` no auto-fix, no suggestions, as we don't know what `a` is.
6. Trailing comma will be inserted if last array literal argument or original array literal has it.

Fixes #1041


test/snapshots/prefer-spread.js.md
```diff
@@ -531,7 +531,7 @@ Generated by [AVA](https://avajs.dev).
 > Output
 
     `␊
-      1 | [1, 2]␊
+      1 | [1, 2,]␊
```
sindresorhus, line 534: Shouldn't this be:
```suggestion
      1 | [1, , 2]␊
```

?

fisker: No,

```text
> [1,].concat(2)
[ 1, 2 ]
> [1, , 2]
[ 1, <1 empty item>, 2 ]
```

It fixed in this way, because `2` inserted after `1`, and the old `[1,]` has trailing comma, we keep it, so `[1, 2,]` is correct.

sindresorhus: Ok. I misunderstood how concat works with holes. I assumed `[1,]` would create `1` + a hole, but you need `[1,,]` to get a hole.

PR #1045, opened 2021-01-21
Add `prefer-destructuring-in-parameters` rule

### LINKED ISSUE #954: Rule proposal: `prefer-destructuring-in-parameters`
I think it's nicer, especially in simple functions.

## Fail

```js
foo.map(element => element.id);
```

```js
Object.entries(foo).filter(pair => pair[1]);
```

## Pass

```js
foo.map(({id}) => id);
```

```js
Object.entries(foo).filter(([, value])=> value);
```

WDYT?

fisker:
Questions: 

1. Should we allow rename, maybe under an option

```diff
- const foo = (bar, x) => bar.x === x
+ const foo = ({x: xOfBar}, x) => xOfBar === x
```

2. Not sure if we should only allow `0`/`1` or just make sure the fixed code not against [`no-unreadable-array-destructuring`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/no-unreadable-array-destructuring.md) rule

```diff
- const foo = bar => bar[1] + bar[2]
+ const foo = ([, second, third] => second + third)
```

3. Should we add a limitation on numbers of properties

See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1045/files#diff-e9f53b43e2668561c178de9cab0e6745a034fee9e0c5321f93c83750b6c73553R740, this case creates a lot of variables.

----

Did not fix codebase, since the diff is big, I want keep this PR small. But you can check changes https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1046/commits/9887f564f74342c7e40c6b3ae36aeff9ee511b86

Fixes #954


fisker:
Another question, currently

```diff
- const foo = (bar, baz) => bar.x === baz.x;
+ const foo = ({x}, baz) => x === baz.x;
```

This is a little a awkward, `baz.x` can't be destructed, because we created a new variable `x`, maybe we want ignore this case "accessing same property on different parameter".

But if we decided allow rename, this won't be a problem, we can fix to 

```diff
- const foo = (bar, baz) => bar.x === baz.x;
+ const foo = ({x: xOfBar}, {x: xOfBaz}) => xOfBar === xOfBaz;
```

sindresorhus:
I think this rule should have an option to only enforce it for inline arrow functions. I find it a bit opinionated to do it for all arrow functions. Constructuring many times leads to less readable code when applied to large functions.

sindresorhus:
> Should we allow rename, maybe under an option

Are you asking whether we should trigger on the case where it needs to be renamed or are you asking whether we should allow renaming at all?

> Not sure if we should only allow 0/1 or just make sure the fixed code not against

Yes, just make sure the fix follows what we enforce in `no-unreadable-array-destructuring`.

> Should we add a limitation on numbers of properties

Yes, but not sure what number makes sense. Should it be configurable?

fregante:
> > Should we add a limitation on numbers of properties
> 
> Yes, but not sure what number makes sense. Should it be configurable?

When I decide whether to destructure or just use the object, I also think of:

- which one is shorter

    this is fine:
    ```diff
    - element => element.id
    + ({id}) => id
    ```

    this not so much:

    ```diff
    - user => user.isHomeDirectoryOpen
    + ({isHomeDirectoryOpen}) => isHomeDirectoryOpen
    ```

- does the property make sense on its own?

    `element.id` makes perfect sense on its own, while `id`… what id?

While it's a nice rule, I'm not sure that it can consistently improve the readability. More like 50/50

sindresorhus:
@fisker I'm not sure about this rule. It seems to make code more unreadable in many cases. Maybe many of the cases could be detected, but not sure all can. (https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1046)

docs/rules/prefer-destructuring-in-parameters.md
```diff
@@ -0,0 +1,38 @@
+# Prefer destructuring in parameters over accessing properties
+
+Makes your code shorter and nicer.
+
+This rule is fixable.
```
bmish, line 5: There are some new requirements for rule docs, be sure to rebase on the latest `main` to pick them up.

rules/prefer-destructuring-in-parameters.js
```diff
@@ -0,0 +1,270 @@
...
+					parameter,
+					properties,
+					type: propertyType
+				})
+			});
+		}
+	};
+};
+
+module.exports = {
```
bmish, line 260: We'll want to add a `type` annotation here (more info in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1606):

```suggestion
/** @type {import('eslint').Rule.RuleModule} */
module.exports = {
```

sindresorhus:
Closing as this has been stale for 2 years.

PR #1046, opened 2021-01-21
[DO NOT MERGE] Auto fix `prefer-destructuring-in-parameters` on codebase

fisker:
Changes a lot, I'd like to fix codebase after #1045 get merged.


rules/consistent-destructuring.js
```diff
@@ -98,8 +98,8 @@ const create = context => {
 			const member = source.getText(node.property);
 
 			// Member might already be destructured
-			const destructuredMember = destructurings.find(property =>
-				property.key.name === member
+			const destructuredMember = destructurings.find(({key}) =>
```
sindresorhus, line 101: The problem with destructuring is that you lose some readability information. It's no longer clear when you read this that `key` is the key of a **property**...

rules/consistent-function-scoping.js
```diff
@@ -90,18 +90,18 @@ const reactHooks = new Set([
...
-	scope.block.parent.callee &&
-	scope.block.parent.callee.type === 'Identifier' &&
-	reactHooks.has(scope.block.parent.callee.name);
-
-const isArrowFunctionWithThis = scope =>
-	scope.type === 'function' &&
-	scope.block &&
-	scope.block.type === 'ArrowFunctionExpression' &&
-	(scope.thisFound || scope.childScopes.some(scope => isArrowFunctionWithThis(scope)));
+const isReactHook = ({block}) =>
```
sindresorhus, line 93: It's no longer possible to infer from the function what type it supports (that it supports a **scope**...

rules/consistent-function-scoping.js
```diff
@@ -90,18 +90,18 @@ const reactHooks = new Set([
...
+	block.parent &&
+	block.parent.callee &&
+	block.parent.callee.type === 'Identifier' &&
+	reactHooks.has(block.parent.callee.name);
+
+const isArrowFunctionWithThis = ({type, block, thisFound, childScopes}) =>
+	type === 'function' &&
+	block &&
+	block.type === 'ArrowFunctionExpression' &&
+	(thisFound || childScopes.some(scope => isArrowFunctionWithThis(scope)));
```
sindresorhus, line 104: I think the previous version was more readable, and not much longer.

sindresorhus (changes requested):

sindresorhus:
Gonna close this for now as it's uncertain what we'll do with this rule.

PR #1048, opened 2021-01-22
Refactor: remove `eslint-ast-utils` package

fisker:
We only used a few lines from it, and the complicated `computeStaticExpression` on can be replaced by `getStaticValue` from `eslint-utils`.


rules/prefer-reflect-apply.js
```diff
@@ -24,9 +24,27 @@ const getReflectApplyCall = (sourceCode, target, receiver, argumentsList) => (
 	`Reflect.apply(${sourceCode.getText(target)}, ${sourceCode.getText(receiver)}, ${sourceCode.getText(argumentsList)})`
 );
 
+const getPropertyName = node => {
```
sindresorhus, line 27: I think this should be in `utils`.

PR #1059, opened 2021-01-24
 Add `prefer-array-index-of` rule

Fabrice-TIERCELIN:
Add `prefer-indexof` rule

I didn't write the snapshot test as I don't know how to get the file `.js.snap`.

Fixes #1058

PR #1062, opened 2021-01-24
Add `prefer-array-includes` rule

### LINKED ISSUE #1058: Rule proposal: `prefer-array-includes`
Some built-ins have [`Array#includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) in addition to [`Array#some()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some). Prefer `Array#includes()` over use of `Array#some()` passing a function that tests equality as parameter.

This rule is partly fixable.

## Fail

```js
[].some(x => x === 'foo');
values.some(x => 'foo' === x);
values.some(x => {return x === 'foo';});
['foobar'].some(x => {return 'foo' === x;});
values.some(function (x) {return x === 'foo';});
['foobar'].some(function (x) {return 'foo' === x;});
```

## Pass

```js
[].some(i => i === list[i]);
[].some(x => x === 'foo' && isValid);
[].some(x => y === 'foo');
[].some(x => y.x === 'foo');
[].some(x => x == 'foo');
['foobar'].some(function (x) {return 'foo' == x;});
[].some(x => x !== 'foo');
values.some(function (x) {return x !== 'foo';});
```

Fabrice-TIERCELIN:
Add `prefer-array-includes` rule

I didn't write the snapshot test as I don't know how to get the file `.js.snap`.

Fixes #1058


yakov116:
> Update/create snapshot, use `npx ava -u` or `npx ava test/xxxx.js -u`.

_Originally posted by @fisker in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1059#issuecomment-766370794_

fisker:
@Fabrice-TIERCELIN Have you seen [this](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1058#issuecomment-766369222)? And we can't copy some many lines from `prefer-array-index-of.js`, please reuse them.

rules/prefer-array-includes.js
```diff
@@ -0,0 +1,162 @@
...
+	const [parameter] = callback.params;
+	if (
+		getVariableIdentifiers(findVariable(scope, parameter))
+			.some(identifier => identifier !== parameter && identifier !== parameterInBinaryExpression)
+	) {
+		return true;
+	}
+
+	// `call` is done
+	if (getReferences(scope).some(reference => reference.identifier.name === 'call' && !reference.resolved)) {
```
fisker, line 68: And I don't understand this, this line is new, why `call` special?

Fabrice-TIERCELIN:
```JavaScript
const isFound = foo.some(x => x === call());
```

This call has "side effects".

Fabrice-TIERCELIN: So it's not true? Or it's not a good solution?

fisker: I don't understand what are you doing, this one should have exactly same logic as `prefer-array-index-of`.

And please read other comments!

Fabrice-TIERCELIN:
> 
> 
> > Update/create snapshot, use `npx ava -u` or `npx ava test/xxxx.js -u`.
> 
> _Originally posted by @fisker in [#1059 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1059#issuecomment-766370794)_

I will do this.

rules/prefer-array-includes.js
Fabrice-TIERCELIN: > > However, if the expression you are looking for has side effects, relies on element related to the function (the name of the function, its arguments, etc.), the case is still valid.
> 
> This part seems not right, the searching value has "side effects" not valid.
> 
> ```js
> const index = foo.findIndex(x => x === call());
> ```
> 
> This `call` has "side effects", but it will still reported without "auto-fix".

I'm just relying on what you have said on `prefer-array-index-of` rule. If I understand well what you have said, `call()` has side effects, so it should be valid. Here, I'm trying to handle this. I know `prefer-array-index-of` rule has been merged without it but it might be taken into account.

Actually if it's not the case or if you want to remove it, I will just remove this part.

> And please read other comments!

I haven't answered because I haven't any answer yet. I should have added acknowledges.

fisker: You misunderstood, that case is "invalid", but can't auto-fix. Of cause it's handled https://github.com/sindresorhus/eslint-plugin-unicorn/blob/df09d31511825ba383efa466e317f735e35cb45e/test/prefer-array-index-of.js#L105-L108

`call` is just an example, you can't just ignore that name, it can be any function.

fisker:
I really don't understand what are you doing, these two points in this comment https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1062#issuecomment-766753503, did  you missed or is there something you don't understand?

Fabrice-TIERCELIN:
I don't have already handled this.

Fabrice-TIERCELIN:
OK, so I have looked at [`prefer-includes`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-includes.md). Here is how I see the situation and then you will tell me what you decide to do.

❌**No common code:** One rule analyzes the inner code of a function and [the other rule analyzes the outter code around a function](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-includes.md) in order that they do not share any code. So they can't be merged. They can only be put together.

❌**No big process:** Although they produce the same result with the same function `includes()`, the both rules are independent and unrelated. Let's imagine a case where a rule converts _A_ to _B_ and another rule converts _B_ to _C_. So let's create a big process that converts _A_ to _C_. Actually, we have a rule that converts _A_ to _C_ and a rule that converts _B_ to _C_. So there is no point to merge.

✔️**Simpler for the user:** It's an _"All in one"_ for the user. So when the user has to select the rules them wants, it's faster and simpler to choose.

**You can decide what you want.** It's only a point of view. If you want me to merge the rules, I'm OK. It's up to you. Tell me what you want.

fisker:
Extract code from `prefer-array-index-of.js` to a function in a shared file, call that function, return `selector`, `listener`, `messages` etc. Use them to create `prefer-array-index-of` rule.

Call that function with different `method` and `replacement`, add `selector`, `listener`, `messages` to `prefer-include`.

Do the same with tests(we don't even need write new tests, we can use the same one).

Fabrice-TIERCELIN:
Have you got a good code example?

fisker:
Do you want me do it? Won't take long.

Fabrice-TIERCELIN:
Either you give an example, or you do it. As you prefer. It will give me an example for the future features :)

fisker:
Check #1097

fisker:
@Fabrice-TIERCELIN Better check commits one by one, you'll understand how it's done.

PR #1069, opened 2021-01-26
`prefer-spread`: Ignore `Buffer.concat()`

### LINKED ISSUE #1068: Buffer.concat incorrectly flagged as prefer-spread
The NodeJS Buffer class has a static `.concat` function that is being flagged as a violation of `prefer-spread - Array#concat()`

Example usage (taken from https://nodejs.org/api/buffer.html#buffer_static_method_buffer_concat_list_totallength)

```
// Create a single `Buffer` from a list of three `Buffer` instances.

const buf1 = Buffer.alloc(10);
const buf2 = Buffer.alloc(14);
const buf3 = Buffer.alloc(18);
const totalLength = buf1.length + buf2.length + buf3.length;

console.log(totalLength);
// Prints: 42

const bufA = Buffer.concat([buf1, buf2, buf3], totalLength);

console.log(bufA);
// Prints: <Buffer 00 00 00 00 ...>
console.log(bufA.length);
// Prints: 42
```

The Buffer class should be ignored by the `prefer-spread` rule.

fisker:
Ignored all possible static methods of a class.

Fixes #1068


test/prefer-spread.js
```diff
@@ -146,7 +146,11 @@ test.snapshot({
 		outdent`
 			const string = 'foo';
 			foo = string.concat("bar");
-		`
+		`,
+		// #1068
+		'const bufA = Buffer.concat([buf1, buf2, buf3], totalLength);',
+		'Foo.concat(1)',
+		'A.concat(1)'
```
sindresorhus, line 153: Can you also add `FooBar.concat(1)`?

fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1069/commits/0a1e32361fb25c4b2ce6f5529d5c3a4f5b8bb01b

PR #1071, opened 2021-01-26
Refactor `prefer-array-index-of` and `no-array-for-each`

fisker:
Simplify `prefer-array-index-of` with an `isFunctionSelfUsedInside` function extracted from `no-array-for-each` rule.


rules/utils/is-function-self-used-inside.js
```diff
@@ -0,0 +1,43 @@
...
+
+const getReferences = (scope, nodeOrName) => {
+	const {references = []} = findVariable(scope, nodeOrName) || {};
+	return references;
+};
+
+/**
+Check if `this`, `arguments`, or `id` of function is used inside it self.
+
+@param {Node} node - The function node.
```
sindresorhus, line 12: These should match the parameter names in JS.

rules/utils/is-function-self-used-inside.js
```diff
@@ -0,0 +1,43 @@
...
+@param {Scope} [scope] - The scope of the function node.
```
sindresorhus, line 13: This is marked as optional, but doesn't look like it is.

rules/utils/is-function-self-used-inside.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
+const {findVariable} = require('eslint-utils');
+
+const getReferences = (scope, nodeOrName) => {
+	const {references = []} = findVariable(scope, nodeOrName) || {};
+	return references;
+};
+
+/**
+Check if `this`, `arguments`, or `id` of function is used inside it self.
```
sindresorhus, line 10:
```suggestion
Check if `this`, `arguments`, or `id` of the function is used inside of itself.
```

sindresorhus: Not clear from reading here what the "id" of the function is.

sindresorhus (changes requested):

rules/utils/is-function-self-used-inside.js
fisker: Right, I'll fix.

rules/utils/is-function-self-used-inside.js
fisker: Changed to ``Check if `this`, `arguments`, or the function name is used inside of itself.``

rules/utils/is-function-self-used-inside.js
fisker: Another thing in this function, I ask to pass two arguments, but acutally `functionNode` is alway `functionScope.block`, I just don't want people get the wrong scope. 

The scope can also get by calling `scopeManager.acquire(node)`, but in rule if it's in function, `context.getScope()` is easier to get the function scope, so I was considering accept `scope/scopeManager` as the second parameter.

Any better idea?

sindresorhus: I think what you already have is the best solution.

Issue #1075, opened 2021-01-27
Fix smoke test failure


fisker:
Job log https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/506558511

fisker:
1.

<details>
<summary>Details</summary>

```log
Repository: jaredLunde/masonic
Rule: utils/should-add-parentheses-to-member-expression-object
Message: Cannot read property 'type' of null
Occurred while linting <text>:70
Path: jaredLunde/masonic/src/interval-tree.test.ts
Link:
https://github.com/jaredLunde/masonic/blob/HEAD/src/interval-tree.test.ts#L70

      [-2, -1, ''],
    ]

    for (let i = 0; i < 10000; ++i) {
      for (const [, , id] of shuffle(list)) {
        tree.remove(id)
      }

      expect(tree.size).toBe(0)


Error:
TypeError: Cannot read property 'type' of null
Occurred while linting <text>:70
    at shouldAddParenthesesToMemberExpressionObject
(/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/rules/utils/shoul
d-add-parentheses-to-member-expression-object.js:21:15)
    at Object.ArrayPattern[elements.length>=3].problem.fix
(/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/rules/no-unreadab
le-array-destructuring.js:50:8)
    at ArrayPattern[elements.length>=3].problem.fix.next (<anonymous>)
    at Function.from (<anonymous>)
    at normalizeFixes (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-uni
corn/node_modules/eslint/lib/linter/report-translator.js:182:33)
    at /home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_module
s/eslint/lib/linter/report-translator.js:347:49
    at Object.report (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unic
orn/node_modules/eslint/lib/linter/linter.js:920:41)
    at ArrayPattern[elements.length>=3] (/home/runner/work/eslint-plugin-unicorn
/eslint-plugin-unicorn/rules/no-unreadable-array-destructuring.js:61:12)
    at /home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_module
s/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
```

</details>

[...]

PR #1077, opened 2021-01-27
Ignore unexpected case in `catch-error-name` and `prefer-set-has`

fisker:
Ref #1075


fisker:
Tried a lot, can't reproduce https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1075#issuecomment-768072967

rules/catch-error-name.js
```diff
@@ -69,6 +69,13 @@ const create = context => {
 		const scope = context.getScope();
 		const variable = findVariable(scope, node);
 
+		// This was reported https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1075#issuecomment-768072967
+		// But can't reproduce, just ignore this case
```
sindresorhus, line 73: Does this still apply? I noticed you added some tests.

fisker: I don't know, the smoke test failed 3days ago, I can't reproduce.  Very strange,  same thing happen few days ago https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1045/files#diff-10f1d5918272c29964d48236585e1df3b35709537364651b22f63e9a7e21a3c7R174

PR #1084, opened 2021-01-29
Add `prefer-https` rule

### LINKED ISSUE #971: Rule proposal: `no-http`
Enforce using HTTPS URLs over HTTP.

## Fail

```js
const foo = 'http://sindresorhus.com';

// Hello: http://sindresorhus.com
```

## Pass

```js
const foo = 'https://sindresorhus.com';

// Hello: https://sindresorhus.com
```

---

Alternative name: `prefer-https`. Which one do people prefer?

I'm aware this could be partly solved with the `string-content` rule, but I don't think most people would think of this and it also wouldn't work in comments.

himynameisdave:
Adds `prefer-https` rule.

Checks string literals, as well as comments.

Resolves #971


himynameisdave:
I set the [`meta.type`](https://eslint.org/docs/developer-guide/working-with-rules#rule-basics) to `'problem'`, not sure if it should be `'suggestion'` instead though.

docs/rules/prefer-https.md
```diff
@@ -0,0 +1,24 @@
+# Enforce using HTTPS URLs over HTTP
+
+Enforce using HTTPS URLs over HTTP.
```
sindresorhus, line 3: This should describe in what places this is enforced.

rules/prefer-https.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+
+const MESSAGE_ID = 'prefer-https';
+const messages = {
+	[MESSAGE_ID]: 'Prefer `{{replacement}}` over `{{value}}`.'
+};
+
+//	Helper to determine if a string is an unsafe URL.
+const containsHttp = nodeValue => /http:\/\/(www\.)?[\w#%+.:=@~-]{2,256}\.[a-z]{2,4}\b([\w#%&+./:=?@~-]*)/.test(nodeValue);
```
sindresorhus, line 10: We don't need to match more after the TLD.

rules/prefer-https.js
```diff
@@ -0,0 +1,55 @@
...
+
+const create = context => {
+	const sourceCode = context.getSourceCode();
+	const reportPreferHttp = node => {
+		const fixed = node.value.replace('http', 'https');
```
sindresorhus, line 15: What if the comment is `// Foo bar http is good: https://asdasd.com`?

I think just replacing `http` substring is too fragile.

rules/prefer-https.js
```diff
@@ -0,0 +1,55 @@
...
+				reportPreferHttp(node);
+			}
+		}
+	};
+};
+
+module.exports = {
+	create,
+	meta: {
+		type: 'problem',
```
sindresorhus, line 48: This is not correct.

test/prefer-https.js
```diff
@@ -0,0 +1,14 @@
...
+	valid: [
+		'const foo = \'https://sindresorhus.com\';',
+		'new URL(\'https://sindresorhus.com\');',
+		'const foo = \'http://localhost\''
+	],
+	invalid: [
+		'const foo = \'http://sindresorhus.com\'',
+		'new URL(\'http://sindresorhus.com\')'
+	]
+});
```
sindresorhus, line 14: You need to add a *lot* more tests.

sindresorhus (changes requested):

fisker:
Is it possible to disable this rule on comments? Just a question, I never disabled problems in comments.

```js
// eslint-disable-next-line
// http://www.example.com
````

himynameisdave:
Hi @sindresorhus.

I've addressed your comments but am still struggling with the tests. They are currently failing in CI due to some weird module error: `'import' and 'export' may appear only with 'sourceType: module'` (see [here](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1084/checks?check_run_id=1943974197)).

In general I found it quite difficult to debug this new rule, specifically debugging the output of the `fix`. Not sure what your prescribed method for debugging this would be, but usually for an eslint rule I would just run `eslint ./some-test-file.js --fix` to understand what's happening, although we don't seem to use eslint in that way for `run-rules-on-codebase`.

Another roadblock that I have run into is that, because of the nature of this rule, I needed to disable the rule on the test (because the test cases were being flagged when running `run-rules-on-codebase`). However when I then went to run `npm test`, it complains that there is no definition for the rule!?

All this to say is that I am happy to continue working on this rule, but I am will need some guidance as this is non-obvious to me. It feels much like I am coding in the dark, unable to easily debug what is happening when my rule is run (as well as how the infrastructure that is used to run the tests in this codebase actually works).

sindresorhus:
> They are currently failing in CI due to some weird module error: 'import' and 'export' may appear only with 'sourceType: module' (see here).

You need to rename the file extension to `.mjs`.

sindresorhus:
@fisker can probably give you some tips on debugging `fix`.

fisker:
You can disable it run on codebase by adding it [here](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c5a43459e881d36e1aa79a7e41a0ec0a05b7e7b7/test/run-rules-on-codebase/lint.mjs#L23), we can enable it when `xo` updated.

fisker:
> It feels much like I am coding in the dark

I'm sorry it makes you feel bad, but `run-rules-on-codebase` is not a required step during implementing, it only need to be done before merge, mentioned in last step here https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/new-rule.md

sindresorhus:
@himynameisdave Still interested in finishing this? :)

himynameisdave:
@sindresorhus yep I'm still interested, although happy to hand this off to someone else if they have more time to work on it. 👍

sindresorhus:
Closing this to let someone else be able to continue it: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/971#issuecomment-933228408

PR #1085, opened 2021-01-29
`prefer-string-slice`: Keep optional chaining in autofix

### LINKED ISSUE #999: Broken autofix in prefer-string-slice with optional chaining
When fixing code with the `prefer-string-slice` rule, code which contains the optional chaining operator can break, as the operator is discarded in the output.

This occurs when using the recommended rules.

**Example input and configuration**
```js
const foo = {};
console.log(foo.bar?.substring(2));
```
```json eslintrc
{
	"plugins": [
		"unicorn"
	],
	"parserOptions": {
		"ecmaVersion": 2020
	},
	"rules": {
		"unicorn/prefer-string-slice": "error"
	}
}
```

Expected output after the auto fix runs is `foo.bar?.slice(2)`, and the program should output `undefined`.

Observed output does not retain optional chaining operator, and after fixing is `foo.bar.slice(2)`. This causes the above program to throw an error on execution instead of outputting `undefined`.

Resources:
* [Optional chaining operator on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
* [OptionalExpression Specification](https://tc39.es/ecma262/#prod-OptionalExpression)
* [Feature support](https://node.green/#ES2020-features-optional-chaining-operator-----) (tldr: all current Node versions)

noftaly:
Keep optional chaining in auto fix

Fixes #999

### DIFF
--- a/test/prefer-string-slice.js
+++ b/test/prefer-string-slice.js
@@ -24,6 +24,7 @@ test({
 		'foo.slice(0)',
 		'foo.slice(1, 2)',
 		'foo?.slice(1, 2)',
+		'foo?.slice?.(1, 2)',
 		'foo?.bar.baz.slice(1, 2)',
 		'foo.slice(-3, -2)'
 	],
@@ -42,6 +43,21 @@ test({
 		{
 			code: 'foo.bar?.substring()',
 			output: 'foo.bar?.slice()',
+			errors: errorsSubstring
+		},
+		{
+			code: 'foo?.[0]?.substring()',
+			output: 'foo?.[0]?.slice()',
+			errors: errorsSubstring
+		},
+		{
+			code: 'foo.bar.substr?.()',
+			output: 'foo.bar.slice?.()',
+			errors: errorsSubstr
+		},
+		{
+			code: 'foo.bar?.substring?.()',
+			output: 'foo.bar?.slice?.()',
 			errors: errorsSubstring
 		},
 		{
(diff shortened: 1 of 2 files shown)


fisker:
What about  `foo?.slice?.(1, 2)`

fisker:
And  `foo?.bar.slice(1, 2)`, I believe it's different `callee` type.

noftaly:
> And  `foo?.bar.slice(1, 2)`, I believe it's different `callee` type.

This is already tested

> What about foo?.slice?.(1, 2)

I'll add this one in an hour or so

sindresorhus (approved):

test/prefer-string-slice.js
```diff
@@ -32,6 +35,46 @@ test({
...
+			errors: errorsSubstring
+		},
+		{
+			code: 'foo?.[0]?.substring()',
+			output: 'foo?.[0]?.slice()',
+			errors: errorsSubstring
+		},
+		{
+			code: 'foo.bar.substr?.()',
+			output: 'foo.bar.slice?.()',
```
fisker, line 55: I'm not sure about this, the old one is testing object has `.substr`, but new one testing `.slice` exists, if someone mean to test it is a `string` and `foo.bar` is an `array`, it's broken.

fisker: Hope no one use this...

fisker (approved):

test/prefer-string-slice.js
noftaly: As you want

PR #1088, opened 2021-01-30
`no-array-for-each`: Ignore `React.Children.forEach`

### LINKED ISSUE #1087: False positive on no-array-for-each
React.Children.forEach is being wrongly flagged.

`no-array-for-each`

```js
React.Children.forEach(children, (child) => {<child />});
```

fisker:
Fixes #1087


sindresorhus:
I think this rule should share code with https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e54dc6668c818f25dba52b0eee4886ba620075f5/rules/no-array-callback-reference.js#L71-L80

test/no-array-for-each.js
```diff
@@ -5,7 +5,9 @@ test.snapshot({
 	valid: [
 		'new foo.forEach(element => bar())',
 		'forEach(element => bar())',
-		'foo.notForEach(element => bar())'
+		'foo.notForEach(element => bar())',
+		// #1087
+		'React.Children.forEach(children, (child) => {});'
```
sindresorhus, line 10: This could also be `Children.forEach(children, (child) => {});` with named import.

fisker:
I wrote a new function and applied to `no-array-callback-reference` too, the logic changed a little bit, old one ignores `AnyThing.Children`, I don't think it mean to do that. New logic is stricter.

rules/no-array-for-each.js
```diff
@@ -315,6 +316,11 @@ function isFixable(callExpression, sourceCode, {scope, functionInfo, allIdentifi
 	return true;
 }
 
+const ignoredObjects = [
+	'React.Children',
+	'Children'
+];
```
sindresorhus, line 322: I was thinking we could have a shared utility method that is like `isNotNativeArrayMethod`. In case there are other false-positives in the future.

rules/utils/is-node-matches.js
```diff
@@ -0,0 +1,29 @@
...
+			return false;
+		}
+
+		node = node.object;
+	}
+
+	return true;
+}
+
+function isNodeMatches(node, descriptions) {
```
sindresorhus, line 25: This needs a short doc comment on what `descriptions` is.

test/no-array-for-each.js
```diff
@@ -5,7 +5,10 @@ test.snapshot({
 	valid: [
 		'new foo.forEach(element => bar())',
 		'forEach(element => bar())',
-		'foo.notForEach(element => bar())'
+		'foo.notForEach(element => bar())',
+		// #1087
+		'React.Children.forEach(children, (child) => {});',
+		'React.Children.forEach(children, (child) => {});'
```
sindresorhus, line 11:
```suggestion
		'Children.forEach(children, (child) => {});'
```

rules/utils/is-node-matches.js
sindresorhus: `keyPath` is a commonly used term for `foo.bar` descriptions of a method chain.

rules/no-array-callback-reference.js
```diff
@@ -70,7 +71,7 @@ const iteratorMethods = [
 
 const ignoredCallee = [
 	'Promise',
-	'React.children',
+	'React.Children',
```
fisker, line 74: I think this mean to be `React.Children`.

rules/no-array-for-each.js
fisker: That's a  good idea, like the [`not-dom-node.js`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/not-dom-node.js), but I was hoping get some help from  [`eslint-utils`](https://github.com/mysticatea/eslint-utils/issues/17), anyway I'll keep this in mind, maybe try in another PR.

fisker:
My 11 month daughter is coding for me... I'm fix after she sleep.

rules/no-array-for-each.js
sindresorhus: That package isn't very actively maintained. I suggest we just do the utilities here for now and we can move them there later on if it becomes maintained again. Alternatively, you could fork `eslint-utils` for now.

PR #1091, opened 2021-01-30
`no-array-callback-reference`: Update examples

fisker:
Make them more real.

### DIFF
--- a/docs/rules/no-array-callback-reference.md
+++ b/docs/rules/no-array-callback-reference.md
@@ -83,7 +83,7 @@ array.forEach(callback, thisArgument);
 ## Pass
 
 ```js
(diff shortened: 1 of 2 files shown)


docs/rules/no-array-callback-reference.md
```diff
@@ -33,129 +33,130 @@ const unicorn = require('unicorn');
...
+array.forEach(callback, thisArgument);
 ```
 
 ## Pass
 
 ```js
-const fn = x => x + 1;
-
-[1, 2, 3].map(x => fn(x));
+const foo = array.map(element = > callback(element));
```
noftaly, line 86:
```suggestion
const foo = array.map(element => callback(element));
```

Issue #1099, opened 2021-02-13
Missing dependency on `eslint-visitor-keys`


SirWindfield:
I migrated to yarn v2 plug'n'play and I get errors about this package missing a dependency declaration.

The package misses a dependency declaration on `eslint-visitor-keys`, which is used inside of the rules.

I can open up a PR if you want. Since the dependency is used directly, it has to be inside of `dependencies`, right?

fisker:
> Since the dependency is used directly, it has to be inside of dependencies, right?

Yes.

> I can open up a PR if you want.

PR welcome.

PR #1100, opened 2021-02-14
Fix missing `eslint-visitor-keys` dependency

Mesteery:
The dependency `eslint-visitor-keys`
was not declared in the `package.json`. This produced an error.

Closes: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1099


SirWindfield:
Dang, you were faster! I looked the issue up again so I could reference it in the commit message :) Thank you very much!

fisker (approved):

PR #1117, opened 2021-02-19
`prefer-spread`: Add `Array#slice()` check

### LINKED ISSUE #737: `prefer-spread`: Add prefer spread over `.slice()/.concat()`
Currently, [`prefer-spread`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-spread.md) only check `Array.from()`, how about add prefer spread over `.slice()/.concat()`?

# Fail

```js
const foo = array.slice();
```

```js
const foo = array.concat(item);
```

```js
const foo = array.concat(anotherArray);
```

# Pass

```js
const foo = [...array];
```

```js
const foo = array.slice(1);
```

The `concat()` case, we can't know `item` is array or not. 

Maybe with suggestions? 
Suggestion 1: `[...array, itemOrArray]`
Suggestion 2: `[...array, ...itemOrArray]`

Another problem is `TypedArray#slice()` and `String#concat()`. `String#slice` should not a problem. Nobody use `string.slice()`(no arguments).

---
Any other method people use to clone an array?

fisker:
Let's not add `options` for this to see if people complains?

Fixes #737


sindresorhus:
> Let's not add options for this to see if people complains?

👍

PR #1119, opened 2021-02-19
Add `class-reference-in-static-methods` rule

### LINKED ISSUE #1050: Disallow super() and .this in static methods
The `this` keyword on static methods refers the class instance (the constructor). This can be confusing.

## Fail

```js
class A {
    static foo() {
        doSomething()
    }

    static bar() {
        this.foo()   //ERROR: Unexpected 'this'.
    }
}

class B extends A {
    static foo() {
        super.foo()  //ERROR: Unexpected 'super'.
    }
}
```

## Pass

```js
class A {
    static foo() {
        doSomething()
    }

    static bar() {
        A.foo()
    }
}

class B extends A {
    static foo() {
        A.foo()
    }
}
```

fisker:
Fixes #1050


sindresorhus:
>  so I think the rule should let you enforce either this or "class name" in static methods. Same with super.  - https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1050

fisker:
Sorry, I missed that.

What the option name should be? `preferThis`?

Also, note that replace `superClassName` with `super`  has lots of  limitation.

```js
class A extends B {
	static foo() {
		console.log(B);  // Can't use `super`
		const x = {B};  // Can't use `super`
		B = C; // Can't use `super`
		return class extends B {} // Can't use `super`
	}
}
```

sindresorhus:
> What the option name should be? preferThis?

`preferThisAndSuper`

fisker:
Due to limitation mentioned above, maybe we should config `this` and `super` separately.

sindresorhus:
Sure

sindresorhus:
So two options then:

- `preferThis`
- `preferSuper`

?

rules/class-reference-in-static-methods.js
```diff
@@ -0,0 +1,77 @@
...
+	if (!node || node.type !== 'Identifier') {
+		return false;
+	}
+
+	const variable = findVariable(scope, node);
+
+	return !variable || getVariableIdentifiers(variable).includes(node);
+}
+
+function create(context) {
```
bmish, line 23:
```suggestion
/** @param {import('eslint').Rule.RuleContext} context */
function create(context) {
```

sindresorhus:
Closing for lack of activity.

PR #1120, opened 2021-02-20
Add `no-static-only-class` rule

### LINKED ISSUE #973: Rule proposal: No static-only class
A class with only static members could just be an object instead.

## Fail

```js
class X {
	static foo = false;
	static bar() {};
}
```

## Pass

```js
const X = {
	foo: false,
	bar() {}
};
```

```js
class X {
	static foo = false;
	static bar() {};

	unicorn() {}
}
```

fisker:
Fixes #973


rules/no-static-only-class.js
```diff
@@ -0,0 +1,238 @@
+'use strict';
+const {isSemicolonToken} = require('eslint-utils');
+const getDocumentationUrl = require('./utils/get-documentation-url');
+const getClassHeadLocation = require('./utils/get-class-head-location');
+const removeSpacesAfter = require('./utils/remove-spaces-after');
+
+const MESSAGE_ID = 'no-static-only-class';
+const messages = {
+	[MESSAGE_ID]: 'Use object instead of class with only static members.'
```
sindresorhus, line 9:
```suggestion
	[MESSAGE_ID]: 'Use an object instead of a class with only static members.'
```

rules/no-static-only-class.js
```diff
@@ -0,0 +1,238 @@
...
+	/* istanbul ignore next */
+	if (
+		type !== expected.type ||
+		value !== expected.value
+	) {
+		const issueLink = 'https://github.com/sindresorhus/eslint-plugin-unicorn/issues/new?title=%60no-static-only-class%60%3A%20Unexpected%20token%20error';
+		throw new Error(
+			`Expect token '${JSON.stringify(expected)}', got '${JSON.stringify({type, value})}', Please open an issue at ${issueLink}.`
+		);
+	}
```
sindresorhus, line 30: Can you make this an utility function?

Could be used in these places too:

```
rules/prefer-optional-catch-binding.js
36-					const tokenAfter = context.getTokenAfter(node);
37-
38-					/* istanbul ignore next */
39-					if (!isOpeningParenToken(tokenBefore) || !isClosingParenToken(tokenAfter)) {
40:						throw new Error('Unexpected token.');
41-					}
42-
43-					yield fixer.remove(tokenBefore);
44-					yield fixer.remove(node);

rules/no-array-for-each.js
92-		const returnToken = sourceCode.getFirstToken(returnStatement);
93-
94-		/* istanbul ignore next: `ReturnStatement` firstToken should be `return` */
95-		if (returnToken.value !== 'return') {
96:			throw new Error(`Unexpected token ${returnToken.value}.`);
97-		}
98-
99-		if (!returnStatement.argument) {
100-			yield fixer.replaceText(returnToken, 'continue');
```

rules/no-static-only-class.js
```diff
@@ -0,0 +1,238 @@
...
+const assertToken = (token, expected) => {
+	const {type, value} = token;
+	/* istanbul ignore next */
+	if (
+		type !== expected.type ||
+		value !== expected.value
+	) {
+		const issueLink = 'https://github.com/sindresorhus/eslint-plugin-unicorn/issues/new?title=%60no-static-only-class%60%3A%20Unexpected%20token%20error';
+		throw new Error(
+			`Expect token '${JSON.stringify(expected)}', got '${JSON.stringify({type, value})}', Please open an issue at ${issueLink}.`
```
sindresorhus, line 28:
```suggestion
			`Expected token '${JSON.stringify(expected)}', got '${JSON.stringify({type, value})}', Please open an issue at ${issueLink}.`
```

rules/no-static-only-class.js
```diff
@@ -0,0 +1,238 @@
...
+				const openingBraceToken = sourceCode.getFirstToken(body);
+				yield fixer.remove(openingBraceToken);
+			} else {
+				yield fixer.replaceText(classToken, '');
+
+				// Avoid breaking case like
+				// ```js
+				// return class
+				// {};
+				// ```
```
sindresorhus, line 192: This would be better as a `/* */` type comment. Same with the one above.

rules/no-static-only-class.js
```diff
@@ -0,0 +1,238 @@
...
+	// 	{
+	// 		static a() {}
+	// 	}
+	// }
+	// ```
+	const classToken = sourceCode.getFirstToken(node);
+	/* istanbul ignore next */
+	assertToken(classToken, {type: 'Keyword', value: 'class'});
+
+	let needMoveBodyOpeningBraceToken = false;
```
sindresorhus, line 168:
```suggestion
	let shouldMoveBodyOpeningBraceToken = false;
```

?

sindresorhus (changes requested):

rules/no-static-only-class.js
fisker: Done.

sindresorhus:
Looks great 👍🏻

PR #1127, opened 2021-02-23
Support `PropertyDefinition`

fisker:
https://github.com/babel/babel/pull/12370


fisker:
babel [switched type back to `ClassProperty` by default](https://github.com/babel/babel/pull/12867), but it's in estree spec, it will still happen someday.

JLHwung:
Yeah, I plan to enable it by default in Babel 8. Unfortunately we could not enable it on 7.13 since many eslint plugins still depend on Babel AST.

fisker:
I checked the babel PR  above, seems it use `eslint` as  babel plugin in test, how do I enable `classFeatures` in `@babel/eslint-parser`? So I can test with the new tree.

JLHwung:
You can pin `@babel/eslint-parser` to `7.13.0`. In the future it can be toggled on by `BABEL_8_BREAKING=true` env.

fisker:
@JLHwung Do you think we can add an option for it  in `@babel/eslint-parser`? Maybe through `parseOptions`.

I think this can help other eslint plugins easily test/prepare for both these two node types.

sindresorhus:
@JLHwung ^

JLHwung:
@sindresorhus Thanks for the ping!

The upstream Babel issue has been resolved. See https://github.com/babel/babel/pull/12891 for an example config enabling the `PropertyDefinition` in `@babel/eslint-parser`.

sindresorhus:
@fisker Is this ready to be merged?

PR #1131, opened 2021-02-23
Add `no-complex-iteratee-expression` rule

### LINKED ISSUE #1044: Rule proposal: `no-complex-iteratee-expression`
> How about forcing the looped expression to be one of:
> * a variable (`x`)
> * a property access (`x.a.b`)
> * a single function call with no arguments (`x()`, `x.a.b()`)
> * one of built-in function calls commonly used for iteration (`Object.keys(x)`/`Object.values(x)`/`Object.entries(x)`)
> 
> Otherwise force the user to define a variable before the loop (may be possible to auto-fix, the iteratee variable name could be the pluralized name of the element variable).

Originally from https://github.com/sindresorhus/eslint-plugin-unicorn/issues/846#issuecomment-703317406

## Fail

```js
for (const x of xs.filter(x => shouldKeep(x))) {
	console.log(x)
}
```

## Pass

```js
const filteredXs = xs.filter(x => shouldKeep(x))

for (const x of filteredXs) {
	console.log(x)
}
```

```js
for (const x of Object.values(xs)) {
	console.log(x)
}
```

noftaly:
Add the `no-complex-iteratee-expression` rule which forbids complex expressions in `for-of` headers.

To fix before merge:
- Fix using already used identifiers for fix
- Use visualise tests

Thanks!

Fixes #1044


fisker:
> JS node selection or esquery selectors?

They are the same, you can  choose one, personally I prefer selector only because some time the check will be like `node && Array.isArray(node.elements) && node.elements.length > 2 &&  node.element[0] && node.elemens[0].type === "Identifier"`, while `[elements.length>2][elements.0.type="Identifier"]` do the  same thing but much nicer.

fisker:
> but I don't know how to do it with selectors

You are  not doing right, you should write a selector that matches the forbidden node and wrap them with `:not()` then, like `:not(OBJECT_KEYS, OBJECT_VALUES, SIMPLE_CALL)`

fisker:
> to insert a line before the node

User `fixer.insertTextBefore(node, 'your code')`

noftaly:
> > but I don't know how to do it with selectors
> 
> You are not doing right, you should write a selector that matches the forbidden node and wrap them with `:not()` then, like `:not(OBJECT_KEYS, OBJECT_VALUES, SIMPLE_CALL)`

@fisker So we do a blacklist then? The issue suggested doing a whitelist (only allowing variables, property access, functions with no arguments).



> > to insert a line before the node
> 
> User `fixer.insertTextBefore(node, 'your code')`

Thanks, will do

fisker:
`whitelist` list all the nodes allowed, but use `:not()` to ignore them, you get all other nodes.

noftaly:
Hey, sorry but I need help, I don't know why but I can't figure out why the fixer doesn't work (it's just the very beginning, but still, the replaceText doesn't do anything even though the fixer is called...)
I also did not find any examples on how to add semicolon at the end, only at the beginnning.
Thanks

rules/no-complex-iteratee-expression.js
```diff
@@ -0,0 +1,83 @@
...
+
+				if (!resolveVariableName(iterateeName, context.getScope())) {
+					const iteratee = source.getText(node.right);
+
+					context.report({
+						node: node.right,
+						messageId: MESSAGE_ID,
+						fix: fixer => {
+							// FIXME: handle ";"
+							fixer.insertTextBefore(node, `const ${iterateeName} = ${iteratee};`);
```
fisker, line 54: `fixer.replaceText` returns an fix object, you need return the result or array of these objects returned by `fix.xxx` call.


You can also use generator function to yield these objects. See fix in other rules, just search `yield`.

noftaly: Thanks, it worked

noftaly:
What do we do about arrays?
In the codebase, there are these cases:
```
for (const {test, openingParenthesisToken, closingParenthesisToken} of [outer, inner]) // ...
for (const rule of [
	'plugin/rule',
	'@scope/plugin/rule-name',
	'@scope/rule-name',
	'@scopewithoutplugin'
]) // ...
```
The first one doesn't bother me at all, and I even think it looks cleaner that way. However, I agree the second one should be forbidden.

Also, I think that it could be great if we allowed users to add selectors to the whitelist, but maybe selectors are not user-friendly enough to do that.

fisker:
Why are you using `test()` instead of `test.snapshot()`? Isn't it much easier to write tests? (Yes, harder to review, but worth.)

noftaly:
To be fair, I don't know 😄 I'll update them

fisker:
Note: the fix is not that easy

```js
for (const values of array.filter(element => shouldKeep(element))) {}
```

```js
function a() {
	const values = [];
	for (const value of array.filter(element => shouldKeep(element))) {}
}
```

```js
const values = [];
function a() {
	for (const value of array.filter(element => shouldKeep(element))) {}
}
```

Notice that `values` conflicts.

fisker:
> To be fair, I don't know 😄 I'll update them

You can keep it as you already done. I just want remind there is a easy way.

noftaly:
> Note: the fix is not that easy
> 
> ```js
> for (const values of array.filter(element => shouldKeep(element))) {}
> ```
> 
> ```js
> function a() {
> 	const values = [];
> 	for (const value of array.filter(element => shouldKeep(element))) {}
> }
> ```
> 
> ```js
> const values = [];
> function a() {
> 	for (const value of array.filter(element => shouldKeep(element))) {}
> }
> ```
> 
> Notice that `values` conflicts.

Yes I tried to find a function to know if the variable name is already used, but I find so many different functions in `rules/utils` that I don't know which one to chose, and I pick one, but it might not be the best. Which one should I take?

And I'll add test for that.

noftaly:
> > To be fair, I don't know 😄 I'll update them
> 
> You can keep it as you already done. I just want remind there is a easy way.

No problem, it does less test-duplication anyway :\)

noftaly:
Ohh I just realised, I should also add parenthesis if the for-loop is inlined (in such a case:
```js
if (something)
  for (const value of invalid(arg)) {}
```

fisker:
> I should also add parenthesis if the for-loop is inlined

Yes, [this](https://eslint.org/docs/rules/no-case-declarations) may also want be considered

> Which one should I take?

`avoidCapture`

noftaly:
> Yes, this may also want be considered

Nice catch

> avoidCapture

Thanks

noftaly:
> Why are you using test() instead of test.snapshot()? Isn't it much easier to write tests? (Yes, harder to review, but worth.)

I think I found out why, it doesn't check if the auto-fix is correct, does it?
I just wrote tests for captures and braces, but it shows me that they pass (but they clearly don't)

fisker:
> I think I found out why, it doesn't check if the auto-fix is correct, does it?

Of course it does, when you wrote something, the test result is in the `snapshots/*.mjs.md` file, you need check what's in there, and make sure it's expected.

noftaly:
> Of course it does, when you wrote something, the test result is in the `snapshots/*.mjs.md` file, you need check what's in there, and make sure it's expected.

Oh so I have to do it manually 😕

fisker:
Never mind, just use the way you prefer to test.

test/no-complex-iteratee-expression.mjs
```diff
@@ -0,0 +1,240 @@
...
+		{
+			code: 'for (const value of array.prop.prop.filter(shouldKeep)) {}',
+			errors: [{messageId: MESSAGE_ID}],
+			output: outdent`
+				const values = array.prop.prop.filter(shouldKeep);
+				for (const value of values) {}
+			`
+		},
+		{
+			code: 'for (const value of array.prop.prop.method(...someVariable)) {}',
```
fisker, line 32: `snapshot` test don't need write output, that's why it's easier to test..

noftaly: Ohh ok, I'll look into that after I'm done with the braces 😄

noftaly:
Really sorry I think I might have taken a rule a bit too hard for the little experience I have 😄
I can't find a way to insert a token on the same line as the if
```js
if (something) // <-- right here, after the ")"
  blablah();
```
If I do `insertTextAfter(node.parent, '{')` it inserts it after the whole if block. Same with insertTextBefore, it inserts it before the if.
The best I could come with is `insertTextAfter(node.parent.consequent, '{')`, but it is not on the same line...
I also tried a replace text but I can't only replace the `if ()` line...

fisker:
`insertTextAfter(node， '{')` is the right way.

fisker:
```js
function shouldInsertBraces(node) {
	return node.parent.type === 'IfStatement' && (node.parent.consequent === node || node.parent.alternate === node) // And other cases
}

* fix() {
	if (shouldInsertBraces(node)) {
		yield fixer.insertTextBefore(node, '{');
		yield fixer.insertTextAfter(node, '}');
	}
}
```

For the cases should insert braces, you can search `adjustClause` on this page https://github.com/prettier/prettier/blob/5682d2f4405d843ed77d8c13b8048d7b7dfd8b49/src/language-js/printer-estree.js, that's the cases need this.

noftaly:
>```js
>if (shouldInsertBraces(node)) {
>		yield fixer.insertTextBefore(node, '{');
>		yield fixer.insertTextAfter(node, '}');
>}
>```

If I do that, then the indentation is incorrect... Is it ok and will be handled by another rule, or should I fix it? It outputs this, for example:
```js
// From
if (something) for (const value of values) {}
// To
if (something) {const values = invalid(var_);
 for (const value of values) {}} // Notice only 1 space, not the correct indentation
// Instead of 
if (something) {
	const values = invalid(var_);
	for (const value of values) {}
}
```
That was the main concern I had.
I just commit this really naive implementation that adds braces but not newlines nor indents.

rules/no-complex-iteratee-expression.js
```diff
@@ -0,0 +1,115 @@
...
+	const before = lines[line - 1].slice(0, column);
+
+	return before.match(/\s*$/)[0];
+};
+
+const scopeStatements = new Set(['DoWhileStatement', 'ForInStatement', 'ForOfStatement', 'ForStatement', 'WhileStatement', 'WithStatement']);
+const shouldAddBraces = node => {
+	if (node.parent.type === 'IfStatement') {
+		const nodeType = node.parent.consequent === node ? 'consequent' : 'alternate';
+		return node.parent[nodeType].type !== 'BlockStatement';
```
fisker, line 48: Do not check type, check `parent.consequent === node`, not the same thing.

fisker:
> What do we do about arrays?
> In the codebase, there are these cases:
> 
> ```js
> for (const {test, openingParenthesisToken, closingParenthesisToken} of [outer, inner]) // ...
> for (const rule of [
> 	'plugin/rule',
> 	'@scope/plugin/rule-name',
> 	'@scope/rule-name',
> 	'@scopewithoutplugin'
> ]) // ...
> ```
> 
> The first one doesn't bother me at all, and I even think it looks cleaner that way. However, I agree the second one should be forbidden.
> 
> Also, I think that it could be great if we allowed users to add selectors to the whitelist, but maybe selectors are not user-friendly enough to do that.

I think we should ignore array literals.

rules/no-complex-iteratee-expression.js
```diff
@@ -0,0 +1,115 @@
...
+			'[right.type="Identifier"]',
+			'[right.type="MemberExpression"]',
+			'[right.type="CallExpression"][right.arguments.length=0]',
+			[
+				'[right.type="CallExpression"]',
+				'[right.callee.type="MemberExpression"]',
+				'[right.callee.object.type="Identifier"]',
+				'[right.callee.object.name="Object"]',
+				'[right.callee.property.type="Identifier"]',
+				'[right.callee.property.name=/(keys|values|entries)/]'
```
fisker, line 25:
```suggestion
				'[right.callee.property.name=/^(?:keys|values|entries)$/]'
```

rules/no-complex-iteratee-expression.js
noftaly: I don't see why, because if the for loop is in the alternate, we want to format the alternate. So if you check whether it is the consequent, we will miss the case where it is in an `else`.
Am I right?

fisker: Your code works fine in current esree spec, only because the `IfStatement` can't be other child of the listing parents.

But imagine if your checking node is an `Identifier`, your test will all wrong, because if can be other child, safest way is to check `parent.xxx === node` to make sure it is correct child of parent. Even if the spec add other child.

rules/no-complex-iteratee-expression.js
fisker: Need tests for it, previous version will ignore


```js
for (const foo of Object.xx_keys());
```

noftaly: Added in [`25b0af`](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1131/commits/25b0af0ab70d5aab923f37ad77f4ff9ed7b5f544)

rules/no-complex-iteratee-expression.js
noftaly: Oh ok. So I should just return `node.parent.consequent === node`? Should I do the same for `SwitchCase`, or only for the `If`?

fisker: All of them.  Just use

```js
return node.parent.type === "XXX" && node.parent.yyy ==== node
```

Means node is parent with type `XXX`'s property `yyy`

noftaly:
> > Which one should I take?
>
> `avoidCapture`

Are you sure? The integration test seems to be failing with multiple "Identifier XXX has already been declared"...

---

I also noticed a problem with TS, I'll fix it:
```ts
for (const query of (reachableQueries: Array<ReactFlightReachableQuery>)) {

// Gets fixed to
const queries = reachableQueries: Array<ReactFlightReachableQuery>;
for (const query of (queries)) {

// But should be
const queries: Array<ReactFlightReachableQuery> = reachableQueries;
for (const query of queries) {
```

---

And just a little bump for the braces placement, I can't figure out a way to place them properly... It seems they are correctly placed in the integration test, I assume because ESLint goes back to it and another rule fix them. Should we let it like this (with "weird" placement in the tests), or should we fix it? If we have to fix it, how?

fisker:
Same thing 


```js
for (const query of (0, queries)) {}
```

```js
for (const query of (0, queries = x)) {}
```

noftaly:
> Same thing
> 
> ```js
> for (const query of (0, queries)) {}
> ```
> 
> ```js
> for (const query of (0, queries = x)) {}
> ```

This outputs an Error in Node.js
```js
> for (const a of (0, 1)) { console.log(a) }
// Uncaught TypeError: (0 , 1) is not iterable
```
Do we really need to support it?

fisker:
```
> queries = [1];for (const query of (0, queries)) {console.log(query)}
1
```

fisker:
Do you know what `(0, [1])` mean? Your fix fails on `(    0, [1])` and `((0, [1]))`. You can use this function https://github.com/sindresorhus/eslint-plugin-unicorn/blob/2397d940572c74a48e00be11c307e821f9f1664b/rules/prefer-spread.js#L72

noftaly:
> I also noticed a problem with TS, I'll fix it:
> ```js
> for (const query of (reachableQueries: Array<ReactFlightReachableQuery>)) {
>
> // Gets fixed to
> const queries = reachableQueries: Array<ReactFlightReachableQuery>;
> for (const query of (queries)) {
> 
> // But should be
> const queries: Array<ReactFlightReachableQuery> = reachableQueries;
> for (const query of queries) {
> ```


Update: we can't fix this, because even though `typescript` parses `for (const a of (b: SomeType))`, `@typescript-eslint/parser` (tested with `4.15.2`) does not: `"=> expected"`.
Should I add it as a TODO ?

---

> Do you know what (0, [1]) mean? Your fix fails on ( 0, [1]) and ((0, [1])). You can use this function

Nope, not really, never seen it in real code... I know it will iterate over the last expression in the sequence, but nothing much. Will look into this function

fisker:
> I know it will iterate over the last expression in the sequence

This is correct, no one really use `SequenceExpression`, but it's most special node that always need parens, it's best choice when testing parens logic, but logic should not relay on the type.

In this rule anything can be wrapped in parens

```js
for (const foo of ((a.map(x=>x))));
```

noftaly:
> > I know it will iterate over the last expression in the sequence
> 
> This is correct, no one really use `SequenceExpression`, but it's most special node that always need parens, it's best choice when testing parens logic, but logic should not relay on the type.
> 
> In this rule anything can be wrapped in parens
> 
> ```js
> for (const foo of ((a.map(x=>x))));
> ```

Good to know, thanks 😄
Because of the `getParenthesisedRange` function, the output gets modified as you can see in tests:
`( 0, [1])` -> `(0, [1])`
`((0, [1]))` -> `(0, [1])`
Is it ok ?

noftaly:
Hey, just a little bump with a summary of things currently blocking me:
- Braces placement, I can't figure out a way to place them properly... It currently add them around the block, but not on the same line as the if (or whatever outer statement there was). See tests. It's seems to be good on integrations tests, maybe another ESLint rule fixes this?
- Should we simplify sequence expressions? Because of the `getParenthesisedRange` function, the output gets modified as you can see in tests: `( 0, [1])` becomes `(0, [1])` and `((0, [1]))` becomes `(0, [1])`
- Currently breaking integration: The fix will use the same names if there are 2 for loops to fix with the same iteratee name. I tried to use the `scopeToNamesGeneratedByFixer` that I found in other files, but no matter how hard I try, it didn't work (iirc, the variable was not really added to `scopeToNamesGeneratedByFixer`)

fisker:
> Braces placement

Just keep it as it is.

> Should we simplify sequence expressions? Because of the getParenthesisedRange function, the output gets modified as you can see in tests: ( 0, [1]) becomes (0, [1]) and ((0, [1])) becomes (0, [1])

This not important too, we only need make sure it won't fix to invalid code. And it's changed because you use `const iteratee = sourceCode.getText(node.right);` instead of parenthesized range.

> Currently breaking integration

Use `scopeToNamesGeneratedByFixer` is the right way, you also need `extendFixRange`

fisker:
I think checking `Object.keys()` is not the right way, 


```js
for (const x of Object.keys(Object.fromEntries(a(foo).map(b).filter(c))) {}
```

It's still on `Object.keys()`, but too "complicated".

sindresorhus:
Bump :)

noftaly:
> Bump :)

Hey, really sorry but I don't know how to fix/implement the last missing things (listed in the main PR post) ; and I won't have the time to look further into it anytime soon.

I hand it over to whoever wants to finish it.

Sorry for not giving updates and letting this PR down :\(

sindresorhus:
No worries. I'm going to close this to let someone else be able to continue it: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1044#issuecomment-933227578

PR #1133, opened 2021-02-25
`better-regex`: Report parsing errors

### LINKED ISSUE #1081: `better-regex` enhancement: report invalid regexes (when exceptions thrown by the parser)
Invalid regexes (which can nevertheless pass through some parsers) are silently passed over despite the opportunity to report them.

`better-regex`

This should be reported even if the parser does not catch it. It is reported by the default `espree`, but not by `babel-eslint` or `@babel/eslint-parser`, `@typescript-eslint/parser`, etc.

```js
const c = /(/;
```

(I've also filed https://github.com/DmitrySoshnikov/regexp-tree/issues/220 such that the following should also fail.)

```js
const c = /{/u;
```

There appear to be [other cases](https://github.com/DmitrySoshnikov/regexp-tree/search?q=throw) which throw as well.

brettz9:
Fixes #1081


fisker:
It was reported https://github.com/sindresorhus/eslint-plugin-unicorn/pull/454/files

brettz9:
IIUC, you're indeed avoiding errors with the u flag already, but as per my PR here, there can be other errors that get through (at least if a non-default parser is being used).

fisker:
I'm trying to remember the reason behind that change, I think I was trying to avoid cases that `regexp-tree` can understand and fix to something the parser can't understand, it will cause crash.

fisker:
Anyway, because I can't come up with some case causes probelom, I think maybe we can give this a shot, if something came up, we can ignore it again.

brettz9:
Sure... It looks like at least the reported cases that caused crashes were both u flag specific.

With #446 , a single curly bracket `{` is problematic unescaped with the u flag but not otherwise.

With #448, though bad sorting doesn't seem to occur now (the hyphen appears to go to the end where it is safe (or no sorting if the `sortCharacterClasses` option is set)), I imagine previously (perhaps before https://github.com/DmitrySoshnikov/regexp-tree/issues/163 ), the hyphen in the middle of `[a-zA-Z-\d]` was not sorted or not sorted properly.

If for example, it had been sorted to `/[\d-a-zA-Z]/`, the hyphen will still be safely treated as a literal without the Unicode flag but with the Unicode flag a problem is reported because `\d` contains more than one character (see https://stackoverflow.com/questions/54205197/why-is-w-a-valid-regex-but-w-u-invalid ). (And FWIW, the original regex isn't problematic because `Z` is not the beginning of a range, but is part of `A-Z`; if the original had instead been `/[Z-\d]/u`, this would have been problematic.)

brettz9:
Also, and more importantly, I didn't change your fix to catch `optimize` errors. I'm merely reporting them now.

sindresorhus:
@fisker Good to merge?

PR #1140, opened 2021-03-17
`prefer-number-properties`: Refactor

fisker:
I was working on `prefer-module` rule, I need similar track global variable `__dirname` and `__filename`, just like this rule tracking `NaN` and `Infinity`.

Feel it's easier to use `ReferenceTracker`, [the ESLint core rules use this a lot](https://github.com/eslint/eslint/blob/6f4abe5d5ade2711cc4c21bc8485af952763c2d3/lib/rules/require-unicode-regexp.js#L51).


fisker:
It seems this is a bad idea, it relay on `globals`

PR #1141, opened 2021-03-17
Add `prefer-module` rule

### LINKED ISSUE #1067: Rule proposal: `prefer-module`
Should fail on all legacy commonjs stuff.

## Fail

```js
const foo = require('./foo');
```

```js
const foo = path.join(__dirname, './foo.js');
```

```js
module.exports = {};
```

```js
exports.foo = {};
```

## Pass

```js
export default foo;
```

```js
const foo = path.join(path.dirname(url.fileURLToPath(import.meta.url)), './foo.js');
```

fisker:
Fixes #1067


fisker:
I'm almost done here, questions:

1. Should we ignore `.cjs` file? Or let user to choose?
2. I used `suggestion` instead of auto-fix, I think all these fix breaks code. (expect removing "use strict")
3. Anything missing?

fisker:
@sindresorhus ^

rules/prefer-module.js
```diff
@@ -0,0 +1,318 @@
...
+		}
+	};
+}
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Prefer ES modules over CommonJS.',
```
sindresorhus, line 312:
```suggestion
			description: 'Prefer JavaScript modules over CommonJS.',
```

Should we go with this naming?

sindresorhus:
Should we also enforce using `"type": "module"` in package.json?

sindresorhus:
1. I would not. We could publish and see if someone opens an issue requesting it before we add an option.
2. Makes sense. Even `use strict` is not technically safe to remove as it can change semantics if not in an ESM context, as it then makes the code non-strict.
3. I think you covered it all.

docs/rules/prefer-module.md
```diff
@@ -0,0 +1,103 @@
+# Prefer ES modules over CommonJS
```
sindresorhus, line 1:
```suggestion
# Prefer JavaScript modules (ESM) over CommonJS
```

I think this is a better naming.

fisker:
> Even use strict is not technically safe to remove as

You are right, but it most likely not going to cause problem, let's keep it this way for now.

fisker:
> Should we also enforce using "type": "module" in package.json?

I don't think so, `package.json` self is not JavaScript file, it will not be linted normally, and if we report this problem in a `.js` file seems weird.

fisker:
Maybe we can create `eslint-plugin-package-json` :smile:

sindresorhus:
> Maybe we can create eslint-plugin-package-json 😄

I think it's too much work to maintain a plugin just for that. How about an opt-in `package-json` rule in this repo?

sindresorhus:
I pushed some docs improvements.

sindresorhus:
Did you implement https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1067#issuecomment-809626154?

fisker:
Yes, all access of global `require`, `module`, and `exports` are reported.

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1141/files#diff-13455d67606e537047e216aceb85af848608cfa605ddcbb5e1c4f0c55c15f076R261

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1141/files#diff-13455d67606e537047e216aceb85af848608cfa605ddcbb5e1c4f0c55c15f076R1276

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1141/files#diff-13455d67606e537047e216aceb85af848608cfa605ddcbb5e1c4f0c55c15f076R1296

sindresorhus:
Awesome!

yoyo837:
Excuse me, How can i disable this rule in nodejs code file?

fisker:
```js
{
	overrides: [
		{
			files: [],
			rules: {
				'unicorn/prefer-module': 'off'
			}
		}
	]
}
```

https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns

yoyo837:
Thanks.

PR #1149, opened 2021-03-24
Improve `no-array-for-each` rule docs

bmish:


docs/rules/no-array-for-each.md
```diff
@@ -1,9 +1,15 @@
...
-A [`for…of` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is more readable than [`Array#forEach(…)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).
+Benefits of [`for…of` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) over [`Array#forEach(…)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) can include:
+
+* Better readability
+* No additional scope, can mean easier debugging
+* Ability to early exit with `break` / `return`
 
 This rule is partly fixable.
 
+Note that this rule may erroneously catch/fix loops with arrays that are not iterable and thus cannot be converted to for-of loops.
```
papb, line 11: What would be an example of this?

bmish: I had experienced issues where this rule (and also the [unicorn/prefer-spread](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-spread.md) rule) erroneously converted `someObj.forEach(...)` to a for-of loop in my Ember app leading to the [x is not iterable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/is_not_iterable) error, but it turns out `someObj` was just an object that implemented array functions like forEach (see [Ember.ArrayProxy#forEach](https://api.emberjs.com/ember/release/classes/ArrayProxy/methods?anchor=forEach)) and was not actually iterable.

I removed this warning for now.

docs/rules/no-array-for-each.md
```diff
@@ -1,6 +1,10 @@
 # Prefer `for…of` over `Array#forEach(…)`
 
-A [`for…of` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) is more readable than [`Array#forEach(…)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).
+Benefits of [`for…of` statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) over [`Array#forEach(…)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) can include:
+
+- Better readability
+- No additional scope (easier debugging)
```
sindresorhus, line 6: This is not entirely correct. It does have a scope, just a block scope instead of a function scope.

bmish: Good point, removed this line.

PR #1158, opened 2021-04-07
`prefer-set-has`: Use suggestions if node has type annotation

### LINKED ISSUE #1148: prefer-set-has doesn't autofix types
When "autofixing" an array to a set, the type definition should be fixed too (or the fix should be skipped).

prefer-set-has

```ts
const a: Array<'foo' | 'bar'> = ['foo', 'bar']

for (let i = 0; i < 3; i++) {
  if (a.includes(someString)) {
    console.log(123)
  }
}
```

This is corrected to:

```ts
const a: Array<'foo' | 'bar'> = new Set(['foo', 'bar'])

for (let i = 0; i < 3; i++) {
  if (a.has(someString)) {
    console.log(123)
  }
}
```

Which is confusing and wrong.

fisker:
Fixes #1148

### DIFF
--- a/rules/prefer-set-has.js
+++ b/rules/prefer-set-has.js
@@ -4,9 +4,11 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const getVariableIdentifiers = require('./utils/get-variable-identifiers');
 const methodSelector = require('./utils/method-selector');
 
-const MESSAGE_ID = 'preferSetHas';
+const MESSAGE_ID_ERROR = 'error';
+const MESSAGE_ID_SUGGESTION = 'suggestion';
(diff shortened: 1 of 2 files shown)


sindresorhus:
Wouldn't it be better to just fix the type?

fisker:
I don't have the knowledge to fix it.  As I understand array could be  `foo[]` or `Array<foo>`?

fisker:
I'll let someone else to fix it then.

sindresorhus:
Ok. I think this PR is a good workaround for now though. Could you reopen? I can just make an issue afterwards for actually fixing the type and someone can pick it up if they want.

PR #1167, opened 2021-04-08
`no-array-for-each`: Do not fix if callback parameter has type annotation

### LINKED ISSUE #1107: [no-array-for-each] incorrect autofix for ts files
```js
// const elements: NodeListOf<Element> = ...

elements.forEach((el: HTMLDivElement) => {})

// autofixed
for (const el: HTMLDivElement of elements) {
}
```

Errors:

```log
error TS2483: The left-hand side of a 'for...of' statement cannot use a type annotation.

error TS2488: Type 'NodeListOf<Element>' must have a '[Symbol.iterator]()' method that returns an iterator.
```

fisker:
Fixes #1107


sindresorhus:
Integration tests seems to be failing.

fisker:
Not related, I've fixed in the main branch.

PR #1168, opened 2021-04-08
`no-array-push-push`: Add `ignore` option

### LINKED ISSUE #1057: unicorn/no-array-push-push - breaks stream.push
`unicorn/no-array-push-push` combines `.push`, but it can't do this if the `.push` is for something that is not an array, like `stream`.

## Before
```ts
const stream = new Readable();
stream.push('attachment content');
stream.push(null); // end of file
```

## After
```ts
const stream = new Readable();
stream.push('attachment content', null); // end of file
//                                ^^^^ null is not a valid encoding
```

fisker:
Added `ignore` option, and ignore `stream`, `this`, `this.stream` by defualt.

Ideally, I want ignore `**.stream`, but seems little hard to implement, maybe I'll try another time.

Fixes #1057


sindresorhus:
> nd ignore stream, this, this.stream by default.

I think that should be documented.

PR #1173, opened 2021-04-08
`prefer-regexp-test`: Do not fix if regexp has `g` flag

### LINKED ISSUE #1114: [prefer-regexp-test] should not apply when the regex has the global modifier (g)
The rule [prefer-regexp-test] suggests that it is equivalent to replace a `match` method call on a string to a `test` method call on a regex when the `match` call is only ever used to see if the regex matches.

This is only correct when the regex in question does not have the global modifier (g). I added an example at the bottom where you can see a different result for the second call to `test`. [The example is also available as a fiddle here.](https://jsfiddle.net/0dbkq5v6/)

You can also see another example in the [MDN documentation on `test`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test).

Removing the global modifier in the example below fixes the issue.

```js
const regex = /weird/g;

const testString = 'well this is weird!';

console.log(`First match execution: ${!!testString.match(regex)}`);
// Prints: First match execution: true

console.log(`Second match execution: ${!!testString.match(regex)}`);
// Prints: Second match execution: true

console.log(`First test execution: ${regex.test(testString)}`);
// Prints: First test execution: true

console.log(`Second test execution: ${regex.test(testString)}`);
// Prints: Second test execution: false
```

fisker:
But only if it's a reference, so `if (foo.match(/a/g));` and  `if (foo.match(new RegExp('a', 'g')));`still fixable.

Fixes #1114

---

Update: This PR was checking `g` and `y` flag, after second thought the `y` flag seems not effect `.test()` result.


yangmingshan:
Should update docs from `This rule is fixable.` to `This rule is partly fixable.` ?

fisker:
My idea is remove the term `partly`, almost every rule can't fix all cases.

yangmingshan:
It would be easier to maintain [eslint-config-silent](https://github.com/yangmingshan/eslint-config-silent), if I can learn from the docs whether a rule can be fixed in all cases.

fisker:
Core rule never mention what's not fixable, and many rules not really fixable for edge cases https://github.com/eslint/eslint/search?q=fixer+return+null

PR #1179, opened 2021-04-13
Use `@babel/eslint-parser` in integration tests

fisker:
Removed `react` and `relay` project since they use `flow`, and I don't think we should enable `flow` plugin for `babel` parser.


fisker:
@futpib You are right we need both `configFile: false` and `babelrc: false` to prevent babel load config.

PR #1180, opened 2021-04-14
`no-keyword-prefix`: Rename `blacklist` option to `disallowedPrefixes`

### LINKED ISSUE #942: Improve option naming
I don't think we should use `whitelist`/`blacklist` type naming. They have negative connotations for some.

whitelist => allowList
blacklist => denyList

This applies to:

- `blacklist` option in `no-keyword-prefix`
- `whitelist` option in `prevent-abbreviations`
- `extendDefaultWhitelist` option in `prevent-abbreviations`

---

I'm not really interested in discussing the merits of this. It's the right thing to do.

https://thenextweb.com/dd/2020/07/13/linux-kernel-will-no-longer-use-terms-blacklist-and-slave/

fisker:
Fixes #942

### DIFF
--- a/docs/rules/no-keyword-prefix.md
+++ b/docs/rules/no-keyword-prefix.md
@@ -22,15 +22,15 @@ const fooNew = 'foo';
 
 ## Options
 
-### `blacklist`
+### `keywords`
 
-If you want a custom list of forbidden prefixes you can set them with `blacklist`:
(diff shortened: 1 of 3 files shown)


sindresorhus:
I don't think `keywords` is the correct naming here. It does describe the default values, but not the intent of the option. The prefix could be for anything, not just keywords. How about `disallowedPrefixes`?

PR #1181, opened 2021-04-14
Add `prefer-switch` rule

### LINKED ISSUE #1051: Rule proposal: `prefer-switch`
A switch statement is easier to read than multiple if statements with simple equality comparisons.

## Fail

```js
if (foo === bar) {
    /* ... */
}
else if (foo === baz) {
    /* ... */
}
else if (foo === arglebargl) {
    /* ... */
}
else if (foo === glopglyph) {
    /* ... */
}
```

## Pass

```js
if (foo === bar) {
    /* ... */
}
else if (foo === baz) {
    /* ... */
}
```

```js
switch (foo) {
    case bar:
        /* ... */
        break;
    case baz:
        /* ... */
        break;
    case arglebargl:
        /* ... */
        break;
    case glopglyph:
        /* ... */
        break;
}
```
This could have a ```minimumCases``` option, defaulting to 3.

fisker:
Fixes #1051

/cc @Torhal


docs/rules/prefer-switch.md
```diff
@@ -0,0 +1,120 @@
...
+	Insert `default` case and add `// Do nothing` comment.
+- `'no-default-case'`
+	Don't insert default case or comment.
+
+#### `breakStatementInDefaultCase`
+
+Type: `boolean`\
+Default: `false`
+
+Insert `break;` into the default case.
```
sindresorhus, line 120: What's the point of this option? It doesn't seem related to this rule. I don't think this rule should enforce any style on the switch.

docs/rules/prefer-switch.md
```diff
@@ -0,0 +1,120 @@
...
+// Fails
+```
+
+#### `emptyDefaultCase`
+
+Type: `string`\
+Default: `'no-default-comment'`
+
+- `'no-default-comment'` (default)
+	Insert `// No default` comment after last case.
```
sindresorhus, line 109: I don't think it's clear to the user what the comments do. It should refer to https://eslint.org/docs/rules/default-case

sindresorhus (changes requested):

docs/rules/prefer-switch.md
fisker: Removed.

sindresorhus:
Looks great 👍

PR #1184, opened 2021-04-15
Update `create-rule` script

fisker:
- Remove `partly`
- Try open VSCode after rule created

### DIFF
--- a/scripts/create-rule.js
+++ b/scripts/create-rule.js
@@ -5,6 +5,7 @@ const fs = require('fs');
 const path = require('path');
 const enquirer = require('enquirer');
 const {template} = require('lodash');
+const execa = require('execa');
 
 const ROOT = path.join(__dirname, '..');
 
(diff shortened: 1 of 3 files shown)


scripts/create-rule.js
```diff
@@ -199,6 +192,16 @@ function updateReadme(data) {
 	});
 	updateIndex(id);
 	updateReadme(data);
+
+	try {
+		await execa('code', [
```
sindresorhus, line 197: I don't think you should assume VSCode. How about using https://github.com/sindresorhus/open-editor ?

fisker: It seems open-editor  can't start my editor on windows, I'll look into it.

fisker: Found: vscode can't use `--goto` when opening multiple files.

fisker: https://github.com/sindresorhus/open-editor/issues/15

PR #1197, opened 2021-04-20
Increase width of readme rules table name column

bmish:
When there are excessively-long strings in the rule description column, the description column takes up too much width and causes the strings in the rule name column to split into multiple lines which hurts readability and aesthetics. We can truncate the descriptions when they are too long to solve this.

Notes:
* Based on my experimentation, there's no other way to force a wider rule name column width.
* `100` was the largest description length I could allow without resulting in the rule name column width shrinking.

Alternative to #1201.


scripts/generate-rules-table.mjs
```diff
@@ -50,3 +53,17 @@ writeFileSync(
...
+		return string;
+	}
+
+	const trimmed = string.slice(0, maxLength).trim();
+
+	// If we break the string in the middle of a code sample, be sure to close it.
+	const countBackticks = (trimmed.match(/`/g) || []).length;
+	const hasUnclosedBackticks = countBackticks % 2 === 1;
+
+	return trimmed + (hasUnclosedBackticks ? '`' : '') + '...';
```
sindresorhus, line 68:
```suggestion
	return trimmed + (hasUnclosedBackticks ? '`' : '') + '…';
```

scripts/generate-rules-table.mjs
```diff
@@ -50,3 +53,17 @@ writeFileSync(
 		`
 	)
 );
+
+function trimString(string, maxLength) {
```
sindresorhus, line 57:
```suggestion
function truncateString(string, maxLength) {
```

PR #1201, opened 2021-04-21
Increase width of readme rules table name column (with HTML table)

bmish:
Alternative to #1197.


bmish:
It looks like this doesn't help us because github markdown strips the style tag from `<td style="white-space: nowrap">`.

fisker:
Does `<td width="150">` work?

fisker:
Or [`<col>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/col)

bmish:
> Does `<td width="150">` work?

Looks like `<td width="250">` works and is a good width.

bmish:
@fisker markdown doesn't work inside HTML, mainly affecting our backtick code samples in the rule descriptions which no longer render.

Looks like I'd have to manually replace the backticks with `<code>` to fix it, or ideally find a markdown-to-HTML converter which could handle any other markdown (bold, italics) that happened to be used in rule descriptions.

fisker:
Lets move it to the first `<th>`, should work too.

fisker:
Try wrap descriptions with `<span>`, should work.

fisker:
Or `markdown="1"`, found here https://stackoverflow.com/questions/29368902/how-can-i-wrap-my-markdown-in-an-html-div

fisker:
Haha, out of ideas.

bmish:
@fisker the markdown renders correctly when the rule description is separated by blank lines.

fisker:
Let's use markdown for link too? And align code with `outdent`.

fisker:
I wonder if this file can be rendered correctly on npmjs.com.

bmish:
> I wonder if this file can be rendered correctly on npmjs.com.

Good question. That's one reason why I would slightly prefer avoiding the custom HTML solution and going with #1197 instead.

fisker:
I don't have strong opinion. But worth to try, learned 😄

fisker:
Found another solution https://stackoverflow.com/questions/36121672/set-table-column-width-via-markdown/51701842

bmish:
> Found another solution https://stackoverflow.com/questions/36121672/set-table-column-width-via-markdown/51701842, adding `&nbsp;` seems nice.

I tried those solutions earlier and didn't have luck with them.

bmish:
Shall we make a call on going with this PR or #1197 before I keep fixing this PR?

fisker:
> I tried those solutions earlier and didn't have luck with them.

The `&nbsp;` trick works https://github.com/fisker/eslint-plugin-unicorn/tree/readme-rule-name-size#rules

fisker:
`&emsp;` work too

bmish:
> > I tried those solutions earlier and didn't have luck with them.
> 
> The `&nbsp;` trick works https://github.com/fisker/eslint-plugin-unicorn/tree/readme-rule-name-size#rules

Oh nice! That's a bit different from what I tried. Want to just open a PR with your change and I'll close mine?

PR #1203, opened 2021-04-22
Add `prefer-node-protocol` rule

### LINKED ISSUE #1198: Rule proposal: `prefer-node-protocol`
Node.js has added `node:` support for builtin modules https://nodejs.org/api/esm.html#esm_node_imports, I think add it is more clear that module is builtin.

## Fail

```js
import fs from 'fs';
```

```js
export {promises as default} from 'fs'
```

```js
async function foo() {
	const fs = await import('fs');
	// ...
}
```

## Pass

```js
import fs from 'node:fs';
```

```js
export {promises as default} from 'node:fs'
```

```js
async function foo() {
	const fs = await import('node:fs');
	// ...
}
```

fisker:
Fixes #1198


docs/rules/prefer-node-protocol.md
```diff
@@ -0,0 +1,45 @@
+# Prefer use `node:` protocol when importing Node.js builtin modules
```
sindresorhus, line 1:
```suggestion
# Prefer using the `node:` protocol when importing Node.js builtin modules
```

sindresorhus (changes requested):

fisker:
CI seems not running.

sindresorhus:
GH Actions is currently under cryptocurrency attacks. I think that might be related.

https://github.com/search?q=%22Demo+title+Add+files+via+upload%22+is:pr+user:sindresorhus&type=issues

rubiin:
how to use this type of import in typescript? Am I missing something. Using nodejs V14+

sindresorhus:
@rubiin This is not the place to ask TypeScript questions. Try https://github.com/microsoft/TypeScript/issues

PR #1206, opened 2021-04-23
`prefer-node-protocol`: Add `checkRequire` option

fisker:
Requested https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1198#issuecomment-825470155


sindresorhus:
I would personally have just waited for Node.js to backport the feature. Seems like a bit of waste of time to add an option that will soon be moot. But the work is already done, so why not.

karlhorky:
So looks like the `node:` scheme / prefix was backported to Node.js 14.13.1 / 12.20.0 🎉

https://nodejs.org/api/esm.html -> link "node: Imports" (links with anchors don't scroll properly right now)

Screen Shot 2021-05-30 at 11 04 39

karlhorky:
Ah, maybe `require` / CommonJS is not quite there yet:

> In Node 16 `require('node:fs')` works, but it looks like it still doesn't in 14.17.0.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52595#issuecomment-851015583

PR #1209, opened 2021-04-25
Use `null` to assert no autofixes tests

fisker:
It's more clear and has a better message, https://github.com/eslint/eslint/blob/8a77b661bc921c3408bae01b3aa41579edfc6e58/lib/rule-tester/rule-tester.js#L846-L852.

Used for core rules https://github.com/eslint/eslint/search?q=output%3A+null

### DIFF
--- a/package.json
+++ b/package.json
@@ -121,6 +121,14 @@
 				"parserOptions": {
 					"sourceType": "module"
 				}
+			},
+			{
+				"files": [
+					"test/*.mjs"
+				],
+				"rules": {
+					"unicorn/no-null": "off"
+				}
 			}
 		],
 		"rules": {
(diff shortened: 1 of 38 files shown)


sindresorhus:
I don't find it intuitive to use `null` for this at all. Can we use `hasOutput: false` or something and then in `getTester` replace it with `output: null` to make it rule-tester compatible?

fisker:
Sounds good, I'll try

fisker:
I'm going to remove `output` from tests, and check `if (output === code)` before sending to eslint tester, if they are the same, enforce to remove it. And then add `output: null` send to tester.

fisker:
Close infavor of #1259

PR #1211, opened 2021-04-25
Drop support for Node.js v10 and `babel-eslint`

### LINKED ISSUE #1040: Remove support for `babel-eslint`
It's [not maintained anymore](https://github.com/babel/babel-eslint), so I don't think it makes sense for us to support it.

- [ ] Document that `babel-eslint` is not supported.
- [x] Remove `test.babelLegacy()`.
- [ ] Detect and throw at runtime?
- [ ] Remove workarounds that are fixed in `@babel/eslint-parser`, like: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1036/files#diff-66bd496c5747eb2ce80ffc1d5e923d08c0bc538e37af77e1078e106e992d7a4aR11-R13, and [`ExperimentalRestProperty` node](https://github.com/sindresorhus/eslint-plugin-unicorn/search?q=ExperimentalRestProperty)

I think we can do `1` right now, and do the rest a little bit later on to give people time to migrate.

@fisker Thoughts?

fisker:
I think we can close #1040, see https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1040#issuecomment-826300148

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
 		"@lubien/fixture-beta-package": "^1.0.0-beta.1",
 		"@typescript-eslint/parser": "^4.21.0",
 		"ava": "^3.15.0",
-		"babel-eslint": "^10.1.0",
 		"chalk": "^4.1.0",
 		"enquirer": "2.3.6",
 		"eslint": "^7.23.0",
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"engines": {
-		"node": ">=10"
+		"node": ">=12"
 	},
 	"scripts": {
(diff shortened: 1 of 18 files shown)


fisker:
I believe `TypeAlias` and `TypeParameter` was used for `flow` in `babel-eslint`, they may still in use when `flow` plugin enabled in `@babel/estree-parser`, should we care?

Removed in this commit https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1211/commits/c65c2254e331a6a37eb6a8dba321024cf8602b58, they are dead, since we don't run tests for `flow`.

sindresorhus:
> should we care?

No

fisker:
All done.

PR #1220, opened 2021-04-27
Refactor utils

fisker:
Refactor and rename some utility functions, it's more correct.


rules/utils/is-assignment-pattern-shorthand-property-identifier.js
```diff
@@ -1,17 +0,0 @@
...
-module.exports = identifier =>
-	identifier.parent.type === 'AssignmentPattern' &&
-	identifier.parent.left === identifier &&
-	identifier.parent.parent.type === 'Property' &&
-	(
-		identifier === identifier.parent.parent.key ||
-		// In `@babel/eslint-parser` the `variable.reference.identifier`
-		// is neither reference of `identifier.parent.key` nor `identifier.parent.value`
-		// https://github.com/babel/babel/issues/13205
-		hasSameRange(identifier, identifier.parent.parent.key)
```
fisker, line 14: Turns out this check is not needed. 

```js
const {foo = 1} = {}
```

```js
const {foo: bar = 1} = {}
```

In the second one, `bar` is not the `left`, so check `left` is enough.

PR #1221, opened 2021-04-27
`prefer-ternary`: Check `if + return/throw`

### LINKED ISSUE #1116: Can `prefer-ternary` be merged into eslint core rule `no-else-return`?
I think they achieve similar functionality.

fisker:
Fixes #1116


fisker:
//cc @fregante  @JounQin https://github.com/fisker/eslint-plugin-unicorn/pull/282/files I run this on our codebase, look good to you?

fregante:
Indeed there’s an existing issue with that rule:

- prefer-ternary should probably not apply to long lines, it tends to make them less readable

a possible improvement would be to put the condition on its own constant.

JounQin:
LGTM except @fregante mentioned.

fisker:
```js
if (a) {
	return 1;
}

if (b) {
	return 2;
}

if (c) {
	return 3;
} else {
	return 4;
} 
```

This seems fine to fix to
```js
if (a) {
	return 1;
}

if (b) {
	return 2;
}

return c ? 3 : 4;
```


But

```js
if (a) {
	return 1;
}

if (b) {
	return 2;
}

if (c) {
	return 3;
}

return 4;
```

seems bad to fix to

```js
if (a) {}

if (b) {}

return c ? 3 : 4
```

fisker:
Example this line https://github.com/fisker/eslint-plugin-unicorn/pull/282/files#diff-b6b472744fff513bca9c7577f37bc75b74371f8f7fab1621e4aae16ce15793dcL53

sindresorhus:
> seems bad to fix to

Agreed

sindresorhus:
Closing as this has been open for multiple years now without any activity.

PR #1222, opened 2021-04-27
Add title to rule proposal template

fisker:
This front matter should work https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository


fisker:
Works!

PR #1226, opened 2021-04-28
`explicit-length-check`: Check `Set#size`

### LINKED ISSUE #958: `explicit-length-check`: Check `Set#size`
Make [`explicit-length-check`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/explicit-length-check.md) rule work for `Set#size`

We can't tell `foo` in `foo.size` is a `Set`, but it seems safe to assume `.size` is an `integer`.

Ref: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/952#issuecomment-749907925

fisker:
Fixes #958


sindresorhus:
You need to update the readme too.

wooorm:
I just now ran into this being incorrect. I have an [arbitrary object with a `size` field](https://github.com/syntax-tree/xast-util-feed/blob/b19f79a8e82483fc43930475bb18e2dc3b5c8c4b/lib/rss.js#L153).

Maybe I’m alone but I think `size` is a bit more common for other things, compared to `length`?

fisker:
@wooorm  It seems `size` is [a required number field](https://github.com/syntax-tree/xast-util-feed/blob/b19f79a8e82483fc43930475bb18e2dc3b5c8c4b/lib/types.js#L9), and you are checking existence?  I think better use `Reflect.has` in your case.

PR #1234, opened 2021-04-29
`prefer-switch`: Avoid conflict with `no-unreachable` rule

### LINKED ISSUE #1232: `prefer-switch` rule causes unreachable break statement when return statement present
The autofixer of the [unicorn/prefer-switch](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-switch.md) rule adds a `break` statement to the end of each case. This `break` statement could be considered redundant or an error/violation when coming after a `return` statement. Likely the `break` should be omitted after a `return`?

Before autofix:

```js
function getSomething() {
  if (p.type === 'PathExpression') {
    return 1;
  } else if (p.type === 'StringLiteral') {
    return 2;
  } else if (p.type === 'SubExpression') {
    return 3;
  } else {
    return 4;
  }
}
```

After autofix:

```js
function getSomething() {
  switch (p.type) {
    case 'PathExpression': {
      return 1;

      break;
    }
    case 'StringLiteral': {
      return 2;

      break;
    }
    case 'SubExpression': {
      return 3;

      break;
    }
    default: {
      return 4;
    }
  }
}
```

Causes two errors on the `break` lines:

```
Unreachable code detected. ts(7027)
```

```
Unreachable code. eslint no-unreachable
```

https://eslint.org/docs/rules/no-unreachable (recommended rule)

fisker:
I believe there should be still missing cases, but we can simply add more checks later.

// @bmish 

Fixes #1232


sindresorhus:
This fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1232, correct?

fisker:
Yes, forgot link issue.

PR #1235, opened 2021-04-29
`prefer-switch`: Do not add braces to the default case

fisker:


fisker:
Bad idea, this violate `no-case-declarations`

PR #1242, opened 2021-05-06
`no-for-loop`: Ignore known non-array loop variables

### LINKED ISSUE #738: `no-for-loop` rule produces invalid autofix in trying to use `.entries()` to iterate a string
In situations like this where we can tell that a variable being iterated is a string (not an array), we should not convert the loop to use the `.entries()` function which does not exist on strings.

Before:
```js
const myString = 'helloworld';
for (let i = 0; i < myString.length; i++) {
    console.log(i, myString[i])
}
```

After autofix (invalid):
```js
const myString = 'helloworld';
for (const [i, element] of myString.entries()) {
    console.log(i, element);
}
```

bmish:
In situations like this where we can tell that a variable being iterated is not an array (i.e. a string), we should not convert the loop to use the `.entries()` function which would not exist on non-arrays.

Fixes #738.

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -4,6 +4,7 @@ const {isClosingParenToken} = require('eslint-utils');
 const getDocumentationUrl = require('./utils/get-documentation-url');
 const isLiteralValue = require('./utils/is-literal-value');
 const avoidCapture = require('./utils/avoid-capture');
+const {getStaticValue} = require('eslint-utils');
 
 const MESSAGE_ID = 'no-for-loop';
 const messages = {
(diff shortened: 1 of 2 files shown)


fisker (approved):
LGTM

bmish:
Force-pushed to update commit author, no code change.

sindresorhus:
@bmish Can you fix the merge conflict?

bmish:
> @bmish Can you fix the merge conflict?

Fixed.

rules/no-for-loop.js
```diff
@@ -5,6 +5,7 @@ const isLiteralValue = require('./utils/is-literal-value');
 const avoidCapture = require('./utils/avoid-capture');
 const getChildScopesRecursive = require('./utils/get-child-scopes-recursive');
 const singular = require('./utils/singular');
+const {getStaticValue} = require('eslint-utils');
```
sindresorhus, line 8: See the first require statement.

bmish: Fixed. I'm used to the [import/no-duplicates](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md) rule catching/autofixing that for me in some of my own codebases.

PR #1244, opened 2021-05-08
Add `no-document-cookie` rule

### LINKED ISSUE #695: Rule proposal: `no-document-cookie`
It's not recommended to use `document.cookie` directly as it's easy to get the string wrong. See https://github.com/sindresorhus/eslint-plugin-unicorn/issues/299 and https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie Instead, the user should use a cookie library.

I guess we should only prevent setting `document.cookie`? There might be valid reasons to read it?

fisker:
Fixes #695


docs/rules/no-document-cookie.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Fail
+
+```js
+document.cookie =
+	'foo=bar' +
+	'; Path=/' +
+	'; Domain=example.com' +
+	'; expires=Fri, 31 Dec 9999 23:59:59 GMT' +
+	'; Secure';
```
sindresorhus, line 13: Maybe use a multi line template literal here?

sindresorhus (changes requested):

docs/rules/no-document-cookie.md
fisker: Copied from https://wicg.github.io/cookie-store/#intro-proposed-change

PR #1247, opened 2021-05-08
Add `prefer-prototype-methods` rule

### LINKED ISSUE #1223: Rule proposal: `prefer-prototype-methods`
### Fail

```js
[].push.apply(foo, elements);
```

```js
[].slice.call(foo);
```

```js
({}).toString.call(foo);
```

```js
({}).hasOwnProperty.call(foo, 'bar');
```

### Pass

```js
Array.prototype.push.apply(foo, elements);
```

```js
Array.prototype.slice.call(foo);
```

```js
Object.prototype.toString.call(foo);
```

```js
Object.prototype.hasOwnProperty.call(foo, 'bar');
```

Alter name `no-literal-prototype`

fisker:
Fixes #1223


docs/rules/prefer-prototype-methods.md
```diff
@@ -0,0 +1,49 @@
+# Prefer use methods from the prototype instead of methods from an instance.
```
bmish, line 1: This rule is an extension of the recommended eslint core rule [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins) right? It would be good to link to that as a related rule in the rule doc.

fisker: I don't think they are related.

rules/prefer-prototype-methods.js
```diff
@@ -0,0 +1,110 @@
+'use strict';
+const getDocumentationUrl = require('./utils/get-documentation-url');
+const methodSelector = require('./utils/method-selector');
+const getPropertyName = require('./utils/get-property-name');
+
+const messages = {
+	'known-constructor-known-method': 'Prefer use `{{constructorName}}.prototype.{{methodName}}`.',
```
sindresorhus, line 7:
```suggestion
	'known-constructor-known-method': 'Prefer using `{{constructorName}}.prototype.{{methodName}}`.',
```

Same with the others.

rules/prefer-prototype-methods.js
```diff
@@ -0,0 +1,110 @@
...
+		[functionMethodsSelector]: check
+	};
+}
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Prefer use methods from the prototype instead of methods from an instance.',
```
sindresorhus, line 103:
```suggestion
			description: 'Prefer using methods from the prototype instead of methods from an instance.',
```

sindresorhus, line 103: Hmm, maybe this makes the intent clearer?

```suggestion
			description: 'Prefer borrowing methods from the prototype instead of the instance.',
```

docs/rules/prefer-prototype-methods.md
```diff
@@ -0,0 +1,49 @@
...
+
+When "borrow" a method from different objects (especially generic methods from `Array`), it's more clear to use it from the constructor prototype.
```
sindresorhus, line 3:
```suggestion
When “borrowing” a method from a different object (especially generic methods from `Array`), it‘s clearer to get it from the prototype than from an instance.
```

sindresorhus (changes requested):

PR #1257, opened 2021-05-10
Update `xo` to v0.40

fisker:
Should be bug in `xo`


sindresorhus:
https://github.com/xojs/xo/releases/tag/v0.40.1

PR #1264, opened 2021-05-11
`explicit-length-check`: Ignore known, non-number length properties

### LINKED ISSUE #1262: [explicit-length-check]: misunderstanding a string an index
eslint-plugin-unicorn is fantastic! Keep up the good work.

image
In this screenshot you can see property 'size' is a string. I think it's safe as even `"0"` resolves to truthy.

explicit-length-check

```js
const blah = { size: 'small' }
blah.size && true
```

bmish:
Fixes #1262.


fisker (changes requested):
Looks good, but seems you forgot to update snapshot.

rules/explicit-length-check.js
```diff
@@ -154,6 +154,12 @@ function create(context) {
 			let node;
 			let autoFix = true;
 
+			const staticValue = getStaticValue(lengthNode, context.getScope());
+			if (staticValue && typeof staticValue.value !== 'number') {
+				// Ignore known, non-number length properties.
```
fisker, line 159: Let's get this even stricter, only allow positive integers.

bmish: Good point! Fixed.

fisker: Actually, I mean use `Number.isInteger()` exclude `1.2` too.

bmish: Fixed.

fisker (approved):

fisker:
@bmish  Can you add `NaN` and `Infinity` to the test?

PR #1275, opened 2021-05-14
Minor refactor

fisker:
- Add `toLocation` to simplify report on range
- Use `Node.loc` where possible, [loc info is required by ESLint](https://github.com/eslint/eslint/blob/c981fb1994cd04914042ced1980aa86b68ba7be9/lib/source-code/source-code.js#L38), safe to use, don't need `getLocFromIndex`


sindresorhus:

PR #1277, opened 2021-05-14
`prefer-string-starts-ends-with`: add suggestions for safely handling non-strings

### LINKED ISSUE #1267: `prefer-string-starts-ends-with`: incorrect autofix when string does not exist
### Problem

If the string variable is null or undefined, the autofix will cause an exception.

Before autofix:

```js
/^en/.test(lang)
// Returns `false` if `lang` is `undefined` / `null`
```

After autofix:

```js
lang.startsWith('en')
// Uncaught TypeError: Cannot read property 'startsWith' of undefined
```

### Fix

A year from now when we drop support for Node 12, we can use optional chaining to autofix this safely:

```js
lang?.startsWith('en')
```

Until then, we have a few options:
1. Do nothing since the autofixer has been around or a year (#711) and no one else has complained
2. Temporarily turn the autofixer into an automated suggestion until we can use the optional chaining autofix
3. Warn about this situation in the [rule doc](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-string-starts-ends-with.md)

bmish:
Fixes #1267.


test/prefer-string-starts-ends-with.mjs
```diff
@@ -70,29 +73,109 @@ test({
...
 		{
 			code: '(/^b/).test(("a"))',
 			output: '("a").startsWith((\'b\'))',
-			errors: [{messageId: MESSAGE_STARTS_WITH}]
+			errors: [{
+				messageId: MESSAGE_STARTS_WITH,
+				suggestions: [
+					{
+						messageId: SUGGEST_STRING_CAST,
+						output: '(String("a")).startsWith((\'b\'))' // TODO: remove extra parens around String()
```
bmish, line 125: TODO: remove extra parens around String()

rules/prefer-string-starts-ends-with.js
```diff
@@ -64,33 +70,62 @@ const create = context => {
...
+				) {
+					targetString = `(${targetString})`;
+				}
+
+				if (useNullishCoalescing) {
+					// (target ?? '').startsWith(pattern)
+					targetString = (isRegexParenthesized ? '' : '(') + targetString + ' ?? \'\'' + (isRegexParenthesized ? '' : ')');
+				} else if (useStringCasting) {
+					// String(target).startsWith(pattern)
+					const isTargetStringParenthesized = targetString.startsWith('(');
```
bmish, line 93: This check needs work.

test/prefer-string-starts-ends-with.mjs
fisker: I don't think we should, just keep it.

rules/prefer-string-starts-ends-with.js
fisker: Let's use `getParenthesizedText` here, so this case `.test( (0, undefined) )`, we don't need add check when move to `String( (0, undefined) )`

fisker:
Use generator function instead of return array, should make function more clear

rules/prefer-string-starts-ends-with.js
```diff
@@ -64,33 +70,62 @@ const create = context => {
...
+				const method = result.messageId === MESSAGE_STARTS_WITH ? 'startsWith' : 'endsWith';
+				const [target] = node.arguments;
+				let targetString = sourceCode.getText(target);
+				const isRegexParenthesized = isParenthesized(regexNode, sourceCode);
+				const isTargetParenthesized = isParenthesized(target, sourceCode);
+
+				if (
+					// If regex is parenthesized, we can use it, so we don't need add again
+					!isRegexParenthesized &&
+					(isTargetParenthesized || shouldAddParenthesesToMemberExpressionObject(target, sourceCode))
```
fisker, line 83: And this need change, because the MemberExpression.object is not the old `target` any more, for example.

```js
/^a/.test(1)
```


```js
(1).startsWith()
```
this need add `()`

```js
String(1).startsWith()
```

this don't need.

docs/rules/prefer-string-starts-ends-with.md
```diff
@@ -4,6 +4,8 @@ Prefer [`String#startsWith()`](https://developer.mozilla.org/en/docs/Web/JavaScr
 
 This rule is fixable.
 
+Note: the autofix will throw an exception when the string being tested is `null` or `undefined`. Several safer but more verbose automatic suggestions are provided for this situation.
```
fisker, line 7: It not only about `null` and `undefined`, could be anything that's not a string

```js
/^1/.test(1);
```

```js
/^t/.test(true);
```

```js
/^1/.test(10n);
```

```js
/^1,2/.test([1,2,3,])
```

```js
/^\[object/.test({})
```

```js
/^a/.test(async () => {})
```

test/prefer-string-starts-ends-with.mjs
bmish: Okay.

rules/prefer-string-starts-ends-with.js
bmish: I significantly refactored and improved the parenthesis logic.

docs/rules/prefer-string-starts-ends-with.md
bmish: Updated messaging.

rules/prefer-string-starts-ends-with.js
fisker: I'm really sorry that I didn't make it clear  I  don't mean that one , should use [this one ](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/0b980dd719a9fdf6d3e62db17a9259190d3a30da/rules/utils/parentheses.js#L70). The one in `prefer-ternary` adding parens only fit to fix into ternary. We need write a new one here .

test/snapshots/prefer-string-starts-ends-with.mjs.md
```diff
@@ -34,6 +46,18 @@ Generated by [AVA](https://avajs.dev).
     `␊
     > 1 | /^a/.test((0, "string"))␊
         | ^^^^^^^^^^^^^^^^^^^^^^^^ Prefer \`String#startsWith()\` over a regex with \`^\`.␊
+    ␊
+    --------------------------------------------------------------------------------␊
+    Suggestion 1/3: For strings that may be \`undefined\` / \`null\`, use string casting.␊
+      1 | String(0, "string").startsWith(('a'))␊
```
fisker, line 52: As you can see here , this one is not correct.

rules/prefer-string-starts-ends-with.js
bmish: Updated to use:

```
const {getParenthesizedText} = require('./utils/parentheses');
```

fisker: I'm sent a PR to update `getParenthesizedText` in `prefer-ternary`, #1280

You need do similar thing here, 

````js
let text = getParenthesizedText(target, ...)
if (useNullishCoalescing) {
	if (
		!isParenthesized(target) &&
		shouldAddParenthesesToLogicalExpressionLeft(target)
	) 
		text = `(${text})`
	}

	// then we need use a fake LogicalExpression to check `shouldAddParenthesesToMemberExpressionObject`
	// Or just add parens to `${text} ?? ''` since it's always needed.
}

Comment (bmish) [reply rules/prefer-string-starts-ends-with.js]: Will try...

File: test/snapshots/prefer-string-starts-ends-with.mjs.md (hunk 5 above)
Comment (bmish) [reply test/snapshots/prefer-string-starts-ends-with.mjs.md]: Fixed this

File: rules/prefer-string-starts-ends-with.js (hunk 2 above)
Comment (bmish) [reply rules/prefer-string-starts-ends-with.js]: I added a new function `shouldAddParenthesesToLogicalExpressionLeft` which is a copy of `shouldAddParenthesesToConditionalExpressionChild`. Not entirely sure if it should be different at all...

Comment (bmish) [reply rules/prefer-string-starts-ends-with.js]: Should this function be `shouldAddParenthesesToLogicalExpressionLeft` or `shouldAddParenthesesToLogicalExpressionChild`? Trying to figure out if the left side has different needs...

Comment (fisker) [reply rules/prefer-string-starts-ends-with.js]: I think left side is different from the right side, eg `a ?? b` to be `??`'s left don't need but if it's right, it's needed. We can still call that function `...Node` but only implement `left` part logic, because we only need that now, we also need pass operator so we should call it in this way `shouldAddParenthesesToLogicalExpressionChild(node, {operator:"??", property:"left"})`.

Comment (fisker) [reply rules/prefer-string-starts-ends-with.js]: > Not entirely sure if it should be different at all...

They are total different, you should consider all nodes has lower precedence according to [ this rable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#Table).

File: rules/prefer-string-starts-ends-with.js [hunk 6]
```diff
@@ -4,12 +4,20 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const methodSelector = require('./utils/method-selector');
 const quoteString = require('./utils/quote-string');
 const shouldAddParenthesesToMemberExpressionObject = require('./utils/should-add-parentheses-to-member-expression-object');
+const shouldAddParenthesesToLogicalExpressionChild = require('./utils/should-add-parentheses-to-logical-expression-child');
+const {getParenthesizedText} = require('./utils/parentheses');
 
 const MESSAGE_STARTS_WITH = 'prefer-starts-with';
 const MESSAGE_ENDS_WITH = 'prefer-ends-with';
+const SUGGEST_STRING_CAST = 'suggest-string-cast';
```
fisker, line 12: Let's give these two constants a  better name, so we can use  

```js
suggest: [A, B, C].map(type => ({messageId: type, fix: fix(node, type}))
``

And also use these constants to compare `type` in `fix` function.

Comment (bmish) [reply rules/prefer-string-starts-ends-with.js]: Cool idea, updated.

Comment (fisker) [reply rules/prefer-string-starts-ends-with.js]: Not the thing in my mind, let me commit my idea.

Comment (fisker) [reply rules/prefer-string-starts-ends-with.js]: Check https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1277/commits/cb5e41dbcede4f2a0f4a49cb7aedcce6cef3cdff, better?

Comment (bmish) [reply rules/prefer-string-starts-ends-with.js]: That works, I just pushed a commit to finish that refactoring.

**Comment by fisker** [comment]
The code fix part  still need more work , if you don't mind, I can fix it later, unless you prefer do it yourself, I can guide.

Another thing, I think we should remove auto fix for unknow non-strings, yes auto-fix will throw on runtime, but not every line is test covered, or even not reachable. You need fix this.

**Comment by bmish** [comment]
@fisker I opened a separate PR for ignore known non-strings: #1281 I think we should merge that first since that can be done independently of this suggestion PR.

**Comment by bmish** [comment]
@fisker feel free to push tweaks if you want. I will keep working on it too, appreciate your guidance!

File: rules/prefer-string-starts-ends-with.js (hunk 3 above)
Comment (bmish) [reply rules/prefer-string-starts-ends-with.js]: I think this comment is outdated.

**Comment by fisker** [comment]
I pushed some changes, [this](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1277/commits/52f3d666162ff909c8e6c3b73b18933bb551b8d1) corrects the fix logic, [this](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1277/commits/c1e4502ed19920835796fccf15da021909aff4b6) add missing parens when the target is ternary `(a ? b : c ?? '').startsWith` -> `((a ? b : c) ?? '').startsWith`, see https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1277#discussion_r632980437.

File: rules/utils/should-add-parentheses-to-logical-expression-child.js [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+		node.type === 'AssignmentExpression' ||
+		node.type === 'YieldExpression' ||
+		node.type === 'SequenceExpression'
+	) {
+		return true;
+	}
+
+	/* istanbul ignore next: When operator or property is different, need more logic here, not implemented */
+	if (operator !== '??' || property !== 'left') {
+		throw new Error('Not supported.');
```
bmish, line 32: Shouldn't we throw this exception at the beginning of the function so people don't try to use it for this without implementing it fully?

fisker: I was hoping this function can return earlier, and if the operator or property is different. logic should be write here, but your concern make sense too, I don't have strong opinion.

fisker:
@bmish Can you fix conflicts?

fisker:
Removed suggestions for known string.

Updated message, I'm not good at English, but I think it's more useful.

docs/rules/prefer-string-starts-ends-with.md
```diff
@@ -4,6 +4,8 @@ Prefer [`String#startsWith()`](https://developer.mozilla.org/en/docs/Web/JavaScr
 
 This rule is fixable.
 
+Note: the autofixed code will throw an exception when the value being tested is not a string. Several safer but more verbose automatic suggestions are provided for this situation.
```
fisker, line 7: How about a simple one

```suggestion
This rule is fixable, unless the matching object is known not a string.
```

fisker (approved):
The code looks good, few other comments.

Thanks @bmish

rules/prefer-string-starts-ends-with.js
```diff
@@ -64,40 +72,102 @@ const create = context => {
...
+						if (!isRegexParenthesized) {
+							targetText = addParentheses(targetText);
+						}
+
+						break;
+
+					// Goal: `String(target).startsWith(pattern)`
+					case FIX_TYPE_STRING_CASTING:
+						// `target` was a call argument, don't need check parentheses
+						targetText = `String(${targetText})`;
```
bmish, line 127: This adds extra parenthesis that aren't needed sometimes. I think I had code to handle this before.

`(/^b/).test((a))`
`((a)).startsWith('b')`

rules/utils/should-add-parentheses-to-logical-expression-child.js
bmish: I think it's generally better to throw early. Otherwise, someone might use this, and not test it fully, and a consumer may hit the exception in the wild.

bmish:
@fisker looks good, thanks for helping to finish it up.

rules/prefer-string-starts-ends-with.js
fisker: I know that, but better "add missing parens" instead of "remove extra parens".

PR #1281, opened 2021-05-15
`prefer-string-starts-ends-with-rule`: don't autofix known non-string values

bmish:
Known, non-string values don't have a startsWith/endsWith function.


fisker:
I don't think we should ignore, `String(foo).startsWith('foo')` still better than `/^foo/.test(foo)`

bmish:
@fisker updated to only remove the autofix for this situation.

fisker (approved):

fisker:
Thanks

PR #1284, opened 2021-05-18
Add `array-join-separator` rule

### LINKED ISSUE #1246: Rule proposal: `array-join-separator`
Require [`separator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#parameters) when calling [`Array#join()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join). Even mean to use `','` to join, `array.join(',')` is more clear.

## Fail

```js
const string = array.join();
```

## Pass

```js
const string = array.join(',');
```
```js
const string = array.join('|');
```

fisker:
Fixes #1246


docs/rules/array-join-separator.md
```diff
@@ -0,0 +1,37 @@
+# Enforce using the separator argument when concatenating elements in an array
```
sindresorhus, line 1:
```suggestion
# Enforce using the separator argument with `Array#join()`
```

?

docs/rules/array-join-separator.md
```diff
@@ -0,0 +1,37 @@
...
+
+Enforce using [the separator argument](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join#parameters) when calling [Array#join()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join), it's more obvious what the separator is.
```
sindresorhus, line 3:
```suggestion
It's better to make it clear what the separator is when calling [Array#join()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join).
```

sindresorhus (changes requested):

PR #1288, opened 2021-05-18
Add `require-number-to-fixed-digits-argument` rule

### LINKED ISSUE #1245: Rule proposal: `number-to-fixed-digits`
Require pass [`digits`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed#parameters) when calling [`Number#toFixed()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed), most time user just forgot `digits` not mean to `0`.

## Fail

```js
(1.23).toFixed();
```

## Pass

```js
(1.23).toFixed(0);
```

fisker:
Fixes #1245


bmish:
Some rule names in this plugin could use the common `require-` rule name prefix. I.e. `require-number-to-fixed-digit-argument` or `require-array-join-separator`.

fisker:
I thought about it and checked core rules and other plugins, but seems `require-` not very common, but I'm fine to add `require-`.

bmish:
I try to name almost all my lint rules with the following prefixes to make it clear what they are actually doing:
* `require-`
* `no-`
* `prefer-`

docs/rules/number-to-fixed-digits.md
```diff
@@ -0,0 +1,37 @@
+# Enforce using the digits argument with `Number#toFixed()`
+
+It's better to make it clear what the value of the `digits` argument is when calling [Number#toFixed()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed).
```
sindresorhus, line 3: We should mention the default here too, like https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1294

sindresorhus:
I agree. I don't think it's worth renaming existing rules, but we should try to follow that in the future.

sindresorhus:
> require-number-to-fixed-digit-argument

👍

sindresorhus:
> require-array-join-separator

We could do this one too, since it's not released yet.

bmish:
> > require-array-join-separator
> 
> We could do this one too, since it's not released yet.

Rename PR: #1295

sindresorhus:
> I try to name almost all my lint rules with the following prefixes to make it clear what they are actually doing:

Feel free to open a new issue with your suggestions on what rules should be renamed and we can consider it.

fisker:
I forgot another reason I didn't add `require-` at first place, if we want to do more in this rule, eg: check `digits` type, check it's `0-100` then the rule name won't fit. Same concern for `require-array-join-seperator`.

bmish:
> I forgot another reason I didn't add `require-` at first place, if we want to do more in this rule, eg: check `digits` type, check it's `0-100` then the rule name won't fit. Same concern for `require-array-join-seperator`.

@fisker makes sense, I agree that if there's any chance we would want to generalize what the rule does, then your original name would be better. Feel free to revert if that's the case. In the past, I've also named rules `require-valid-foo-argument` in order to cover various possible checks that we could perform on that argument.

fisker:
Another example in core rules. The [`radix`](https://eslint.org/docs/rules/radix) rule added check that number need to be integer `2-36`, https://github.com/eslint/eslint/pull/12608, the rule name didn't block that change.

fisker:
> require-valid-foo-argument

👍 , but for this rule,  `require-valid-number-to-fixed-digits-argument` seems way too long

sindresorhus:
`npm run generate-rules-table && git diff --exit-code` is failing.

fisker:
I'll fix, what's your opinion about https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1288#issuecomment-843714971

sindresorhus:
> I forgot another reason I didn't add require- at first place, if we want to do more in this rule, eg: check digits type, check it's 0-100 then the rule name won't fit. Same concern for require-array-join-seperator.

I think in general we should try to think upfront before adding a rule whether we're likely to add more stuff to it in the future, if not, I think it's fine to use a specific name with `require-*`.

> eg: check digits type

We can still do this even when it's named `require-number-to-fixed-digits-argument`.

PR #1296, opened 2021-05-18
`explicit-length-check`: Ignore objects with non-Array/Set properties/functions

### LINKED ISSUE #1266: `explicit-length-check`: ignore variables with non-array properties?
```js
if (
  packagingData.dimensions.width &&
  packagingData.dimensions.height &&
  packagingData.dimensions.length // violation reported by unicorn/explicit-length-check
) { }
```

In this example, `packagingData.dimensions` is an object, and `length` could be a string. 

As a heuristic to avoid false positives with detecting `length` or `size` on variables that are not actually arrays, could we ignore variables that have non-array properties? For example, an array does not have a `width` property, so if we see that used, we would assume `dimensions` is not an array, and ignore it.

bmish:
Fixes #1266.

Cases that aren't ignored:

* Nested objects: `const foo = { bar: getBar() }; if (foo.bar.width && foo.bar.length) {}`
* When we can't find the variable declaration: `if (foo.width && foo.length) {}`


rules/explicit-length-check.js
```diff
@@ -161,6 +167,32 @@ function create(context) {
...
 			}
 
+			const variable = findVariable(context.getScope(), lengthNode.object);
+			if (variable) {
+				const identifiers = getVariableIdentifiers(variable).filter(identifier => identifier !== lengthNode.object);
+				let expectedPrototype;
+				if (lengthNode.property.name === 'length') {
+					expectedPrototype = Array.prototype;
+				} else if (lengthNode.property.name === 'size') {
+					expectedPrototype = Set.prototype;
```
fisker, line 177: This maybe hard than I thought, I don't think we mean to only check `Array#length` and `Set#size`, there are many builtin objects we want check,eg: `Map#size` `{File, Blob}.size`.

bmish: Oh, I see. If we are just intending to check these built-in JS objects, shall I update this to check those prototypes too?

fisker: Problem: too many builtins has `.length`.

I searched around, but didn't see a data with all those objects and their properties.


Image we have a list

```
File: ['size', 'method1', 'property1'],
Foo: ['size', 'method1', 'property1'],
Bar: ['noSize', 'method1', 'property1'],
```

We got all objects can has `.size` and all their methods and properties.

bmish: Okay so you recommend closing this?

We should be able apply this strategy to other rules though which are limited to specific data structures, right?

fisker: Les't make a list ourself. List `Map, Set Blob`'s property, and only check `.size` for now.

fisker: And `fs.Stats` too.

fisker:
```js
Object.keys(Object.getOwnPropertyDescriptors(Set.prototype))
(10) ["constructor", "has", "add", "delete", "clear", "entries", "forEach", "size", "values", "keys"]

Object.keys(Object.getOwnPropertyDescriptors(Map.prototype))
(11) ["constructor", "get", "set", "has", "delete", "clear", "entries", "forEach", "keys", "size", "values"]

Object.keys(Object.getOwnPropertyDescriptors(Blob.prototype))
(7) ["size", "type", "arrayBuffer", "slice", "stream", "text", "constructor"]

Object.keys(Object.getOwnPropertyDescriptors(File.prototype))
(5) ["name", "lastModified", "lastModifiedDate", "webkitRelativePath", "constructor"]

Object.keys(Object.getOwnPropertyDescriptors(require('fs').Stats.prototype))
[ 'constructor', 'isFile', '_checkModeProperty' ]
```

sindresorhus:
This needs to be rebased from the main branch.

sindresorhus:
@bmish Still interested in finishing this?

PR #1299, opened 2021-05-19
`prefer-array-flat-map`: Remove check on `[].concat(...array.map(…))`

### LINKED ISSUE #1298: Simplify `prefer-array-flat-map` rule
This.rule also [check `[].concat(...arr.map())`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/be512266bf4ac1e4c9153fd5a8946f1c1f90156b/rules/prefer-array-flat-map.js#L14-L48), and don't have a fix for it.

This also checked in `prefer-array-flat`, and doing a better job, it checkes `[].concat(array)` , `[].concat(...array)`, `[].concat.apply([], array)` , and `Array.prototype.concat.apply([], array)`.

We should remove that case, let `prefer-array-flat` handle that, when it fixed to `.map().flat()`, will be caught by `prefer-flat-map`. I don't think anyone enables `prefer-array-flat-map` but don't enable `prefer-array-flat`.

I'll do this after #1297, going to use code from that.

fisker:
Fixes #1298


docs/rules/prefer-array-flat-map.md
```diff
@@ -8,14 +8,28 @@ This rule is fixable.
...
+const foo = bar.map(element => unicorn(element)).flat(2);
+```
+
+```js
+const foo = bar.map(element => unicorn(element)).foo().flat();
+```
+
+```js
+const foo = bar.flat().map(element => unicorn(element));
 ```
```
bmish, line 35: There should be a `## Related Rules` section that links to the `prefer-array-flat` rule. And vice versa.

PR #1301, opened 2021-05-20
`new-for-builtins` & `no-new-buffer`: Handle line breaks after `new` keyword

fisker:
When class name is not on the same line with `return`, auto-fix can break code.

```js
() => {
	return new //
		Buffer()
}
```

```js
() => {
	return // returns `undefined` here
		Buffer.from()
}
```

Added parentheses after `return` to fix this.


rules/fix/index.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = {
+	switchNewExpressionToCallExpression: require('./switch-new-expression-to-call-expression')
+};
```
fisker, line 5: I'm adding more files to this dir.

sindresorhus:

PR #1302, opened 2021-05-20
Add `appendArgument` fix utility

fisker:
I was planning to more "fix" related changes in this PR, but found bug #1301, I think better make PRs smaller should be easier to review, and if bug found during refactor, I don't need to split code.


fisker:
I'm going merge this as base for other fix related PRs.

PR #1303, opened 2021-05-20
Update selectors with boolean attributes

fisker:
Use `!=` seems safer than `=`, prevent cases like #1302

When running with lower ecma version , some property didn't present, so it could be `undefined`.

### DIFF
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -15,7 +15,7 @@ const declaratorSelector = [
 
 const memberSelector = [
 	'MemberExpression',
-	'[computed=false]',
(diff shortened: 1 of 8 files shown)


sindresorhus:
> When running with lower ecma version , some property didn't present, so it could be undefined.

I don't think we should have to deal with legacy cases like that. Is there any way we can enforce the ECMAScript version?

fisker:
> Is there any way we can enforce the ECMAScript version?

I'm not sure. 

But since the  ESTree spec are changing, even latest version, new features maybe not fully supported by parser, there still can be undefined properties. I think `!=` should be prefered.

PR #1305, opened 2021-05-20
Add `callOrNewExpressionSelector`

fisker:


sindresorhus:
Since we're making reusable tools, wouldn't it be easier to work with the AST directly in these tools rather than generate strings for the query syntax?

fisker:
Here is why I prefer use selector,

selector

```js
'[a.b.c="foo"]'
```

js code

```js
if (a && a.b && a.b.c === 'foo');
```

waiting for this support

```js
if (a?.b?.c === 'foo'):
```

sindresorhus:
Yeah, I agree. So I think we should use the AST more when we can use `?` (Node.js 14). It's much easier to make typo mistakes with string construction.

fisker:
Yes, we should use more ast by that time.

There is also a trouble when use selector, it's hard to select `MemberExpression` and `CallExpression` in 
 optional chain, because `ChainExpression` can be any level, makes the selector very difficult to match, if use js code, we can easily skip it. Just let you know.

fisker:
I hate that AST design.

PR #1306, opened 2021-05-20
Indicate which rules provide suggestions in readme rules table

### LINKED ISSUE #1292: Add `suggestion` property to rule meta
1. Add `suggestion: true` to meta, this may change in ESLint V8 https://github.com/eslint/eslint/issues/14312
2. Add suggestion info to rule table https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/scripts/generate-rules-table.mjs
3. Enable eslint rule https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/pull/105

bmish:
Fixes #1292.

This is another useful point of information about rules that we can include in our readme rules table.

Note that this property name/location will likely change slightly in ESLint 8 and become enforced so that it will be more reliable: https://github.com/eslint/eslint/issues/14312

### DIFF
--- a/scripts/generate-rules-table.mjs
+++ b/scripts/generate-rules-table.mjs
@@ -17,6 +17,7 @@ const tablePlaceholder = /<!-- RULES_TABLE_START -->.*<!-- RULES_TABLE_END -->/s
 // Config/preset/fixable emojis.
 const EMOJI_RECOMMENDED = '✅';
 const EMOJI_FIXABLE = '🔧';
+const EMOJI_HAS_SUGGESTIONS = '💡';
 
 // Generate rule table contents.
 const ruleNames = Object.keys(rules).filter(ruleName => !rules[ruleName].meta.deprecated).sort();
@@ -25,13 +26,14 @@ const rulesTableContent = ruleNames
 		// Check which emojis to show for this rule.
 		const isRecommended = configs.recommended.rules[`unicorn/${ruleName}`] === 'error';
 		const isFixable = rules[ruleName].meta.fixable;
+		const hasSuggestions = rules[ruleName].meta.docs.suggest; // Property likely to change in ESLint 8: https://github.com/eslint/eslint/issues/14312
 
 		const url = `docs/rules/${ruleName}.md`;
 		const link = `[${ruleName}](${url})`;
 
 		const {description} = rules[ruleName].meta.docs;
 
-		return `| ${link} | ${description} | ${isRecommended ? EMOJI_RECOMMENDED : ''} | ${isFixable ? EMOJI_FIXABLE : ''} |`;
+		return `| ${link} | ${description} | ${isRecommended ? EMOJI_RECOMMENDED : ''} | ${isFixable ? EMOJI_FIXABLE : ''} | ${hasSuggestions ? EMOJI_HAS_SUGGESTIONS : ''} |`;
 	})
 	.join('\n');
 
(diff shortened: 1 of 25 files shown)


fisker:
Actually, we can test this, some code here  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/utils/test.mjs should work.

bmish:
> Actually, we can test this, some code here [https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/utils/test.mjs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/utils/test.mjs?rgh-link-date=2021-05-20T23%3A28%3A28Z) should work.

Added a test to ensure `meta.docs.suggest` is present in suggestable rules, is that what you had in mind?

fisker:
Yes, but add it to snapshot tester too, please.

scripts/generate-rules-table.mjs
```diff
@@ -25,13 +26,14 @@ const rulesTableContent = ruleNames
...
 		const isFixable = rules[ruleName].meta.fixable;
+		const hasSuggestions = rules[ruleName].meta.docs.suggestion; // Property likely to change in ESLint 8: https://github.com/eslint/eslint/issues/14312
 
 		const url = `docs/rules/${ruleName}.md`;
 		const link = `[${ruleName}](${url})`;
 
 		const {description} = rules[ruleName].meta.docs;
 
-		return `| ${link} | ${description} | ${isRecommended ? EMOJI_RECOMMENDED : ''} | ${isFixable ? EMOJI_FIXABLE : ''} |`;
+		return `| ${link} | ${description} | ${isRecommended ? EMOJI_RECOMMENDED : ''} | ${isFixable ? EMOJI_FIXABLE : ''} | ${hasSuggestions ? EMOJI_HAS_SUGGESTIONS : ''} |`;
```
fisker, line 36:
```suggestion
		return `| ${[
			link,
			description,
			isRecommended ? EMOJI_RECOMMENDED : '',
			isFixable ? EMOJI_FIXABLE : '',
			hasSuggestions ? EMOJI_HAS_SUGGESTIONS : ''
		].join(' | ')} |`;
```

scripts/generate-rules-table.mjs
```diff
@@ -42,8 +44,8 @@ writeFileSync(
 		outdent`
 			<!-- RULES_TABLE_START -->
 
-			| Name${'&nbsp;'.repeat(40)} | Description | ${EMOJI_RECOMMENDED} | ${EMOJI_FIXABLE} |
-			| :-- | :-- | :-- | :-- |
+			| Name${'&nbsp;'.repeat(40)} | Description | ${EMOJI_RECOMMENDED} | ${EMOJI_FIXABLE} | ${EMOJI_HAS_SUGGESTIONS} |
+			| :-- | :-- | :-- | :-- | :-- |
```
fisker, line 48:
```suggestion
			|${' :-- |'.repeat(5)}
```

test/utils/test.mjs
```diff
@@ -112,6 +117,7 @@ class Tester {
 		const tester = snapshotRuleTester(test, {
 			parserOptions: defaultParserOptions
 		});
+		verifySuggestionPropertyConsistency(this.rule, tests.invalid);
```
fisker, line 120: This won't work because snapshot tester don't need set `errors`, should check here https://github.com/sindresorhus/eslint-plugin-unicorn/blob/bb0446884238745fd565b3ebc64e6c0de3f2c31b/test/utils/snapshot-rule-tester.mjs#L105

test/utils/test.mjs
```diff
@@ -8,19 +8,23 @@ import defaultParserOptions from './default-parser-options.mjs';
 
 const require = createRequire(import.meta.url);
 
-function normalizeInvalidTest(test, rule) {
-	const {code, output, errors} = test;
+function verifySuggestionPropertyConsistency(rule, invalidTests) {
+	for (const invalidTest of invalidTests) {
+		if (Array.isArray(invalidTest.errors) && invalidTest.errors.some(error => error.suggestions) && !(rule.meta && rule.meta.docs && rule.meta.docs.suggestion === true)) {
```
fisker, line 13: Let's use `rule.meta.docs.suggestion` directly, because `rule.meta.docs` is required field.

bmish: Updated.

test/utils/test.mjs
bmish: Fixed.

fisker (approved):

PR #1308, opened 2021-05-21
Add `prefer-object-from-entries` rule

### LINKED ISSUE #1260: Rule proposal: `prefer-object-from-entries`
## Fail

```js
const foo = _.fromPairs(pairs);
```
Configurable function list like  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-flat.md#functions

```js
const foo = pairs.reduce((object, [key, value]) => Object.assign(object, {[key]: value}), {});
```

```js
const foo = pairs.reduce((object, [key, value]) => Object.assign(object, {[key]: value}), Object.create(null));
```

```js
const foo = pairs.reduce((object, [key, value]) => ({...object, [key]: value}), {});
```

```js
const foo = pairs.reduce((object, [key, value]) => {
	object[key] = value;
	return object;
}, {});
```

```js
const foo = {};
for (const [key, value] of pairs) {
	foo[key] = value;
}
```

## Pass

```js
const foo = Object.fromEntries(pairs);
```

fisker:
Currently, 

1. Simple cases `pairs.reduce(object => ({...object, key}), {})` and `pairs.reduce(object => Object.assign(object, {key}), {})` are fixable
1. Any `Array#reduce()` with inital value `{}` or `Object.create(null)` (except fixable cases above) are reported
1. `_.fromPairs()` are reported, and a configurable `functions` list.

---

Question:

I'm not sure about the case mentioned in the proposal now, 

```js
const foo = {};
for (const [key, value] of pairs) {
	foo[key] = value;
}
```

do we really want to check it? 

Funny thing, [this rule self violate it](https://github.com/fisker/eslint-plugin-unicorn/blob/f9db2924f109e69c2e279bedbb1f7da653e43a7b/rules/prefer-object-from-entries.js#L176-L208), but I'm not sure I want use `Object.fromEntries()`.

Fixes #1260


sindresorhus:
> Funny thing, this rule self violate it, but I'm not sure I want use Object.fromEntries().

Why not? And would the rule even be able to handle such an advanced case anyway?

sindresorhus:
> I'm not sure about the case mentioned in the proposal now,

I think it should handle simple cases like that, but maybe stop if the inner scope refers to variables outside it or some other heuristic.

fisker:
> Why not? 

Can't tell, personal feeling, get used to use `for...of`.

> would the rule even be able to handle such an advanced case anyway?

The case is the simplest case, only one assignment 😄 , just a little long.

sindresorhus:
I tried rewriting it to using `Object.fromEntries` and it looks ok to me:

```js
const arrayReduce = new Map();

const listeners = Object.fromEntries(fixableArrayReduceCases.map(({selector, test, getKey, getValue}) => {
	const value = function (node) {
		// If this listener exit without adding fix, the `arrayReduceWithEmptyObject` listener
		// should still add it into the `arrayReduce` map, to be safer, add it here too
		arrayReduce.set(node);

		const [callbackFunction] = node.arguments;
		if (!test(callbackFunction)) {
			return;
		}

		const [firstParameter] = callbackFunction.params;
		const variables = context.getDeclaredVariables(callbackFunction);
		const firstParameterVariable = variables.find(variable => variable.identifiers.length === 1 && variable.identifiers[0] === firstParameter);
		if (!firstParameterVariable || firstParameterVariable.references.length !== 1) {
			return;
		}

		arrayReduce.set(
			node,
			// The fix function
			fixReduceAssignOrSpread({
				sourceCode,
				node,
				key: getKey(callbackFunction),
				value: getValue(callbackFunction)
			})
		);
	};

	return [selector, value];
}));
```

fisker:
Yes, if we decide to catch this, it will be fixed like this.

fisker:
But do you think it's better than `for...of`?

fisker:
Let's do this, only report cases don't need to do `.map()`, like the case in proposal, it's already `[key, value]`

```js
// Report this
const foo = {};
for (const [key, value] of pairs) {
	foo[key] = value;
}
```

```js
// Not this
const foo = {};
for (const [key, value] of pairs) {
	foo[key] = doSomething(value);
}
```

sindresorhus:
> WDYT?

👍

fisker:
Close for now, I'll reopen when it's ready.

fisker:
@sindresorhus

After a long time thinking about the `for...of` case, I still can't find a good solution for it.
It's really complicated, I suggest we merge this (only checking `.reduce()`) first, and improve it in future?

sindresorhus:
> It's really complicated, I suggest we merge this (only checking .reduce()) first, and improve it in future?

👍 Can you open an issue about it?

fisker:
#1449

docs/rules/prefer-object-from-entries.md
```diff
@@ -0,0 +1,80 @@
...
+
+Type: `object`
+
+### functions
+
+Type: `string[]`
+
+You can also check custom functions that transforms pairs.
+
+`lodash.fromPairs()` and `_.fromPairs()` are checked by default.
```
sindresorhus, line 58: It's not clear from the docs here whether setting this option overrides the default checked functions or extends it. That should be made more clear.

rules/prefer-object-from-entries.js
```diff
@@ -0,0 +1,274 @@
...
+				fixReduceAssignOrSpread({
+					sourceCode,
+					node,
+					property: getProperty(callbackFunction),
+				}),
+			);
+		};
+	}
+
+	listeners[arrayReduceWithEmptyObject] = function (node) {
```
sindresorhus, line 216:
```suggestion
	listeners[arrayReduceWithEmptyObject] = node => {
```

Same with the other ones.

sindresorhus (changes requested):

docs/rules/prefer-object-from-entries.md
fisker: How about 

> `lodash.fromPairs()` and `_.fromPairs()` are always checked.


Note to myself, [`prefer-array-flat` use the same](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-flat.md#functions).

sindresorhus: 👍

PR #1319, opened 2021-05-27
`no-array-for-each`: Handle only return statement inside `if` or `else`

### LINKED ISSUE #1318: `no-array-for-each` invalid fix
Consider foreach calls some logic as return side-effect:
```js
function handleTwo() {}
function handleNotTwo() {}

[ 1, 2, 3, 4 ].forEach(item => {
    if (item === 2) return handleTwo();
    handleNotTwo();
});
```

after fixing by ```no-array-for-each``` next snippet is generated:

```js
function handleTwo() {}
function handleNotTwo() {}

for (const item of [ 1, 2, 3, 4 ]) {
    if (item === 2)  handleTwo(); continue;
    handleNotTwo();
}
```

here ```handleTwo``` is being called in case item equals 2 as expected, but then continue is called out of if scope.
as a result  ```handleNotTwo```  becomes  unreachable, and snippet **works in different way** then initial  forEach.

Consider addding scope here should work as expected:

```return handleTwo();``` => ```{ handleTwo(); continue; }```

so snippet becomes:
```js
function handleTwo() {}
function handleNotTwo() {}

for (const item of [ 1, 2, 3, 4 ]) {
    if (item === 2)  { handleTwo(); continue; }
    handleNotTwo();
}
```

fisker:
Fixes #1318


rules/utils/needs-semicolon.js
```diff
@@ -43,12 +43,44 @@ function needsSemicolon(tokenBefore, sourceCode, code) {
...
 
 	const {type, value, range} = tokenBefore;
+	const lastBlockNode = sourceCode.getNodeByRangeIndex(range[0]);
 	if (type === 'Punctuator') {
 		if (value === ';') {
 			return false;
 		}
 
-		if (value === ']' || value === ')') {
+		if (value === ']') {
```
fisker, line 52: This part not related to this PR, but found during fix this.

```js
foo.forEach(_ => {
	if (true) return {};
})
```

was fixed to 

```js
foo.forEach(_ => {
	if (true) ;({}); continue;
              ^ Should not insert
})
```

I'll send another PR to add tests cover this part.

PR #1320, opened 2021-06-01
Ensure no unexpected properties present in snapshot tests

bmish:
The snapshot tests in this plugin have a custom implementation and don't benefit from all the verification checks that ESLint performs on standard ESLint tests to ensure they are properly formed ([example from ESLint 7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#rule-tester-strict)).

We can help ensure developers working in this plugin do not make mistakes while writing snapshot tests by throwing an exception when unexpected properties (like `errors`) are included in the snapshot test cases.

Example test cases with unexpected properties:

```js
import {getTester} from './utils/test.mjs';

const {test} = getTester(import.meta);

test.snapshot({
	foo: 123,
	bar: 456,
	valid: [
        {code: 'abc', foo: 123}
    ],
	invalid: [
		{code: 'abc', foo: 123},
    ]
});
```

Example exception output:

```js
Uncaught exception in test/prefer-string-starts-ends-with.mjs

Error: Unexpected snapshot test properties: foo,bar
```

### DIFF
--- a/test/prefer-module.mjs
+++ b/test/prefer-module.mjs
@@ -61,12 +61,6 @@ test({
 
 // `__dirname` and `__filename`
 test.snapshot({
-	testerOptions: {
-		globals: {
-			__dirname: true,
-			__filename: true
-		}
-	},
 	valid: [
 		outdent`
 			const __filename = 1;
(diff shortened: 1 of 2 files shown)


test/prefer-module.mjs
```diff
@@ -61,12 +61,6 @@ test({
 
 // `__dirname` and `__filename`
 test.snapshot({
-	testerOptions: {
```
bmish, line 64: This isn't a real option in snapshot tests.

fisker (approved):
Good idea, thanks!

PR #1326, opened 2021-06-02
Add `require-post-message-target-origin` rule

### LINKED ISSUE #1307: Rule proposal: `require-post-message-target-origin`
When calling [`window.postMessage()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) without `targetOrigin` argument, browser won't throw any error, but message can't receive by any window. I remember the first time I got into this, really drive me crazy.

## Fail

```js
foo.postMessage('hello!');
```

## Pass

```js
foo.postMessage('hello!', '*');
```
```js
foo.postMessage('hello!', 'http://example.com');
```

fisker:
Fixes #1307


docs/rules/require-post-message-target-origin.md
```diff
@@ -0,0 +1,19 @@
+# Enforce using the targetOrigin argument with `Window.postMessage()`
```
sindresorhus, line 1:
```suggestion
# Enforce using the `targetOrigin` argument with `window.postMessage()`
```

fisker: MDN wrote `Window`, I wonder where is the `postMessage` method, it's not in `Window.prototype`

PR #1331, opened 2021-06-03
Add `prefer-at` rule

### LINKED ISSUE #983: Rule proposal: Prefer `.at()` method for negative index access
https://github.com/tc39/proposal-relative-indexing-method

~~We cannot do this one yet because ESLint only supports stage 4 features.~~

What should the rule name be? `prefer-at-method-for-negative-index-access` is the most correct, but also a mouthful.

## Fail

```js
array[array.length - 1];
array.slice(-1)[0];
```

## Pass

```js
array.at(-1);
```

fisker:
Fixes #983


docs/rules/prefer-at.md
```diff
@@ -0,0 +1,122 @@
+# Prefer `.at()` method for index accessing
```
sindresorhus, line 1:
```suggestion
# Prefer `.at()` method for index access
```

test/prefer-at.mjs
```diff
@@ -0,0 +1,202 @@
...
+		'array[0]',
+		'array[1]',
+		'array[5 + 9]',
+		'const offset = 5;array[offset + 9]',
+		'array[array.length - 1]',
+		// `charAt` don't care about value
+		'string.charAt(9)',
+		'string.charAt(5 + 9)',
+		'const offset = 5;string.charAt(offset + 9)',
+		'string.charAt(unknown)',
```
sindresorhus, line 195: Also `string.charAt()`

PR #1334, opened 2021-06-04
Refactor: Auto report problems

fisker:
I tried to make this change into ESLint core, but has been rejected https://github.com/eslint/eslint/issues/14661. I pick some rule to make the change.

---

The first rule change https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1334/commits/534297ff91bd988537ffba159d533f01d8bb3235,

The old one added `getProblem` function, only because we have to use

```js
context.report(problem); 
return;
```

And there are two places should report and return.

---

The second rule change https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1334/commits/3c5ad855bd7446bb2fa64ccacb4ff3429e2795af

I didn't use "early return" before for the same reason, and without "early return", some calculation has to do before the `if-else` block;

---

This rule change https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1334/commits/d0e0db2a6bee94f6ce5b657a20aa647472cccb00, only show needs for generator support.


fisker:
Another pain point, we've started to use generator function to fix code.

Before

```js
function fix(fixer) {
	const fixes = [];
	if (foo) {
		fixes.push(fixer.remove(node));
	}

	// Other fixes

	if (hasComments(node.arguments[0].object.callee)) {
		return [];  // <- very simple to abort fix.
	}

	return fixes;
}
````

Now

```js
function * fix(fixer) {
	if (foo) {
		yield fixer.remove(node);
	}

	// Other fixes

	if (hasComments(node.arguments[0].object.callee)) {
		// <- There is no way to abort fix here
	}
}
````

So I do fixable check before actual fix https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/no-array-for-each.js#L406-L407

In this way, the fix function relay on logic in `isFixable` check, otherwise some check need to do twice. 

After this PR, it's possible to do something like


```js
function * fix(fixer, abort) {
	return abort;
}
````

```js
function * fix(fixer, abort) {
	abort();
}
````

```js
function * fix(fixer, abortSignal) {
	abortSignal.abort();
}
````

I'm not sure about the API yet, but all possible.

**Comment by fisker** [comment]
@sindresorhus I'd like to hear your thoughts.

File: rules/no-array-for-each.js [hunk 1]
```diff
@@ -407,22 +407,23 @@ const create = context => {
...
-	meta: {
-		type: 'suggestion',
-		docs: {
-			description: 'Prefer `for…of` over `Array#forEach(…)`.',
-			url: getDocumentationUrl(__filename)
-		},
-		fixable: 'code',
-		schema: [],
-		messages
+module.exports = createRule(
```
bmish, line 416: It's unfortunate that this would prevent the [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) lint rules from linting the rules.

fisker: I understand that, but we can do things in `createRule` ourself. For example we don't need it enforce `schema`.

fisker: Or, if we move this call to `index.js` and `test/`, the plugin can still recognize the rule.

sindresorhus:
Seems like a good improvement 👍

I do agree it would be nice to retain support for the rule listing.

sindresorhus:
> I'm not sure about the API yet, but all possible.

Some other alternatives:

- Returns a special symbol.
- Throw a special error.

PR #1336, opened 2021-06-04
`filename-case`: Enforce lowercased file extension

### LINKED ISSUE #1268: Enforce extensions case  in `filename-case`
Enforce file extensions to be in lower case, especially on Windows, it files accidentally use an upper case extension, it's hard to find the problem.

This not really similar to the problems this rule currently catching, but I guess it fits.

fisker:
Fixes #1268


rules/filename-case.js
```diff
@@ -5,8 +5,10 @@ const getDocumentationUrl = require('./utils/get-documentation-url');
 const cartesianProductSamples = require('./utils/cartesian-product-samples');
 
 const MESSAGE_ID = 'filename-case';
+const MESSAGE_ID_EXTENSION = 'filename-extension';
 const messages = {
-	[MESSAGE_ID]: 'Filename is not in {{chosenCases}}. Rename it to {{renamedFilenames}}.'
+	[MESSAGE_ID]: 'Filename is not in {{chosenCases}}. Rename it to {{renamedFilenames}}.',
+	[MESSAGE_ID_EXTENSION]: 'File extension `{{extension}}` is not in lower-case. Rename it to `{{filename}}`.'
```
sindresorhus, line 11:
```suggestion
	[MESSAGE_ID_EXTENSION]: 'File extension `{{extension}}` is not in lowercase. Rename it to `{{filename}}`.'
```

Both are allowed, but I think `lowercase` is better.

PR #1340, opened 2021-06-05
Recommend using `unicorn/no-lonely-if` together with `no-lonely-if`

zloirock:
https://github.com/zloirock/core-js/commit/8acc71b00b2102eb49bc40c04f084b20c670bdf1#commitcomment-51752858


fisker (approved):
Thanks!

PR #1349, opened 2021-06-12
Use `meta.hasSuggestions` for suggestable rules to prepare for ESLint 8

bmish:
The change to require suggestable rules to have `meta.hasSuggestions` has been accepted and mentioned in the blog post for the upcoming ESLint 8 breaking changes. So we should adopt this change now to ensure we are compatible with ESLint 8 as soon as possible. The old property `meta.docs.suggestion` was unused anyway.

https://eslint.org/blog/2021/06/whats-coming-in-eslint-8.0.0#rules-with-suggestions-now-require-the-metahassuggestions-property

Follow-up to #1306 / #1292.

### DIFF
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -159,14 +159,14 @@ module.exports = {
 		type: 'suggestion',
 		docs: {
 			description: 'Use destructured variables over properties.',
-			url: getDocumentationUrl(__filename),
-			suggestion: true
+			url: getDocumentationUrl(__filename)
 		},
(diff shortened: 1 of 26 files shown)


fisker:
@bmish https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1349/checks?check_run_id=2810397188#step:9:4

bmish:
@fisker fixed, sorry accidentally forced pushed.

fisker (approved):
Thanks!

PR #1350, opened 2021-06-15
Simplify `catch-error-name` rule

fisker:


sindresorhus:
Can you rebase from the main branch?

PR #1352, opened 2021-06-15
Simplify `no-array-reduce` rule

fisker:


test/utils/not-function-types.mjs
```diff
@@ -33,7 +33,7 @@ const notFunctionTypes = [
 	'~ fn',
 	'typeof fn',
 	'void fn',
-	'delete fn',
+	// 'delete fn',
```
sindresorhus, line 36: Intentional?

fisker: Yah, it need non-strict mode to run test, no one uses `array.reduce(delete foo)`, not worth add a seperate test.

fisker: Added a comment

fisker: I can use delete property to test :)

PR #1357, opened 2021-06-16
Add `no-array-method-this-argument` rule

### LINKED ISSUE #974: Rule proposal: No `this` argument in array methods when using arrow functions
Prevent specifying the `this` argument to array methods when using an arrow function, since it has no effect.

Applies to the following instance methods:

```
map
filter
some
find
forEach
findIndex
every
```

And also `Array.from`.

Only `Array.from` can be safely auto-fixed. The others should use suggestions.

## Fail

```js
array.filter(x => x === 1, this);
```

## Pass

```js
array.filter(x => x === 1);
```

fisker:
Fixes #974


test/snapshots/no-array-method-this-argument.mjs.md
```diff
@@ -0,0 +1,379 @@
...
+    > 1 | array.map(callback, thisArgument)␊
+        |                     ^^^^^^^^^^^^ Do not use this argument in \`Array#map()\`.␊
+    ␊
+    --------------------------------------------------------------------------------␊
+    Suggestion 1/2: Remove the second argument.␊
+      1 | array.map(callback)␊
+    ␊
+    --------------------------------------------------------------------------------␊
+    Suggestion 2/2: Use bound function.␊
+      1 | array.map(callback.bind(thisArgument))␊
```
fisker, line 180: I didn't auto-fix this case to use `callback.bind`, since it's not the best option, people may won't get attention with auto-fix.

docs/rules/no-array-method-this-argument.md
```diff
@@ -0,0 +1,49 @@
+# Forbid using this argument in array methods
```
sindresorhus, line 1:
```suggestion
# Disallow using the `this` argument in array methods
```

docs/rules/no-array-method-this-argument.md
```diff
@@ -0,0 +1,49 @@
...
+
+Forbid using the `thisArg` argument in array methods:
```
sindresorhus, line 3:
```suggestion
The rule forbids using the `thisArg` argument in array methods:
```

rules/no-array-method-this-argument.js
```diff
@@ -0,0 +1,120 @@
...
+const {getParentheses, getParenthesizedText} = require('./utils/parentheses.js');
+const shouldAddParenthesesToMemberExpressionObject = require('./utils/should-add-parentheses-to-member-expression-object.js');
+
+const ERROR = 'error';
+const SUGGESTION_BIND = 'suggestion-bind';
+const SUGGESTION_REMOVE = 'suggestion-remove';
+const messages = {
+	[ERROR]: 'Do not use this argument in `Array#{{method}}()`.',
+	[SUGGESTION_REMOVE]: 'Remove the second argument.',
+	[SUGGESTION_BIND]: 'Use bound function.'
```
sindresorhus, line 14:
```suggestion
	[SUGGESTION_BIND]: 'Use a bound function.'
```

sindresorhus (changes requested):

PR #1360, opened 2021-06-17
`explicit-length-check`: Support Vue SFC files

fisker:
Can we accept this?

### DIFF
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -45,7 +45,7 @@ const isEmptySet = !foo.size;
 
 ```vue
 <template>
-	<div v-if="foo.length"></div>
+	<div v-if="foo.length">Vue</div>
 </template>
 ```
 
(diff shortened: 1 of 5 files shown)


sindresorhus:
Sure!

fisker:
Do you think we should make all rule work for Vue SFC? We can add one line [here](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/91996110a45ecb35c0c14f98dae90d88b904e863/rules/utils/load-rule.js#L34).

fisker:
Let's start with some other rules first, I'll send more PRs

sindresorhus:
> Do you think we should make all rule work for Vue SFC?

I don't mind. Vue is quite popular, so it's gonna help a lot of people.

fisker:
I'm worried about auto-fix for some rule, for example `prefer-string-replace-all`, we'll fix `<div v-if='foo.replace(/a/g, "")'` to `<div v-if='foo.replaceAll('a', "")'`, but we can't use single quote for the fix, it's in single quoted attribute, so let's start with some safe rules first.

PR #1363, opened 2021-06-17
`explicit-length-check`: Support `v-show` directive

fisker:
I forgot that `v-show` is also a boolean directive.


fisker:
Funny, `v-Show="fasle"` is not the same as `v-show="fasle"`, component doesn't hide. But in AST, the have the same name.

PR #1375, opened 2021-06-22
`expiring-todo-comments`: More helpful error when no conditions

### LINKED ISSUE #744: `expiring-todo-comments` - more specific error message for missing conditions
Rather than falling back to `no-warning-comments` when "useless" todos are found, could the rule explicitly report something like `TODO comment should have a condition or due date`?

When I first saw `Unexpected "todo" comment` I thought "oh, this is a rule which aggressively says 'todos are bad'", and I turned it off. Later, I spent a bunch of time implementing a worse version of `expiring-todo-comments`, which parsed ISO dates. Eventually I stumbled across the docs and realised how useful this rule really is.

Basically the problem is that it's piggy-backing off eslint's error message, which is too broad and doesn't really nudge the user into fixing the problem - and risks pushing them into deleting the todo or turning off the rule.

mmkal:
Fixes #744

This overrides the `no-warning-comments` default message for `TODO`s without any conditions, when `allowWarningComments` is set to false, to give a hint that the `TODO` would be allowed if it had a condition.

When I first saw `Unexpected "todo" comment` I thought "oh, this is a rule which aggressively says 'todos are bad'", and I turned it off. (Later, I spent a bunch of time implementing a worse version of `expiring-todo-comments`, which parsed ISO dates. Eventually I stumbled across the docs and realised how useful this rule really is.)

This change reduces the risk of the above scenario.

### DIFF
--- a/docs/rules/expiring-todo-comments.md
+++ b/docs/rules/expiring-todo-comments.md
@@ -141,7 +141,7 @@ You can also use block comments to specify TODOs with conditions. Each line can 
 
 This rule implements [`eslint/no-warning-comments`](https://eslint.org/docs/rules/no-warning-comments).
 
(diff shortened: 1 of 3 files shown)


sindresorhus (approved):

fisker (approved):
Thank you!

PR #1386, opened 2021-06-30
`no-array-method-this-argument`: Fix false positives

### LINKED ISSUE #1385: [no-array-method-this-argument] - false positives, lodash map, foreach, filter and React.Chidlren.map
```js
          const files = _.filter(data, (file) => _.endsWith(file, '.pdf')); //  Do not use the `this` argument in `Array#filter()`  unicorn/no-array-method-this-argument
        
          {React.Children.map(children, (child) => // Do not use the `this` argument in `Array#map()`  unicorn/no-array-method-this-argument
            React.cloneElement(child, { className: classes.nested })
          )}
          
          
```

fisker:
- Ignore methods from `lodash`, `underscore`, `jQuery`, and `React`
- Ignore cases that first argument is not a function

Fixes #1385


fisker:
Why don't you send a PR, add them to [the list](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/93b34793d470e40b5f53eb08854e7f50795e15d5/rules/no-array-method-this-argument.js#L18), I think we can accept it.

PR #1388, opened 2021-06-30
Add `prefer-nullish-coalescing` rule

### LINKED ISSUE #471: Rule Proposal: prefer-nullish-coalescing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

https://github.com/tc39/proposal-nullish-coalescing

```js
let foo = (undefined || null) || bar;

// equals

let foo = (undefined || null) ?? bar;
```

anyway, it not safe for `foo || bar` -> `foo ?? bar`, we should [use ESLint suggestion api](#441)

fisker:
Fixes #471


fisker:
After excluding many possible boolean values, there still false positive, because we can't know the `||` left side value type

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-922f3a7977dfb855fdc9b3a7ab921bf1de948d8b27ac6e2c8286b089db47b899L255

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-929e56747d8ebca03aa53d228a84d03e8d3ba3cf2c63adafaa9dc0eac82b471eR97

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-1cfe9745542244677eac6e608d3b38f06600a40c3f092c1535e220114dabcfe3R137

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-4f3314711d30c02b6d078f0a04c3fe5fd455eb17d68185c4a9f4b7a792d4a25eR40

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-ffbf303eac24990fbfa25f4240a4e5f8c88f86e4cc15aa4d45207b323db554f3R36

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-d40e13bf614e2b95c6ee4e14a2ea3d2aa286efec3da4f87dd39947a1b294ab63R29

https://github.com/fisker/eslint-plugin-unicorn/pull/296/files#diff-4057a0454b81be8653d171ac0c8bcd6b183f35d8c233dd81208827b387dd5a54R11

fregante:
Is this even safely possible without types?

fisker:
> Is this even safely possible without types?

It's not safe, as you can see the changes, I was thinking maybe we can set this rule to `warn` by default, there is no harm without fix.

There are same concerns in the ESLint issue https://github.com/eslint/eslint/issues/14037, and I don't think they will accept it.

I've already used this rule at my work with `warn`, this reminds me to use `??` every time when I use `||`, and it did help.

fregante:
Unless the user disables warnings, warnings are still noisy enough that makes people wary of using `||` at all. Correct, common code should not cause warnings IMHO

fisker:
That's all I got in mind.

docs/rules/prefer-nullish-coalescing.md
```diff
@@ -0,0 +1,27 @@
+# Prefer the nullish coalescing operator(`??`) over the logical OR operator(`||`)
+
```
bmish, line 2: There are some new requirements for rule docs, be sure to rebase on the latest `main` to pick them up.

rules/prefer-nullish-coalescing.js
```diff
@@ -0,0 +1,79 @@
...
+						data: messageData,
+						fix: fixer => fixer.replaceText(operatorToken, replacementOperator)
+					}
+				]
+			};
+		}
+	};
+};
+
+module.exports = {
```
bmish, line 69: We'll want to add a `type` annotation here (more info in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1606):

```suggestion
/** @type {import('eslint').Rule.RuleModule} */
module.exports = {
```

sindresorhus:
I think this rule would be useful as something I would run once-in-a-while to catch cases where I forgot to use `??`, but I agree, even warnings can be annoying when there are a lot of false-positives. So I think this rule should be off by default.

Alternatively, it would use type information when available: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/347

sindresorhus:
Closing for lack of activity.

Issue #1250, opened 2021-05-09
Rule proposal: Array#push + rest parameter instead of Arrray#concat


fregante:
Just a minor suggestion, not sure if it's actually any better or preferred. Maybe:

- `array` can be a `const`
- no repetition
- shorter

## Fail

```js
array = array.concat(otherArray);
```

## Pass

```js
array.push(...otherArray);
```

sindresorhus:
We already fix `array = array.concat(otherArray)` to `array = [...array, ...otherArray]` in the `prefer-spread` rule.

fregante:
I suppose my “real-life example” wasn’t caught because it was incorrect. But I still think the .push version is better in every way (shorter, no repetition, less looping via spread, no creation of another array in memory)

fregante:
I suppose the difference would be that with `.push`, `array` would be modified, so if it's a parameter function it's not a safe fix:

```js
function spread(array) {
	const otherArray = [1, 2, 3];
	array = [...array, ...otherArray];
	return array;
}
const initial = [0];
const newOne = spread(initial);

// initial => [0]
// newOne => [0, 1, 2, 3]
```


```js
function push(array) {
	const otherArray = [1, 2, 3];
	array.push(…otherArray);
	return array;
}
const initial = [0];
const newOne = append(initial);

// initial => [0, 1, 2, 3]
// newOne => [0, 1, 2, 3]
```

papb:
I've heard that JavaScript can pass at most 30 thousand arguments to a function, could `.push(...foo)` (or any other function call with the spread operator) hit this limitation as well if the spreaded array has length above 30000?

fisker:
```js
[].push(...Array.from({length: 3e6}))

Uncaught RangeError: Maximum call stack size exceeded
    at <anonymous>:1:4
```


```js
[...Array.from({length: 3e6})] // can run
```

fisker:
Check this https://github.com/prettier/prettier/pull/7587#discussion_r379907099

fisker:
https://dev.to/uilicious/javascript-array-push-is-945x-faster-than-array-concat-1oki

fregante:
> I've heard that JavaScript can pass at most 30 thousand arguments to a function

Could this be part of the error message, without an autofix?

```
Prefer array.push to .concat if the array is smaller than 5000 items // for safety
```

sindresorhus:
I don't think we should do this

[...]

PR #1392, opened 2021-06-30
[docs] Add call stack size benefit to `no-array-for-each`

bartveneman:
Large arrays will eventually run into `Maximum call stack size exceeded` errors, whereas a `for .. of` will not, since it doesn't add a function call to the stack for every iteration.

### DIFF
--- a/docs/rules/no-array-for-each.md
+++ b/docs/rules/no-array-for-each.md
@@ -5,6 +5,7 @@ Benefits of [`for…of` statement](https://developer.mozilla.org/en-US/docs/Web/
 - Faster
 - Better readability
 - Ability to exit early with `break` or `return`
+- Less likely to run into `Maximum call stack size exceeded` errors
 
 This rule is partly fixable.
 


fisker:
I don't think it's true...

sindresorhus:
Source?

bartveneman:
Ran into [that exact error](https://github.com/lukeed/uvu/pull/117) with uvu and simply replacing the `.forEach()` or `.map()` fixed the issue.

bartveneman:
Created [a REPL](https://replit.com/@bartveneman/CallStackExceededLoop#index.js) to demonstrate the issue and found out that I was wrong, so my apologies. 😞 
Apparently spreading 15000 items in `Math.min()` is too much for Node to handle. Again, my apologies and thanks for the quick responses.

fisker:
Same thing here https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1250#issuecomment-842812531

PR #1398, opened 2021-07-02
Add `no-useless-length-check` rule

### LINKED ISSUE #897: Rule proposal: `no-unnecessary-length-check-before-some-every`
I havn't think through the details, but cases:

## Fail

```js
if (array.length === 0 || array.every(Boolean)) {}
```

```js
if (array.length !== 0 && array.some(Boolean)) {}
```

## Pass

```js
if (array.every(Boolean)) {}
```

```js
if (array.some(Boolean)) {}
```

```js
if (array.length !== 0 && array.every(Boolean)) {}
```

```js
if (array.length === 0 || array.some(Boolean)) {}
```

fisker:
~Not finished yet, nested logic operators are complicated.~

Fixes #897


rules/no-useless-length-check.js
```diff
@@ -0,0 +1,148 @@
+'use strict';
+const {methodCallSelector, matches, memberExpressionSelector} = require('./selectors/index.js');
+const isSameReference = require('./utils/is-same-reference.js');
+const {getParenthesizedRange} = require('./utils/parentheses.js');
+
+const messages = {
+	'non-zero': '`Array#some()` returns `false` on empty array, the empty check is not needed.',
+	zero: '`Array#every()` returns `true` on empty array, the non-empty check is not needed.'
```
sindresorhus, line 8:
```suggestion
	'non-zero': 'The empty check is useless as `Array#some()` returns `false` for an empty array.',
	zero: 'The non-empty check is useless as `Array#every()` returns `true` for an empty array.'
```

?

sindresorhus (changes requested):

rules/no-useless-length-check.js
fisker: The message seems not right, I'll double-check the logic.

fisker: The message was wrong about the length check node, fixed.

PR #1399, opened 2021-07-03
Refactor `avoidCapture` function

fisker:
Reason behind this change, the new version of ESLint allows [`ecmaVersion: "latest"`](https://github.com/eslint/eslint/pull/14720), and this part don't understand it.

Let's not care about `ecmaVersion` and strict mode, just use the safest strategy.

WDYT? @futpib

### DIFF
--- a/rules/utils/avoid-capture.js
+++ b/rules/utils/avoid-capture.js
@@ -6,13 +6,78 @@ const {
 }= require("@babel/helper-validator-identifier");
 const resolveVariableName = require('./resolve-variable-name.js');
 
+// https://github.com/microsoft/TypeScript/issues/2536#issuecomment-87194347
+const typescriptReservedWords = new Set([
+	'break',
+	'case',
+	'catch',
+	'class',
+	'const',
+	'continue',
+	'debugger',
+	'default',
+	'delete',
+	'do',
+	'else',
+	'enum',
+	'export',
+	'extends',
+	'false',
+	'finally',
+	'for',
+	'function',
+	'if',
+	'import',
+	'in',
+	'instanceof',
+	'new',
+	'null',
+	'return',
+	'super',
+	'switch',
+	'this',
+	'throw',
+	'true',
+	'try',
+	'typeof',
+	'var',
+	'void',
+	'while',
+	'with',
+	'as',
+	'implements',
+	'interface',
+	'let',
+	'package',
+	'private',
+	'protected',
+	'public',
+	'static',
+	'yield',
+	'any',
+	'boolean',
+	'constructor',
+	'declare',
+	'get',
+	'module',
+	'require',
+	'number',
+	'set',
+	'string',
+	'symbol',
+	'type',
+	'from',
+	'of'
+]);
+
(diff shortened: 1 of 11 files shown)


fisker:
On second thought, it's not safe anyway, running ESLint with specific ecma version, doesn't mean scripts always run on the same environment. I always use my shared config with `ecmaVersion: 2021`, but my code can run on old engine.

rules/utils/avoid-capture.js
```diff
@@ -1,34 +1,7 @@
 'use strict';
-const reservedWords = require('reserved-words');
+const {isValidES3Identifier} = require('@babel/types');
```
sindresorhus, line 2: I don't think we should support ES3 kinda identifiers. The minimum should be ES5. Unless, I'm misunderstanding this.

fisker: Here is the code, https://github.com/babel/babel/blob/main/packages/babel-types/src/validators/isValidES3Identifier.ts#L36, I don't think you want allow `public` `static` use as a variable.

sindresorhus: > I don't think you want allow public static use as a variable.

Why not?

futpib:
My implementation tried to match the target ecma version closely with regards to reserved words and allowed `arguments` as a variable name where language allows it without shadowing. This rewrite bans reserved words ignoring ecma version and bans `arguments` name ignoring complex rules around it.

If original approach is not a valuable enough feature, this PR provides a more practical approach that will be easier to maintain.

rules/utils/avoid-capture.js
fisker: I think there is noting wrong to be safer, also people may use `typescript` parser to lint js file. So maybe I need to check if there are extra reserved words for typescript since we are also not  (and can't ) checking parsers.

But if you insist , I can switch to use `isValidIdentifier`.

sindresorhus: Why not support ES1 reserved keywords then? Safe is good, but you have to draw the line somewhere. I think ES5 is a good line. And we could add TS reserved keywords too.

fisker:
I also found potential threat, for example we used it in `catch-error-name`, if some bad shared config, config it as 

```js
"unicorn/catch-error-name": [
	"error",
	{
		"name": "){} evilCode; if(false"
	}
]
```

```js
try {} catch (e) {}
```

will fix to 

```js
try {} catch (){} evilCode; if(false) {}
```

And it may run into infinity loop in all used rules,

for example: 

```js
	"error",
	{
		"replacements":  {
			"a" : {
				123: true
			}
		}
	}
]
```

current implementation will not find a usable variable name and hang forever, I'll fix.

sindresorhus:
> I also found potential threat, for example we used it in catch-error-name, if some bad shared config, config it as

I wouldn't consider that a vulnerability. It applies to a lot of ESLint rules. Rules make no guarantee about being resistant to evil config, nor should they.

fisker:
But the infinity one, people may config it by mistake, for example, added a space after the replacement.

sindresorhus:
The infinity case I agree we should fix.

PR #1402, opened 2021-07-06
`prevent-abbreviations`: Add `def`

dangowans:
Just added eslint-plugin-unicorn to my project, and while straightening out my code, I noticed I used "def" as an abbreviation.  Not sure if it means "definition" to everyone else, but it does to me, : )

### DIFF
--- a/rules/shared/abbreviations.js
+++ b/rules/shared/abbreviations.js
@@ -41,6 +41,9 @@ module.exports.defaultReplacements = {
 		database: true
 	},
 	def: {
+		defer: true,
+		deferred: true,
+		define: true,
 		definition: true
 	},
 	dest: {
--- a/rules/shared/abbreviations.js
+++ b/rules/shared/abbreviations.js
@@ -40,6 +40,9 @@ module.exports.defaultReplacements = {
 	db: {
 		database: true
 	},
+	def: {
+		definition: true
+	},
 	dest: {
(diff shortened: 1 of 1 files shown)


fisker:
I think it could be [`jQuery.Deferred`](https://github.com/search?l=JavaScript&q=jquery+def+Deferred&type=Code) too.

fisker:
Maybe also `define()`?

dangowans:
I could see `define`.  Maybe less on `defer` or `deferred`, but that's just me.

I can update the pull request with the others, or is there kind of a limit like, if there's three or more possibilities, the abbreviation shouldn't be included because it's no longer useful?

fisker:
We can add them all, still useful, just can't do auto fix.

fisker (approved):

dangowans:
On a side note, I see there is `el` and `elem` for `element`, but not `ele`.  Not sure that's on purpose, but I can add it as well in this or another pull request.

sindresorhus:
> On a side note, I see there is el and elem for element, but not ele. Not sure that's on purpose, but I can add it as well in this or another pull request.

I have never seen `ele`, personally.

dangowans:
> I have never seen `ele`, personally.

No worries.  It's all over my codebase, but it's easy enough to add my own replacement. : )

PR #1415, opened 2021-07-13
Disable `require-post-message-target-origin` for TypeScript files

### LINKED ISSUE #1396: `require-post-message-target-origin` shouldn't trigger for incompatible APIs
## What's wrong 
I'm currently developing a browser extension that communicates with native software:

```ts
import { browser, Runtime } from 'webextension-polyfill-ts'
const nativePort: Runtime.Port = browser.runtime.connectNative(APPLICATION_ID)
nativePort.postMessage({ cmd: 'Version' }) // (x) Missing the `targetOrigin` argument.
```

## Which rule is buggy
`require-post-message-target-origin` is triggered but the method only accepts one argument (see [runtime.Port](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/Port)).

## Ideas 

1. Enforce the name of the variable to be `window`, `document` or the likes.
2. Check the type of the variable or the existence of the import `'webextension-polyfill'`/`'webextension-polyfill-ts'`.
3. Add a warning for web extension developers here: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v34.0.1/docs/rules/require-post-message-target-origin.md

JounQin:
close #1396

### DIFF
--- a/index.js
+++ b/index.js
@@ -135,7 +135,7 @@ module.exports = {
 				{
 					files: ['*.ts', '*.tsx'],
 					rules: {
-						'unicorn/require-post-message-target-origin': 0,
+						'unicorn/require-post-message-target-origin': 'off',
 					},
 				},
 			],
--- a/index.js
+++ b/index.js
@@ -34,7 +34,9 @@ module.exports = {
 				ecmaVersion: 2021,
 				sourceType: 'module',
 			},
-			plugins: ['unicorn'],
+			plugins: [
+				'unicorn',
+			],
 			rules: {
 				'unicorn/better-regex': 'error',
 				'unicorn/catch-error-name': 'error',
--- a/index.js
+++ b/index.js
@@ -34,9 +34,7 @@ module.exports = {
 				ecmaVersion: 2021,
 				sourceType: 'module',
 			},
-			plugins: [
-				'unicorn',
-			],
+			plugins: ['unicorn'],
 			rules: {
 				'unicorn/better-regex': 'error',
 				'unicorn/catch-error-name': 'error',
@@ -131,6 +129,14 @@ module.exports = {
 				'unicorn/string-content': 'off',
 				'unicorn/throw-new-error': 'error',
 			},
+			overrides: [
+				{
+					files: ['*.ts', '*.tsx'],
+					rules: {
+						'unicorn/require-post-message-target-origin': 0,
+					},
+				},
+			],
 		},
 	},
 };
--- a/package.json
+++ b/package.json
@@ -102,6 +102,7 @@
 			"eslint-remote-tester-results"
 		],
 		"rules": {
+			"quotes": ["error", "single"],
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -34,9 +34,7 @@ module.exports = {
 				ecmaVersion: 2021,
 				sourceType: 'module',
 			},
-			plugins: [
-				'unicorn',
-			],
+			plugins: ['unicorn'],
```
JounQin, line 37: It is fixed by `yarn xo --fix`.

package.json
```diff
@@ -102,6 +102,7 @@
 			"eslint-remote-tester-results"
 		],
 		"rules": {
+			"quotes": ["error", "single"],
```
JounQin, line 105: It seems the default value is `double`.

fisker:
Let's skip it in code, like this, https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c193e1e84a077e1f1cd8df19ab64730d038ee5ab/rules/prefer-module.js#L217

index.js
fisker: I don't think `xo` do this, you must enabled prettier or something in your editor.

JounQin:
Users can still enable `require-post-message-target-origin` for `ts` files if they want.

Personally, I don't think hard-coded extensions in `rule` is good. (`.cjs` for `prefer-module` is an exception).

JounQin:
I have no idea why `readme.md` has diff on CI.

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1415/checks?check_run_id=3055591613#step:9:4

fisker:
Because we use this recommended config to generate readme

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c193e1e84a077e1f1cd8df19ab64730d038ee5ab/scripts/generate-usage-example.mjs#L27

So overrides should be in it, this is also why I suggest to use `getPhysicalFilename`

JounQin:
Then I'd prefer regenerating `readme.md` in this PR. @sindresorhus Any thoughts?

JounQin:
@fisker Besides, `getPhysicalFilename` seems not good for `prefer-module` too.

We can have a virtual filename `test.md/0.cjs` in a Markdown code block, why it should be reported error?

fisker:
Good point. 👍

index.js
```diff
@@ -131,6 +131,14 @@ module.exports = {
 				'unicorn/string-content': 'off',
 				'unicorn/throw-new-error': 'error',
 			},
+			overrides: [
+				{
+					files: ['*.ts', '*.tsx'],
+					rules: {
+						'unicorn/require-post-message-target-origin': 0,
```
sindresorhus, line 138:
```suggestion
						'unicorn/require-post-message-target-origin': 'off',
```

Same with the readme.

PR #1418, opened 2021-07-13
Add `allowSimpleOperations` option to `no-array-reduce` rule

### LINKED ISSUE #1348: Ignore simple sum in `no-array-reduce`
I still think 

```js
foo.reduce((total, element) => total + element.bar, 0);
```

should be preferred over

```js
let total = 0;
for (const element of foo) {
	total += element.bar;
}
```

Especially, where `reduce` is not an expression, like this case

```js
if (foo.reduce((total, element) => total + element.bar, 0) > 100) {};
```

Or at least allow to ignore it.

manovotny:
This is my first time contributing something to this project besides from minor documentation tweaks, so don't feel bad telling me I did something incorrectly. 

I took a stab at the open issue to allow simple sums when using `reduce`. I am not married to the option name, so please offer better suggestions if you have them. 

Fixes #1348

### DIFF
--- a/docs/rules/no-array-reduce.md
+++ b/docs/rules/no-array-reduce.md
@@ -65,7 +65,7 @@ Set it to `false` to disable reduce completely.
 
 ```js
 // eslint unicorn/no-array-reduce: ["error", {"allowSimpleOperations": true}]
-arr.reduce((total, item) => total + item) // Passes
+array.reduce((total, item) => total + item) // Passes
 ```
 
 ```js
--- a/docs/rules/no-array-reduce.md
+++ b/docs/rules/no-array-reduce.md
@@ -1,6 +1,8 @@
 # Disallow `Array#reduce()` and `Array#reduceRight()`
 
-`Array#reduce()` and `Array#reduceRight()` usually result in [hard-to-read](https://twitter.com/jaffathecake/status/1213077702300852224) and [less performant](https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern) code. In almost every case, it can be replaced by `.map`, `.filter`, or a `for-of` loop. It's only somewhat useful in the rare case of summing up numbers.
+`Array#reduce()` and `Array#reduceRight()` usually result in [hard-to-read](https://twitter.com/jaffathecake/status/1213077702300852224) and [less performant](https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern) code. In almost every case, it can be replaced by `.map`, `.filter`, or a `for-of` loop.
+
+It's only somewhat useful in the rare case of summing up numbers, which is allowed by default.
 
 Use `eslint-disable` comment if you really need to use it.
 
@@ -40,6 +42,10 @@ array.reduce(reducer, initialValue);
 ```
 
 ```js
+array.reduce((total, value) => total + value, 0);
+```
+
+```js
 let result = initialValue;
 
 for (const element of array) {
@@ -51,9 +57,11 @@ for (const element of array) {
 ### allowNumericInitialValue
 
 Type: `boolean`\
-Default: `false`
+Default: `true`
 
-Can be enabled to allow `reduce` to be used to sum numbers.
+Allow a numeric `initialValue` in a `reduce` call.
+
+Pass `"allowNumericInitialValue": false` to disable reduce completely.
 
 ```js
 // eslint unicorn/no-array-reduce: ["error", {"allowNumericInitialValue": true}]
--- a/docs/rules/no-array-reduce.md
+++ b/docs/rules/no-array-reduce.md
@@ -61,7 +61,7 @@ Default: `true`
 
 Allow simple operations (like addition, subtraction, etc.) in a `reduce` call.
 
-Pass `"allowSimpleOperations": false` to disable reduce completely.
+Set it to `false` to disable reduce completely.
 
 ```js
 // eslint unicorn/no-array-reduce: ["error", {"allowSimpleOperations": true}]
--- a/docs/rules/no-array-reduce.md
+++ b/docs/rules/no-array-reduce.md
@@ -46,3 +46,21 @@ for (const element of array) {
 	result += element;
 }
 ```
+## Options
+
+### allowNumericInitialValue
+
+Type: `boolean`\
+Default: `false`
+
+Can be enabled to allow `reduce` to be used to sum numbers.
+
+```js
+// eslint unicorn/no-array-reduce: ["error", {"allowNumericInitialValue": true}]
+arr.reduce((total, item) => total + item, 0) // Passes
+```
+
+```js
+// eslint unicorn/no-array-reduce: ["error", {"allowNumericInitialValue": false}]
+arr.reduce((total, item) => total + item, 0) // Fails
+```
--- a/docs/rules/no-array-reduce.md
+++ b/docs/rules/no-array-reduce.md
@@ -54,21 +54,21 @@ for (const element of array) {
 ```
 ## Options
 
-### allowNumericInitialValue
+### allowSimpleOperations
 
 Type: `boolean`\
 Default: `true`
 
-Allow a numeric `initialValue` in a `reduce` call.
+Allow simple operations (like addition, subtraction, etc.) in a `reduce` call.
 
-Pass `"allowNumericInitialValue": false` to disable reduce completely.
+Pass `"allowSimpleOperations": false` to disable reduce completely.
 
 ```js
-// eslint unicorn/no-array-reduce: ["error", {"allowNumericInitialValue": true}]
-arr.reduce((total, item) => total + item, 0) // Passes
+// eslint unicorn/no-array-reduce: ["error", {"allowSimpleOperations": true}]
+arr.reduce((total, item) => total + item) // Passes
 ```
 
 ```js
-// eslint unicorn/no-array-reduce: ["error", {"allowNumericInitialValue": false}]
-arr.reduce((total, item) => total + item, 0) // Fails
+// eslint unicorn/no-array-reduce: ["error", {"allowSimpleOperations": false}]
+arr.reduce((total, item) => total + item) // Fails
 ```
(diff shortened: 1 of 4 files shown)


sindresorhus:
I think it should be `true` by default. This is the one case where "reduce" 100% makes sense.

manovotny:
@sindresorhus I am totally open to that and can start whipping up changes. I figured it was best to start it off as a non-breaking change until a conversation could be had.

docs/rules/no-array-reduce.md
```diff
@@ -46,3 +46,21 @@ for (const element of array) {
...
 }
 ```
+## Options
+
+### allowNumericInitialValue
+
+Type: `boolean`\
+Default: `false`
+
+Can be enabled to allow `reduce` to be used to sum numbers.
```
sindresorhus, line 56: I think the description should first describe the option and then afterwards what it enables (the above).

manovotny:
@sindresorhus with the flipping of default to `true` it uncovered a use case that I didn't have right originally and could use your advice of how you want to handle it.

It's this from [MDN's reduce documentation on parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#parameters) (emphasis added):

> A value to use as the first argument to the first call of the `callbackFn`. _**If no `initialValue` is supplied, the first element in the array will be used as the initial `accumulator` value and skipped as `currentValue`**_. 

For example:

```js
const nums = [1, 2, 3, 4, 5];
console.log(nums.reduce((total, item) => total + item, 0)); // -> 15
console.log(nums.reduce((total, item) => total + item)); // -> 15
```

I am unsure of how to check the assumed values that are being iterated on without the ability to check `initialValue`. 

Any great ideas? Or should the rule only work / pass when `initialValue` is present, thus forcing people to use an `initialValue` to get started?

sindresorhus:
> Or should the rule only work / pass when initialValue is present, thus forcing people to use an initialValue to get started?

I think an initial value helps readability, so I agree we can just enforce that if they want it to pass.

manovotny:
@sindresorhus I've changed the default value to `true` and updated the documentation. Please let me know if there are any other changes you'd like me to make.

fisker:
Do you think the following cases are simply enough to ignore?

```js
const getValueByPath = (object, path) =>
	path.split('.').reduce((value,key) => value[key], object);
```

```js
const queue = steps =>
	steps.reduce((promise, step) => promise.then(step), Promise.resolve());
```

sindresorhus:
> Do you think the following cases are simple enough to ignore?

IMHO, no.

fisker:
I think we should also check the callback function body, only allow simple `+` or `*` (maybe `**` too?). And let's remove this option, just allow these cases.  WDYT?

fisker:
We can also change these two lines in our codebase back

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/704/files#diff-2c0df1d17a5ab16085062d0ff80023998f8e23ef6d48af8bc8bb14a35e569dadL4

And

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/704/files#diff-962f4629ce8856ba31836e3e55a7b5f5b2d6ad52cc7fe5dcdeecb05d025b3ba4L32

sindresorhus:
> only allow simple + or * (maybe ** too?)

Yeah, and `-` and `/`.

manovotny:
@sindresorhus @fisker both your comments make sense and seem doable. I was in all day meetings today and will be again tomorrow. I should be able to circle back to this on Friday or into the weekend. Thanks for the feedback!

manovotny:
I've updated the PR to allow a direct `BinaryExpression` as the body of the function. That way we shouldn't need to check for specific operators. Let me know if I am not thinking of something by allowing this. 

I also renamed the lint rule option to `allowSimpleOperations`. I am not completely happy with it, but it was the best I could think of. Completely open to suggestions.

And @fisker, I tried changing the [lines of code from your comment](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1418#issuecomment-879821447) back to using `reduce`, but then `npm test` failed. I think it's because `xo` needs to update to a newer release of `eslint-plugin-unicorn` once this rule is merged and updated in `xo` for it to work. Let me know if I am mistaken in that thinking.

manovotny:
@sindresorhus @fisker do either of you have any feedback you'd like me to address?

rules/no-array-reduce.js
```diff
@@ -34,14 +36,50 @@ const selector = matches([
...
+				return problem;
+			}
+
+			if (callback.type === 'ArrowFunctionExpression' && callback.body.type === 'BinaryExpression') {
+				return;
+			}
+
+			if ((callback.type === 'ArrowFunctionExpression' || callback.type === 'FunctionExpression') &&
+				callback.body.type === 'BlockStatement' &&
+				callback.body.body[0].type === 'ReturnStatement' &&
```
fisker, line 73:
```suggestion
				callback.body.body.length === 1 &&
				callback.body.body[0].type === 'ReturnStatement' &&
```

I think we should only allow one `return`

```js
foo.reduce(function (a, b) {
	return a + doingComplicatedThings();

	function doingComplicatedThings() {
       // ...
	}
});
```

sindresorhus: Make sure to add a test for this too.

manovotny: Excellent idea / catch! Length check added along with corresponding test.

sindresorhus:
Can you fix the lint violation?

docs/rules/no-array-reduce.md
```diff
@@ -39,10 +41,34 @@ Array.prototype.reduce.call(array, reducer);
 array.reduce(reducer, initialValue);
 ```
 
+```js
+array.reduce((total, value) => total + value;
```
sindresorhus, line 45: This is invalid code.

docs/rules/no-array-reduce.md
```diff
@@ -39,10 +41,34 @@ Array.prototype.reduce.call(array, reducer);
...
+## Options
+
+### allowSimpleOperations
+
+Type: `boolean`\
+Default: `true`
+
+Allow simple operations (like addition, subtraction, etc.) in a `reduce` call.
+
+Pass `"allowSimpleOperations": false` to disable reduce completely.
```
sindresorhus, line 64:
```suggestion
Set it to `false` to disable reduce completely.
```

docs/rules/no-array-reduce.md
```diff
@@ -39,10 +41,34 @@ Array.prototype.reduce.call(array, reducer);
...
+
+```js
+// eslint unicorn/no-array-reduce: ["error", {"allowSimpleOperations": true}]
+arr.reduce((total, item) => total + item) // Passes
```
sindresorhus, line 68:
```suggestion
array.reduce((total, item) => total + item) // Passes
```

docs/rules/no-array-reduce.md
```diff
@@ -39,10 +41,34 @@ Array.prototype.reduce.call(array, reducer);
...
+```
+
+```js
+// eslint unicorn/no-array-reduce: ["error", {"allowSimpleOperations": false}]
+arr.reduce((total, item) => total + item) // Fails
```
sindresorhus, line 73:
```suggestion
array.reduce((total, item) => total + item) // Fails
```

rules/utils/numeric.js
```diff
@@ -8,7 +8,7 @@ const isDecimalIntegerNode = node => isNumber(node) && isDecimalInteger(node.raw
 
 const isNumber = node => typeof node.value === 'number';
 const isBigInt = node => Boolean(node.bigint);
-const isNumeric = node => isNumber(node) || isBigInt(node);
+const isNumeric = node => node ? isNumber(node) || isBigInt(node) : false;
```
sindresorhus, line 11: Wrap the middle part in parens for readability.

test/no-array-reduce.mjs
```diff
@@ -93,12 +93,21 @@ test({
 		// Second argument is not a function
 		...notFunctionTypes.map(data => `Array.prototype.reduce.call(foo, ${data})`),
 
-	].flatMap(code => [code, code.replace('reduce', 'reduceRight')]),
-	invalid: [
+		// Option: allowSimpleOperations
 		'arr.reduce((total, item) => total + item)',
```
sindresorhus, line 97:
```suggestion
		'array.reduce((total, item) => total + item)',
```

(I know the existing code uses `arr`, but better to use something readable going forward)

manovotny: You want to keep the diff clean and just update the new lines added or would you like me to make the rest more readable as well?

docs/rules/no-array-reduce.md
manovotny: Addressed in this [commit](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1418/commits/eb8b2260c5bf0fc0924d1ebce81ab50f1ce4e2bf).

rules/utils/numeric.js
manovotny: Addressed in this [commit](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1418/commits/960ed280fcc23a1aaac0cc8c4f1e65a416241fda).

manovotny:
> Can you fix the lint violation?

Unless I am misunderstanding the failure, I believe it's coming from `prevent-abbreviations.js`. Perhaps an oversight of #1445?

I can address it here if you'd like or in a separate PR. Your call. Either way. It take a little more work (not a ton) because it's a dynamically generated message.

manovotny:
@sindresorhus aside from the lint issue and [my comment above]() regarding the issue, I believe everything else on the PR has been addressed.

test/no-array-reduce.mjs
sindresorhus: Make the rest of them readable too.

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1418#discussion_r680418508

test/no-array-reduce.mjs
manovotny: Addressed in this [commit](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1418/commits/d0a00c0337e90fab916c275c436f570db44d3a08).

manovotny:
@sindresorhus all `arr` abbreviations removed in the rest file. Please let me know if / how you'd like me to resolve the lint issue.

sindresorhus (approved):

fisker (approved):

PR #1419, opened 2021-07-14
`prefer-dom-node-text-content`: Test destructuring with default value

fisker:
This is not done, there is something wrong in the parser, I'll confirm.


fisker:
Raised an issue in ESLint codebase https://github.com/eslint/eslint/issues/14799

fisker:
Blocker #1488

PR #1425, opened 2021-07-14
`prefer-string-slice`: Use `getParenthesizedText`

fisker:


test/prefer-string-slice.mjs
```diff
@@ -150,11 +150,11 @@ test({
 		{
 			code: outdent`
 				const uri = 'foo';
-				(uri || '').substr(1)
+				((uri || '')).substr(1)
```
sindresorhus, line 153: Is the thinking that a different rule will remove the moot parens?

fisker: Not really, we keeps multiple parens only if the original code wrap in multiple parens. When fixing code, it's more important to know where the outermost parens are. This is the reason I made `getParenthesizedText` and `getParenthesizedRange`, checking `isParenthesized` then remove one pair of parens is not right in some rule, so I always test parens with `(( ))`. Update this rule to make sure this won't happen.

The parens is not the problem this rule should care.

sindresorhus: Would be nice to not include moot parens when it's 100% statically detectable though, but not a big deal.

PR #1433, opened 2021-07-19
`no-object-as-default-parameter`: Forbid destructuring

fisker:
I checked the original proposal and the  implementation PR, it seems we missed this, and I don't think we should allow this pattern.


sindresorhus:
Good catch 👍

PR #1436, opened 2021-07-19
Fix cases without space after keywords

fisker:
JS allows no space after keywords in some cases, this PR fixes them.


sindresorhus:
> this PR fixes them.

What exactly does it fix? From my understanding, you're adding support for code where there's no space after a keyword, correct? That's quite an edge-case, and shouldn't that be the responsibility of the JS parser? I might be misunderstanding this though.

fisker:
```js
const[,,thirdElement] = foo;
```

Without this PR(incorrect),

```js
constthirdElement = foo[2];
```


Now

```js
const thirdElement = foo[2];
```

sindresorhus:
But what is the reason we need special handling for it? It seems like something the JS parser and ESLint should handle.

fisker:
No, they parse the original code correctly, it's valid code, engine can run it.

sindresorhus:
I feel like `fixer.replaceText` should do what `fixSpaceAroundKeyword` does built-in. Maybe you could open an ESLint issue about that.

PR #1445, opened 2021-07-25
`prevent-abbreviations`: Ignore `i18n` and `l10n`

### LINKED ISSUE #1188: prevent-abbrevations and i18n
When using `i18n` as a file or variable name, `unicorn/prevent-abbreviations` will report a problem seemingly because of the `i`:

> The variable `i18n` should be named `index18n`. A more descriptive name will do too.

This is also not easily fixed by overriding replacements since the rule seems to catch on the `i`, so using `i18n:` does not work. Instead, you have to have this replacement: `{ i: { index: false } }`.

I think that `i18n` is well-known and popular enough that it doesn't merit triggering this rule.

Divlo:
Hey :wave:,

This PR fixes #1188
It ignores the `i18n` keyword for the `prevent-abbreviations` rule, as it is well-known and popular enough.

I also like the idea of, explicit > implicit, but sometimes, things are done like that, so we can relax a bit the rule.

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -8,7 +8,7 @@ const isShorthandPropertyValue = require('./utils/is-shorthand-property-value.js
 const isShorthandImportLocal = require('./utils/is-shorthand-import-local.js');
 const getVariableIdentifiers = require('./utils/get-variable-identifiers.js');
 const isStaticRequire = require('./utils/is-static-require.js');
-const {defaultReplacements, defaultAllowList} = require('./shared/abbreviations.js');
+const {defaultReplacements, defaultAllowList, defaultIgnore} = require('./shared/abbreviations.js');
 const {renameVariable} = require('./fix/index.js');
 const getScopes = require('./utils/get-scopes.js');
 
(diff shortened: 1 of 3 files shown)


rules/prevent-abbreviations.js
```diff
@@ -96,6 +96,10 @@ const getNameReplacements = (name, options, limit = 3) => {
 		return {total: 0};
 	}
 
+	if (name.includes('i18n')) {
```
sindresorhus, line 99: It should not be hard-coded directly here. There should be a list like `allowList` that is checked in the above if-statement.

fisker:
Let's ignore `l10n` too?

rules/prevent-abbreviations.js
Divlo: Done!
I added `i18n` and `l10n` to `defaultAllowList`.

But we should probably move it to `defaultReplacements`, because something like `const i18nData = {}` will not work, and it will still suggests to refactor it to `index18nData` but that's wrong.

What do you think ? :smile:

fisker: It's known https://github.com/sindresorhus/eslint-plugin-unicorn/issues/752, you'll have to use `ignore` option for now.

Divlo: Right, so I added `defaultIgnore`, should we also, add a new option, `extendDefaultIgnore` ?

sindresorhus: > Right, so I added defaultIgnore, should we also, add a new option, extendDefaultIgnore ?

Nah. We can add it when it's needed. I see no reason to not want these ignores.

sindresorhus (approved):

fisker (approved):

Issue #463, opened 2019-11-30
Proposal changes on `prevent-abbreviations`


fisker:
I did a little research on how people use `prevent-abbreviations`, found some reasonable usecase

I think we should add them to whitelist

`EmberENV`
https://api.emberjs.com/ember/3.10/classes/Ember.EmberENV/properties
https://github.com/ClarkSource/eslint-config/blob/3f6ca81b44bddbfd6a5cd4a9e00950aa3f2fb213/packages/eslint-config-node/index.js#L28
also `launch_in_dev`
https://guides.emberjs.com/v1.10.0/testing/test-runners/  

`devDependencies`
https://github.com/ClarkSource/eslint-config/blob/3f6ca81b44bddbfd6a5cd4a9e00950aa3f2fb213/packages/eslint-config-node/index.js#L41

`getInitialProps`
https://github.com/alienvspredator/typescript-project-template/blob/3b1b19e5bd5bbe6de7c7aadfd4f5c95b238b4e0e/.eslintrc.js#L27
https://github.com/alienvspredator/node-project-boilerplate/blob/15fa866e549258fead6dfbfcabb37fdcdcb38a44/.eslintrc.js#L88
https://nextjs.org/learn/basics/fetching-data-for-pages

`devServer`
https://github.com/mhxbe/react-suspense-experiment/blob/220a75e0c409de0f3b5322ed8b0013815a472fda/.eslintrc.json#L38
https://webpack.js.org/configuration/dev-server/#devserver

`setupFilesAfterEnv`
https://github.com/dbpiper/Euclid/blob/d3c036fa2da48cc67eecc9a616d37dae7b8405fd/src/client/jest.config.js#L35-L37
https://jestjs.io/docs/en/configuration#setupfilesafterenv-array

sindresorhus:
I agree with everything except for `launch_in_dev` and `devServer`. Those are too generic.

sindresorhus:
One more: `getInitialProps`

fisker:
envs -> environments
app -> application

what do you think

sindresorhus:
Yes to the first two.

Not sure about the last one. I have never seen it used before and it could be ambiguous.

fisker:
https://github.com/search?q=tit+%E6%A0%87%E9%A2%98&type=Code

sindresorhus:
Alright, we can try adding `tit`.

PR #1446, opened 2021-07-25
feat(prevent-abbreviations): add `app` => `application`

Divlo:
Hey :wave:,

The word `app` is really famous in programming, but in fact, I'm surprised it is not yet included in the list of abbreviations of this `eslint` plugin.

I tried to see if there was issues or PRs talking about that, here is what I found:

> The second one is actually a valid word. I'd rather do application => app.

From @sindresorhus in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/463#issuecomment-560086267

In fact, it is not a "valid" word, `app` is an abbreviation.

From https://searchmobilecomputing.techtarget.com/definition/app (and many others sources saying that `app` is an abbreviation).
> App is an abbreviated form of the word "application."  An application is a software program  that's designed to perform a specific function directly for the user or, in some cases, for another application program.

### DIFF
--- a/rules/shared/abbreviations.js
+++ b/rules/shared/abbreviations.js
@@ -4,6 +4,9 @@ module.exports.defaultReplacements = {
 	acc: {
 		accumulator: true,
 	},
+	app: {
+		application: true,
+	},
 	arg: {
 		argument: true,
 	},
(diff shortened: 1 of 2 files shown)


sindresorhus:
"app" is actually a word in the dictionary, even though its origin is indeed as an abbreviation form of "application".

- https://english.stackexchange.com/a/282984
- https://trends.google.com/trends/explore?date=all&geo=NO&q=app,application

PR #1453, opened 2021-07-27
Add `prefer-export-from` rule

### LINKED ISSUE #334: Rule proposal: `prefer-export-from`
There are lots of ways to `export` but one pattern that is unnecessary is importing something and then immediately exporting it.

## Fail

```js
import _foo from './foo';
import { bar as _bar } from './bar';

export const foo = _foo;
export const bar = _bar;
```

## Pass

```js
export { default as foo } from './foo';
export { bar } from './bar';
```

fisker:
Fixes #334


sindresorhus:
With this case:

```js
import foo from 'foo';

unicorn(foo);

export default foo;
```

I would personally prefer this instead (re-exporting, even if it was imported):

```js
import foo from 'foo';

unicorn(foo);

export default from 'foo';
```

As it's IMHO more readable and also makes a guarantee that the export is just re-exported and not modified on the way.

Thoughts?

fisker:
Personally, I only use `export...from` if it's used(at least not modified), imaging this

```js
import foo from 'foo';

foo.bar = 1;

export default from 'foo';
```

sindresorhus:
> Still prefer export...from ?

Yes. It's still more readable. You can immediately see what is exported from where instead of having to read and understand all the code in the file.

docs/rules/prefer-export-from.md
```diff
@@ -0,0 +1,61 @@
+# Prefer `export…from` when re-exporting
+
+When re-exporting from module, it's unnecessary to import and then export, it can be done in one `export…from` declaration.
```
sindresorhus, line 3:
```suggestion
When re-exporting from a module, it's unnecessary to import and then export. It can be done in a single `export…from` declaration.
```

rules/prefer-export-from.js
```diff
@@ -0,0 +1,287 @@
...
+
+	for (const {identifier} of references) {
+		const exported = getExported(identifier, context);
+
+		if (!exported) {
+			continue;
+		}
+
+		/*
+		There is no substitution for
```
sindresorhus, line 224:
```suggestion
		There is no substitution for:

```

sindresorhus (changes requested):

test/prefer-export-from.mjs
```diff
@@ -0,0 +1,287 @@
...
+		`,
+	],
+	invalid: [
+		outdent`
+			import * as namespace from 'foo';
+			export {namespace};
+			export default namespace;
+		`,
+	],
+});
```
fisker, line 287: These seem strange, I don't have time to confirm how should them work, but I guess we are fine to ignore it for now.

fisker:
This PR has been a while, I forgot if I left something to do, but seems fine to merge.

test/prefer-export-from.mjs
sindresorhus: Seems like a bug in ESLint. Both of those should be valid.

sindresorhus:
Thanks for finishing this 👍

Issue #1376, opened 2021-06-23
`no-array-callback-reference` false positive with jQuery methods


fregante:
jQuery methods accept both functions and strings, but `no-array-callback-reference` is reporting errors even if the type is `string`. For some reason `lodash.find(string)` is not affected

<img width="730" alt="Do not pass function `submit` directly to `.find(…)`.eslint unicorn/no-array-callback-reference" src="https://user-images.githubusercontent.com/1402241/123024788-d0eb3d00-d403-11eb-9439-3808f6b3bc3c.png">

fregante:
I think `.find()` was not the only method that was reported. Others where some, filter, map, etc

PR #1457, opened 2021-07-28
`no-array-callback-reference`: Ignore jQuery methods

pustovalov:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1376

### DIFF
--- a/rules/no-array-callback-reference.js
+++ b/rules/no-array-callback-reference.js
@@ -83,6 +83,7 @@ const ignoredCallee = [
 	'async',
 	'this',
 	'$',
+	'jQuery',
 ];
 
 function getProblem(context, node, method, options) {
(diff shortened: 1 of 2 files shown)


sindresorhus (approved):

rules/no-array-callback-reference.js
```diff
@@ -164,6 +165,10 @@ const create = context => {
 				return;
 			}
 
+			if (node.callee.object.callee && isNodeMatches(node.callee.object.callee, ignoredCallee)) {
```
fisker, line 168:
```suggestion
			if (node.callee.object.type === 'CallExpression' && isNodeMatches(node.callee.object.callee, ignoredCallee)) {
```

rules/no-array-callback-reference.js
```diff
@@ -82,6 +82,7 @@ const ignoredCallee = [
 	'Async',
 	'async',
 	'this',
+	'$',
```
fisker, line 85:
```suggestion
	'$',
	'jQuery',
```

fisker (changes requested):

fisker (approved):
Nice work!

PR #1478, opened 2021-08-11
Template indent

### LINKED ISSUE #580: Rule proposal: `indent` for template literals
This looks fine:

ok

But is it?

ko

## Description

Template literals are awkward because their indentation may be significant, but I think:

1. We have tools to make indentation at least partially insignificant, like https://github.com/sindresorhus/strip-indent
2. If leading/trailing whitespace is significant, it shouldn't be invisible. Use `\n` and `\t` if you _really_ want spaces in the output.

The second point could be an **option** that still forces indentation that's removable by `strip-indent` while still allowing _invisible but significant_ whitespace.

### Good 👍 

```js
function a () {
	return `
		Hello
	`;
}

function a () {
	return `
	  Hello
	`; // Mixed spacing *may* be ok
}
```

### Not good 👎 

```js
function a () {
	return `
	Hello
	`; // Wrong expected indentation of 'Hello'
}

function a () {
	return `
		Hello`; // Wrong position of closing "`"
}

function a () {
	return `
    Hello
    `; // 4 spaces instead of tab
}
```

## Related

Trailing spaces should also be dropped, but perhaps as a separate rule.

mmkal:
Fixes #580 

Todo:

- [ ] Fill out markdown docs
- ~[ ] Support `comments`~ - I would rather descope this from the first iteration. It's much more complex and I don't see why it couldn't go in separately.

### DIFF
--- a/docs/rules/template-indent.md
+++ b/docs/rules/template-indent.md
@@ -1,0 +1,15 @@
+# Fix whitespace-insensitive template indentation.
+
+Indents things
+
+## Fail
+
+```js
+const foo = 'unicorn';
+```
+
+## Pass
+
+```js
+const foo = '🦄';
+```
--- a/rules/template-indent.js
+++ b/rules/template-indent.js
@@ -1,9 +1,11 @@
 'use strict';
 const stripIndent = require('strip-indent');
 
-const MESSAGE_ID = 'template-indent';
+const MESSAGE_ID_IMPROPERLY_INDENTED_TEMPLATE = 'template-indent';
+const MESSAGE_ID_INVALID_NODE_TYPE = 'invalid-node-type';
 const messages = {
-	[MESSAGE_ID]: 'Templates should be properly indented.',
+	[MESSAGE_ID_IMPROPERLY_INDENTED_TEMPLATE]: 'Templates should be properly indented. Selector: {{ selector }}',
+	[MESSAGE_ID_INVALID_NODE_TYPE]: 'Invalid node type matched by selector {{ selector }}. Expected TemplateLiteral, got {{ type }}'
 };
 
 /** @param {import('eslint').Rule.RuleContext} context */
@@ -11,27 +13,38 @@ const create = context => {
 	const options = {
 		tags: ['outdent', 'dedent', 'gql', 'sql', 'html', 'styled'],
 		functions: ['dedent', 'stripIndent'],
+		selectors: [],
 		...context.options[0],
 	};
-	return {
-		/** @param {import('@babel/core').types.TemplateLiteral} node */
-		TemplateLiteral(node) {
-			const parentTag = node.parent.type === 'TaggedTemplateExpression' && node.parent.tag.type === 'Identifier'
-				? node.parent.tag.name
-				// eslint-disable-next-line unicorn/no-null
-				: null;
-			const parentFunction = node.parent.type === 'CallExpression' && node.parent.callee.type === 'Identifier'
-				? node.parent.callee.name
-				// eslint-disable-next-line unicorn/no-null
-				: null;
 
-			const fixable = options.tags.includes(parentTag) || options.functions.includes(parentFunction);
-			if (!fixable) {
+	const selectors = [
+		...options.tags.map(tag => `TaggedTemplateExpression[tag.name="${tag}"] > .quasi`),
+		...options.functions.map(fn => `CallExpression[callee.name="${fn}"] > .arguments:first-child`),
+		...options.selectors,
+	];
+
+	/** @param {string} selector */
+	const getTemplateLiteralHandler = selector => {
+		/** @param {import('@babel/core').types.TemplateLiteral} node */
+		return node => {
+			if (node.type !== 'TemplateLiteral') {
+				context.report({
+					node,
+					messageId: MESSAGE_ID_INVALID_NODE_TYPE,
+					data: {
+						selector,
+						type: node.type,
+					}
+				})
 				return;
 			}
 
 			const delimiter = '__PLACEHOLDER__' + Math.random();
 			const joined = node.quasis.map(q => q.value.raw).join(delimiter);
+
+			if (!joined.includes('\n')) {
+				return
+			}
 
 			const dedented = stripIndent(joined).trim();
 
@@ -69,11 +82,16 @@ const create = context => {
 			});
 			context.report({
 				node,
-				messageId: MESSAGE_ID,
+				messageId: MESSAGE_ID_IMPROPERLY_INDENTED_TEMPLATE,
+				data: {
+					selector,
+				},
 				fix: fixer => replacements.reverse().map(r => fixer.replaceTextRange(r.range, r.value)),
 			});
-		},
-	};
+		};
+	}
+
+	return Object.fromEntries(selectors.map(selector => [selector, getTemplateLiteralHandler(selector)]));
 };
 
 /** @type {import('json-schema').JSONSchema7[]} */
@@ -84,7 +102,19 @@ const schema = [
 			tags: {
 				type: 'array',
 				items: {
-					oneOf: [{type: 'string'}, {type: 'null'}],
+					type: 'string',
+				},
+			},
+			functions: {
+				type: 'array',
+				items: {
+					type: 'string',
+				},
+			},
+			selectors: {
+				type: 'array',
+				items: {
+					type: 'string',
 				},
 			},
 		},
--- a/rules/template-indent.js
+++ b/rules/template-indent.js
@@ -1,0 +1,106 @@
+'use strict';
+const stripIndent = require('strip-indent');
+
+const MESSAGE_ID = 'template-indent';
+const messages = {
+	[MESSAGE_ID]: 'Templates should be properly indented.',
+};
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => {
+	const options = {
+		tags: ['outdent', 'dedent', 'gql', 'sql', 'html', 'styled'],
+		functions: ['dedent', 'stripIndent'],
+		...context.options[0],
+	};
+	return {
+		/** @param {import('@babel/core').types.TemplateLiteral} node */
+		TemplateLiteral(node) {
+			const parentTag = node.parent.type === 'TaggedTemplateExpression' && node.parent.tag.type === 'Identifier'
+				? node.parent.tag.name
+				// eslint-disable-next-line unicorn/no-null
+				: null;
+			const parentFunction = node.parent.type === 'CallExpression' && node.parent.callee.type === 'Identifier'
+				? node.parent.callee.name
+				// eslint-disable-next-line unicorn/no-null
+				: null;
+
+			const fixable = options.tags.includes(parentTag) || options.functions.includes(parentFunction);
+			if (!fixable) {
+				return;
+			}
+
+			const delimiter = '__PLACEHOLDER__' + Math.random();
+			const joined = node.quasis.map(q => q.value.raw).join(delimiter);
+
+			const dedented = stripIndent(joined).trim();
+
+			const source = context.getSourceCode().getText();
+
+			const preamble = source.slice(0, node.range[0]).split('\n');
+			const line = preamble.length;
+
+			const marginMatch = preamble[line - 1].match(/^(\s*)\S/);
+			const parentMargin = marginMatch ? marginMatch[1] : '';
+			const tabs = parentMargin.startsWith('\t');
+			const indent = tabs ? '\t' : '  ';
+			const templateMargin = parentMargin + indent;
+
+			const fixed = '\n' + dedented.split('\n').map(line => templateMargin + line).join('\n').trimEnd() + '\n' + parentMargin;
+
+			if (fixed === joined) {
+				return;
+			}
+
+			const replacements = fixed.split(delimiter).map((section, i, {length}) => {
+				const range = [...node.quasis[i].range];
+
+				// Add one either for the "`" or "}" prefix character
+				range[0] += 1;
+
+				// Subtract one at the end for the "`" or two in the middle for the "${"
+				const last = i === length - 1;
+				range[1] -= last ? 1 : 2;
+
+				return {
+					range,
+					value: section,
+				};
+			});
+			context.report({
+				node,
+				messageId: MESSAGE_ID,
+				fix: fixer => replacements.reverse().map(r => fixer.replaceTextRange(r.range, r.value)),
+			});
+		},
+	};
+};
+
+/** @type {import('json-schema').JSONSchema7[]} */
+const schema = [
+	{
+		type: 'object',
+		properties: {
+			tags: {
+				type: 'array',
+				items: {
+					oneOf: [{type: 'string'}, {type: 'null'}],
+				},
+			},
+		},
+		additionalProperties: false,
+	},
+];
+
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Fix whitespace-insensitive template indentation.',
+		},
+		fixable: 'code',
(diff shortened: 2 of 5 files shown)


fisker:
>  I don't see why it couldn't go in separately.

Because your template literal can be anywhere, use something like `/* HTML */` is easier than write a selector.

Imaging

```js
foo.innerHTML = /* HTML */ `
// ...
`;
```

```js
$(foo).html(/* HTML */ `
// ...
`)
```

```js
div.insertAdjacentHTML(
	'beforebegin',
	  /* HTML */ `
	  // ...
	  `
);
```

```js
Object.assign(div, {
	innerHTML: /* HTML */ `
	`,
});
```

You don't have to list all those cases in `selectors`, and it can work with Prettier .



> It's much more complex

It's very simple, use `sourceCode.getTokenBefore(..., {includeComments: true})`, and check that comment.

rules/template-indent.js
```diff
@@ -0,0 +1,135 @@
...
+				context.report({
+					node,
+					messageId: MESSAGE_ID_INVALID_NODE_TYPE,
+					data: {
+						selector,
+						type: node.type,
+					},
+				});
+				return;
+			}
```
fisker, line 40: Let's just ignore it.

User can use `CallExpression[callee.name="foo"] > .arguments`, but `foo` may also accept other types, so we only check

```js
foo(`
	// A template string
`)
```

and ignore

```js
foo('A string not a template')
```

User don't have to use `CallExpression[callee.name="foo"] > TemplateLiteral.arguments` selector.

rules/template-indent.js
```diff
@@ -0,0 +1,135 @@
...
+			const delimiter = '__PLACEHOLDER__' + Math.random();
+			const joined = node.quasis.map(q => q.value.raw).join(delimiter);
+
+			if (!joined.includes('\n')) {
+				return;
+			}
+
+			const dedented = stripIndent(joined).trim();
+
+			const source = context.getSourceCode().getText();
```
mmkal, line 51: there might be a better way to get the line the template start on than getting the whole source file

docs/rules/template-indent.md
```diff
@@ -0,0 +1,91 @@
...
+```js
+function foo() {
+  const sqlQuery = sql`
+select *
+from students
+where first_name = ${x}
+and last_name = ${y}
+  `;
+
+  // if you "fix" indentation manually, then copy code somewhere else, it can look stupid
```
fregante, line 18: We can probably avoid this comment, even though I found it funny 😃 👍

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+	/** @param {string} selector */
+	const getTemplateLiteralHandler = selector =>
+		/** @param {import('@babel/core').types.TemplateLiteral} node */
+		node => {
+			if (node.type !== 'TemplateLiteral') {
+				return;
+			}
+
+			const delimiter = '__PLACEHOLDER__' + Math.random();
+			const joined = node.quasis.map(q => q.value.raw).join(delimiter);
```
fisker, line 37: Can we use `sourceCode.getText(node)`?

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+				return;
+			}
+
+			const dedented = stripIndent(joined).trim();
+
+			const startLine = context.getSourceCode().lines[node.loc.start.line - 1];
+			const marginMatch = startLine.match(/^(\s*)\S/);
+			const parentMargin = marginMatch ? marginMatch[1] : '';
+			const tabs = parentMargin.startsWith('\t');
+			const indent = tabs ? '\t' : '  ';
```
fisker, line 49: Let's add an option to allow set indention, accepts string and number, like [`space` for `JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+			const templateMargin = parentMargin + indent;
+
+			const fixed = '\n'
+				+ dedented
+					.split('\n')
+					.map(line => line && line !== '\r' ? templateMargin + line : line)
+					.join('\n')
+					.trimEnd()
+				+ '\n'
+				+ parentMargin;
```
fisker, line 59: sindresorhus have package [`indent-string`](https://github.com/sindresorhus/indent-string), maybe we can use it? And I think it can handle `\r` `\r\n` better. Or maybe just [`redent`](https://github.com/sindresorhus/redent). I havn't test any of them, just an idea. We need tests for `\r` `\r\n` and mixed line breaks.

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+					.map((replacement, index) => replaceTemplateElement(fixer, node.quasis[index], replacement)),
+			});
+		};
+
+	return {
+		...Object.fromEntries(selectors.map(selector => [selector, getTemplateLiteralHandler(selector)])),
+		/**
+		 * @param {import('@babel/core').types.TemplateLiteral} node
+		 */
+		TemplateLiteral: node => {
```
fisker, line 82: If user have `TemplateLiteral` selector, it should report on any template, but I think current implementation won't, because it's overridden.

Is it possible to add only one `TemplateLiteral` listener, and use `esquery.matches` to test node with  `options.selectors`?

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+	const selectors = [
+		...options.tags.map(tag => `TaggedTemplateExpression[tag.name="${tag}"] > .quasi`),
+		...options.functions.map(fn => `CallExpression[callee.name="${fn}"] > .arguments:first-child`),
+		...options.selectors,
+	];
+
+	/** @param {string} selector */
+	const getTemplateLiteralHandler = selector =>
+		/** @param {import('@babel/core').types.TemplateLiteral} node */
+		node => {
```
fisker, line 31: We need a cache to cache reported `node`s, we don't want report twice on same node.

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+		};
+
+	return {
+		...Object.fromEntries(selectors.map(selector => [selector, getTemplateLiteralHandler(selector)])),
+		/**
+		 * @param {import('@babel/core').types.TemplateLiteral} node
+		 */
+		TemplateLiteral: node => {
+			const previous = context.getSourceCode().getTokenBefore(node, {includeComments: true});
+			if (previous.type === 'Block' && options.comments.includes(previous.value.trim().toLowerCase())) {
```
fisker, line 84: `previous` should name as `previousToken`, and it can be nullish

```js
`
// ...
`;
```

Let's add an extra check make sure the comment is on the same line or one line above the template.

```js
const html = ''; /* HTML */

`
// ...
`.split('').forEach(/*...*/)
```

rules/template-indent.js
```diff
@@ -0,0 +1,137 @@
...
+				type: 'array',
+				items: {
+					type: 'string',
+				},
+			},
+			comments: {
+				type: 'array',
+				items: {
+					type: 'string',
+				},
```
fisker, line 119: Maybe JSON schema support unique array?

test/template-indent.mjs
```diff
@@ -0,0 +1,390 @@
+import stripIndent from 'strip-indent';
+import {getTester} from './utils/test.mjs';
+
+/**
+ * The interesting things to test for this rule are whitespace and multiline templates. Both of those are _very_ hard to see in a
+ * normal text editor, so replace spaces with •, tabs with →→, backticks with @, and $ (template parameter prefix symbol) with #.
+ *
+ * @param {string} text
+ */
+const fixInput = text => stripIndent(text).split('@').join('`').split('#').join('$').split('•').join(' ').split('→→').join('\t');
```
fisker, line 10: Let's use `String.replace` here.

fisker (changes requested):

rules/template-indent.js
mmkal: I tried it, but it includes the backtick prefix and `${` suffix. Here's what I did (open to other suggestions if you know a way to avoid this problem, but also `.value.raw` seems appropriate to me:

```js
			const joined = node.quasis.map(q => sourceCode.getText(q)).join(delimiter);
```

rules/template-indent.js
mmkal: I used indent-string, but something seems to be stripping out `\r` from the input somewhere along the way. I'm hopeful that indent-string and strip-indent do "the right thing", but any ideas of how I can get this working?

rules/template-indent.js
mmkal: I reduced down to one listener for functions/tags/custom selectors, and one for comments (where the comments one is not added if `options.comments` is empty). The single listener for functions/tags/custom selectors uses `const {matches} = require('./selectors/index.js');` to generate one big selector matcher string.

rules/template-indent.js
mmkal: Done, and added a test for it.

rules/template-indent.js
mmkal: Done + test case for template literal being the very first token

test/template-indent.mjs
mmkal: Done

rules/template-indent.js
mmkal: Apparently they do. Eslint rule tester can't test this though 🤷 https://github.com/eslint/eslint/issues/13434

rules/template-indent.js
mmkal: Done + tests + docs

rules/template-indent.js
fisker: I don't understand, what's the problem if text including `${`, the expressions should indent too, are you considering nested template literal?

rules/template-indent.js
```diff
@@ -0,0 +1,164 @@
...
+		} else if (typeof options.indent === 'number') {
+			indent = ' '.repeat(options.indent);
+		} else {
+			const tabs = parentMargin.startsWith('\t');
+			indent = tabs ? '\t' : '  ';
+		}
+
+		const dedented = stripIndent(joined);
+		const fixed
+			= os.EOL
```
fisker, line 68: Use `os.EOL` is a bad idea.

rules/template-indent.js
```diff
@@ -0,0 +1,164 @@
...
+				.split(delimiter)
+				.map((replacement, index) => replaceTemplateElement(fixer, node.quasis[index], replacement)),
+		});
+	};
+
+	const matchers = {
+		[matches(selectors)]: templateLiteralHandler,
+	};
+
+	if (options.comments.length > 0 && !selectors.includes('TemplateLiteral')) {
```
fisker, line 90: This is not good enough, because `TemplateLiteral` can be written in other way, `[type="TemplateLiteral"]`, `[type='TemplateLiteral']`, `[type=/^TemplateLiteral$/]` etc, can you try the solution I proposed https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1478#discussion_r687536214?

mmkal: This check is just to make sure we don't override the existing handler, I was trying to avoid adding a handler for every template literal when not necessary. But yes I can switch to your suggestion, since there are comments in the default config we will need to look at every template literal almost all the time.

rules/template-indent.js
mmkal: strip-indent would get confused by the leading and trailing characters, at least with my current implementation of taking the whole template with placeholder parameters. Are you suggesting a different approach entirely?

mmkal:
I haven't forgotten about this, I'm on vacation until Sep 5 so likely will not get a chance to update until then.

rules/template-indent.js
mmkal: changed to `\n`

rules/template-indent.js
mmkal: Switched to `sourceCode.getText(...)` which incidentally seems to respect `\r\n` newlines unlike `.raw`.

mmkal:
@fisker anything else you feel should be added here?

fisker:
I'm take a look ASAP.

rules/template-indent.js
```diff
@@ -0,0 +1,167 @@
...
+		selectors: [],
+		comments: ['HTML', 'indent'],
+		...context.options[0],
+	};
+
+	options.comments = options.comments.map(comment => comment.toLowerCase());
+
+	const selectors = [
+		...options.tags.map(tag => `TaggedTemplateExpression[tag.name="${tag}"] > .quasi`),
+		...options.functions.map(fn => `CallExpression[callee.name="${fn}"] > .arguments:first-child`),
```
fisker, line 27: We should mention that only the first child is checked, or make change here to check all arguments.

Personally I prefer check all arguments.

rules/template-indent.js
```diff
@@ -0,0 +1,167 @@
...
+		...options.selectors,
+	];
+
+	/** @param {import('@babel/core').types.TemplateLiteral} node */
+	const indentTemplateLiteralNode = node => {
+		const delimiter = '__PLACEHOLDER__' + Math.random();
+		const joined = node.quasis
+			.map((quasi, index) => {
+				const untrimmedText = sourceCode.getText(quasi);
+				const last = node.quasis.length - 1;
```
fisker, line 37: You can use `quasi.tail` here.

rules/template-indent.js
```diff
@@ -0,0 +1,167 @@
...
+				return untrimmedText.slice(1, index === last ? -1 : -2);
+			})
+			.join(delimiter);
+
+		const eolMatch = joined.match(/\r?\n/);
+		if (!eolMatch) {
+			return;
+		}
+
+		const eol = eolMatch[0];
```
fisker, line 47: We already have configurable `indent`, do you think it's a good idea to allow config `eol` too?

test/template-indent.mjs
```diff
@@ -0,0 +1,495 @@
...
+ * The interesting things to test for this rule are whitespace and multiline templates. Both of those are _very_ hard to see in a
+ * normal text editor, so replace spaces with •, tabs with →→, backticks with @, and $ (template parameter prefix symbol) with #.
+ *
+ * @param {string} text
+ */
+const fixInput = text => stripIndent(text)
+	.replace(/@/g, '`')
+	.replace(/#/g, '$')
+	.replace(/•/g, ' ')
+	.replace(/→→/g, '\t');
```
fisker, line 14: The `•` and `→→` are nice, but `@` and `#` are not very clear when reading the test cases, let's use `` \` `` and `\$` directly?

fisker (changes requested):

rules/template-indent.js
fisker: Bad idea, please ignore.

rules/template-indent.js
mmkal: Done, and added a test case.

fisker:
We need test for

```js
foo = outdent`
before
before${
expression
}after
after
`;
```

```js
foo = outdent`
  before
  before${
expression
}after
  after
`;
```

```js
foo = outdent`
  before
  before${
	outdent`
inner
	`
}after
  after
`;
```

```js
foo = outdent`
  before
  before${
	normalTemplate`
inner
	`
}after
  after
`;
```

mmkal:
@fisker added tests in 84ada62. Note that this rule is unopinionated about how parent variables are indented, and only cares about the template literals themselves, so 

```js
outdent`
  before
  before${
	outdent`
inner
	`
}after
  after
`;
```

becomes

```js
outdent`
  before
  before${
	outdent`
		inner
	`
}after
  after
`;
```

etc. even though the outer template is indented with spaces and the inner with tabs. It should work nicely with other rules that _do_ care about the indentation level of the parent, like prettier or whatever. Good call to make sure this behaviour doesn't get regressed though. 👍

PR #1483, opened 2021-08-13
Make the example simpler

TuiMao233:


fisker:
Thanks for sending this PR, but the old one is exactly what you should use, and that's the whole point of this rule.

PR #1506, opened 2021-08-30
Add `no-empty-file` rule

### LINKED ISSUE #1477: Rule proposal: `no-empty-file`
There seems to be all kinds of "empty" rules for [block statements](https://eslint.org/docs/rules/no-empty), [destructuring](https://eslint.org/docs/rules/no-empty-pattern), [functions](https://eslint.org/docs/rules/no-empty-function), lines between things, etc., but interestingly nothing for entirely empty files, unless I am completely overlooking something (please correct me if I am wrong and I will humbly 🤦‍♂️).

I am more than happy to submit a PR if approved.

## Description

> Disallow empty files.

Disallow files containing absolutely nothing, only whitespace, or only comments.

### No space or all whitespace

```js

```

### Only Return characters / new lines (or combination thereof) 

```js
\n
```

```js
\r
```

```js
\r\n
```

### Only comments

```js
// comment
```

```js
/* comment */
```

## Pass

```js
const value = 'string';
```

manovotny:
Fixes #1477.


rules/no-empty-file.js
```diff
@@ -0,0 +1,50 @@
...
+		const problem = {
+			node,
+			messageId: MESSAGE_ID,
+			data: {
+				value: 'unicorn',
+				replacement: '🦄',
+			},
+		};
+
+		if (node.body.length === 1 && node.body[0].directive === 'use strict') {
```
fisker, line 19: Should use `node.body.every()` or `node.body.some()`, so `;;'use strict';` will consider empty.

fisker:
Empty `BlockStatement` should also consider empty

```js
{
}
```

```js
{
;;;
}
```


```js
{
	'use strict';
}
```


```js
{
	  {
	  }
}
```

rules/no-empty-file.js
manovotny: Excellent suggestion!

rules/no-empty-file.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+const MESSAGE_ID = 'no-empty-file';
+const messages = {
+	[MESSAGE_ID]: 'Empty files are not allowed.',
+};
+
+const isEmpty = node =>
+	node.type === 'EmptyStatement'
+	|| (node.type === 'ExpressionStatement' && node.expression.value === 'use strict')
```
fisker, line 9: This is wrong, `"use asm"` is also a directive, while `("use strict")` is not.

manovotny:
@fisker all your changes have been incorporated. Please let me know if there is anything else you'd like to see.

test/no-empty-file.mjs
```diff
@@ -0,0 +1,37 @@
...
+
+
+		`,
+		'// comment',
+		'/* comment */',
+		'\'use strict\';',
+		';',
+		';;',
+		';;\'use strict\';',
+		'{}',
```
fisker, line 32: Let's also test `({})`, it should be valid, because it's an object.

test/no-empty-file.mjs
```diff
@@ -0,0 +1,37 @@
...
+		tab,
+		'\n',
+		'\r',
+		'\r\n',
+		outdent`
+
+
+		`,
+		'// comment',
+		'/* comment */',
```
fisker, line 27: Let's also test hashbang.

rules/no-empty-file.js
manovotny: I am not sure what you want me to change here. Are you wanting me to check for `.directive` instead of the expression value?

fisker: Yes. That's correct way to check directives.

test/no-empty-file.mjs
manovotny: Done!

test/no-empty-file.mjs
manovotny: Done!

test/no-empty-file.mjs
fisker: > it should be valid.

manovotny: My apologies. Should be addressed now.

rules/no-empty-file.js
manovotny: I can check `.directive` if `'use strict` is at the top level, however `{'use strict';}` doesn't have a `.directive`, which is why I started checking it this way instead.

fisker: Sorry about this wrong example

```js
{
	'use strict';
}
```

it's not a directive, should be valid.

manovotny: @fisker that means this should be valid too, `;;'use strict';`, correct?

manovotny: This should be fixed.

manovotny:
@fisker everything should once again be addressed. Please let me know if you notice anything else.

fisker:
Everything looks good, I'll take a deeper look ASAP, I've been very busy recently.

manovotny:
@fisker no worries! Was just making sure there wasn't anything else immediate to take care of. Totally understand busyness, so take whatever time you need.

fisker:
Can you merge https://github.com/manovotny/eslint-plugin-unicorn/pull/1, and fix failed tests? I can't commit to your fork.

manovotny:
@fisker I'm not sure the Vue tests that were added are correct... I am not a Vue developer (I use React), but in reading the [Vue Single File Components documentation](https://vuejs.org/v2/guide/single-file-components.html), I think this is what we'd want for tests instead. Do you agree / disagree?

```js
test.snapshot({
	testerOptions: {
		parser: parsers.vue,
	},
	valid: [
		outdent`
			<template></template>
			<style></style>
			<script>const x = 0;</script>
			<custom-block></custom-block>
		`,
	],
	invalid: [
		outdent`
			<template></template>
			<style></style>
			<script></script>
			<custom-block></custom-block>
		`,
		outdent`
			<template></template>
			<style></style>
			<script>;</script>
			<custom-block></custom-block>
		`,
	],
});
```

manovotny:
@fisker @sindresorhus can I get any input on my Vue comment above and if it is correct or incorrect?

fisker:
I'm sorry, somehow I missed this comment.

I'm not sure what you mean. Are you saying we should have all those blocks? If true, I disagree. A SFC file only have `<script>` is valid, the template can be replaced by `render` function in `<script>`, and only have `<template>` is also valid.

manovotny:
@fisker I'm sure you only have one or two comments a day to reply to. 😂 Sincerely no worries. 

Good to know. As I stated, I have little to no experience. I will proceed with your tests, but might need some insight. I'll let you know if I have questions.

manovotny:
@fisker I guess where I'm still slightly confused is that you have all of these marked as `valid`. 

```js
test.snapshot({
	testerOptions: {
		parser: parsers.vue,
	},
	valid: [
		'<template></template>',
		'<style></style>',
		'<script></script>',
		'<script>;</script>',
		'<custom-block></custom-block>',
	],
	invalid: [
		'   ',
	],
});
```

But if they don't contain any logic, markup, or styles, they would be considered "empty", right? Shouldn't these all be considered `invalid` instead?

Also seems like we should add tests for `valid` tests for `<script src="./component.js"></script>` and `<style src="./component.css"></style>` too.

fisker:
> Also seems like we should add tests for valid tests for <script src="./component.js"></script> and <style src="./component.css"></style> too.

👍 

> But if they don't contain any logic, markup, or styles, they would be considered "empty", right?

I'm not really sure about those cases now, maybe we should just ignore `.vue` files.

fisker:
> As far as I can tell, an SFC is invalid unless it has a render function in the script block or a template block. 

I think you are right about this, but there is also an edge case, the empty component is used with a global mixin (with render of cause), then it can be a valid component. I'm just pointing the possibility in theory.

---

I'm also a little worried about svelte SFC files(I don't know about it), and empty code blocks in markdown files.

fisker:
This PR has been blocked for quite a long time.

Let's do this, only check .js/.mjs/.cjs/.ts/.mts/.cts files for now.

I've committed changes, hope it's fine.

@sindresorhus WDYT?

fisker (approved):

manovotny:
@fisker that sounds like a good plan to me. If / When a clearer path forward for Vue, Svelte, etc. becomes more obvious, we can do a follow up PR and address it at that time.

sindresorhus:
Thanks for contributing this rule :)

PR #1509, opened 2021-09-03
Add `prefer-number-is-integer` rule

### LINKED ISSUE #1460: Rule proposal: Use `Number.isInteger`
Use [`Number.isInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger).
Not really sure if there's any benefit to using it. Maybe it's more performant. Makes the code marginally cleaner.

## Fail

```js
number % 2 === 0;
```

## Pass

```js
Number.isInteger(number);
```

tristanHessell:
Fixes #1460.

One thing to note is that I couldn't run `npm run integration` to completion as my computer only has 4GB RAM.


docs/rules/prefer-number-is-integer.md
```diff
@@ -0,0 +1,36 @@
+# Prefer `Number.isInteger()` for integer checking
+
+Enforces the use of [Number.isInteger()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) for checking if a number is an integer.
+
+These different implementations have slightly different behaviours.
+
+For example:
+
+```js
+let number = [['1']];
```
sindresorhus, line 10: It's not immediately clear to the user why this is used as an example.

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,150 @@
...
+].join('');
+
+// Math.round(value) === value
+const mathRoundSelector = [
+	'BinaryExpression',
+	'[left.type="CallExpression"]',
+	'[left.callee.type="MemberExpression"]',
+	'[left.callee.object.name="Math"]',
+	'[left.callee.property.name="round"]',
+	'[left.arguments.0.type="Identifier"]',
```
fisker, line 59: I don't think we should limit this to `Identifier`, it should work for anything, just make sure the argument and the right side are the same reference, you can use [`isSameReference`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/is-same-reference.js).

And don't auto-fix if the expression [has side effect](https://eslint-utils.mysticatea.dev/api/ast-utils.html#hassideeffect).

tristanHessell:
Thanks for the feedback on PR, I'll get onto updating this PR over the next week :)

docs/rules/prefer-number-is-integer.md
tristanHessell: Updated :+1:

rules/prefer-number-is-integer.js
tristanHessell: @fisker I've updated the PR to use `isSameReference`.

Can you give me an example of an expression with a side effect that fits here? So I can TDD against it :)

test/prefer-number-is-integer.mjs
```diff
@@ -0,0 +1,108 @@
...
+			output: 'Number.isInteger(value)',
+		}]}),
+		suggestionCase({code: 'parseInt(value, 10) === value', suggestions: [{
+			messageId: 'preferNumberIsIntegerSuggestion',
+			output: 'Number.isInteger(value)',
+		}]}),
+		suggestionCase({code: 'parseInt(value, 10) == value', suggestions: [{
+			messageId: 'preferNumberIsIntegerSuggestion',
+			output: 'Number.isInteger(value)',
+		}]}),
```
fisker, line 58: I don't think this should be fixed

```js
parseInt('11', 10) == '11'
true
Number.isInteger('11')
false
````

But it's safe if it's `===`, maybe we should only check `===`

Comment (fisker) [reply test/prefer-number-is-integer.mjs:58]: Sorry, I didn't see that this is a suggestion.

**Comment by fisker** [comment]
Sorry for the long delay, overall looks good. Can you fix tests?

File: docs/rules/prefer-number-is-integer.md [hunk 4]
```diff
@@ -0,0 +1,44 @@
+# Prefer `Number.isInteger()` for integer checking
+
```
bmish, line 2: There are some new requirements for rule docs, be sure to rebase on the latest `main` to pick them up.

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+		}),
+		[bitwiseNotSelector]: createNodeListener(sourceCode, node => {
+			if (isSameReference(node.left.argument.argument, node.right)) {
+				return getNodeName(node.right);
+			}
+		}),
+	};
+};
+
+module.exports = {
```
bmish, line 147: We'll want to add a `type` annotation here (more info in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1606):

```suggestion
/** @type {import('eslint').Rule.RuleModule} */
module.exports = {
```

TristanHessellHyper:
Happy to make the changes! Will get it done over the next week or so

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+const MESSAGE_ID = 'preferNumberIsInteger';
+const MESSAGE_ID_SUGGEST = 'preferNumberIsIntegerSuggestion';
+const messages = {
+	[MESSAGE_ID]: 'Replace `{{original}}` with `Number.isInteger({{variable}})`.',
+	[MESSAGE_ID_SUGGEST]: 'Prefer `Number.isInteger()` for integer checks.',
+};
+
+const equalsSelector = ':matches([operator="==="],[operator="=="])';
+
+// Value % 1 === 0
```
fisker, line 14:
```suggestion
// `value % 1 === 0`
```

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+const modulo1Selector = [
+	'BinaryExpression',
+	'[left.type="BinaryExpression"]',
+	'[left.operator="%"]',
+	'[left.right.value=1]',
+	equalsSelector,
+	'[right.value="0"]',
+].join('');
+
+// (value^0) === value OR (value | 0) === value
```
fisker, line 24:
```suggestion
//  `(value ^ 0) === value`
//  `(value | 0) === value`
```

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+const mathOperatorSelector = [
+	'BinaryExpression',
+	'[left.type="BinaryExpression"]',
+	`:matches(${['^', '|'].map(operator => `[left.operator="${operator}"]`).join(',')})`,
+	'[left.right.value=0]',
+	equalsSelector,
+].join('');
+
+// ParseInt(value,10) === value
```
fisker, line 33:
```suggestion
// `Number.parseInt(value, 10) === value`
```

Let's forget about `parseInt`, only check `Number.parseInt`

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+const parseIntSelector = [
+	'BinaryExpression',
+	'[left.type="CallExpression"]',
+	'[left.callee.name="parseInt"]',
+	'[left.arguments.1.value=10]',
+	equalsSelector,
+].join('');
+
+// _.isInteger(value)
```
fisker, line 42: Use `callExpressionSelector`, search how to use it in other rules.

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+const lodashIsIntegerSelector = [
+	'CallExpression',
+	`:matches(${['_', 'lodash', 'underscore'].map(callee => `[callee.object.name="${callee}"]`).join(',')})`,
+	'[callee.property.name="isInteger"]',
+].join('');
+
+// Math.round(value) === value
```
fisker, line 49: Use `methodCallSelector`,

sindresorhus:
Bump :)

tristanHessell:
@sindresorhus appreciated! I had forgot about this PR.

Working on it over the next couple of days

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,157 @@
...
+			return getNodeName(node.expression);
+		}
+
+		case 'MemberExpression': {
+			return `${getNodeName(node.object)}.${getNodeName(node.property)}`;
+		}
+
+		default: {
+			return '';
+		}
```
tristanHessell, line 108: @sindresorhus what should we do here?

I've added the default case as to appease linting, but the calling code protects us from this branch being taken.

How should I approach appeasing Codecov?

sindresorhus: You could use an ignore comment:

```
/* c8 ignore next 3 */
```

tristanHessell:
@sindresorhus et al, this is ready or review :)

rules/prefer-number-is-integer.js
```diff
@@ -0,0 +1,158 @@
...
+	'[left.argument.type="UnaryExpression"]',
+	'[left.argument.operator="~"]',
+	equalsSelector,
+].join('');
+
+function createNodeListener(sourceCode, variableGetter) {
+	return node => {
+		const variable = variableGetter(node);
+
+		if (!variable) {
```
fisker, line 72: We should not limit this to  `Identifier` and `MemberExpression`, should allow anything

```js
_.isInteger(a ? b : c)
```

sindresorhus: // @tristanHessell

sindresorhus:
@tristanHessell Bump

PR #1512, opened 2021-09-05
Add `all`config

### LINKED ISSUE #1510: Export an `all` config
Export an `all` config that contains the default options for all of the rules (except for deprecated ones).

Yash-Singh1:
Fixes #1510


Yash-Singh1:
Lemme split into sep files like: https://github.com/typescript-eslint/typescript-eslint/blob/8218055d6dfd94c9e6c8645848f981d9d51ce08c/packages/eslint-plugin/src/configs/all.ts as @sindresorhus linked earlier (https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1510#issuecomment-913116625).

index.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 const createDeprecatedRules = require('./rules/utils/create-deprecated-rules.js');
 const {loadRules} = require('./rules/utils/rule.js');
+const recommended = require('./configs/recommended.js');
```
sindresorhus, line 4:
```suggestion
const recommendedConfig = require('./configs/recommended.js');
```

sindresorhus:
You need to add the `configs` directory to package.json

configs/eslint-recommended.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+module.exports = {
+	rules: {
+		'no-nested-ternary': 'off',
+	},
+};
```
fisker, line 6: I don't think this file name is a good name, but I don't have a good idea.

`conflicting-rules.js`?

Yash-Singh1: I got the file name from here: https://github.com/typescript-eslint/typescript-eslint/blob/8218055d6dfd94c9e6c8645848f981d9d51ce08c/packages/eslint-plugin/src/configs/eslint-recommended.ts. I think it made sense for them, because they were disabling **and enabling** a rules, so I think `conflicting-rules.js` would work.

Yash-Singh1:
Applied suggestions in b79b233c5d3552a3e2a4d1a944068b32ad438098

AriPerkkio:
Coverage of smoke tests could now be increased.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/7bece2ad3de8686e68ac694aaec3a0f50ae827a1/test/smoke/eslint-remote-tester.config.js#L52

```diff
- extends: ['plugin:unicorn/recommended'], 
+ extends: ['plugin:unicorn/all'], 
```

_(It would also be great to include there even the rules which require options but that's out of scope here)_

PR #1523, opened 2021-09-13
`no-array-for-each`: Don't mention about performance

### LINKED ISSUE #1522: no-array-for-each Slower Performance
I went  to https://jsbench.me/ and put in the following two tests...

**Test  1:**
```js
var x = new Set([1, 2, 3, 4, 5, 6, 7, 8]);
x.forEach(value => {
	console.log(value);
});
```

**Test 2:**
```js
var x = new Set([1, 2, 3, 4, 5, 6, 7, 8]);
for(const p of x) {	
	console.log(p);
}
```

The rule "no-array-for-each" is not better to use.  The documentation indicates that using "for of" has better performance, but I dont believe it does.  I may have better performance for straight arrays, but when it comes to Maps and Sets, I am finding better performance with using the built in "forEach()" method.

fisker:
Fixes #1522


sindresorhus:
While it's not the intent of the rule, it is one of the positive side-effects. I'd like to keep it in.

PR #1526, opened 2021-09-17
`prefer-top-level-await`: Improve top-level expression detection

### LINKED ISSUE #1502: `prefer-top-level-await`: Miss cases
Currently, it only reports top-level expressions, the following cases should be reported too.

```js
if (foo) {
	bar().catch(() => {});
}
```

```js
const promise = Promise.resolve();
for (const foo of bar) {
	promise.then(foo);
}
```

I'll work on it, when I have time.

fisker:
Fix #1502


sindresorhus:
This looks good to me. Is there any reason it is still draft?

fisker:
```js
await foo.then(bar)
```

This will still be reported, but I guess we shouldn't.

And

```js
foo.then(bar).then(bar).then(bar).then(bar).then(bar)
```

Will be many errors, we need to find the promise chain root and only report on the root.

---

I won't have time to work on this recently.

fisker:
I didn't forgot this, I'll continue work on it when I got time.

PR #1540, opened 2021-10-04
Don't break CI with `expiring-todo-comments`

fregante:
Is suddenly-breaking CI ideal? TODO comments that break the build immediately seem ok if condition is "engines", but not dates.

https://github.com/sindresorhus/refined-github/runs/3786952560

The current setting makes sense, but it probably means I can't use dates.

### DIFF
--- a/configs/recommended.js
+++ b/configs/recommended.js
@@ -10,7 +10,12 @@ module.exports = {
 		'unicorn/empty-brace-spaces': 'error',
 		'unicorn/error-message': 'error',
 		'unicorn/escape-case': 'error',
-		'unicorn/expiring-todo-comments': 'warn',
+		'unicorn/expiring-todo-comments': [
+			"error",
+			{
+				"ignoreDatesOnPullRequests": true
+			}
+		],
 		'unicorn/explicit-length-check': 'error',
 		'unicorn/filename-case': 'error',
 		'unicorn/import-index': 'off',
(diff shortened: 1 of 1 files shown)


sindresorhus:
What if we added a `ignoreDatesOnCi` option? https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md#ignoredatesonpullrequests

fregante:
yesplz

fregante:
Also I don't know why it's failing here

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1541

configs/recommended.js
```diff
@@ -10,7 +10,12 @@ module.exports = {
 		'unicorn/empty-brace-spaces': 'error',
 		'unicorn/error-message': 'error',
 		'unicorn/escape-case': 'error',
-		'unicorn/expiring-todo-comments': 'error',
+		'unicorn/expiring-todo-comments': [
+			"error",
+			{
+				"ignoreDatesOnPullRequests": true
```
sindresorhus, line 16: This is already the default: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/expiring-todo-comments.md#ignoredatesonpullrequests

fregante:
Ah true. So let's close this PR while we wait for the option

PR #1547, opened 2021-10-11
Add `id-match` rule

### LINKED ISSUE #1545: Rule proposal: `id-match`
When working with external libraries, it is often undesirable to rename their exports for the sole purpose of conforming with `id-match` rule.

This has been proposed as addition to ESLint in https://github.com/eslint/eslint/issues/14005, but ultimately it was rejected:

> I think this falls under our policies of not adding new options to stylistic rules, so in my opinion the best course of action would be to make a custom rule/plugin.

This rule would be equivalent to `id-match` with an additional option `ignoreImports` that would ignore named imports.

## Fail

Assuming that rule does not permit IDs with `$`:

```js
const ContraImageFragment$key = '';
```

## Pass

Assuming that this rule is configured to ignore imports:

```js
import type { ContraImageFragment$key } from '@/__generated__/ContraImageFragment.graphql';
```

Happy to contribute to the codebase.

gajus:
Fixes #1545

### DIFF
--- a/rules/id-match.js
+++ b/rules/id-match.js
@@ -1,0 +1,289 @@
+/**
+ * @fileoverview Rule to flag non-matching identifiers
+ * @author Matthieu Larcher
+ *
+ * Adapted from https://github.com/eslint/eslint/blob/c4fffbcb089182d425ef1d5e45134fecc0e2da46/lib/rules/id-match.js
+ * Related discussion about not adding this option to ESLint https://github.com/eslint/eslint/issues/14005
+ */
+
+'use strict';
+
+const create = context => {
+	const pattern = context.options[0] || '^.+$';
+	const regexp = new RegExp(pattern, 'u');
+
+	const options = context.options[1] || {};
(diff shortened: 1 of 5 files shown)


rules/id-match.js
```diff
@@ -0,0 +1,289 @@
...
+ * @fileoverview Rule to flag non-matching identifiers
+ * @author Matthieu Larcher
+ *
+ * Adapted from https://github.com/eslint/eslint/blob/c4fffbcb089182d425ef1d5e45134fecc0e2da46/lib/rules/id-match.js
+ * Related discussion about not adding this option to ESLint https://github.com/eslint/eslint/issues/14005
+ */
+
+'use strict';
+
+const create = context => {
```
bmish, line 11:
```suggestion
/** @param {import('eslint').Rule.RuleContext} context */
const create = context => {
```

gajus:
Someone please take over this PR if interested. I ended up releasing an ESLint plugin for all personal-preference rules.

PR #1549, opened 2021-10-15
Update eslint-plugin-eslint-plugin to v4

bmish:
https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/releases/tag/v4.0.0

The `meta.fixable` and `meta.hasSuggestions` properties are now grouped together by the [eslint-plugin/meta-property-ordering](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/meta-property-ordering.md) rule.


fisker:
Thank  you!

PR #1569, opened 2021-11-02
`require-array-join-separator`: Check optional member

### LINKED ISSUE #1568: require-array-join-separator doesn't work with optional chaining
The rule doesn't work if `join` method is called with optional chaining

Rule: require-array-join-separator 

```js
        const array: string[] = ['1', '2', '3'];
        console.log(array.join());  // error
        console.log(array?.join()); // no error
```

fisker:
Fixes #1568


test/require-array-join-separator.mjs
```diff
@@ -71,5 +70,8 @@ test.snapshot({
 					)/**/
 			)
 		`,
+		'foo?.join()',
+		'foo.join?.()',
```
sindresorhus, line 74: I don't think this should be invalid. `String#join` is never optional, so if it is, it's a sign that it's a custom method and we should ignore it.

fisker: You are right. I'll fix.

PR #1573, opened 2021-11-02
Add basic tests of rule docs

bmish:
1. Ensure rule doc title matches rule description (prevents these titles from becoming out-of-date).
2. Ensure rules with options have an `## Options` section and rules without options don't.
3. Ensure rules with options have their named options mentioned in their rule doc.

### DIFF
--- a/test/package.mjs
+++ b/test/package.mjs
@@ -32,6 +32,40 @@ const testSorted = (t, actualOrder, sourceName) => {
 	}
 };
 
+function toSentenceCase(originalString) {
+	let newString = originalString.charAt(0).toUpperCase() + originalString.slice(1); // Capitalize first letter.
+	if (newString.endsWith('.')) {
+		newString = newString.slice(0, -1); // Remove any ending period.
+	}
+
+	return newString;
+}
+
+/**
+ * Get list of named options from a JSON schema (used for rule schemas).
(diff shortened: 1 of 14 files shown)


test/package.mjs
```diff
@@ -32,6 +32,40 @@ const testSorted = (t, actualOrder, sourceName) => {
...
 };
 
+function toSentenceCase(originalString) {
+	let newString = originalString.charAt(0).toUpperCase() + originalString.slice(1); // Capitalize first letter.
+	if (newString.endsWith('.')) {
+		newString = newString.slice(0, -1); // Remove any ending period.
+	}
+
+	return newString;
+}
```
fisker, line 42: Can you apply the same logic to this? 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/scripts/template/documentation.md.jst#L1

Maybe we can add another property here.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c10e2ba6aa700abe90d350f89275f314a8453d84/scripts/create-rule.mjs#L140

bmish: Updated.

sindresorhus (approved):

bmish:
I'll be submitting another improvement after this is merged :)

PR #1581, opened 2021-11-08
Turn of `require-post-message-target-origin` rule in recommended configs by default

### LINKED ISSUE #1396: `require-post-message-target-origin` shouldn't trigger for incompatible APIs

### LINKED ISSUE #1552: *.ts override makes ESLint lint .ts files
By default, ESLint only looks at files with the `.js` extension. However, after installing this plugin, [this override](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/93b34793d470e40b5f53eb08854e7f50795e15d5/configs/recommended.js#L105) causes it to lint `.ts` files too.

Since I am in a JS codebase with a manually written `.d.ts` file, I get parsing errors like this:

```
  12:9  error  Parsing error: Unexpected token const
```

on

_index.d.ts_
```ts
declare const myFunction: (name: string) => Promise<boolean>;
```

I would prefer it doesn't try to lint my `.d.ts` or `ts` files if I don't have the Typescript plugin installed.

Although there are workarounds for this, I think this is unexpected default behavior that shouldn't require patching on user-land.

-----

Workarounds:
- manually adding the rules from _recommended_
- adding `"ignorePatterns": ["*.ts"]`

fisker:
Fixes #1396
Fixes #1552

### DIFF
--- a/configs/recommended.js
+++ b/configs/recommended.js
@@ -96,18 +96,12 @@ module.exports = {
 		'unicorn/prevent-abbreviations': 'error',
 		'unicorn/require-array-join-separator': 'error',
 		'unicorn/require-number-to-fixed-digits-argument': 'error',
-		'unicorn/require-post-message-target-origin': 'error',
+		// Turned off because we can't distinguish `widow.postMessage` and `{Worker,MessagePort,Client,BroadcastChannel}#postMessage()`
+		// See #1396
+		'unicorn/require-post-message-target-origin': 'off',
 		'unicorn/string-content': 'off',
 		'unicorn/template-indent': 'warn',
 		'unicorn/throw-new-error': 'error',
 		...require('./conflicting-rules.js').rules,
 	},
-	overrides: [
-		{
-			files: ['*.ts', '*.tsx'],
-			rules: {
-				'unicorn/require-post-message-target-origin': 'off',
-			},
-		},
-	],
 };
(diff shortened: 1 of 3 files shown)


jdanil:
Just FYI, this rule is still flagged as recommended in the rule's doc file.
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/require-post-message-target-origin.md?plain=1#L3

fisker:
@bmish Can you add a test for that?

bmish:
@fisker fixed! https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1619

PR #1586, opened 2021-11-09
Add `no-await-expression-member` rule

### LINKED ISSUE #1458: Rule proposal: `no-nested-await`
Nesting await inside a statement often gets messy. It’s better to split the line using an intermediate variable.

This is potentially autofixable. The only issue is that this would need to generate a name for an intermediate variable.

## Fail

```js
const filtered = (await Promise.all(promises)).filter(Boolean);
```

## Pass

```js
const unfiltered = await Promise.all(promises);
const filtered = unfiltered.filter(Boolean);
```

fisker:
Fixes #1458


rules/no-await-expression-member.js
```diff
@@ -0,0 +1,84 @@
+'use strict';
+const {
+	removeParentheses,
+	removeMemberExpressionProperty,
+} = require('./fix/index.js');
+
+const MESSAGE_ID = 'no-await-expression-member';
+const messages = {
+	[MESSAGE_ID]: 'Do not access member directly from an await expression.',
```
sindresorhus, line 9:
```suggestion
	[MESSAGE_ID]: 'Do not access a member directly from an await expression.',
```

sindresorhus (changes requested):

PR #1590, opened 2021-11-10
`prefer-export-from`: Add `ignoreUsedVariables` option

### LINKED ISSUE #1577: `prefer-export-from` shouldn't report re-export of variable that is used in other statements
```js
import { a } from './my-module.js';

export const exportA = a;
export const exportB = a.b;
```

The `prefer-export-from` rule warns about the first export and wants to change it to:

```js
import { a } from './my-module.js';


export const exportB = a.b;
export { a as exportA } from './my-module.js';
```

I think that this duplication is more confusing: `a` imported as `a`, but also re-exported with a different name.

I therefore suggest to ignore exports of variables that are used in other statements.

Real-life reproduction: [OpenLightingProject/open-fixture-library@`dependabot/npm_and_yarn/eslint-plugin-unicorn-38.0.1`/lib/schema-properties.js](https://github.com/OpenLightingProject/open-fixture-library/blob/dependabot/npm_and_yarn/eslint-plugin-unicorn-38.0.1/lib/schema-properties.js)

fisker:
Fixes #1577


test/prefer-export-from.mjs
```diff
@@ -289,3 +289,108 @@ test.typescript({
...
+		`,
+		outdent`
+			import defaultExport, {named1, named2} from 'foo';
+			use(named1);
+			export {defaultExport, named2};
+		`,
+		outdent`
+			import {notUsedNotExported, exported} from 'foo';
+			export {exported};
+		`,
```
fisker, line 364: In this case,  `notUsedNotExported` is not really used, but also not exported,

I think the main purpose for this option is to avoid multiple `from ...`, if variable is not used, it's possible this file not done yet. 

Once user fixes [`no-unused-vars`](https://eslint.org/docs/rules/no-unused-vars), we'll know it will be used or exported.

sindresorhus: Can you add what you commented here as a comment to that test case? For posterity.

fisker: On second thought, we can just skip auto-fix, seems better.

docs/rules/prefer-export-from.md
```diff
@@ -61,3 +61,41 @@ export {
...
 ```
+
+## Options
+
+### ignoreUsedVariables
+
+Type: `boolean`\
+Default: `false`
+
+When pass `"ignoreUsedVariables": true`, if any variable is used more than just re-export or not exported, all variables in the import declaration will be ignored.
```
sindresorhus, line 72:
```suggestion
When `true`, if an import is used in other places than just a re-export or is not exported, all variables in the import declaration will be ignored.
```

?

sindresorhus (changes requested):

test/prefer-export-from.mjs
fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1590/commits/a58be750ceffd0528db2140d45a5e563f6b8bb13

docs/rules/prefer-export-from.md
fisker: Fixed.

sindresorhus:
Lint is failing

fregante:
No default? 😇 🙏

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e05fbe749df4d502be6ba3c8ca0309460903086c/configs/recommended.js#L69

fisker:
Sindre prefer to check them all.

PR #1604, opened 2021-11-11
Use `npm-run-all` to run `test` and `lint` scripts

bmish:
* Reorganizes `test` and `lint` NPM scripts with popular tool [npm-run-all](https://github.com/mysticatea/npm-run-all)
* Adds convenient scripts for running different types of linting or tests
* Runs linting in parallel but keeps output of linters separate
* Does not run linting and testing in parallel, since this would break the real-time/live output from ava
* Shows complete output from all types of linting and tests, even if one type fails (previously, the `test` script would exit early if one type failed, so you wouldn't be able to see all the failures)
* Avoids using `&&` which might not have as good of cross-platform support according to the [npm-run-all](https://github.com/mysticatea/npm-run-all) readme

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 	},
 	"scripts": {
 		"lint": "run-p --continue-on-error \"lint:!(fix)\"",
-		"lint:fix": "run-p --continue-on-error \"lint:*:fix\"",
+		"lint:fix": "run-p --continue-on-error lint:*:fix",
 		"lint:js": "xo",
 		"lint:js:fix": "npm run lint:js -- --fix",
 		"lint:md": "markdownlint \"**/*.md\"",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
 	},
 	"scripts": {
 		"lint": "run-p --continue-on-error \"lint:!(fix)\"",
-		"lint:fix": "run-p \"lint:*:fix\"",
+		"lint:fix": "run-p --continue-on-error \"lint:*:fix\"",
 		"lint:js": "xo",
 		"lint:js:fix": "npm run lint:js -- --fix",
 		"lint:md": "markdownlint \"**/*.md\"",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
 	"scripts": {
 		"lint": "npm-run-all --continue-on-error --aggregate-output --parallel lint:*",
 		"lint:js": "xo",
-		"lint:md": "markdownlint '**/*.md'",
+		"lint:md": "markdownlint \"**/*.md\"",
 		"test": "npm-run-all --continue-on-error lint test:*",
 		"test:js": "nyc ava",
 		"create-rule": "node ./scripts/create-rule.mjs && npm run generate-rules-table && npm run generate-usage-example",
--- a/package.json
+++ b/package.json
@@ -14,12 +14,12 @@
 		"node": ">=12"
 	},
 	"scripts": {
-		"lint": "run-p --continue-on-error \"lint:!(fix)\"",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -14,7 +14,11 @@
 		"node": ">=12"
 	},
 	"scripts": {
-		"test": "xo && nyc ava && markdownlint '**/*.md'",
+		"lint": "npm-run-all --continue-on-error --aggregate-output --parallel lint:*",
```
fisker, line 17: I used `npm-run-all` for a long time, I think simple `run-p lint:*` is good enough.

package.json
```diff
@@ -14,7 +14,11 @@
...
+		"lint:js": "xo",
+		"lint:md": "markdownlint '**/*.md'",
```
fisker, line 19: I believe only double quote is supported on Windows.

fisker (changes requested):
I suggest also add a `fix` version.

package.json
bmish: I can switch to `run-p` but don't you think `--continue-on-error --aggregate-output` have benefits as mentioned in PR description?

package.json
bmish: Fixed.

package.json
fisker: I don't have strong opinion, I can give it a shot.

bmish: Actually `--aggregate-output` causes it to lose the output formatting so I removed that. But switched to `run-p`.

bmish:
> I suggest also add a `fix` version.

Added `:fix` scripts.

fisker:
Sorry, I didn't make myself clear.

I prefer

```json
{
	"lint": "",
	"lint:a": "",
	"lint:b": "",
	"fix": "",
	"fix:a": "",
	"fix:b": ""
}
```

`yarn fix` and `yarn fix:js` is simpler than `yarn lint:fix` and `yarn lint:js:fix`

bmish:
@fisker that's fine with me, updated.

fisker (approved):

fisker (changes requested):
Can you also update the action config file? https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/.github/workflows/main.yml#L28

bmish:
> Can you also update the action config file? [`main`/.github/workflows/main.yml#L28](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/.github/workflows/main.yml?rgh-link-date=2021-11-12T01%3A46%3A52Z#L28)

Updated to have CI run `npm run lint` and `npm run test:js`.

However, this is why I personally prefer to have the test and lint scripts completely independent, so that CI can just run the `lint` and `test` scripts separately.

fisker:
Me too. But I guess sindresorhus prefers it in one.

fisker (approved):

bmish:
It looks like Windows CI doesn't like something about running `nyc ava` even though the tests appear to have passed.

```
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! eslint-plugin-unicorn@38.0.1 test:js: `nyc ava`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the eslint-plugin-unicorn@38.0.1 test:js script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

fisker:
Try switch nyc to c8?

bmish:
> Try switch nyc to c8?

I doubt `nyc` itself is the problem as I've used that in other repos with Windows CI.

fisker:
I ran into similar thing, and fixed by using `c8`. But I'm not sure, we can figure later.

PR #1605, opened 2021-11-11
`no-empty-file`: Fix false positive with triple-slash directives

### LINKED ISSUE #1598: False positives with `no-empty-files`
The rule `no-empty-files` triggers false positives for [`.d.ts` modules](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) that contain only [triple-slash directives](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html).

This is used, for example, by vite in lieu of `compilerOptions.types` in `jsconfig.json` or `tsconfig.json` ([ref](https://github.com/vitejs/vite/tree/main/packages/create-vite/template-svelte-ts#technical-considerations))

Example:

```js
/// <reference types="svelte" />
/// <reference types="vite/client" />
```

manovotny:
The simplest fix seems to be to ignore `.d.ts` files as `@babel/eslint-parser` and `@typescript-eslint/parser` still view [triple-slash directives](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html) as regular comments that produce an empty `body`.

Fixes #1598

### DIFF
--- a/rules/no-empty-file.js
+++ b/rules/no-empty-file.js
@@ -16,7 +16,7 @@ const isEmpty = node =>
 const create = context => {
 	const filename = context.getPhysicalFilename().toLowerCase();
 
-	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename)) {
+	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename) || filename.endsWith('.d.ts')) {
 		return {};
 	}
 
--- a/rules/no-empty-file.js
+++ b/rules/no-empty-file.js
@@ -17,7 +17,7 @@ const isTripleSlashDirective = node =>
 	node.type === 'Line' && node.value.startsWith('/');
 
 const hasTripeSlashDirectives = comments =>
-	!comments.every(currentNode => !isTripleSlashDirective(currentNode));
+	comments.some(currentNode => isTripleSlashDirective(currentNode));
 
 /** @param {import('eslint').Rule.RuleContext} context */
 const create = context => {
--- a/rules/no-empty-file.js
+++ b/rules/no-empty-file.js
@@ -5,18 +5,29 @@ const messages = {
 	[MESSAGE_ID]: 'Empty files are not allowed.',
 };
 
+const isTripleSlashDirective = node =>
+	node.type === 'Line' && node.value.startsWith('/');
+
 const isEmpty = node =>
 	(
-		(node.type === 'Program' || node.type === 'BlockStatement')
+		node.type === 'Program'
+		&& node.body.every(currentNode => isEmpty(currentNode))
+		&& node.comments.every(currentNode => !isTripleSlashDirective(currentNode))
+	)
+	|| (
+		node.type === 'BlockStatement'
 		&& node.body.every(currentNode => isEmpty(currentNode))
 	)
-	|| node.type === 'EmptyStatement'
-	|| (node.type === 'ExpressionStatement' && 'directive' in node);
+	|| (
+		node.type === 'ExpressionStatement'
+		&& 'directive' in node
+	)
+	|| node.type === 'EmptyStatement';
 
 const create = context => {
 	const filename = context.getPhysicalFilename().toLowerCase();
 
-	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename) || filename.endsWith('.d.ts')) {
+	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename)) {
 		return {};
 	}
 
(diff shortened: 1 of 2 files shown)


fisker:
I think it's possible to get all comments and test that it's start with a `/`.

manovotny:
@fisker I am trying the "starts with `/`" route, but shebangs are giving me trouble as they meet that criteria too. I'll keep fiddling.

fisker:
I 'm not familiar with this syntax, but I see this line https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/rules/triple-slash-reference.ts#L96

Maybe we can ignore file, if
1. Starts with `/` and not one  first line, shebang has to be on first line.
2. Starts with `/\s*<reference\s*`, we can add missing cases later.

fisker:
Wait, I don't think `.d.ts` can use shebang, so filename check  push first char  should enough ?

manovotny:
@fisker yeah, never mind. I was making it more difficult than it needed to be. PR has been updated.

rules/no-empty-file.js
```diff
@@ -5,13 +5,24 @@ const messages = {
...
 
+const isTripleSlashDirective = node =>
+	node.type === 'Line' && node.value.startsWith('/');
+
 const isEmpty = node =>
 	(
-		(node.type === 'Program' || node.type === 'BlockStatement')
+		node.type === 'Program'
+		&& node.body.every(currentNode => isEmpty(currentNode))
+		&& node.comments.every(currentNode => !isTripleSlashDirective(currentNode))
```
fisker, line 15: I don't think we can use `node.comment`. See https://eslint.org/docs/developer-guide/working-with-rules#sourcecodegetallcomments

I also found [shebang is stored as `Shebang`](https://eslint.org/docs/developer-guide/working-with-rules#accessing-shebangs).

manovotny: Yep! `comments` are `Line`, `Block`, and `Shebang`. I was messing with that before I simplified. 

The comments are showing up in `node.comments` as I'd expect if you debug or `console.log`, but I can look into switching it over to use `sourceCode.getAllComments()` instead, if you prefer.

manovotny: @fisker PR updated again.

rules/no-empty-file.js
```diff
@@ -13,17 +13,25 @@ const isEmpty = node =>
 	|| node.type === 'EmptyStatement'
 	|| (node.type === 'ExpressionStatement' && 'directive' in node);
 
+const isTripleSlashDirective = node =>
+	node.type === 'Line' && node.value.startsWith('/');
+
+const hasTripeSlashDirectives = comments =>
+	!comments.every(currentNode => !isTripleSlashDirective(currentNode));
```
fisker, line 20: Use `.some()`?

manovotny: Yes! Definitely reads better. 😅 Updated!

rules/no-empty-file.js
fisker: > The comments are showing up in node.comments 

It's not safe for other nodes except `Program`.  

> For example, the @babel/eslint-parser can set attachComment to false since the linter does not care comment attachment. We even delete such comments in @babel/eslint-parser.

Quote from https://github.com/babel/babel/pull/13229#issue-871065828

rules/no-empty-file.js
```diff
@@ -13,17 +13,25 @@ const isEmpty = node =>
...
+	const comments = sourceCode.getAllComments();
 
 	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename)) {
 		return {};
 	}
 
 	return {
 		Program(node) {
-			if (!isEmpty(node)) {
+			if (!isEmpty(node) || hasTripeSlashDirectives(comments)) {
```
fisker, line 34: Can you check how `getAllComments()` works? Is it expensive to access? If true, we can inline that, so we don't need access it before `isEmpty()` check.

manovotny: I wrapped the `no-empty-file` test with `console.time()` and tried it as is and then with the looking for comments after the `isEmpty` check. There was no discernible difference in time. Granted, this is a super small use case.

Digging into ESLint, `getAllComments` is a [simple lookup](https://github.com/eslint/eslint/blob/dd58cd4afa6ced9016c091fc99a702c97a3e44f0/lib/source-code/source-code.js#L299-L301) on `ast.comments`, which [comes from the `SourceCode` constructor](https://github.com/eslint/eslint/blob/dd58cd4afa6ced9016c091fc99a702c97a3e44f0/lib/source-code/source-code.js#L176). 

It doesn't seem too overtly expensive, but more than happy to split it out after `isEmpty` to be safe.

manovotny: [Done](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1605/commits/d9f76cd8e51b3b90704290e4b608a5a77645738c)!

fisker (approved):
Good job! Thank you!

manovotny:
@fisker you're welcome!

PR #1606, opened 2021-11-11
Add jsdoc `type` annotation to rules

bmish:
Supported by VSCode/Webstorm/other code editors. Uses TypeScript declaration to give JavaScript hints so code editors can provide information/autocomplete about various rule fields.

https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#type

Idea from: https://github.com/eslint/generator-eslint/pull/113

133895350-a19d0daf-593e-44b9-835f-bebf8b6a3d30

### DIFF
--- a/rules/better-regex.js
+++ b/rules/better-regex.js
@@ -111,6 +111,9 @@ const schema = [
 	},
 ];
 
+/**
+ * @type {import('eslint').Rule.RuleModule}
+ */
 module.exports = {
 	create,
 	meta: {
--- a/rules/better-regex.js
+++ b/rules/better-regex.js
@@ -112,8 +112,8 @@ const schema = [
 ];
 
 /**
- * @type {import('eslint').Rule.RuleModule}
- */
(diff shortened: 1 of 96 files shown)


rules/better-regex.js
```diff
@@ -111,6 +111,9 @@ const schema = [
 	},
 ];
 
+/**
+ * @type {import('eslint').Rule.RuleModule}
+ */
```
sindresorhus, line 116:
```suggestion
/**
@type {import('eslint').Rule.RuleModule}
*/
```

I prefer this style. I find the `*` inside the comment to be pure noise and makes editing harder than it should be.

bmish: Updated, that's fine with me. It can also be a single line comment if desired.

bmish: Updated again to single-line. No need for it to take up additional space.

fisker (approved):

PR #1608, opened 2021-11-12
Use `c8` instead of `nyc`

fisker:


sindresorhus:
https://github.com/bcoe/c8#ignoring-uncovered-lines-functions-and-blocks

fisker:
It supposed to support `istanbul ignore` too. And `c8 ignore` seems don't support ignore reason https://github.com/bcoe/c8/issues/181

PR #1612, opened 2021-11-16
Add test for rule jsdoc type comment

bmish:
Follow-up to https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1606

Ensures each rule has this jsdoc type comment. Otherwise, it would be easily forgotten in new rules.

### DIFF
--- a/rules/utils/rule.js
+++ b/rules/utils/rule.js
@@ -74,6 +74,7 @@ function checkVueTemplate(create, options) {
 	return wrapped;
 }
 
+/** @returns {import('eslint').Rule.RuleModule} */
 function loadRule(ruleId) {
 	const rule = require(`../${ruleId}`);
 
--- a/rules/utils/rule.js
+++ b/rules/utils/rule.js
@@ -74,7 +74,6 @@ function checkVueTemplate(create, options) {
 	return wrapped;
 }
 
-/** @returns {import('eslint').Rule.RuleModule} */
 function loadRule(ruleId) {
 	const rule = require(`../${ruleId}`);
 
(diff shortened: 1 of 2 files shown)


fisker:
I'm 👎 for this, this type info supposed to help rule author write the `module.export`, users can not import these files directly.

fisker:
Maybe add this jsdoc to this function is useful. 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/rule.js#L77

bmish:
> I'm 👎 for this, this type info supposed to help rule author write the `module.export`, users can not import these files directly.

The jsdoc comment is only intended for the convenience of internal developers. What's the problem with testing to ensure each of our rules has it?

bmish:
> Maybe add this jsdoc to this function is useful.
> 
> [`main`/rules/utils/rule.js#L77](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/rule.js?rgh-link-date=2021-11-16T23%3A54%3A30Z#L77)

Good idea, added (can be moved to separate PR if desired).

bmish:
> > Maybe add this jsdoc to this function is useful.
> > [`main`/rules/utils/rule.js#L77](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/rule.js?rgh-link-date=2021-11-17T00%3A05%3A09Z#L77)
> 
> Good idea, added (can be moved to separate PR if desired).

Extracted to separate PR: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1613

sindresorhus:
I agree we should enforce consistency here.

bmish:
Thanks. By the way, can we enable the GitHub setting "Automatically delete head branches", so we don't have to manually click "Delete branch" each time?

sindresorhus:
Weird. It's already enabled.

PR #1616, opened 2021-11-22
Add `no-thenable` rule

### LINKED ISSUE #1596: Rule proposal: `no-thenable`
### Description

If object is defined as thenable, once it's accidentally used in an await expression, your program may hangs. [Airbnb has similar rule for module exports.](https://github.com/airbnb/javascript/blob/366bfa66380c08304101c6add46355696e90b348/packages/eslint-config-airbnb-base/rules/es6.js#L68)

I want forbid more cases.

Maybe we don't even need care it's a function or not. Objects except Promise shouldn't have a `then` property.

### Fail

```js
const foo = {then() {}}
```

```js
const foo = {get then() {}}
```

```js
export const then = () => {};
```

```js
export {then}
```

```js
class Foo {
	then;
}
```

```js
class Foo {
	static then;
}
```

```js
class Foo {
	then() {}
}
```

```js
class Foo {
	static then() {}
}
```

```js
class Foo {
	get then() {}
}
```

```js
class Foo {
	static get then() {}
}
```

```js
foo.then = () => {};
```

```js
Foo.prototype.then = () => {};
```

```js
Object.defineProperty(foo, 'then', {value(){}})
```

```js
Reflect.defineProperty(foo, 'then', {value(){}})
```

```js
Object.fromEntries(['then', () => {}])
```

### Pass

```js
const foo = {notThen() {}}
```

fisker:
Fixes #1596


docs/rules/no-thenable.md
```diff
@@ -0,0 +1,63 @@
+# Disallow `then` property
+
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+If an object is defined as "thenable", once it's accidentally used in an await expression, it may cause problems.
```
sindresorhus, line 5: It should describe what kind of problem.

docs/rules/no-thenable.md
```diff
@@ -0,0 +1,63 @@
...
+
+If a module have exported `then`, dynamic `import()` may won't works as expected.
```
sindresorhus, line 7:
```suggestion
If a module has an export named `then`, dynamic `import()` may not work as expected.
```

This also should describe how it will not work.

sindresorhus (changes requested):

docs/rules/no-thenable.md
bmish, line 5: Any docs to link to about this for context?

fisker:
I've added some example to show the problems.

docs/rules/no-thenable.md
fisker: I can't find any link for this.

docs/rules/no-thenable.md
```diff
@@ -0,0 +1,93 @@
...
+
+If an object is defined as "thenable", once it's accidentally used in an await expression, it may cause problems.
+
+```js
+const foo = {
+	unicorn: 1,
+	then() {},
+};
+const {unicorn} = await foo;
+console.log('after'); //<- This will nerve execute
```
fisker, line 13:
```suggestion
console.log('after'); //<- This will never execute
```

PR #1619, opened 2021-11-23
Fix test for `recommended` rule doc notice

bmish:
There was a bug in this test so we mistakenly added the `recommended` rule notice to some rule docs that are not in the `recommended` config.

The test will now correctly verify that the `recommended` rules include this notice and non-recommended rules don't.

### DIFF
--- a/test/package.mjs
+++ b/test/package.mjs
@@ -236,7 +236,7 @@ test('Every rule has a doc with the appropriate content', t => {
 		// Decide which notices should be shown at the top of the doc.
 		const expectedNotices = [];
 		const unexpectedNotices = [];
-		if (index.configs.recommended.rules[`unicorn/${ruleName}`] !== 'off') {
+		if (['error', 'warn'].includes(index.configs.recommended.rules[`unicorn/${ruleName}`])) {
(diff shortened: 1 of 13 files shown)


test/package.mjs
```diff
@@ -236,7 +236,7 @@ test('Every rule has a doc with the appropriate content', t => {
 		// Decide which notices should be shown at the top of the doc.
 		const expectedNotices = [];
 		const unexpectedNotices = [];
-		if (index.configs.recommended.rules[`unicorn/${ruleName}`]) {
+		if (index.configs.recommended.rules[`unicorn/${ruleName}`] === 'error') {
 			expectedNotices.push('configRecommended');
```
fisker, line 240: `!= 'off'`，, I believe `template-indent` was configured as `warn`.

bmish: Thanks, fixed

bmish:
This is ready to be merged.

test/package.mjs
bmish: Had to switch to `includes` to avoid [no-negated-condition](https://eslint.org/docs/rules/no-negated-condition).

PR #1623, opened 2021-11-28
Add `no-useless-promise-resolve-reject` rule

### LINKED ISSUE #1609: Rule proposal: No unnecessary `Promise.resolve/reject()` in async functions
### Description

The rule should prevent using moot `Promise.reject` and `Promise.resovle` calls in async functions.

Rule name suggestion: `no-useless-promise-resolve-reject`

### Fail

```js
const main = async foo => {
	if (foo > 4) {
		return Promise.reject(new Error('🤪'));
	}

	return Promise.resolve(result);
};
```

### Pass

```js
const main = async foo => {
	if (foo > 4) {
		throw new Error('🤪');
	}

	return result;
};
```

cherryblossom000:
Fixes #1609

### DIFF
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -91,7 +91,7 @@ const create = context => {
 		},
 		[yieldExpressionSelector](node) {
 			const parentFunction = getParentFunction(node);
-			if (!parentFunction || !parentFunction.async || !parentFunction.generator) {
+			if (!parentFunction || !parentFunction.async) {
 				return;
 			}
 
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -57,9 +57,11 @@ const create = context => {
 		return {
 			node: node.callee,
 			messageId: isResolve ? resolveMessage : rejectMessage,
-			fix: (!isResolve && isInTryStatement) || node.arguments.length > 1
-				? undefined
-				: createFix(isResolve, value ? sourceCode.getText(value) : 'undefined', value),
+			fix: node.arguments.length <= 1
+					&& (!value || value.type !== 'SpreadElement')
+					&& (isResolve || !isInTryStatement)
+				? createFix(isResolve, value ? sourceCode.getText(value) : 'undefined', value)
+				: undefined,
 		};
 	};
 
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -12,6 +12,10 @@ const messages = {
 	[YIELD_REJECT]: 'Prefer `throw value` over `yield Promise.reject(value)`.',
 };
 
+const getMessageId = (isYield, isResolve) => isYield
+	? (isResolve ? YIELD_RESOLVE : YIELD_REJECT)
+	: (isResolve ? RETURN_RESOLVE : RETURN_REJECT);
+
 const SELECTOR = [
 	methodCallSelector({
 		object: 'Promise',
@@ -93,9 +97,7 @@ const create = context => {
 
 			return {
 				node: node.callee,
-				messageId: isResolve
-					? (isYield ? YIELD_RESOLVE : RETURN_RESOLVE)
-					: (isYield ? YIELD_REJECT : RETURN_REJECT),
+				messageId: getMessageId(isYield, isResolve),
 				fix: node.arguments.length <= 1
 						&& (!value || value.type !== 'SpreadElement')
 						// Can't fix if the Promise.reject is inside a try block as a
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -54,14 +54,17 @@ const create = context => {
 		[asyncArrowFunctionReturnSelector](node) {
 			return createProblem(
 				node,
-				(isResolve, value) => fixer =>
+				(isResolve, valueString, value) => fixer =>
 					fixer.replaceText(
 						node,
 						isResolve
 							// Turns `=> Promise.resolve(value)` into `=> value`
-							? value
+							? (value.type === 'ObjectExpression' || value.type === 'SequenceExpression'
+								? `(${valueString})`
+								: valueString
+							)
 							// Turns `=> value` into `=> { throw value }`
-							: `{ throw ${value}; }`,
+							: `{ throw ${valueString}; }`,
 					),
 			);
 		},
@@ -73,11 +76,11 @@ const create = context => {
 
 			return createProblem(
 				node,
-				(isResolve, value) => fixer => isResolve
+				(isResolve, valueString) => fixer => isResolve
 					// Turns `return Promise.resolve(value)` into `return value`
-					? fixer.replaceText(node, value)
+					? fixer.replaceText(node, valueString)
 					// Turns `return Promise.reject(value)` into `throw value`
-					: fixer.replaceText(node.parent, `throw ${value};`),
+					: fixer.replaceText(node.parent, `throw ${valueString};`),
 			);
 		},
 		[yieldExpressionSelector](node) {
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -94,9 +94,15 @@ function fix(callExpression, isInTryStatement, sourceCode) {
 			// `return Promise.reject()` -> `throw undefined`
 			text = text || 'undefined';
 			text = `throw ${text}`;
-			if (!isYieldExpression) {
-				text += ';';
+
+			if (isYieldExpression) {
+				return fixer.replaceTextRange(
+					getParenthesizedRange(parent, sourceCode),
+					text,
+				);
 			}
+
+			text += ';';
 
 			// `=> Promise.reject(error)` -> `=> { throw error; }`
 			if (isArrowFunctionBody) {
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -48,12 +48,11 @@ function getFunctionNode(node) {
 function createProblem(callExpression, fix) {
 	const {callee, parent} = callExpression;
 	const method = callee.property.name;
-	const messageId = method === 'reject' ? MESSAGE_ID_REJECT : MESSAGE_ID_RESOLVE;
 	const type = parent.type === 'YieldExpression' ? 'yield' : 'return';
 
 	return {
 		node: callee,
-		messageId,
+		messageId: method,
 		data: {type},
 		fix,
 	};
@@ -69,14 +68,13 @@ function fix(callExpression, isInTryStatement, sourceCode) {
 		return;
 	}
 
-	const method = callee.property.name;
-	const isReject = method === 'reject';
+	const isReject = callee.property.name === 'reject';
 	const isYieldExpression = parent.type === 'YieldExpression';
 	if (
 		isReject
 		&& (
-			isInTryStatement ||
-			(isYieldExpression && parent.parent.type !== 'ExpressionStatement')
+			isInTryStatement
+			|| (isYieldExpression && parent.parent.type !== 'ExpressionStatement')
 		)
 	) {
 		return;
@@ -87,19 +85,13 @@ function fix(callExpression, isInTryStatement, sourceCode) {
 
 		let text = errorOrValue ? sourceCode.getText(errorOrValue) : '';
 
-		if (
-			errorOrValue &&
-			(
-				errorOrValue.type === 'SequenceExpression'
-				|| (!isReject && isArrowFunctionBody && errorOrValue.type === 'ObjectExpression')
-			)
-		) {
+		if (errorOrValue && errorOrValue.type === 'SequenceExpression') {
 			text = `(${text})`;
 		}
 
 		if (isReject) {
 			// `return Promise.reject()` -> `throw undefined`
-			text ||= 'undefined';
+			text = text || 'undefined';
 			text = `throw ${text}`;
 			if (!isYieldExpression) {
 				text += ';';
@@ -110,19 +102,24 @@ function fix(callExpression, isInTryStatement, sourceCode) {
 				text = `{ ${text} }`;
 			}
 		} else {
+			// eslint-disable-next-line no-lonely-if
 			if (isYieldExpression) {
 				text = `yield${text ? ' ' : ''}${text}`;
 			} else if (parent.type === 'ReturnStatement') {
 				text = `return${text ? ' ' : ''}${text};`;
 			} else {
+				if (errorOrValue && errorOrValue.type === 'ObjectExpression') {
+					text = `(${text})`;
+				}
+
 				// `=> Promise.resolve()` into `=> {}`
-				text ||= `{}`;
+				text = text || '{}';
 			}
 		}
 
 		return fixer.replaceText(
 			isArrowFunctionBody ? callExpression : parent,
-			text
+			text,
 		);
 	};
 }
@@ -140,7 +137,7 @@ const create = context => {
 
 			return createProblem(
 				callExpression,
-				fix(callExpression, isInTryStatement, sourceCode)
+				fix(callExpression, isInTryStatement, sourceCode),
 			);
 		},
 	};
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -1,5 +1,5 @@
 'use strict';
-const {methodCallSelector} = require('./selectors/index.js');
+const {matches, methodCallSelector} = require('./selectors/index.js');
 
 const RETURN_RESOLVE = 'return-resolve';
 const RETURN_REJECT = 'return-reject';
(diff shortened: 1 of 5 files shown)


rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
+'use strict';
+const {methodCallSelector} = require('./selectors/index.js');
+
+const RETURN_RESOLVE = 'return-resolve';
+const RETURN_REJECT = 'return-reject';
+const YIELD_RESOLVE = 'yield-resolve';
+const YIELD_REJECT = 'yield-reject';
+const messages = {
+	[RETURN_RESOLVE]: 'Prefer `return value` over `return Promise.resolve(value)`.',
+	[RETURN_REJECT]: 'Prefer `throw value` over `return Promise.reject(value)`.',
```
sindresorhus, line 10:
```suggestion
	[RETURN_REJECT]: 'Prefer `throw error` over `return Promise.reject(error)`.',
```

I know you can technically throw/reject any value, but the common case is an error, so we  should optimize the message for that.

sindresorhus (changes requested):

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+	[YIELD_RESOLVE]: 'Prefer `yield value` over `yield Promise.resolve(value)`.',
+	[YIELD_REJECT]: 'Prefer `throw value` over `yield Promise.reject(value)`.',
+};
+
+const promiseResolveRejectSelector = methodCallSelector({
+	object: 'Promise',
+	methods: ['resolve', 'reject'],
+	allowComputed: true,
+});
+const asyncArrowFunctionReturnSelector = `ArrowFunctionExpression[async=true] > ${promiseResolveRejectSelector}`;
```
fisker, line 20:
```suggestion
const asyncArrowFunctionReturnSelector = `ArrowFunctionExpression[async=true] > ${promiseResolveRejectSelector}.body`;
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+const returnStatementSelector = `ReturnStatement > ${promiseResolveRejectSelector}`;
```
fisker, line 21:
```suggestion
const returnStatementSelector = `ReturnStatement > ${promiseResolveRejectSelector}.argument`;
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+const yieldExpressionSelector = `YieldExpression > ${promiseResolveRejectSelector}`;
```
fisker, line 22:
```suggestion
const yieldExpressionSelector = `YieldExpression > ${promiseResolveRejectSelector}.argument`;
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+const YIELD_RESOLVE = 'yield-resolve';
+const YIELD_REJECT = 'yield-reject';
+const messages = {
+	[RETURN_RESOLVE]: 'Prefer `return value` over `return Promise.resolve(value)`.',
+	[RETURN_REJECT]: 'Prefer `throw error` over `return Promise.reject(error)`.',
+	[YIELD_RESOLVE]: 'Prefer `yield value` over `yield Promise.resolve(value)`.',
+	[YIELD_REJECT]: 'Prefer `throw value` over `yield Promise.reject(value)`.',
+};
+
+const promiseResolveRejectSelector = methodCallSelector({
```
fisker, line 15:
```suggestion
const promiseResolveOrRejectSelector = methodCallSelector({
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+	object: 'Promise',
+	methods: ['resolve', 'reject'],
+	allowComputed: true,
```
fisker, line 18: `allowComputed` should be false, use `true` will matches `Promise[resolve]`, `resolve` may not equals to `'resolve'`.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+	return {
+		[asyncArrowFunctionReturnSelector](node) {
+			return createProblem(
+				node,
+				(isResolve, value) => fixer =>
+					fixer.replaceText(
+						node,
+						isResolve
+							// Turns `=> Promise.resolve(value)` into `=> value`
+							? value
```
fisker, line 62: In some case, the value may need parens. Eg:

```js
async () => Promise.resolve({})
```

```js
async () => ({})
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+	[RETURN_REJECT]: 'Prefer `throw error` over `return Promise.reject(error)`.',
+	[YIELD_RESOLVE]: 'Prefer `yield value` over `yield Promise.resolve(value)`.',
+	[YIELD_REJECT]: 'Prefer `throw value` over `yield Promise.reject(value)`.',
+};
+
+const promiseResolveRejectSelector = methodCallSelector({
+	object: 'Promise',
+	methods: ['resolve', 'reject'],
+	allowComputed: true,
+});
```
fisker, line 19: We should limit `argumentsLength` to `1`

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+	const sourceCode = context.getSourceCode();
+
+	/** @param {(isResolve: boolean, valueString: string, value) => import('eslint').Rule.ReportFixer} createFix */
+	const createProblem = (node, createFix, resolveMessage = RETURN_RESOLVE, rejectMessage = RETURN_REJECT) => {
+		const isResolve = node.callee.property.name === 'resolve';
+		const value = node.arguments.length === 1 ? node.arguments[0] : undefined;
+		return {
+			node,
+			messageId: isResolve ? resolveMessage : rejectMessage,
+			fix: value === undefined ? undefined : createFix(isResolve, sourceCode.getText(value), value),
```
fisker, line 49: `sourceCode.getText(value)` should try keep parens.

```js
async () => Promise.resolve(  (0, 1) )
```

Should fix to 

```js
async () => (0, 1)
```
Not

```js
async () => 0, 1
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+			const parentFunction = getParentFunction(node);
+			if (!parentFunction || !parentFunction.async || !parentFunction.generator) {
+				return;
+			}
+
+			return createProblem(
+				node,
+				(isResolve, valueString, value) => fixer => isResolve
+					// Turns `yield Promise.resolve(value)` into `yield value`
+					? fixer.replaceText(node, value.type === 'SequenceExpression' ? `(${valueString})` : valueString)
```
fisker, line 93: If https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1623/files#r771216741 is addressed, this `SequenceExpression` should not needed, in most case, because it's originally a call argument, we just keep the parens, won't need insert. But I'm not sure if there are edge cases don't need parens in argument, but neeeded in `yield`.

fisker: Check `should-add-parentheses-to-*.js` in `rules/utils` directory, need implment similar logic.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,117 @@
...
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => {
+	const sourceCode = context.getSourceCode();
+
+	/** @param {(isResolve: boolean, valueString: string, value) => import('eslint').Rule.ReportFixer} createFix */
+	const createProblem = (node, createFix, resolveMessage = RETURN_RESOLVE, rejectMessage = RETURN_REJECT) => {
+		const isResolve = node.callee.property.name === 'resolve';
+		const value = node.arguments.length === 1 ? node.arguments[0] : undefined;
+		return {
+			node,
```
fisker, line 47: Let's report on the `callee`, not the whole `CallExpression`.

fisker (changes requested):
Check all comments.

rules/no-useless-promise-resolve-reject.js
cherryblossom000: Should parentheses be preserved even if they are not required (e.g. `async () => Promise.resolve((foo))` → `async () => (foo)`?

fisker: I guess `ObjectExpression` and `SequenceExpression` check is good enough, we can remove or keep, they are both valid.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,120 @@
...
+					: fixer.replaceText(node.parent, `throw ${valueString};`),
+			);
+		},
+		[yieldExpressionSelector](node) {
+			const parentFunction = getParentFunction(node);
+			if (!parentFunction || !parentFunction.async || !parentFunction.generator) {
+				return;
+			}
+
+			return createProblem(
```
fisker, line 92: We need check `YieldExpression.delegate` here, we can't fix

```js
yield* Promise.resolve([])
```

fisker: We also need check if `YieldExpression` is an `ExpressionStatement`, because `YieldExpression` can be any where

```js
async function * a() {
  alert(yield Promise.resolve(1));
}
```

```js
async function * a() {
  const foo = yield Promise.resolve(1);
}
```

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,120 @@
...
+	[RETURN_RESOLVE]: 'Prefer `return value` over `return Promise.resolve(value)`.',
+	[RETURN_REJECT]: 'Prefer `throw error` over `return Promise.reject(error)`.',
+	[YIELD_RESOLVE]: 'Prefer `yield value` over `yield Promise.resolve(value)`.',
+	[YIELD_REJECT]: 'Prefer `throw value` over `yield Promise.reject(value)`.',
+};
+
+const promiseResolveOrRejectSelector = methodCallSelector({
+	object: 'Promise',
+	methods: ['resolve', 'reject'],
+	argumentsLength: 1,
```
fisker, line 18: On second thought, we should not limit arguments length to `1`, but make sure handle them correctly

```js
return Promise.resolve()
// -> 
return;

yield Promise.resolve()
// -> 
yield;

return Promise.reject()
// -> 
throw undefind; // Can't fix to `throw;` it's invalid, we can throw undefined or not fix at all.
```

fisker:
Another edge case, if the return statement is in `try/catch`, we may breaking them

```js
async function  a() {
  try {
    return Promise.reject(1)
  } catch {
    return 2
  }
}

a() // Reject with 1
```

```js
async function  a() {
  try {
    throw 1
  } catch {
    return 2
  }
}

a() // Resolve with 2
```

cherryblossom000:
In that case should we report an error or not?

fisker:
Report but no fix? I'm not sure, either. We can just skip fix if these calls in `try`(with or without `catch`), it's fine to miss cases.

rules/no-useless-promise-resolve-reject.js
cherryblossom000: What about multiple arguments? Should the fix be a suggestion instead so the user can correct their mistake (the extra args could cause side effects despite being unnecessary for Promise.resolve/reject)?

fisker: yes, multiple arguments, report but no fix.

rules/no-useless-promise-resolve-reject.js
cherryblossom000: Why do we need to check if it's in an `ExpressionStatement`? `f` and `g` below seem to behave the same:

```js
async function* f() {
  console.log('next:', yield Promise.resolve(1))
}

async function* g() {
  console.log('next:', yield 1)
}

const a = f()
const a1 = await a.next()
console.log(a1)            // {value: 1, done: false}
const a2 = await a.next(2) // next: 2
console.log(a2)            // {value: undefined, done: true}

const b = g()
const b1 = await b.next()
console.log(b1)            // {value: 1, done: false}
const b2 = await b.next(2) // next: 2
console.log(b2)            // {value: undefined, done: true}
```

fisker: Sorry, I mean `Promise.reject()`

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,138 @@
...
+				break;
+			}
+
+			parent = parent.parent;
+		}
+
+		return {
+			node: node.callee,
+			messageId: isResolve ? resolveMessage : rejectMessage,
+			fix: (!isResolve && isInTryStatement) || node.arguments.length > 1
```
fisker, line 60: We need also check the only argument's type, it can't be `Promise.resolve(...args)`

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,138 @@
...
+
+		let isInTryStatement = false;
+		let {parent} = node;
+		while (parent && !functionTypes.has(parent.type)) {
+			if (parent.type === 'TryStatement') {
+				isInTryStatement = true;
+				break;
+			}
+
+			parent = parent.parent;
```
fisker, line 54: Loop until the function? If function is in `try`, it doesn't matter.

fisker: Sorry, mistake.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,138 @@
...
+	[YIELD_REJECT]: 'Prefer `throw value` over `yield Promise.reject(value)`.',
+};
+
+const promiseResolveOrRejectSelector = methodCallSelector({
+	object: 'Promise',
+	methods: ['resolve', 'reject'],
+});
+const asyncArrowFunctionReturnSelector = `ArrowFunctionExpression[async=true] > ${promiseResolveOrRejectSelector}.body`;
+const returnStatementSelector = `ReturnStatement > ${promiseResolveOrRejectSelector}.argument`;
+const yieldExpressionSelector = `YieldExpression[delegate=false] > ${promiseResolveOrRejectSelector}.argument`;
```
fisker, line 21: Can you try combine them into one selector, there are duplicate lines in those functions.

```js
[
	promiseResolveOrRejectSelector,
	matches([
		'ArrowFunctionExpression[async=true] > .body',
		'ReturnStatement > .argument',
		'YieldExpression[delegate=false] > .argument',
	])
].join('')
```

Should work, we can use `parent.type` to distinguish them in the listener.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -0,0 +1,130 @@
+'use strict';
+const {matches, methodCallSelector} = require('./selectors/index.js');
+
+const RETURN_RESOLVE = 'return-resolve';
+const RETURN_REJECT = 'return-reject';
+const YIELD_RESOLVE = 'yield-resolve';
+const YIELD_REJECT = 'yield-reject';
+const messages = {
+	[RETURN_RESOLVE]: 'Prefer `return value` over `return Promise.resolve(value)`.',
```
fisker, line 9: We can use just one message, 

```text
'Prefer `{{replacement}} {{errorOrValue}}` over `{{original}} Promise.reject({{errorOrValue}})`.'
```
```js
return {
	messageId: MESSAGE_ID,
	data: {
		original: 'yield',
		replacement: 'throw',
		errorOrValue: 'error', // or 'value'
	}
}
```

fisker:
@cherryblossom000 I did a refactor, I this the fix logic is more clear, hope it's fine.

cherryblossom000:
@fisker Looks good, thanks for the refactor.

fisker:
One problem to fix

```js
async () => (( Promise.reject(error) ));
```

fisker:
Sorry. one more

```js
async function * foo() {
	(( yield Promise.reject(error) ));
}
```

fisker (approved):
Looks good, thank you!

sindresorhus:
@cherryblossom000 Thanks for contributing :)

diegocr:
> The use of Promise.reject isn't 'useless' here, but IMO it's better to refactor the code so it uses throw

What about cases where we may want to reject the async with a literal, but having `no-throw-literal` enabled doesn't allow us to use `throw` here? :)

cherryblossom000:
@diegocr I think you should only reject a promise with an `Error`, for the same reasons as detailed in the [`no-throw-literal` docs](https://eslint.org/docs/latest/rules/no-throw-literal):

> It is considered good practice to only `throw` the `Error` object itself or an object using the `Error` object as base objects for user-defined exceptions. The fundamental benefit of `Error` objects is that they automatically keep track of where they were built and originated.

A rejected promise should indicate an error. Rejecting a promise is the async equivalent of `throw`ing an error.

If you’re doing something like `return Promise.reject('oops an error!')`, you can do `throw new Error('oops an error!')` instead.

diegocr:
Hi @cherryblossom000, i actually meant a numerical literal, so that later we can do quick strict comparisons.

[for example](https://github.com/meganz/webclient/blob/4dc5a92109a247b9751a21a4325af7f172a7b4b4/js/utils/network.js#L27):

```js
async foo(offset) {
  if (offset < 1) return Promise.reject(ERANGE);
  ....
}
...
foo().catch((err) => {
  if (err === ERANGE) bar();
  ....
});
```

Perhaps the rule could be improved through options to whitelist certain uses, meanwhile i've enabled it as a Warning only, but would like it as an Error to enforce on my team mates getting rid of Promise.resolve() there.

fisker:
`const ERANGE = new RangeError('Messge')`?

fisker:
```js
class CustomError extends Error {
  ERANGE = true
}

throw new CustomError()


foo().catch((err) => {
  if (err?.ERANGE) bar();
  ....
})

diegocr:
Nice idea @fisker, except `ERANGE` ([amongst others](https://github.com/meganz/webclient/blob/62c9020f761a0ef479da440986a38b25134efa7d/js/crypto.js#L72)) could also be returned from server-side indicating a failure to an specific API request, and tbh i am not a big fan of the overhead towards creating custom classes for something like this :)

cherryblossom000:
If I were you I would disable the `no-throw-literal` rule.

diegocr:
Yeah, that might be worth considering as well.

Alright, thanks guys for your time and help in any case :)

PR #1628, opened 2021-12-02
`filename-case`: Ignore `$` in filenames

### LINKED ISSUE #1627: $file-name.tsx fails the kebab-case rule (filename-case)
I have a file `$file-name.tsx` and `filename-case` rule turned on.

I have an error saying: Filename is not in kebab case. Rename it to `file-name.tsx`

matmilbury:
Fixes #1627 

There is no test case that demonstrates why `$` should not be ignored.

### DIFF
--- a/test/filename-case.mjs
+++ b/test/filename-case.mjs
@@ -379,6 +379,16 @@ test({
 			undefined,
 			'Filename is not in kebab case. Rename it to `[foo-bar].js`.',
 		),
+		testCase(
+			'src/foo/$foo_bar.js',
+			undefined,
+			'Filename is not in kebab case. Rename it to `$foo-bar.js`.',
+		),
+		testCase(
+			'src/foo/$fooBar.js',
+			undefined,
+			'Filename is not in kebab case. Rename it to `$foo-bar.js`.',
+		),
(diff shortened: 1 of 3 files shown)


test/filename-case.mjs
```diff
@@ -79,6 +79,7 @@ test({
 		testCase('src/foo/___foo-bar.js', 'kebabCase'),
 		testCase('src/foo/_FooBar.js', 'pascalCase'),
 		testCase('src/foo/___FooBar.js', 'pascalCase'),
+		testCase("src/foo/$foo.js"),
```
fisker, line 82: Can you also add an invalid case with filename contains `$`?

matmilbury: `$` in the filename will never cause the filename to be invalid because this character has nothing to do with casing.

sindresorhus (approved):

test/filename-case.mjs
fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1628/commits/67737fd0bdcae1457e6572dce38fb6e57fb27dc1?

PR #1635, opened 2021-12-14
Add `no-duplicate-literal-elements` rule

### LINKED ISSUE #1559: Rule proposal: `no-set-duplicates`
Sometimes a hardcoded `Set` grows large enough for it to be hard to see if there is already a value in it. One could miss that a value one wants to add is already there and add a duplicate.

This rule should at minimum report hardcoded sets of literals and `const`s with duplicates.

This rule may also report consecutive `.add` calls with same arguments (this is probably much harder to implement). Or maybe this could be a separate rule like `unicorn/no-array-push-push`.

## Fail

```js
const allowed = new Set([
	'foo',
	'bar',
	'buz',
	'qux',
	'bar', // ← Remove duplicate value `'bar'` from the Set
]);
```

## Pass

```js
const allowed = new Set([
	'foo',
	'bar',
	'buz',
	'qux',
]);
```

SidStraw:
add `no-duplicate-literal-elements` rule to fixes #1559


docs/rules/no-duplicate-literal-elements.md
```diff
@@ -0,0 +1,35 @@
+# Hardcoded `Set` and `Map` of literals and consts with duplicates
+
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+Sometimes a hardcoded Set grows large enough for it to be hard to see if there is already a value in it. One could miss that a value one wants to add is already there and add a duplicate.
+
+This rule report hardcoded `Set` and `Map` of literals and consts with duplicates.
```
sindresorhus, line 7: Not clear what you mean by "consts" here.

docs/rules/no-duplicate-literal-elements.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Fail
+
+```js
+const foo = new Set([1, 2, 1]);
+const foo = new Set([1, null, null]);
+const foo = new Set(["1", "2", "1"]);
```
sindresorhus, line 14: Use single-quotes

docs/rules/no-duplicate-literal-elements.md
```diff
@@ -0,0 +1,35 @@
+# Hardcoded `Set` and `Map` of literals and consts with duplicates
```
sindresorhus, line 1: Needs a better description.

sindresorhus:
Thanks for working on this.

sindresorhus:
It should handle array too: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1559#issuecomment-951072358

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
+const MESSAGE_ID = 'no-duplicate-literal-elements';
+const messages = {
+	[MESSAGE_ID]: 'Hardcoded `{{type}}` of `{{value}}` with duplicates.',
+};
+
+const setSselector = [
```
sindresorhus, line 7: Typo?

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,73 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Hardcoded `Set` and `Map` of literals and consts with duplicates',
```
sindresorhus, line 69: This does not describe what the suggestion does. Please look at other rules for inspiration.

test/no-duplicate-literal-elements.mjs
```diff
@@ -0,0 +1,27 @@
...
+			const map = new Map(Object.entries(obj))
+		`,
+	],
+	invalid: [
+		'const foo = new Set([1, 2, 1])',
+		'const foo = new Set([1, null, null])',
+		'const foo = new Set(["1", "2", "1"])',
+		'const foo = new Map([["a", 1], ["a", 2]])',
+	],
+});
```
sindresorhus, line 27: Need a lot more tests.

sindresorhus (changes requested):

SidStraw:
> It should handle array too: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1559#issuecomment-951072358

OK, thx.

I will make up.

docs/rules/no-duplicate-literal-elements.md
kayac-chang: @SidStraw 
你這邊寫的 "consts" 是指什麼，意思不太明確

docs/rules/no-duplicate-literal-elements.md
kayac-chang: @SidStraw 
請使用 單引號

docs/rules/no-duplicate-literal-elements.md
kayac-chang: @SidStraw 
這邊可以想一個更好的描述

rules/no-duplicate-literal-elements.js
kayac-chang: @SidStraw 
打錯字了

rules/no-duplicate-literal-elements.js
kayac-chang: @SidStraw 
這段 建議的內文 沒有辦法描述的很清楚
可以參考其他規則來刺激一下靈感

test/no-duplicate-literal-elements.mjs
kayac-chang: @SidStraw 
需要建立更多的測試

kayac-chang:
Hello @sindresorhus, 
We have some question about `CI / lint-test` and `CI / integration` which includes some failing test.
But not sure how to fix.
Could you provide some suggestion for that, Thank you every much!!!

sindresorhus:
@kayac-chang Please write your comments in English. It's fine to use Google Translate, but you should not force everyone else to have to do that.

docs/rules/no-duplicate-literal-elements.md
SidStraw: I have updated the content of the file

docs/rules/no-duplicate-literal-elements.md
SidStraw: Adjusted and added case

docs/rules/no-duplicate-literal-elements.md
SidStraw: I have updated the description

rules/no-duplicate-literal-elements.js
SidStraw: Yes, it's a typo.
Fixed.

test/no-duplicate-literal-elements.mjs
SidStraw: Test case has been added

rules/no-duplicate-literal-elements.js
SidStraw: I have updated the description

SidStraw:
@sindresorhus  Thank you for your review.
I have made corrections and added Array checks, doc updates, and new tests.


But I have the same problem as @kayac-chang 
It's about `CI / lint-test` and `CI / integration` which includes some failing tests.
But not sure how to fix it.
Could you provide some suggestions for that, Thank you very much.


By the way, my English is not good.
So if there is something wrong with the text,
please let me know or provide suggestions,
thank you very much.

test/snapshots/no-duplicate-literal-elements.mjs.md
```diff
@@ -0,0 +1,329 @@
...
+Generated by [AVA](https://avajs.dev).
+
+## Invalid #1
+      1 | const foo = new Set([1, 2, 1])
+
+> Error 1/1
+
+    `␊
+    > 1 | const foo = new Set([1, 2, 1])␊
+        |                     ^^^^^^^^^ Hardcoded \`Set\` of \`1\` with duplicates.␊
```
fisker, line 14: Can we just point to the duplicate item?

```text
    > 1 | const foo = new Set([1, 2, 1])␊
        |                            ^ A message says it's duplicated.
```

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,110 @@
...
+		const duplicatedData = checkArrayHasDuplicatedValue(arrayValue);
+		if (duplicatedData.length > 0) {
+			return {
+				node,
+				messageId: MESSAGE_ID,
+				data: {
+					type: 'Map',
+					value: duplicatedData
+						.map(value => value === null ? 'null' : value)
+						.join(', '),
```
fisker, line 93: We can use the raw string of the literal here, but if we decided to do https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1635/files#r770189903, it won't be necessary.

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,110 @@
...
+	'ArrayExpression',
+].join('');
+
+const setSelector = [
+	'NewExpression',
+	'[arguments.0.type="ArrayExpression"]',
+	'[callee.name="Set"]',
+	' > ',
+	'ArrayExpression',
+].join('');
```
fisker, line 27: You can use `newExpressionSelector` from `rules/selectors/index.js`, it's stricter than this.

test/snapshots/no-duplicate-literal-elements.mjs.md
SidStraw: In that case I have to check for duplicates in the selector stage.

I have no idea about this part.

Can you provide some direction?

test/empty-brace-spaces.mjs
```diff
@@ -22,7 +22,6 @@ const cases = [
 	'while (foo){/* */}',
 	'foo = () => {/* */}',
 	'foo = function (){/* */}',
-	'function foo(){/* */}',
```
fisker, line 25: Please add this back.

rules/no-duplicate-literal-elements.js
SidStraw: Thanks, I will go to confirm

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,110 @@
...
+	'ArrayExpression',
+	' , ',
+	'ExpressionStatement',
+	' > ',
+	'ArrayExpression',
+	' , ',
+	'AssignmentExpression',
+	' > ',
+	'ArrayExpression',
+].join('');
```
fisker, line 19: This is really bad... I didn't understand it at first.

```suggestion
const arraySelector = [
	'VariableDeclarator > ArrayExpression',
	'ExpressionStatement > ArrayExpression',
	'AssignmentExpression > ArrayExpression',
].join(',');
```

test/empty-brace-spaces.mjs
SidStraw: This is based on the message of `npm run run-rules-on-codebase`
This line is repeated on line 9
Why do I need to add it back？

rules/no-duplicate-literal-elements.js
SidStraw: This is really better, thanks.

test/snapshots/no-duplicate-literal-elements.mjs.md
fisker: Select `ArrayExpression > Literal`, check if value exists in siblings before it.

rules/no-duplicate-literal-elements.js
SidStraw: Okay, thank you

test/snapshots/no-duplicate-literal-elements.mjs.md
fisker: This will cause multiple messages on the same array, maybe we can simply change your current implementation to report on the first duplicated item.

You can use `.find()` to find the first one.

SidStraw: Great!
I will try it.

test/empty-brace-spaces.mjs
fisker: Sorry, didn't know that. Seems this rule already helping! 👍

test/snapshots/no-duplicate-literal-elements.mjs.md
SidStraw: @fisker  just report first duplicate index now.

SidStraw:
This rule just reported the first duplicate index now.

And I have tried to fix the CI fail.

SidStraw (commented):
These problems have been fixed in the current version.

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,116 @@
...
+const MESSAGE_ID = 'no-duplicate-literal-elements';
+const messages = {
+	[MESSAGE_ID]: 'Remove duplicate {{valueType}} `{{value}}` from the `{{objectType}}`.',
+};
+
+const arraySelector = [
+	'VariableDeclarator > ArrayExpression > Literal',
+	'ExpressionStatement > ArrayExpression > Literal',
+	'AssignmentExpression > ArrayExpression > Literal',
+].join(',');
```
fisker, line 13: I'm not sure about this, maybe we should check all arrays? Not just declaration and assignments?

@sindresorhus @futpib

rules/no-duplicate-literal-elements.js
```diff
@@ -0,0 +1,116 @@
...
+].join('');
+
+const mapSelector = [
+	newExpressionSelector('Map'),
+	'[arguments.0.type="ArrayExpression"]',
+	' > ',
+	'ArrayExpression > ArrayExpression > Literal:first-child',
+].join('');
+
+const checkArrayHasDuplicatedValue = array => array.filter((element, index, array_) => array_.indexOf(element) !== index);
```
fisker, line 29: This function can directly use `array.find` to find the duplicated element, so we don't need to do `array.includes()`.

rules/no-duplicate-literal-elements.js
SidStraw: If I check all arrays, do I need to rule out certain situations?

For example: 
`Object.entries()` method returns an array of a given object's own enumerable string-keyed property [key, value] pairs

fisker: I understand that, I'm not sure.

fisker:
I don't think this rule should limited to Literals,

```js
const set = new Set([
	undefined,
	undefined,
	foo,
	foo,
	foo.bar,
	foo.bar,
])
```

This is obviously duplicated, I think we can try analyze the value, instead of only checking literals.

rules/no-duplicate-literal-elements.js
SidStraw: > This function can directly use `array.find` to find the duplicated element, so we don't need call `array.includes()` and `getFirstDuplicateIndex()`.

Oh! I see now.
Ok, I will try it later.

SidStraw:
> I don't think this rule should limited to Literals,
> 
> ```js
> const set = new Set([
> 	undefined,
> 	undefined,
> 	foo,
> 	foo,
> 	foo.bar,
> 	foo.bar,
> ])
> ```
> 
> This is obviously duplicated, I think we can try analyze the value, instead of only checking literals.

This is an interesting suggestion, we need to confirm with @sindresorhus and @futpib .

And do I need to update the rule name?

rules/no-duplicate-literal-elements.js
SidStraw: I have refactored.
thx for your suggestion.

sindresorhus:
> This is an interesting suggestion, we need to confirm with @sindresorhus and @futpib .

Yeah, makes sense.

> And do I need to update the rule name?

Maybe `no-duplicate-elements`.

rules/no-duplicate-literal-elements.js
sindresorhus: > I'm not sure about this, maybe we should check all arrays

👍

docs/rules/no-duplicate-literal-elements.md
```diff
@@ -0,0 +1,149 @@
+# Disallow duplicate literal elements in `Array`, `Set` or `Map` key
+
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+It is hard to know if there is duplicate in hardcoded value when `Array`, `Set`, or `Map` key grows larger.
```
haykam821, line 5: This description seems vague to me. Sets and maps make sense for this rule since they do not allow duplicate elements or keys respectively, so this description should mention that. What would be the reason for arrays?

SidStraw: Actually, I’m not sure why I need to check the arrays...

There are many cases where the contents of the array are duplicated.
like this: `((x, y) => Math.pow(x, y)).apply(null, [2, 2])`

I will confirm how to rewrite the description, thx~

sindresorhus:
Bump :)

SidStraw:
Sorry, I didn't ignore this PR.
It's just that I don't have a way forward.

sindresorhus:
> It's just that I don't have a way forward.

What are you stuck on?

sindresorhus:
Closing for lack of activity.

PR #1642, opened 2021-12-17
`no-array-callback-reference`: Ignore primitive wrappers in `Array#map()`

fisker:
All these primitive wrappers only accept one argument, they will never change, it's safe to use in `Array#map()`

### DIFF
--- a/rules/no-array-callback-reference.js
+++ b/rules/no-array-callback-reference.js
@@ -49,14 +49,10 @@ const iteratorMethods = [
 	],
 	[
 		'flatMap',
-		{
-			ignore: [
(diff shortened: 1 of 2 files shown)


rules/no-array-callback-reference.js
```diff
@@ -49,14 +49,10 @@ const iteratorMethods = [
 	],
 	[
 		'flatMap',
-		{
-			ignore: [
-				'Boolean',
-			],
-		},
```
fisker, line 56: I made a mistake, this is not reasonable to use in `Array#flatMap()`, it should return an array.

PR #1648, opened 2021-12-21
`no-new-array` & `no-new-buffer`: Improve argument type detection

### LINKED ISSUE #1534: Improve `no-new-array`
Currently, when we know the argument is a number, we can auto fix the problem. 

But only work for simple cases, we should also smart enough to know cases like `new Array(Math.min(unkown1, unknown2))` is calling with a number.

The following are numbers:

- all `Math` method calls, and `Math` properties
- `Number(unkown)`
- `Number.parseInt` / `Number.parseFloat` (global `parseInt` already checked by `unicorn/prefer-number-properties`)
- `BinaryExpression`s (`-`, `*` , `/`, `%`), but NOT `+` operator, it could be `string`
- `AssignmentOperator`s (`-=`, `*=`, `/=`, `%=`)
- `unknown ++`, `unknown --`, `++ unknown`, `-- unknown`
- `UnaryOperator`s (`+unknown`, `-unknown`, `~unknown`)
- Bitwise operators
- Else ?

This will also make other rules like `no-new-buffer` auto-fix more cases.

fisker:
Fixes #1534


sindresorhus:
Nice improvement 👍

PR #1654, opened 2021-12-22
`no-array-push-push`: Ignore `process.{stdin,stdout,stderr}`

fisker:
@sindresorhus  I'm not familiar with the `Stream` object, are they both reasonable to use?


sindresorhus:
You would only use `Stream`, which is actually `stream.Stream`, for instanceof checking.

Issue #1492, opened 2021-08-18
Rule proposal: `no-typeof-undefined`


B2o5T:
Less code written

## Fail

```js
typeof foo === 'undefined'
```

## Pass

```js
foo === undefined
```

fisker:
How about `consistent-undefined-check`, allow one of these two? Personally I always use `typeof`.

fisker:
When it's undefined variable, `foo === undefiend` will throw 

```text
foo === undefined
VM227:1 Uncaught ReferenceError: foo is not defined
    at <anonymous>:1:1
```

B2o5T:
> When it's undefined variable, `foo === undefiend` will throw
> 
> ```
> foo === undefined
> VM227:1 Uncaught ReferenceError: foo is not defined
>     at <anonymous>:1:1
> ```

in node REPL, yes, but in real projects there is always enabled `no-undef` ESLint rule from `eslint:recommended` config, so this case will be reported

fisker:
I don't think `no-undef` is safe, for example, when you write js for brower, the following won't report, because `AbortController` is considered "defined" as globals.

```js
if (AbortController !== undefined) {}
```

But when you run this code in old browsers, it will throw.

```js
if (typeof AbortController !== 'undefined') {}
```

This won't.

Of cause, `'AbortController' in globalThis` or `globalThis.AbortController` are safer. But nothing wrong to use `typeof`.

B2o5T:
Ok, but we can check if the variable is declared in scope with `var`, `let`, or `const` keywords, if so, inform the suggestion, otherwise do nothing

sindresorhus:
This is now accepted.

sindresorhus:
> How about consistent-undefined-check, allow one of these two? Personally I always use typeof.

You are potentially silently hiding bugs by always using `typeof` though. I think it's generally better to only use it for things that are known to not exist, like when polyfilling. Any other times, IMHO, not using `typeof` is better.

fisker:
I'm going to work on this. And I'm going to add an option to ignore global variables.

fisker:
I've sent #1966, we'll check all `typeof ANYTHING === 'undefined'` by default.

When set `{checkGlobalVariables: false}`, the rule will ignore variables if it's not defined in file, including unresolved and known globals. `typeof foo === 'undefined'` and `typeof Array === 'undefined'` are both valid.

WDYT? @zloirock @papb

[...]

PR #1657, opened 2021-12-22
Add `no-typeof-undefined`

B2o5T:
fix https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1492


test/no-typeof-undefined.mjs
```diff
@@ -0,0 +1,17 @@
...
+		'typeof notDeclared === "undefined"',
+	],
+	invalid: [
+		'var withVar; typeof withVar == "undefined"',
+		'let withLet; "undefined" != typeof withLet',
+		'const withConst = 0; typeof withConst === "undefined"',
+		'import withImport from "foo"; "undefined" !== typeof withImport',
+		'withArrowFnParameter => typeof withArrowFnParameter === "undefined"',
+		'function foo(withFnParameter) { "undefined" !== typeof withFnParameter }',
+	],
```
sindresorhus, line 16: This is gonna need a lot more valid/invalid test cases.

docs/rules/no-typeof-undefined.md
```diff
@@ -0,0 +1,74 @@
...
+
+```js
+let withLet;
+withLet != undefined;
+```
+
+```js
+const withConst = 'foo';
+withConst === undefined;
+```
```
sindresorhus, line 58: This one is enough. No need to show the example with both const, let, var.

docs/rules/no-typeof-undefined.md
```diff
@@ -0,0 +1,74 @@
+# Forbid of comparison `typeof` with `'undefined'` if the variable is declared, imported or variable is a function parameter
```
sindresorhus, line 1: This needs to be better written.

docs/rules/no-typeof-undefined.md
```diff
@@ -0,0 +1,74 @@
...
+
+🔧 _This rule
+is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems)._
```
sindresorhus, line 4: Incorrect linebreak

sindresorhus (changes requested):

sindresorhus:
Thanks for working on this.

docs/rules/no-typeof-undefined.md
B2o5T: removed

docs/rules/no-typeof-undefined.md
B2o5T: Feel free to share your vision :)

test/no-typeof-undefined.mjs
B2o5T: added new valid/invalid tests, you can share your code snippets also

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,61 @@
...
+].join('');
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	* [selector](node) {
+		let variable = node.argument;
+		if (variable.type === 'MemberExpression') {
+			variable = variable.object;
+		} else if (variable.type === 'ChainExpression') {
+			variable = variable.expression.object;
```
B2o5T, line 23: is there is some generic function for this case?

docs/rules/no-typeof-undefined.md
sindresorhus: I suggest you give it a try first.

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,61 @@
+'use strict';
+const {findVariable} = require('eslint-utils');
+const {matches} = require('./selectors/index.js');
+
+const MESSAGE_ID = 'NO_TYPEOF_UNDEFINED';
+const messages = {
+	[MESSAGE_ID]: 'Prefer `{{operator}}` operator instead of typeof with \'undefined\'.',
```
sindresorhus, line 7:
```suggestion
	[MESSAGE_ID]: 'Prefer `{{operator}}` operator instead of `typeof` with `\'undefined\'`.',
```

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const selector = [
+	'BinaryExpression',
+	matches(['[right.value=undefined]', '[left.value=undefined]']),
```
fisker, line 12: We need check it's string literal

```suggestion
matches(['[right.type="Literal"][right.value="undefined"]', '[left.type="Literal"][left.value="undefined"]']),
```

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,61 @@
...
+	' UnaryExpression[operator=typeof]',
```
fisker, line 13: The `UnaryExpression` should  be  a direct child of `BinaryExpression`.

```suggestion
	' > UnaryExpression[operator="typeof"]',
```

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,61 @@
...
+].join('');
```
fisker, line 14: We need check operator too.

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,61 @@
...
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	* [selector](node) {
+		let variable = node.argument;
+		if (variable.type === 'MemberExpression') {
+			variable = variable.object;
+		} else if (variable.type === 'ChainExpression') {
+			variable = variable.expression.object;
+		}
```
fisker, line 24: This should be recursive, `foo?.bar.baz`

Issue #996, opened 2021-01-05
`prefer-array-some`: Work through variable assignment


Exodustx0:
The `prefer-array-some` rule is not triggered when the `.find()` result is assigned to a variable:
```js
const arr = [];
const hasFoo = arr.find(val => val === 'foo');
if (hasFoo) console.log('foo');
```

Obviously it shouldn't be triggered if `hasFoo` were to be used elsewhere for anything other than element existence checks, but if that's all it's used for, I'd expect it to trigger the rule.

Variable assignment of these element existence checks is useful when the predicate function is long, but also to generally make code read more prose-like.

fisker:
If someone want implement this, I prefer update this function https://github.com/sindresorhus/eslint-plugin-unicorn/blob/dc7f79b0e0a87d7dc009905c9dc76719d9f5a3ff/rules/utils/boolean.js#L29

So, other rules can apply the same logic

sindresorhus:
If anyone wants to work on this, see the initial attempt and feedback in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1662.

PR #1662, opened 2021-12-27
improve `isBooleanNode()` to fix false negative error in `prefer-array-some` rule

B2o5T:
fix https://github.com/sindresorhus/eslint-plugin-unicorn/issues/996

### DIFF
--- a/rules/utils/boolean.js
+++ b/rules/utils/boolean.js
@@ -75,9 +75,14 @@ function isBooleanNode(node, scope) {
 
 	if (scope && parent.type === 'VariableDeclarator') {
 		const reference = findReference(scope, parent.id);
+
+		if (!reference) {
+			return false;
+		}
+
 		// Skip variable declaration of current node
-		const references = reference.resolved.references.filter(ref => {
-			const parentNode = ref.identifier.parent;
+		const references = reference.resolved.references.filter(r => {
+			const parentNode = r.identifier.parent;
 			return parentNode.type !== 'VariableDeclarator' || parentNode.id !== parent.id;
 		});
 
@@ -85,7 +90,7 @@ function isBooleanNode(node, scope) {
 			return false;
 		}
 
-		return references.every(ref => isBooleanNode(ref.identifier));
+		return references.every(r => isBooleanNode(r.identifier));
 	}
 
 	return false;
--- a/rules/utils/boolean.js
+++ b/rules/utils/boolean.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const isLogicalExpression = require('./is-logical-expression.js');
+const findReference = require('./find-reference.js');
 
 const isLogicNot = node =>
 	node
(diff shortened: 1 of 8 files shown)


test/prefer-array-some.mjs
```diff
@@ -72,20 +77,14 @@ test({
...
-					message: 'Prefer `.some(…)` over `.find(…)`.',
-					suggestions: [
-						{
-							desc: 'Replace `.find(…)` with `.some(…)`.',
-							output: 'if (foo.some(fn)) {}',
-						},
-					],
-				},
-			],
-		},
```
fisker, line 88: This test mean to test the actual message, please revert.

rules/utils/boolean.js
```diff
@@ -71,6 +73,26 @@ function isBooleanNode(node) {
 		return isBooleanNode(parent);
 	}
 
+	if (scope && parent.type === 'VariableDeclarator') {
+		const reference = findReference(scope, parent.id);
```
fisker, line 77: The `parent.id` can be other things, not only `Identifier`, we need tests.

test/prefer-array-some.mjs
B2o5T: I know, I made a refactoring to use `invalidCase` function instead

fisker: Message not asserted in that function, only messageId.

B2o5T: asserted on line 82

fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/635 This is why we need those tests.

m0dul3 (approved):

m0dul3 (approved):

m0dul3 (approved):

m0dul3 (approved):

sindresorhus:
@B2o5T Friendly bump

rules/utils/boolean.js
B2o5T: Could you suggest me some tests that I need to add?

fisker:
```js
const {} = foo;
const [] = foo;
```

rules/utils/find-reference.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+/**
+ * Finds the eslint-scope reference in the given scope.
+ * @param {Object} scope The scope to search.
+ * @param {ASTNode} node The identifier node.
+ * @returns {Reference|undefined} Returns the found reference or null if none were found.
+ */
+function findReference(scope, node) {
```
fisker, line 9: Better to use `getVariable` from `eslint-utils`.

sindresorhus:
@B2o5T Bump

B2o5T:
@sindresorhus @fisker sorry, don’t have a time to finish it, if someone want give a try, feel free

PR #1666, opened 2021-12-28
Detect unnecessary `Promise.resolve`/`reject` in promise callback functions

### LINKED ISSUE #1609: Rule proposal: No unnecessary `Promise.resolve/reject()` in async functions

cherryblossom000:
Implements https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1609#issuecomment-970571215

Fixes  #1609

### DIFF
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -17,7 +17,7 @@ const selector = [
 	matches([
 		'ArrowFunctionExpression > .body',
 		'ReturnStatement > .argument',
-		'YieldExpression[delegate=false] > .argument',
+		'YieldExpression[delegate!] > .argument',
 	]),
 ].join('');
 
--- a/rules/no-useless-promise-resolve-reject.js
+++ b/rules/no-useless-promise-resolve-reject.js
@@ -40,26 +40,9 @@ function getFunctionNode(node) {
 		}
 	}
 
-	let isInPromiseThenOrCatchCallback = false;
-	if (
-		node
-		&& node.parent
-		&& node.parent.type === 'CallExpression'
-		&& node.parent.callee.type === 'MemberExpression'
-		&& node.parent.callee.property.type === 'Identifier'
-	) {
-		const {callee: {property}, arguments: arguments_} = node.parent;
-		if (property.name === 'then') {
-			isInPromiseThenOrCatchCallback = arguments_[0] === node || arguments_[1] === node;
-		} else if (property.name === 'catch') {
-			isInPromiseThenOrCatchCallback = arguments_[0] === node;
-		}
-	}
-
 	return {
 		functionNode,
 		isInTryStatement,
-		isInPromiseThenOrCatchCallback,
 	};
 }
 
@@ -76,16 +59,17 @@ function createProblem(callExpression, fix) {
 	};
 }
 
-function fix(callExpression, isReject, isInTryStatement, sourceCode) {
+function fix(callExpression, isInTryStatement, sourceCode) {
 	if (callExpression.arguments.length > 1) {
 		return;
 	}
 
-	const {parent, arguments: [errorOrValue]} = callExpression;
+	const {callee, parent, arguments: [errorOrValue]} = callExpression;
 	if (errorOrValue && errorOrValue.type === 'SpreadElement') {
 		return;
 	}
 
+	const isReject = callee.property.name === 'reject';
 	const isYieldExpression = parent.type === 'YieldExpression';
 	if (
 		isReject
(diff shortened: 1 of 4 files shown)


rules/no-useless-promise-resolve-reject.js
```diff
@@ -40,9 +40,26 @@ function getFunctionNode(node) {
...
+		&& node.parent.type === 'CallExpression'
+		&& node.parent.callee.type === 'MemberExpression'
+		&& node.parent.callee.property.type === 'Identifier'
+	) {
+		const {callee: {property}, arguments: arguments_} = node.parent;
+		if (property.name === 'then') {
+			isInPromiseThenOrCatchCallback = arguments_[0] === node || arguments_[1] === node;
+		} else if (property.name === 'catch') {
+			isInPromiseThenOrCatchCallback = arguments_[0] === node;
+		}
```
fisker, line 56: `Promise#finally()` should also checked as welll.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -141,14 +157,15 @@ const create = context => {
 
 	return {
 		[selector](callExpression) {
-			const {functionNode, isInTryStatement} = getFunctionNode(callExpression);
-			if (!functionNode || !functionNode.async) {
+			const isReject = callExpression.callee.property.name === 'reject';
+			const {functionNode, isInTryStatement, isInPromiseThenOrCatchCallback} = getFunctionNode(callExpression);
+			if (!functionNode || (!functionNode.async && (!isInPromiseThenOrCatchCallback || isReject))) {
```
fisker, line 162: I don't understand this part, why we should care if it's reject here?

rules/no-useless-promise-resolve-reject.js
```diff
@@ -40,9 +40,26 @@ function getFunctionNode(node) {
 		}
 	}
 
+	let isInPromiseThenOrCatchCallback = false;
```
fisker, line 43: The logic is the function is async or it's promise callback, we run check.

rules/no-useless-promise-resolve-reject.js
cherryblossom000: My bad, for some reason I thought `promise.catch(() => Promise.reject(error))` would be behave differently to `promise.catch(() => { throw error })`.

rules/no-useless-promise-resolve-reject.js
```diff
@@ -142,7 +142,27 @@ const create = context => {
...
+			if (!functionNode) {
+				return;
+			}
+
+			let isInPromiseThenOrCatchCallback = false;
+			if (
+				functionNode.parent
+				&& functionNode.parent.type === 'CallExpression'
+				&& functionNode.parent.callee.type === 'MemberExpression'
+				&& functionNode.parent.callee.property.type === 'Identifier'
```
fisker, line 155: `&& !functionNode.parent.callee.computed`

rules/no-useless-promise-resolve-reject.js
```diff
@@ -142,7 +142,27 @@ const create = context => {
...
+			) {
+				const {callee: {property}, arguments: arguments_} = functionNode.parent;
+				if (property.name === 'then') {
+					isInPromiseThenOrCatchCallback = arguments_[0] === functionNode || arguments_[1] === functionNode;
+				} else if (property.name === 'catch' || property.name === 'finally') {
+					isInPromiseThenOrCatchCallback = arguments_[0] === functionNode;
+				}
+			}
```
fisker, line 163: Can you extract this to a function? maybe `isPromiseCallback(functionNode)`

fisker: So we can `if (!functionNode || (!functionNode.async && !isPromiseCallback(functionNode)))`

fisker (approved):

fisker:
@cherryblossom000 We may want to mention promise callback functions in the description and the rule docs.

cherryblossom000:
Would returning a value from the `Promise#finally` callback do anything (if the value isn't a rejected promise)?

```js
> await Promise.resolve(1).finally(() => 2)
1
> await Promise.reject(1).finally(() => 2)
Uncaught 1
```

fisker:
> Would returning a value from the Promise#finally callback do anything

I think it's ignored, but we are not breaking things(I think).

sindresorhus:
Linting is failing

rules/no-useless-promise-resolve-reject.js
```diff
@@ -46,6 +46,26 @@ function getFunctionNode(node) {
...
+		&& !node.parent.callee.computed
+		&& node.parent.callee.property.type === 'Identifier'
+	) {
+		const {callee: {property}, arguments: arguments_} = node.parent;
+		if (property.name === 'then') {
+			return arguments_[0] === node || (arguments_[0].type !== 'SpreadElement' && arguments_[1] === node);
+		}
+
+		if (property.name === 'catch' || property.name === 'finally') {
+			return arguments_[0] === node;
```
j-f1, line 62: I don’t know how common non-Promise then/catch/finally methods are in real code. Do you think it would make sense to do a length/type check on the arguments list to confirm it’s likely to be a valid Promise call?

fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1666/commits/21c2773c6e187c9445bac9d7e2c14826119d022d

PR #1673, opened 2021-12-29
`prefer-dom-node-dataset`: Check `.hasAttribute()` and `.getAttribute()`

### LINKED ISSUE #1669: Check `.hasAttribute()` in `prefer-dom-node-dataset` rule
[MDN suggest using `in` operator to check existence](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#accessing_values), I'm not sure what's the best, maybe `hasOwnProperty()`?

## Fail

```js
element.hasAttribute('data-foo');
```

## Pass

```js
"foo" in element.dataset;
```

fisker:
Fixes #1669

I just realized that we are not checking `.getAttribute()`, any reason behind this? I need a break...


rules/prefer-dom-node-dataset.js
```diff
@@ -54,7 +68,7 @@ module.exports = {
 	meta: {
 		type: 'suggestion',
 		docs: {
-			description: 'Prefer using `.dataset` on DOM elements over `.setAttribute(…)` and `.removeAttribute(…)`.',
+			description: 'Prefer using `.dataset` on DOM elements over accessing attributes.',
```
sindresorhus, line 71:
```suggestion
			description: 'Prefer using `.dataset` on DOM elements over calling attribute methods.',
```

?

sindresorhus:
> I just realized that we are not checking .getAttribute(), any reason behind this? I need a break...

No reason other than it not being thought of when the rule was initially added: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/166

rules/prefer-dom-node-dataset.js
```diff
@@ -46,6 +46,7 @@ const create = context => ({
 			}
 
 			case 'hasAttribute':
+				// TODO: Use `Object.hasOwn()` here when targeting Node.js 16.
```
fisker, line 49: Arr you sure you want this comment? This code should running in browser.

sindresorhus: `Object.hasOwn()` is already in Chrome, Safari, and Firefox.

PR #1676, opened 2021-12-30
Add `prefer-json-parse-buffer` rule

### LINKED ISSUE #1273: Rule proposal: `prefer-json-parse-buffer`
When parsing a json file, there is no need to read it as string, because `JSON.parse` can parse `Buffer`.  [Real world case ](https://github.com/fisker/esm-utils/commit/04ea15c4dca43b8ebd6a9665ce34afaf5e86279d)

## Fail

```js
JSON.parse(await fs.readFile(file, 'utf8'));
```

## Pass

```js
JSON.parse(await fs.readFile(file));
```

```js
JSON.parse(await fs.readFile(file, 'gbk'));
```

This is also faster when use `fsPromises`, but this is bug, should not be the main reason for this rule. https://github.com/nodejs/node/issues/37583

fisker:
Fixes #1273


rules/prefer-json-parse-buffer.js
```diff
@@ -0,0 +1,119 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Prefer read JSON file as buffer.',
```
sindresorhus, line 114:
```suggestion
			description: 'Prefer reading a JSON file as a buffer.',
```

sindresorhus:
It also needs to handle `fs.readFile(file, {encoding: 'utf8'})`.

rules/prefer-json-parse-buffer.js
```diff
@@ -0,0 +1,119 @@
+'use strict';
+const {findVariable, getStaticValue} = require('eslint-utils');
+const {methodCallSelector} = require('./selectors/index.js');
+const {removeArgument} = require('./fix/index.js');
+const getKeyName = require('./utils/get-key-name.js');
+
+const MESSAGE_ID = 'prefer-json-parse-buffer';
+const messages = {
+	[MESSAGE_ID]: 'Prefer read JSON file as buffer.',
```
sindresorhus, line 9:
```suggestion
	[MESSAGE_ID]: 'Prefer reading the JSON file as a buffer.',
```

sindresorhus (changes requested):

sindresorhus:
Perfect :+1:

PR #1678, opened 2021-12-31
Remove empty schema & update create-rule script

fisker:
Empty schema is auto added https://github.com/sindresorhus/eslint-plugin-unicorn/blob/b054d650b2f6545a60e440e400aefc2b092d4e88/rules/utils/rule.js#L122

Update create-rule script,

add tailing comma and common script `require()`, support `hasSuggestions`


sindresorhus:

PR #1683, opened 2022-01-06
`expiring-todo-comments`: Add `date` option

mmkal:
This could be useful a few ways:

1. Testing lint configs
2. Finding todos that expire _soon_ but not yet
3. Testing the rule itself
4. Finding _very_ out of date todos, e.g. by setting the date one year in the past

~@fisker @sindresorhus I'll add tests and docs once we've established the feature is something that would be considered~ Docs and tests added. I was going to make an issue but it was about the same amount of effort to make the (runtime) code change as PR. I did it in the GitHub UI 🙃 

👇 something like what I'll add to the docs

### date

Type: `string` (`date` format)\
Default: `<today>`

For TODOs with date deadlines, this option makes them trigger only if the deadline is later than the specified date. You could set this to a date in the future to find TODOs that expire soon, or set it far in the past if you want to ignore recently-expired TODOs.

The format must match [json-schema's date](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times).

### examples

Find tech debt that has grown up and gone to college by triggering the rule for TODOs with old dates:

```js
"unicorn/expiring-todo-comments": [
	"error",
	{
		"date": "2000-01-01"
	}
]
```

Prepare for the future by triggering the rule for TODOs with dates from the next millennium:

```js
"unicorn/expiring-todo-comments": [
	"error",
	{
		"date": "3000-01-01"
	}
]
```

### DIFF
--- a/docs/rules/expiring-todo-comments.md
+++ b/docs/rules/expiring-todo-comments.md
@@ -325,3 +325,36 @@ As an example of this option, if you want this rule to **completely ignore** com
 	}
 ]
 ```
+
+### date
+
+Type: `string` (`date` format)\
+Default: `<today>`
+
+For TODOs with date deadlines, this option makes them trigger only if the deadline is later than the specified date. You could set this to a date in the future to find TODOs that expire soon, or set it far in the past if you want to ignore recently-expired TODOs.
+
+The format must match [json-schema's date](https://json-schema.org/understanding-json-schema/reference/string.html#dates-and-times).
+
+### examples
+
+Find tech debt that has grown up and gone to college by triggering the rule only for incredibly old TODOs:
+
+```js
+"unicorn/expiring-todo-comments": [
+	"error",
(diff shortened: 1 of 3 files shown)


sindresorhus:
Makes sense to me.

---

I use something similar when I make macOS apps where I can pass a flag to see APIs that will be deprecated in the future.

sindresorhus:
Tests are failing

mmkal:
Not sure why it's failing on CI but not locally. Will debug.

fisker:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/328638191a5db837f0e54122dcc4cf2e1d34bddc/rules/expiring-todo-comments.js#L334 caused by this line?

mmkal:
@fisker that was it - thanks!

fisker (approved):

PR #1686, opened 2022-01-10
Replace nyc with c8

XhmikosR:
https://packagephobia.com/result?p=nyc%2Cc8

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,7 +36,7 @@ jobs:
       - run: npm install
       - run: npm run lint
       # Force update snapshots, https://github.com/avajs/ava/discussions/2754
-      - run: npx nyc ava --update-snapshots
+      - run: npx c8 ava --update-snapshots
         env:
           AVA_FORCE_CI: not-ci
       - run: npm run generate-rules-table
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 node_modules
 yarn.lock
-.nyc_output
 coverage
 package-lock.json
 /test/integration/fixtures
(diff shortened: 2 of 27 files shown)


fisker:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1608#issuecomment-966782691

XhmikosR:
Oh, I wasn't aware of that. Let's close it then and someone else can switch it later.

XhmikosR:
Is it such a big deal, though? I still think it's worth it but I'll leave it to you @fisker @sindresorhus to decide.

Feel free to close it yourselves :)

fisker:
Can you replace `istanbul ignore` with `c8 ignore`?

XhmikosR:
I pushed a patch to replace istanbul with c8, but unsure how it'll handle topics...

fisker:
You should move the ignore reason out https://github.com/istanbuljs/v8-to-istanbul/blob/6c2e2ecd2aece8b01543f75dfa203744f8a785b9/lib/source.js#L54

XhmikosR:
I don't think `ignore next` ignores the next block, so the statements are not equivalent. It just ignores the next line. The workaround is to specify the lines to ignore...

fisker:
There are `ignore start` and `ignore end` we can use.

XhmikosR:
In most cases it should be fine doing `c8 ignore next 3` and it's shorter. Are you OK with this?

fisker:
Yes.

XhmikosR:
All right, this is as good as it gets from my side :)

Feel free to tweak it further if needed.

fisker (approved):
LGTM, Thank You!

PR #1687, opened 2022-01-10
Replace `_.uniq` with native

XhmikosR:
At some point, it might be worth trying to replace the other lodash functions too, they should be ~ 6 cases only.

### DIFF
--- a/rules/utils/get-references.js
+++ b/rules/utils/get-references.js
@@ -1,9 +1,9 @@
 'use strict';
-const {uniq} = require('lodash');
+
 const getScopes = require('./get-scopes.js');
 
-const getReferences = scope => uniq(
+const getReferences = scope => [...new Set(
 	getScopes(scope).flatMap(({references}) => references),
-);
+)];
 
 module.exports = getReferences;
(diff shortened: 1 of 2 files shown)


fisker (approved):

XhmikosR:
This is the list of the remaining lodash function:

1. camelCase
2. kebabCase
3. snakeCase
4. defaultsDeep
5. get
6. lowerFirst
7. pick
8. upperFirst

lowerFirst and upperFirst should be easy to replace later if someone wants to try :)

sindresorhus:
I don't really see the need to replace Lodash functions unless it's super easy to write it in plain JS. In this case, it is. But in many of the mentioned cases, it's not.

XhmikosR:
I know, I know, I'm just saying we should keep in mind the deps while not totally reinventing the wheel. lodash is currently the biggest dependency of eslint-plugin-unicorn, and while it's probably used by other dependencies, it's always good to try to reduce the footprint.

Just my 2 cents.

fisker:
We can replace `lodash` with `lodash.*` package

XhmikosR:
IIRC the individual packages are considered deprecated. I still use them for small things, though, and many others do.

PR #1691, opened 2022-01-11
Allow `React.useHook` in `consistent-function-scoping`

### LINKED ISSUE #1685: False positive in `consistent-function-scoping` when using `React.useEffect`
Since #588, `consistent-function-scoping` is not triggered for the following code:

```ts
import { useEffect } from "react";

useEffect(() => {
  function foo() {}
}, []);
```

This makes sense – see #392. However, when React is imported as a namespace, the same code is no longer valid:

```ts
import * as React from "react";

React.useEffect(() => {
  function foo() {} // Move function 'foo' to the outer scope. eslint(unicorn/consistent-function-scoping)
}, []);
```

Reporting React as a namespace [is OK](https://twitter.com/dan_abramov/status/1308739731551858689). I prefer this method because it produces smaller git diffs and prevents `no-unused-var` when refactoring.

Happy to submit a PR with the fix around next weekend. If anyone wants to create one earlier, please feel free to! 🙌

Relevant condition and its test:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d9fd6642312537d8df5002e99022564bfea776bb/rules/consistent-function-scoping.js#L92-L97

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d9fd6642312537d8df5002e99022564bfea776bb/test/consistent-function-scoping.mjs#L266-L271

kachkaev:
Closes #1685

### DIFF
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -89,7 +89,7 @@ const reactHooks = [
 	'useImperativeHandle',
 	'useLayoutEffect',
 	'useDebugValue',
-].flatMap(hookName=> [hookName, `React.${hookName}`]);
+].flatMap(hookName => [hookName, `React.${hookName}`]);
 
 const isReactHook = scope =>
 	scope.block
--- a/rules/consistent-function-scoping.js
+++ b/rules/consistent-function-scoping.js
@@ -89,18 +89,13 @@ const reactHooks = [
 	'useImperativeHandle',
 	'useLayoutEffect',
 	'useDebugValue',
-];
-
-const reactHookNamesOrPaths = [
-	...reactHooks,
-	...reactHooks.map(hookName => `React.${hookName}`),
-];
+].flatMap(hookName=> [hookName, `React.${hookName}`]);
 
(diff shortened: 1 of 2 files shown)


rules/consistent-function-scoping.js
```diff
@@ -88,13 +89,18 @@ const reactHooks = new Set([
 	'useImperativeHandle',
 	'useLayoutEffect',
 	'useDebugValue',
-]);
+];
```
fisker, line 92:
```js
const reactHooks = [

].flatMap(hookName=> [hookName, `React.${hookName}`]);
```

?

fisker (approved):

rules/consistent-function-scoping.js
kachkaev: Thanks! That looks a bit better 👍

kachkaev:
Thanks for merging this PR @sindresorhus! Would it be feasible to make this fix available in a patch release?

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v40.1.0

PR #1709, opened 2022-01-30
Add `switch-braces` rule

### LINKED ISSUE #1702: Rule proposal: Switch braces
### Description

See https://github.com/xojs/xo/issues/650 for additional context.

### Fail

```js
switch (value) {
		case 'unicorn':
			break;
}
```

### Pass

```js
switch (value) {
		case 'unicorn': {
			break;
		}
}
```

Richienb:
Fixes #1702

### DIFF
--- a/configs/recommended.js
+++ b/configs/recommended.js
@@ -104,6 +104,7 @@ module.exports = {
 		// See #1396
 		'unicorn/require-post-message-target-origin': 'off',
 		'unicorn/string-content': 'off',
+		'unicorn/switch-braces': ['error', 'always'],
 		'unicorn/template-indent': 'warn',
 		'unicorn/throw-new-error': 'error',
 		...require('./conflicting-rules.js').rules,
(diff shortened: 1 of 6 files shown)


Richienb:
The fixer will require some tweaks to get it working nicer with both multiline and single-line expressions. Also some more tests may help.

Richienb:
`no-array-for-each` and `no-useless-spread` are both triggering the rule so that parts seems to work

fisker:
You may misunderstood, the braces creates a new scope, when there are variable declare, braces are required. Correct me if I'm wrong.

Richienb:
Yes, you are correct. The way [MDN describes it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#block-scope_variables_within_switch_statements) makes it sound like more of an inconvenience but this will probably make automatic fixing improbable.

fisker:
My personal opinion, we should only report those cases that braces can be removed.

Richienb:
Stylistically speaking, we'd only want for all blocks to either have braces or blocks.

sindresorhus:
I personally think braces should always be used for consistency and safety.

Richienb:
The rule is built to allow for either always braces (default) or never braces to allow for differing opinions. The whole context thing adds extra steps to take into account variables defined outside of break statements which would throw a runtime error which makes fixing it a manual thing.

sindresorhus:
@Richienb Still interested in finishing this?

PR #1717, opened 2022-02-05
Add `no-unnecessary-polyfills` rule

Mesteery:
Fixes: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/36

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#36: Rule proposal: `no-unnecessary-polyfills`](https://issuehunt.io/repos/55832243/issues/36)
---
</details>


zloirock:
For `core-js@3` / `core-js-pure`, use [`core-js-compat`](https://github.com/zloirock/core-js/tree/master/packages/core-js-compat) - `@babel/compat-data` is not the proper data source for them.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,143 @@
+'use strict';
+const semver = require('semver');
+const readPkgUp = require('read-pkg-up');
+const {data: compatTable} = require('core-js-compat');
```
zloirock, line 4: This is not enough. Most of the features depend on some modules. You could check [`entries`](https://unpkg.com/core-js-compat@3.21.0/entries.json). For example, if you wanna polyfill stable ECMAScript `Set`, you load, for example, `core-js/es/set`. This entry contains those modules:

image

so should be checked data for all of those modules.

Mesteery: If I understood correctly, that's what this regex does in a more general way:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/6eba19eac46aad8ec166f6a95118b6c1914681be/rules/no-unnecessary-polyfills.js#L54-L63

zloirock: Yes, I understand. But `data` is about modules, each feature/entry can contain some modules, for the proper result, it should handle data for all required modules. Your approach makes sense, but you can get this data by related entry, not module, name. For example, `es6-array-copy-within` / `array-copy-within` / etc -> `core-js/stable/array/copy-within` and check all required modules.

Mesteery: I got that idea. But the reason why I preferred to keep this solution is that it's more general and it's not reserved to core-js. Unless I do two checks, one with the regex and one with `entries` but I think it would be overkill.

zloirock: I don't think that two such checks are overkill. The problem is that `@babel/compat-data` is oriented to `core-js@2` and incomplete, `core-js-compat` - to `core-js@3`, and we can't get the proper generic data for all possible polyfills from any of those sources. (For one project, I have a plan to add to `core-js-compat` a list of mapping (like `entries`) of popular third-party polyfills -> similar `core-js` modules, but it will be not very soon.)

The most proper way is usage `core-js-compat` for `core-js` and data from `compat-table` and `caniuse` for polyfills that haven't such data, but it's really overkilling.

zloirock: > If I understand correctly your approach is (to be sure):

Yes, something like that.

zloirock:
@Mesteery for `core-js` - yes, but not for the rest polyfills. I meant that you could use `entries` for the rest polyfills too. For example, you can get the list of features like
```js
Object.keys(coreJsEntries).flatMap(it => (it.startsWith('core-js/actual/') && !/\/(virtual|instance)\//.test(it)) ? it.slice(15) : [])
```
and operate with them like with `core-js` entries.

Mesteery:
Sorry, but I didn't quite understand that.

zloirock:
`core-js` modules do not directly correspond to language features. Entry points - correspond. Many language features depend on some modules, like with `Set` example above. A module is an abstract part of a language feature polyfill.

For example, you have an import of `string-at-polyfill`. With your regex patterns, you can understand that it's an `at` method of `String`. However, `core-js` module for this feature is `es.string.at-alternative` - you can't directly rely on it. However, since you know that it's an `at` method of `String`, you can rely on `core-js/actual/string/at` data and you will see that this feature depends on `es.string.at-alternative` module - and check compact data for this module. If it depends on some modules - you can check all of them, like now it's done in the `core-js` case.

Mesteery:
Like this https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1717/commits/707f4f2b3a5c2a24ff651e267039e1c6840dfc8f?

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		return;
+	}
+
+	const polyfill = polyfills.find(({pattern}) => pattern.test(importedModule));
+	if (polyfill) {
+		const [, namespace, method = ''] = polyfill.feature.split('.');
+		const features = coreJsEntries.find(it => it === `core-js/actual/${namespace}${method && '/'}${method}`);
+		if (features.every(feature => compatData[feature].node && checkVersion(nodeVersion, compatData[feature].node))) {
```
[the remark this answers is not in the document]
Mesteery: Does it make a difference? I think it's good right now. By the way I don't know where `setIntersection` comes from.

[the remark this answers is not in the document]
zloirock: I think that's it's better than only Node support. `setIntersection` is just an abstract name of function intersection of sets.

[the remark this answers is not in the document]
Mesteery: I agree that it is better. Maybe `targets` could be used directly as an option instead of the current Node.js specific option (targetVersion)?

```md
### targets

Type: `string`

The `targets` option allows to specify the target versions. This option is a [Browserlist](https://github.com/browserslist/browserslist) query.
If the option is unspecified, the targets are taken from the `package.json`, in the `browserslist` field, or alternatively, in the `engines.node` field (`targets` will be set to `node {engines.node}`).

``js
"unicorn/no-unnecessary-polyfills": ["error", {	"targets": "node >=12" }]
``

``js
"unicorn/no-unnecessary-polyfills": ["error", {	"targets": ["node 14.1.0", "chrome 95"] }]
``

```

What do you think about this?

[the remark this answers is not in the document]
zloirock: As an option -) Note that `core-js-compat` supports not only Browserlist format, it's almost similar to Babel targets format with some extensions. You can just add a link to the format of `targets` option.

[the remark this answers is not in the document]
Mesteery: Thanks you for your feedback :)

> The `targets` option allows to specify the target versions. This option could be [a Browserlist query](https://github.com/browserslist/browserslist) or a targets object, see https://github.com/zloirock/core-js/tree/master/packages/core-js-compat#targets-option for more informations.

?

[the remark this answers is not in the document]
zloirock: 👍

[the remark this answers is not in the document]
Mesteery: Unfortunately, browserlist does not support this syntax `>=4 <5.2 || >6.0.0` (semver):
``Unknown browser query `node >=4 <5.2 || >6.0.0`. Maybe you are using old Browserslist or made typo in query.``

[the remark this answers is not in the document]
zloirock: The upper bound makes no sense for polyfills. You can use, for example, `minVersion` from `semver` package if you wanna use `package.json` `node` field.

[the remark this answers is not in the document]
zloirock: > It's also unfortunately that only non semver-patch Node.js versions are "supported"

What do you mean?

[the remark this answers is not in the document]
Mesteery: `require('core-js-compat/compat')({ targets: 'node 17.3.1' })` -> `Unknown version 17.3.1 of Node.js`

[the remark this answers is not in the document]
Mesteery: Nevermind, v8 updates (and so new features) are always semver major or minor.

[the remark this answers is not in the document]
zloirock: You can use `require('core-js-compat/compat')({ targets: { node: '17.3.1' } })` instead of `browserslis` query. `browserslis` is too limited.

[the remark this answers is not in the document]
Mesteery: Also, `require('core-js-compat').data['es.array.iterator'].node` is `'10.0'` (expected) while `require('core-js-compat/compat')({ targets: {node:'12'} }).list.includes('es.array.iterator')` is `false` (unexpected).
Shouldn't the function return the features of this version as well as those of previous versions?

`require('core-js-compat/compat')({ targets: 'node <= 12' }).list.includes('es.array.iterator')` is `true` which is also IMO strange because `es.array.iterator` is not available in `node<=10` and so should not be present in the result of this query.

Have I misunderstood the purpose of this function?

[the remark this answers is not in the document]
zloirock: `core-js-compat/compat` returns a list of modules that **required** for the target version, `es.array.iterator` is not required for NodeJS 12.

`node <= 12` on the `browserslist` side normalized to NodeJS 0.2, `es.array.iterator` is required for NodeJS 0.2.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,249 @@
...
+		targets = new SemverNodeVersion(targets);
+	}
+
+	if (!targets) {
+		return {};
+	}
+
+	return {
+		'CallExpression[callee.name="require"]'(node) {
+			processRule(context, node, node.arguments[0].value, targets);
```
fisker, line 161:
```suggestion
		[STATIC_REQUIRE_SOURCE_SELECTOR](node) {
```

`require()` arguments can be anything.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,249 @@
...
+	} else if (options.treatsTargetsAsSemver) {
+		targets = new SemverNodeVersion(targets);
+	}
+
+	if (!targets) {
+		return {};
+	}
+
+	return {
+		'CallExpression[callee.name="require"]'(node) {
```
fisker, line 160:
```suggestion
		[STATIC_REQUIRE_SOURCE_SELECTOR](node) {
```

`require()` arguments can be anything.

Mesteery:
PTAL @fisker

fisker:
Sorry for the delay, I'll review ASAP.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,250 @@
...
+};
+
+function getTargetsFromPackage(cwd, useBrowserlistFieldByDefault) {
+	const result = readPkgUp.sync({cwd});
+	if (!result || !result.pkg) {
+		return;
+	}
+
+	const {browserlist} = result.pkg;
+	const nodeEngine = result.pkg.engines && result.pkg.engines.node && new SemverNodeVersion(result.pkg.engines.node);
```
fisker, line 24: Not every package.json have `engine` field.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,250 @@
...
+	} else if (options.treatsTargetsAsSemver) {
+		targets = new SemverNodeVersion(targets);
+	}
+
+	if (!targets) {
+		return {};
+	}
+
+	return {
+		[STATIC_REQUIRE_SELECTOR](node) {
```
fisker, line 161: You can use `STATIC_REQUIRE_SOURCE_SELECTOR`.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,250 @@
...
+	compare(featureVersion) {
+		const supportedNodeVersion = semver.coerce(featureVersion);
+		return this.validNodeVersion
+			? semver.lte(supportedNodeVersion, this.validNodeVersion)
+			: semver.ltr(supportedNodeVersion, this.nodeVersion);
+	}
+}
+
+function processRule(context, node, moduleName, targets) {
+	if (!moduleName || typeof moduleName !== 'string') {
```
fisker, line 105: I think we should only check npm package. So if moduleName starts with `.` should be ignored.

test/no-unnecessary-polyfills.mjs
```diff
@@ -0,0 +1,182 @@
...
+		},
+		{
+			code: 'require("es6-promise")',
+			options: [{targets: 'node >15'}],
+			errors: [{message: 'Use the built-in `Promise`.'}],
+		},
+		{
+			code: 'require("object-assign")',
+			options: [{targets: 'node 6'}],
+			errors: [{message: 'Use the built-in `Object#assign`.'}],
```
fisker, line 119: The message should be ``'Use the built-in `Object.assign` ``, it's a static method.

docs/rules/no-unnecessary-polyfills.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Options
+
+Type: `object`
+
+### targets
+
+Type: `string | string[] | object`
+
+The `targets` option allows to specify the target versions. This option could be a Browserlist query or a targets object, see [core-js-compat `targets` option](https://github.com/zloirock/core-js/tree/HEAD/packages/core-js-compat#targets-option) for more informations. It could also be a Node.js target version (SemVer syntax supported) if the [`treatsTargetsAsSemver` option](#treatsTargetsAsSemver) is set to `true`.
```
fisker, line 50: I think we should only allow `targets`, for `engines.node` field, we can turn it into a browser target. So we don't need `treatsTargetsAsSemver`.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,250 @@
...
+	return {
+		feature,
+		pattern: new RegExp(
+			`^((${prefixes}?`
+				+ `(${ecmaVersion}${delimiter}${constructorName}${delimiter}${methodName}|` // Ex: es6-array-copy-within
+				+ `${constructorName}${delimiter}${methodName}|` // Ex: array-copy-within
+				+ `${ecmaVersion}${delimiter}${constructorName})` // Ex: es6-array
+				+ `${suffixes}?)|`
+				+ `(${prefixes}${methodOrConstructor}|${methodOrConstructor}${suffixes})` // Ex: polyfill-copy-within / polyfill-promise
+				+ `${additionalPolyfillPatterns[feature] || ''})$`,
```
fisker, line 68: Maybe we can use an array of patterns? This is hard to understand.

fisker (commented):
Sorry for the long delay. Feel hard to review, I'll take another round later.

docs/rules/no-unnecessary-polyfills.md
Mesteery: Do you mean to only allow using semver syntax in `targets.node` and remove `treatsTargetsAsSemver`?

```js
"unicorn/no-unnecessary-polyfills": ["error", {	"targets": { "node": "..semver.." } }] // valid
"unicorn/no-unnecessary-polyfills": ["error", {	"targets": "..semver.." }] // invalid
```

rules/no-unnecessary-polyfills.js
Mesteery, line 105:
```suggestion
	if (!moduleName || typeof moduleName !== 'string' || moduleName[0] === '.') {
```

test/no-unnecessary-polyfills.mjs
Mesteery, line 119:
```suggestion
			errors: [{message: 'Use the built-in `Object.assign`.'}],
```
Yes indeed, it was in the initial PR.

rules/no-unnecessary-polyfills.js
Mesteery: Indeed, that is in case the `targets` field is not defined in the config. If the `engines` field is also undefined, the plugin is not executed.

docs/rules/no-unnecessary-polyfills.md
fisker: No, I mean only [core-js "target"](https://github.com/zloirock/core-js/tree/HEAD/packages/core-js-compat#targets-option) , for `engine.node` in package.json, we only need translate to it (maybe?). So we don't need `treatsTargetsAsSemver`

Mesteery: I'm sorry, I'm having trouble understanding, do you have any examples please?

fisker: Sorry I didn't read carefully, yes your example is correct


```
"unicorn/no-unnecessary-polyfills": ["error", {	"targets": { "node": "..semver.." } }] // valid
"unicorn/no-unnecessary-polyfills": ["error", {	"targets": "..semver.." }] // invalid
```

fisker: We don't  need care what targets is, use it directly.

Mesteery: Thanks, it's clearer. PTAL

fisker:
The tests are failing.

Mesteery:
Fixed. The errors is due to `STATIC_REQUIRE_SOURCE_SELECTOR` (maybe i misused it, but it worked correctly with `STATIC_REQUIRE_SELECTOR`), and the Object**.**assign, there is no way to know if it is a static method or not... except hardcoding static methods.

Mesteery:
Thanks you! Is it ready or i missed some changes?

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,248 @@
...
+	context.report({
+		node,
+		messageId: MESSAGE_ID_POLYFILL,
+		data: {
+			featureName: `${constructorCase(namespace)}${method && '#'}${camelCase(method)}`,
+		},
+	});
+}
+
+class SemverNodeVersion {
```
fisker, line 91: Can we pass `node {{engine.node}}` directly to  `core-js-compat` as `targets`?

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,248 @@
...
+		pattern: new RegExp(patterns.join(''), 'i'),
+	};
+});
+
+function report(context, node, feature) {
+	let [, namespace, method = ''] = feature.split('.');
+	if (namespace === 'typed-array' && method.endsWith('-array')) {
+		namespace = method;
+		method = '';
+	}
```
fisker, line 80: > there is no way to know if it is a static method or not

If we can't know, let's just say "use built-in instead", this part seems very risky to me, the module name can be changed in future.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,248 @@
...
+						type: 'array',
+						minItems: 1,
+						items: {
+							type: 'string',
+						},
+					},
+					{
+						type: 'object',
+						minProperties: 1,
+						properties: {
```
fisker, line 192: Let's not list the properties, we don't want to maintain this, or even not list "string"/"array", just pass the `targets` through.

sindresorhus:
@Mesteery Bump

Mesteery:
I'm sorry for all the delay, I was really busy.

fisker:
> I'm not sure if corejs({targets}) supports SemVer syntax with targets = {node:'semver'}

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1717#discussion_r944484182

Mesteery:
If I understand you want `node.engines` to be core-js `targets`?

sindresorhus:
Bump

fisker:
I mean I don't know if core-js accept `node.engines` if not, translate to `targets`, other user input `targets` pass directly to `core-js`.

Mesteery:
I apologize again for the delay. I've slightly refactored the code, and I passed `engines` directly to corejs targets if browserlist is not defined (and I've removed `useBrowserlistFieldByDefault`, as I think it would make more sense to always have engines as a last resort). However, I'm not sure that corejs correctly supports the semver syntax that can be used in `engines`.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,147 @@
...
+		pattern: new RegExp(patterns.join(''), 'i'),
+	};
+});
+
+function getTargets(options) {
+	if (options?.targets) {
+		return options.targets;
+	}
+
+	const result = readPkgUp.sync();
```
sindresorhus, line 60: You need to apply same fix as https://github.com/sindresorhus/eslint-plugin-unicorn/commit/ac51d409b28bedbceee8b3f8f39c2b820dfec280#diff-983077e885cbad3d4bbd1379551c5912e39e4e0685457d3a0a9f8e320e53ac30R59

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,147 @@
...
+	const unavailableFeatures = coreJsCompat({targets: getTargets(context.options[0])}).list;
+	const checkFeatures = features => !features.every(feature => unavailableFeatures.includes(feature));
+
+	return {
+		Literal(node) {
+			if (
+				!(
+					((node.parent.type === 'ImportDeclaration' || node.parent.type === 'ImportExpression')
+            && node.parent.source === node)
+          || (isStaticRequire(node.parent) && node.parent.arguments[0] === node)
```
sindresorhus, line 78: Inconsistent indentation.

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,147 @@
...
+}
+
+const schema = [
+	{
+		type: 'object',
+		additionalProperties: false,
+		required: ['targets'],
+		properties: {
+			targets: {
+				oneOf: [{type: 'string'}, {type: 'array'}, {type: 'object'}],
```
sindresorhus, line 130: Format it on separate lines.

sindresorhus (changes requested):

sindresorhus:
CI is failing

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	/** @type {readPkgUp.ReadResult | undefined} */
+	let packageResult;
+	try {
+		packageResult = readPkgUp.sync({normalize: false, cwd: dirname});
+	} catch {
+		// This can happen if package.json files have comments in them etc.
+		packageResult = undefined;
+	}
```
sindresorhus, line 68:
```suggestion
	try {
		// It can fail if, for example, the package.json file has comments.
		packageResult = readPkgUp.sync({normalize: false, cwd: dirname});
	} catch {}
```

rules/no-unnecessary-polyfills.js
```diff
@@ -0,0 +1,178 @@
...
+
+	if (!packageResult) {
+		return;
+	}
+
+	const {browserlist, engines} = packageResult.packageJson;
+	return browserlist || engines;
```
sindresorhus, line 75:
```suggestion
	return browserlist ?? engines;
```

sindresorhus (changes requested):

PR #1724, opened 2022-02-09
Require ESLint 8

fisker:
- Use `es2021: true` instead of `es6: true`.
- Add tests to validate it actually works
- Drop support for ESLint v7

Can't use `es2022`
```text
Error {
    message: `--config:
        Environment key "es2022" is unknown
    `,
  }
```

Actually, I was going to remove `env`, I thought setting `ecmaVersion` auto enables those envs, I'll investigate.

### DIFF
--- a/configs/recommended.js
+++ b/configs/recommended.js
@@ -1,7 +1,7 @@
 'use strict';
 module.exports = {
 	env: {
-		es6: true,
+		es2021: true,
 	},
 	parserOptions: {
 		ecmaVersion: 'latest',
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
 		"c8": "^7.11.0",
 		"chalk": "^5.0.0",
 		"enquirer": "^2.3.6",
-		"eslint": "^8.6.0",
+		"eslint": "^8.8.0",
 		"eslint-ava-rule-tester": "^4.0.0",
 		"eslint-plugin-eslint-plugin": "^4.1.0",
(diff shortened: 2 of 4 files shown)


fisker:
I was wrong about `env`, https://eslint.org/docs/user-guide/configuring/language-options#specifying-parser-options

> For ES6 syntax, use { "parserOptions": { "ecmaVersion": 6 } }; for new ES6 global variables, use { "env": { "es6": true } }. { "env": { "es6": true } } enables ES6 syntax automatically, but { "parserOptions": { "ecmaVersion": 6 } } does not enable ES6 globals automatically.

sindresorhus:
And since we're dropping ESLint 7: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/f14aa956e8e7403f0dff2b8581e8c572bd1d7a52/rules/utils/get-builtin-rule.js#L4-L5

alfaproject:
I think you mean it requires `eslint@8.9` because according to our CI the update to this version is not compatible with `8.8`...

Can you tell me exactly why `es2022` is needed? Seems like an unnecessary breaking change for no real reason

fisker:
Sorry, this PR was merged before ESLint release `8.9`, add `es2022` was added in that version.

> no real reason

No reason, we only want to support the latest version.

alfaproject:
No problems, we just updated to eslint 8.9 anyway but maybe worth to update the change log. More curiosity than anything

Issue #1146, opened 2021-03-22
[Regression] no-null: error for Object.create(null) with parserOptions.ecmaVersion 2019


kevinoid:
With `.eslintrc.json`

```json
{
  "extends": "plugin:unicorn/recommended",
  "parserOptions": {
    "ecmaVersion": 2019
  }
}
```
the code
```js
module.exports = Object.create(null);
```
is accepted without error when using `eslint-plugin-unicorn@28.0.2`.  When using `eslint-plugin-unicorn@29.0.0` it produces the following error:

    /path/to/file.js:1:32: Use `undefined` instead of `null`. [Error/unicorn/no-null]

I bisected the regression to 2397d94 (presumably the changes in `rules/utils/method-selector.js` although I didn't investigate further).

Thanks,
Kevin

fisker:
@kevinoid What parser are you using?

fisker:
My guess is , you parser don't have `optional` property for `MemberExpression`, maybe we can change `[...=false]` to `[...!=true]`.

The methodSelector function has bug before 2397d94, you can see, [here](https://github.com/sindresorhus/eslint-plugin-unicorn/commit/2397d940572c74a48e00be11c307e821f9f1664b#diff-4dc7a79c5b1abc94f8b81e79d2203d1fca6d935584941ea8fd4bee8566ad601fL27) it mean to check optional, but it checked `computed` instead.

kevinoid:
> @kevinoid What parser are you using?

I'm not sure what you mean.  How would I check this?  For reference, a script to reproduce the issue from scratch:

```sh
cd "$(mktemp -dt)"
npm init -y
npm install --save-dev eslint eslint-plugin-unicorn
cat >.eslintrc.json <<'ESLINTRC'
{
  "extends": "plugin:unicorn/recommended",
  "parserOptions": {
    "ecmaVersion": 2019
  }
}
ESLINTRC
echo 'module.exports = Object.create(null);' >index.js
./node_modules/.bin/eslint index.js
```

`npm ls`:

    tmp.odU4hveyCQ@1.0.0 /tmp/tmp.odU4hveyCQ
    ├── eslint-plugin-unicorn@29.0.0
    └── eslint@7.22.0

fisker:
Thanks, I'll take a look later.

PR #1725, opened 2022-02-09
Add an internal ESLint rule `prefer-negative-boolean-attribute`

fisker:
Use positive `[...=true]` may cause problems, when `ecmaVersion` is not correct. #1146, https://github.com/fisker/eslint-plugin-prettier-doc/commit/f1b1c6f339a77ed306c64172c927849f0d78c7c3

Maybe I can try to make a proposal in [eslint-plugin-eslint-plugin](https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin) later.

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,5 +169,17 @@
 				}
 			}
 		]
+	},
+	"npmpackagejsonlint": {
+		"rules": {
+			"prefer-caret-version-devDependencies": [
+				"error",
+				{
+					"exceptions": [
+						"eslint-plugin-internal-rules"
+					]
+				}
+			]
+		}
 	}
 }
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
 		"eslint": "^8.6.0",
 		"eslint-ava-rule-tester": "^4.0.0",
 		"eslint-plugin-eslint-plugin": "^4.1.0",
+		"eslint-plugin-internal-rules": "file:./scripts/internal-rules/",
 		"eslint-remote-tester": "^2.0.1",
 		"eslint-remote-tester-repositories": "^0.0.3",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -74,6 +74,7 @@
 		"eslint": "^8.6.0",
 		"eslint-ava-rule-tester": "^4.0.0",
 		"eslint-plugin-eslint-plugin": "^4.1.0",
+		"eslint-plugin-internal-rules": "file:./scripts/internal-rules/",
```
silverwind, line 77: Does eslint not offer a way to load a plugin from a file instead of a module?

fisker: This being a loooong story, and the reason is ridiculous, add a plugin path in config, ESLint can't get a plugin name to use...

https://github.com/eslint/eslint/issues/3458

Unless you use the ESLint API, you can't add plugin in other way.

- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e8ab1766f367311b906d198021353c2bbf8df301/test/integration/package.json#L8
- https://github.com/prettier/eslint-plugin-prettier-internal-rules/blob/2c414e5b4dec5b4dca39aec9dcf3cd0a0d603f11/index.js#L5
- https://github.com/mysticatea/eslint-plugin-node/blob/f45c6149be7235c0f7422d1179c25726afeecd83/package.json#L28
- https://www.npmjs.com/package/eslint-plugin-self-dir

silverwind: You could hack the module resolution like [node core does](https://github.com/nodejs/node/blob/91d28d8b46cbb9b101ee2c51a32adafb12bc24b7/.eslintrc.js#L23-L34) but i guess it's even worse of a hack 😉.

PR #1726, opened 2022-02-10
Simplify `englishishJoinWords`

fisker:
https://twitter.com/ericclemmons/status/1488558951008509963

Should we make a rule for it, though it's not easy to detect?


sindresorhus:
> Should we make a rule for it, though it's not easy to detect?

I fear it may be to hard to reliably detect.

fisker:
Agree

PR #1728, opened 2022-02-10
`prefer-export-from`: Fix TypeScript compatibility

### LINKED ISSUE #1727: `unicorn/prefer-export-from` breaks type exports in typescript projects using the `isolatedModules` flag.
When autofixing exports in ts projects `unicorn/prefer-export-from` will change this:

```ts
import UIAvatar, {
  UIAvatarProps,
} from "./UIAvatar";

export type { UIAvatarProps };
export default UIAvatar;
```

into 

```ts
export { default, UIAvatarProps } from "./UIAvatar";
```

Which results in a typescript error `Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.ts(1205)`

The correct fix should be:

```ts
export { default, type UIAvatarProps } from "./UIAvatar";
```

Type only exports should be prefixed with `type`.

I've also seen this problem happen the other way around before where a non type export gets bundled into an `export type { ... } from ""` and has broken internal library code as the autofix is applied using commit hooks.

nrgnrg:
closes #1727

When applying a fix checks the `exportKind` and adds `type` as a prefix to the specifier if it's a type export.


nrgnrg:
@fisker

fisker:
I'm sorry to say this is not correct solution, those "imported" can also variables.

I don't use ts myself, but afaik, `exportKind` can be on specifier and declaration(correct me if I'm wrong). We need get all imported exported kind. And make sure we don't reuse the wrong export.

Example:

```js
import { foo } from "foo";
export type  {bar } from "foo";
```

The export can't be reused, need insert another export.

fisker:
Normally I don't dig deep for the ts parser, we can also simply skip the fix if you see any `importKind` or `exportKind` is "type" if you find hard to fix. So the fix won't break your code until someone make a better fix.

nrgnrg:
I'm not sure I follow exactly.

Given your example the correct fix would be either:

```ts
export { foo, type bar } from "foo";
```

or

```ts
export { foo } from "foo";
export type { bar } from "foo";
```

Which this fix correctly substitutes, I've run this change with `--fix` on our largish codebase at work and it's made the correct substitution every time.

fisker:
What's the result for the code I posted?

nrgnrg:
Ah ok I gotcha, it's reusing the `export type` in this instance and converting `foo` to a type export.

nrgnrg:
This is perhaps not the most perfect solution because we could be smarter about it, but by just ignoring export declarations which are `type` declarations as far as i can see we cannot end up in a scenario where a value is converted to a type.

sindresorhus:
This will need a test.

nrgnrg:
I've corrected the test failures and added a test for the change.

rules/prefer-export-from.js
```diff
@@ -109,6 +109,8 @@ function getFixFunction({
 	const sourceNode = importDeclaration.source;
 	const sourceValue = sourceNode.value;
 	const exportDeclaration = exportDeclarations.find(({source}) => source.value === sourceValue);
+	const isTypeExport = exported.node.parent.exportKind === 'type';
+	const isTypeExportDeclaration = exportDeclaration && exportDeclaration.exportKind === 'type';
```
fisker, line 113: This should check in `.find()` call


```
import { foo } from 'foo';
export {foo};
export type  {bar } from "foo";
export {baz } from "foo"; // Should resue this one
```

rules/prefer-export-from.js
```diff
@@ -109,6 +109,8 @@ function getFixFunction({
 	const sourceNode = importDeclaration.source;
 	const sourceValue = sourceNode.value;
 	const exportDeclaration = exportDeclarations.find(({source}) => source.value === sourceValue);
+	const isTypeExport = exported.node.parent.exportKind === 'type';
```
fisker, line 112:
```difff
const isTypeExport = exported.node.exportKind === 'type' || exported.exportKind === 'type';
```

rules/prefer-export-from.js
```diff
@@ -122,20 +124,22 @@ function getFixFunction({
 				? exported.text
 				: `${imported.text} as ${exported.text}`;
 
-			if (exportDeclaration) {
+			const specifierWithKind = isTypeExport ? `type ${specifier}` : specifier;
```
fisker, line 127: This should not work like this, only if existing export is value export, add `type `, but we should also be able to result `export type {} from ''`, in this case `type ` is not needed.

nrgnrg:
I think I've addressed you comments, I now see the following output for the given inputs (before and after this change).

Previously: value exports converted to type exports breaking code
Now: value exports preserved, type export style preserved (inline or multiple statements)

1)
```ts
// input
import { foo } from "foo";
export { foo };
export type { bar } from "foo";

// Existing Transform:
export type { bar, foo } from "foo";

// new Transform:
export type { bar } from "foo";
export { foo } from "foo";
```

2)
```ts
// input
import { foo } from "foo";
export { foo };
export { type bar } from "foo";

// Existing Transform:
export { type bar, foo } from "foo";

// Updated Transform:
export { type bar, foo } from "foo";
```

3)
```ts
// input
import foo, { bar } from "foo";
export type { bar };
export default foo;

// Existing Transform:
export type { bar, default } from "foo";

// Updated Transform:
export type { bar } from "foo";
export { default } from "foo";
```

4)
```ts
// input
import { foo } from 'foo';
export { foo };
export type { bar } from "foo";
export { baz } from "foo";

// Existing Transform:
export type { bar, foo } from "foo";
export { baz } from "foo";

// Updated Transform:
export type { bar } from "foo";
export { baz, foo } from "foo";
```

5)
```ts
// input
import { foo } from 'foo';
export { foo };
export { type bar } from "foo";
export { baz } from "foo";

// Existing Transform:
export { type bar, foo } from "foo";
export { baz } from "foo";

// Updated Transform:
export { type bar, foo } from "foo";
export { baz } from "foo";
```

rules/prefer-export-from.js
nrgnrg: Now existing export is always a value export

fisker:
Always add tests when you do logic change.

nrgnrg:
@fisker the additional test cases have been added now

fisker:
Does ts allow `export const foo = bar` (`bar` is type)?

nrgnrg:
> Does ts allow export const foo = bar (bar is type)?

no, that results in `'bar' only refers to a type, but is being used as a value here. ts(2693)`

nrgnrg:
> Add more tests, change `export {foo}` to `export type {foo}` in all those tests you added, and test them.

I'm having difficulty correctly testing this but it is working in my project:

For example, in my project

```ts
// input
// foo is exported as a value and imported as a type
// bar is exported as a type and imported as a type
import type { foo } from "foo";
export type { foo };
export type { bar } from "foo";
```

results in:

```ts
export type { bar } from "foo";
export { foo } from "foo";
```

Which is not breaking any code, it's seeing that `foo` is really a value and not a `type` and exporting it as such.

If I flip the order of those imports:
```ts
// input
// foo is exported as a value and imported as a type
// bar is exported as a type and imported as a type
import type { foo } from "foo";
export type { foo };
export type { bar } from "foo";
```

results in:

```ts
export type { foo } from "foo";
export type { bar } from "foo";
```

Which is exactly the same as what we imported.

But when I run this in the rest suite I see this result:

```ts
// input
// foo is exported as a ? and imported as a type
// bar is exported as a ? and imported as a type
import type { foo } from "foo";
export type { foo };
export type { bar } from "foo";
```

```ts
export type { bar } from "foo";
export { foo } from "foo";
```

Which is the same as I see in my project when `foo` is a `value` being imported as a `type`, as far as I can tell there is no way to assert wether `foo` is really a `value` or a `type` in the test, so could technically be correct but the test is misleading.

fisker:
>  there is no way to assert wether foo is really a value or a type in the test

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1728#discussion_r804442450

nrgnrg:
@fisker checking both the `importKind` now as well as the import and export declaration kind.

Repeated all the tests with `export type { foo };`

fisker:
These legal?

```ts
				import { type foo } from 'foo';
				export type { foo };
```

```ts
				import { foo } from 'foo';
				export type { foo };
```

```ts
				import type { foo } from 'foo';
				export { type foo };
```

If so, add them.

rules/prefer-export-from.js
```diff
@@ -109,7 +109,9 @@ function getFixFunction({
 	const sourceNode = importDeclaration.source;
 	const sourceValue = sourceNode.value;
 	const exportDeclaration = exportDeclarations.find(({source, exportKind}) => source.value === sourceValue && exportKind !== 'type');
-	const isTypeExport = exported.node.exportKind === 'type' || exported.exportKind === 'type';
+	const isTypeImport = imported.node.importKind === 'type' || imported.importKind === 'type' || (Boolean(imported.declaration) && imported.declaration.importKind === 'type');
+	const isTypeExport = exported.node.exportKind === 'type' || exported.exportKind === 'type' || (Boolean(exported.declaration) && exported.declaration.exportKind === 'type');
```
fisker, line 113: Let's add this logic to `getExported` and `getImported`?

fisker:
I just noticed, we have `ImportDefaultSpecifier` and `ImportNamespaceSpecifier` can they be type export?

fisker:
FYI: If you feel tests hard to write you can use snapshot test with ts parser, example https://github.com/sindresorhus/eslint-plugin-unicorn/blob/ee9f609b2d17be0da4871631fe4159ecbba6b17b/test/number-literal-case.mjs#L172

nrgnrg:
I don't think we need to worry about namespace imports for types, this kind of thing is not possible:

```ts
import type * as Foo from "foo";
export type Foo
```

This logic works for default type imports, I've added a test.

fisker:
I'm working to make some improvements.

fisker:
@nrgnrg Can you take a look? Main improvements are  in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1728/commits/03ba7c6350b3d315e8b0d72beffe9fff474364ad.

1. Reuse type export declaration
2. Insert a type export instead of a value export

I don't use typescript myself, but I assume 

```js
export type {foo} from 'foo';
```

is prefered over

```js
export {type foo} from "foo";
```

?

fisker:
```ts
import type {foo} from  'foo';
export type bar = foo;
```

This seems not handled by our rule, if you are interested, maybe you can try to fix this in a separate PR?

nrgnrg:
Your changes look good to me 👍

The inline type import/export syntax (`export { type foo } from "foo";`) is a more recent addition to the language which was added to avoid several import/export statements. Personally I find it more concise but the `export type { foo } from 'foo'` syntax would support older versions of TS.

The responsibility of enforcing one style over the other does not feel like the responsibility of this rule, there is a [tracking issue in `typescript-eslint`](https://github.com/typescript-eslint/typescript-eslint/issues/4338) to enforce consistent usage of the inline syntax by making it a part of [import/no-duplicates](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md) so I think as long as this rule plays nicely with changes made there to enforce the inline syntax this is probably the correct way to go here in order to support older typescript versions.

fisker:
Thanks for the information, I guess I was wrong, can you change it to prefer value export first? It should be simple.

rules/prefer-export-from.js
```diff
@@ -118,24 +131,28 @@ function getFixFunction({
 				`\nexport * as ${exported.text} ${getSourceAndAssertionsText(importDeclaration, sourceCode)}`,
 			);
 		} else {
-			const specifier = exported.name === imported.name
+			let specifierText = exported.name === imported.name
 				? exported.text
 				: `${imported.text} as ${exported.text}`;
```
fisker, line 136: This line seems not tested. The following should cover it.

```js
import {type foo} from 'foo';
export {type foo as bar};
```

```js
import {type foo} from 'foo';
export {type foo as bar};
export {type bar} from 'foo';
```

nrgnrg:
@fisker let me know how this looks

fisker:
Actually, I mean https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1728/commits/3020b59597eafd004b40df229eb539716823c587, if `{type foo}` is preferred, we should insert it in a value export, even type export exists.

@sindresorhus What's your opinion on this

```js
import {type foo} from 'foo';
export {type foo}
export type {bar} from 'foo';
export {baz} from 'foo';
```

FIX 1:
```js
export type {bar, foo} from 'foo';
export {baz} from 'foo';
```

FIX 2:

```js
export type {bar} from 'foo';
export {baz, type foo} from 'foo';
```

sindresorhus:
I would go with fix 1. That's what most people would want, I think.

fisker:
Okay, I'll revert.

sindresorhus:
However, my preference would be:

```js
export {baz, type foo, type bar} from 'foo';
```

Since it makes it possible to have a single statement.

I think there should be a TODO comment to switch to that when the TS version with it is more widespread.

fisker:
@nrgnrg already mentioned this in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1728#issuecomment-1038921639, that should be handled a separate rule.

fisker (approved):
I'm fine with those changes, thanks for working on this, good job @nrgnrg .

PR #1733, opened 2022-02-17
Update `eslint-remote-tester.config.js`

fisker:
- Add more extensions
- We already have `env` and `parserOptions` in 'plugin:unicorn/all', we should be able to remove them 

//cc @AriPerkkio


AriPerkkio (approved):
Looks good!

PR #1750, opened 2022-03-18
`prefer-json-parse-buffer`: Remove from `recommended` preset

privatenumber:
Addresses https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1273#issuecomment-1069772100

### DIFF
--- a/docs/rules/prefer-json-parse-buffer.md
+++ b/docs/rules/prefer-json-parse-buffer.md
@@ -2,8 +2,6 @@
 
 <!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
 <!-- RULE_NOTICE -->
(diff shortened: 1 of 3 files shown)


sindresorhus:
The docs should be updated to explain when you would not want to enable the rule (TS, little benefit, etc).

docs/rules/prefer-json-parse-buffer.md
```diff
@@ -17,7 +17,7 @@ const packageJson = JSON.parse(await fs.readFile('./package.json', 'utf8'));
 
 ```js
 const promise = fs.readFile('./package.json', {encoding: 'utf8'});
-const packageJson = JSON.parse(promise);
+const packageJson = JSON.parse(await promise);
```
fisker, line 20: Somehow missed this. This is not nessary forthis change, but since we are here, I just fixed it.

fisker (approved):

Issue #1632, opened 2021-12-09
`no-array-for-each`: is not handling optional chaining when auto-fixing


Primajin:
We have many places in the code where we run `forEach` behind an optional chain.

```js
myIds?.forEach(id => {
  // do something with that id
});
```

This is correctly flagged as `ESLint: Use for…of instead of Array#forEach(…).(unicorn/no-array-for-each)` 👍🏻 
However the autofix is not picking it up.

once I change it to:
```js
if (myIds) {
  myIds.forEach(id => {
    // do something with that id
  });
}
```
it automagically fixes it to:
```js
if (myIds) {
  for (const id of myIds) {
    // do something with that id
  });
}
```

can we maybe add a rule to the autofixer that if there is an optional chain just wrap the block in an `if`? 🤔

fisker:
It make sense. When I implement this rule, I didn't expect the optional version use too much in a codebase, so I skipped that. However, the auto fix part in this rule is already really complated.

PR wellcome.

fisker:
Or maybe fix to this version

```js
if (Array.isArray(myIds)) {
  for (const id of myIds) {
    // do something with that id
  });
}
```

?

Primajin:
OK great, I have never implemented a lint fixer - so let's see if I can come up with a PR 🤞🏻 😅

Primajin:
Great thanks!

PR #1753, opened 2022-03-20
Enhance `no-array-for-each` to handle optional chaining

zaicevas:
Closes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1632 

That's my first contribution, so let me know if I missed anything 😅


rules/no-array-for-each.js
```diff
@@ -175,13 +180,35 @@ function getFixFunction(callExpression, functionInfo, context) {
 			return false;
 		}
 
-		if (callback.body.type !== 'BlockStatement') {
+		if (callback.body.type !== 'BlockStatement' && !isOptionalChaining) {
```
fisker, line 183: Can you explain why we need this here?

rules/no-array-for-each.js
```diff
@@ -175,13 +180,35 @@ function getFixFunction(callExpression, functionInfo, context) {
...
 			return false;
 		}
 
 		return true;
 	};
 
+	function * wrapInIfStatement(fixer) {
+		const isSingleLine = !isBlockStatement || callback.body.loc.start.line === callback.body.loc.end.line;
+
+		yield fixer.insertTextBefore(callExpression, `if (${callExpression.callee.object.name}) {\n`);
```
fisker, line 193: This is not right, `callExpression.callee.object` not always `Indentifier`, `foo.bar()?.forEach(...)`

fisker (changes requested):

rules/no-array-for-each.js
zaicevas: 2 tests would fail. `foo?.forEach((element) => console.log(element));` and `foo?.forEach(element => console.log(element));`

```
    Actual:␊
    if (foo) {␊
    	for (const element of foo) console.log(element)␊
    };␊
    ␊
    Expected:␊
    if (foo) {␊
    	for (const element of foo) console.log(element)␊
    }`,
```

rules/no-array-for-each.js
zaicevas: Good catch! There's an even simpler example: `foo()?.forEach(...)`. The question is how do we handle such cases? I guess we can either auto fix it to:
```
				if (foo()) {
					for (const element of foo()) {
						...
					}
				}
```
which might cause issues if `foo` produces side effects. Alternatively, we can fix it so that `foo()` is stored in a variable or we could NOT fix such cases. WDYT?

rules/no-array-for-each.js
fisker:
```js
if (foo) {
  for (const element of foo) console.log(element);
}
```

This is expected, try to fix.

fisker:
Tests arre not enough, please duplicate all these cases https://github.com/sindresorhus/eslint-plugin-unicorn/blob/93b663719da4389c083e8d34af73efd4ac5e8652/test/no-array-for-each.mjs#L21, change to `?.` and test them all.

rules/no-array-for-each.js
fisker: Use `suggest` for cases has side effect.

zaicevas: How can we know whether there's a side effect? Also, what would be the suggested fix? My above snippet or you're thinking about more auto-fixed suggestions?

zaicevas:
> Tests are not enough, please duplicate all these cases
> 
> https://github.com/sindresorhus/eslint-plugin-unicorn/blob/93b663719da4389c083e8d34af73efd4ac5e8652/test/no-array-for-each.mjs#L21
> 
> , change to `?.` and test them all.

I've faced an issue when tackling this auto fix scenario:
```jsx
foo?.forEach(element => {
    switch (element) {
        case 1:
            break;
        case 2:
            return;
    }
});
```
NEW:
```jsx
if (foo) {
    for (const element of foo) {
        switch (element) {
	    case 1:
	        break;
	    case 2:
	        continue;
        }
    }
}
```

The problem is that we're trying to fix the `return` statement twice. 
1. `return` -> `continue`
```jsx
		for (const returnStatement of returnStatements) {
			yield * replaceReturnStatement(isOptionalChaining ? indentString(returnStatement, 1, {indent: '\t'}) : returnStatement, fixer);
		}
```
2. Indent the switch expression
```jsx
		for (const expression of expressions) {
			yield fixer.replaceText(expression, indentString(sourceCode.getText(expression), 1, {indent: '\t'}));
		}
```

Do you have any ideas on how to elegantly solve this?

rules/no-array-for-each.js
fisker: [`hasSideEffect()`](https://eslint-utils.mysticatea.dev/api/ast-utils.html#hassideeffect).


Store `foo()` in a variable should be the best (and can use auto-fix instead of suggestion), but I'm afraid it's hard to implement.

We can suggest your code above, unless you want do more.

fisker:
Forget about the indention, let other rule handle it.

rules/no-array-for-each.js
```diff
@@ -314,14 +319,7 @@ function isFixable(callExpression, {scope, functionInfo, allIdentifiers, context
...
 	// Check `CallExpression.parent`
-	if (callExpression.parent.type !== 'ExpressionStatement') {
-		return false;
-	}
-
-	// Check `CallExpression.callee`
-	// Because of `ChainExpression` wrapper, `foo?.forEach()` is already failed on previous check keep this just for safety
-	/* c8 ignore next 3 */
-	if (callExpression.callee.optional) {
+	if (callExpression.parent.type !== 'ExpressionStatement' && callExpression.parent.type !== 'ChainExpression') {
```
fisker, line 322: This is not right `() => foo?.forEach()` and many other cases.

fisker:
Is optional call tested? `foo.forEach?.()`

rules/no-array-for-each.js
zaicevas: Should be good now, since I added additional check in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1753/files#diff-0a69d7bd0e300d6d5043f953d861ec9c922c3bdda5bd91a4b5e0040834fd1392R327

zaicevas:
> Is optional call tested? `foo.forEach?.()`

There was a test case prior to my changes https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/no-array-for-each.mjs#L23

fisker:
So, turns out it's quite simple, right? I guess just one issue to solve https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1753#discussion_r830697384

zaicevas:
> So, turns out it's quite simple, right? I guess just one issue to solve [#1753 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1753#discussion_r830697384)

Yep, working on that

rules/no-array-for-each.js
zaicevas: I think I've finished implementing suggestion

rules/no-array-for-each.js
```diff
@@ -314,14 +337,15 @@ function isFixable(callExpression, {scope, functionInfo, allIdentifiers, context
...
 
-	// Check `CallExpression.callee`
-	// Because of `ChainExpression` wrapper, `foo?.forEach()` is already failed on previous check keep this just for safety
-	/* c8 ignore next 3 */
-	if (callExpression.callee.optional) {
+	if (isOptionalChaining && callExpression.parent.parent.type !== 'ExpressionStatement') {
+		return false;
+	}
+
+	if (isOptionalChaining && callExpression.callee.object.type !== 'Identifier' && callExpression.callee.object.type !== 'MemberExpression') {
```
fisker, line 348: Are you serious?

`(a ? b : c)?.forEach()`
`(0, foo)?.forEach()`
`(foo || bar)?.forEach()`

zaicevas: Well, that would involve `(foo && bar())?.forEach()`, `(a ? b : c()).forEach()` and other cases where one of the function calls has a side effect, hence it shouldn't be auto-fixable. AFAICS, it'd considerably complicate the parsing, so I'd say it's reasonable to keep that out of scope for this PR

fisker: I don't think we can ignore them, it's very common.
Would you mind let me work on it?

zaicevas: Sure

rules/no-array-for-each.js
```diff
@@ -314,14 +337,15 @@ function isFixable(callExpression, {scope, functionInfo, allIdentifiers, context
...
-	if (callExpression.parent.type !== 'ExpressionStatement') {
+	if (callExpression.parent.type !== 'ExpressionStatement' && callExpression.parent.type !== 'ChainExpression') {
 		return false;
 	}
 
-	// Check `CallExpression.callee`
-	// Because of `ChainExpression` wrapper, `foo?.forEach()` is already failed on previous check keep this just for safety
-	/* c8 ignore next 3 */
-	if (callExpression.callee.optional) {
+	if (isOptionalChaining && callExpression.parent.parent.type !== 'ExpressionStatement') {
```
fisker, line 344: This is not right, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1772 should fix this.

fisker (approved):

fisker:
@zaicevas Please take a look.

fisker:
I'll fix some edge cases after we merge this.

zaicevas:
> @zaicevas Please take a look, please let me know if you find something wrong.

I'll take a look later today

zaicevas (commented):
Looks much cleaner! I cracked my head trying to understand why you removed most tests. Then I finally found `flatMap(code => [code, code.replace('.forEach', '?.forEach')]),` 😅

rules/no-array-for-each.js
```diff
@@ -411,13 +420,31 @@ const create = context => {
 		},
 		* 'Program:exit'() {
 			for (const {node, scope} of callExpressions) {
+				// TODO: Rename this to iteratable
```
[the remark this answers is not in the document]
fisker: The object not alway array, rename to `interatable` is more correct, see https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1778. But I named it array everywhere, let's do this in a seperate PR, to make this one small.

sindresorhus:
Thanks for contributing :)

sami12000 (commented):
H

sami12000 (commented):
Y

sami12000 (commented):
Smooy55@gmail.com

PR #1756, opened 2022-03-25
Add `dist` and `dst` to abbreviations

privatenumber:
Added `dist` and `dst` to abbreviations.

### DIFF
--- a/rules/shared/abbreviations.js
+++ b/rules/shared/abbreviations.js
@@ -62,16 +62,16 @@ module.exports.defaultReplacements = {
 	dist: {
 		distribution: true,
 	},
-	dst: {
-		destination: true,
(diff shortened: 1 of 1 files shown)


rules/shared/abbreviations.js
```diff
@@ -59,13 +59,20 @@ module.exports.defaultReplacements = {
...
 	doc: {
 		document: true,
 	},
 	docs: {
 		documentation: true,
 		documents: true,
 	},
+	dst: {
+		destination: true,
+		distribution: true,
```
sindresorhus, line 74: DST is also commonly used for "daylight saving time"

sindresorhus (approved):

fisker (approved):

PR #1765, opened 2022-03-29
Add `no-unreadable-iife` rule

### LINKED ISSUE #1730: Rule proposal: `no-unreadable-iife`
### Description

A realworld case 

```js
const leadingSpaceCount = ((match) => (match ? match[1].length : Number.POSITIVE_INFINITY))(
  content.match(/^( *)[^\n\r ]/m)
);
```

I found it's very unreadable, let's forbid this IIFE with parenthesized arrow function body.

### Fail

```js
const foo = ((bar) => (bar ? bar.baz : baz))(getBar());
```

### Pass

```js
const bar = getBar();
const foo = bar ? bar.baz : baz;
```

fisker:
Fixes #1730


docs/rules/no-unreadable-iife.md
```diff
@@ -0,0 +1,36 @@
+# Forbid unreadable IIFEs
```
sindresorhus, line 1: I think we should standardize on the word "disallow" over "forbid". I noticed we have been inconsistent about it, but we mostly use "disallow".

docs/rules/no-unreadable-iife.md
```diff
@@ -0,0 +1,36 @@
...
+
+<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
+<!-- RULE_NOTICE -->
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+<!-- /RULE_NOTICE -->
+
+IIFE with parenthesized arrow function body is considered unreadable.
```
sindresorhus, line 8:
```suggestion
[IIFE](https://en.wikipedia.org/wiki/Immediately_invoked_function_expression) with parenthesized arrow function body is considered unreadable.
```

docs/rules/no-unreadable-iife.md
fisker: I can make a eslint rule for it.

fisker: #1777

PR #1768, opened 2022-03-30
`prefer-string-trim-start-end`: Check optional chaining

fisker:
Report `foo?.trimLeft()`, but not `foo.trimLeft?.()`


sindresorhus:
CI is failing

PR #1769, opened 2022-03-30
Add AST utils

fisker:
As discussed before, we are going to start using AST check instead of selectors since Node.js v14, we'll need lots of functions to do that.


fisker:
I'm going to wait for another release before we drop support for Node.js 12.

PR #1772, opened 2022-03-30
`no-array-for-each`: Handle `ChainExpression` correctly

fisker:
This PR not handling `array?.forEach()` or `array.forEach?.()`, just handle an unrelated `ChainExpression` node correctly.

```js
foo?.bar.forEach(...)
```

Should be fixable, because the array and the `forEach` call  are not optional.


sami12000 (commented):
S_my_1

sami12000 (commented):
G

sami12000 (commented):
G

PR #1775, opened 2022-03-30
Update templates

fisker:
- Add "Additional Info" section to "Rule proposal" template.
- Remove `funding.yml`, shouldn't be needed, https://github.com/sindresorhus/.github/blob/main/funding.yml
- Don't add `bug` label to "Bug report", we'll add it once it's confirmed.
- Comment out the PR notice.


fregante:
> Add "Additional Info" section to "Rule proposal" template.


Why? The "description" section should already be enough

fisker:
Sometimes, I want post something not important, or something I want to say after reading the examples. Like this.

PR #1779, opened 2022-03-31
Add `no-useless-switch-case` rule

### LINKED ISSUE #1698: Rule proposal: `no-useless-switch-case`
### Description

When a switch case without a consequent is directly followed by a default case, it’s redundant. I believe it should be removed. The auto fixer can remove the redundant case.

This has also beem proposed in https://github.com/eslint/eslint/issues/14924, but it was dismissed because it doesn’t meet the criteria for new ESLint core rules.

### Fail

```js
switch (discriminant) {
  case 'redundant':
  default:
    break;
}
```

### Pass

```js
switch (discriminant) {
  case 42:
    break;
  default:
    break;
}
```

fisker:
Fixes #1698.

One known issue, I only checked cases where the `default` case is the last case.

Because

```js
			switch (1) {
					// This is not useless
					case 1:
					default:
							console.log('1')
					case 1:
							console.log('2')
			}
```

```js
			switch (1) {
					// This is useless
					case 1:
					default:
							console.log('1')
					case 2:
							console.log('2')
			}
```

To handle these two cases correctly, I need to compare the conditions, and it can be difficult(event impossible if the test condition is a complex expression).

Let's keep this rule simple for now.


docs/rules/no-useless-switch-case.md
```diff
@@ -0,0 +1,67 @@
+# Disallow useless case in switch statement
+
+<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
+<!-- RULE_NOTICE -->
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+💡 *This rule provides [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).*
+<!-- /RULE_NOTICE -->
+
+Empty case before the last default case is useless.
```
sindresorhus, line 10:
```suggestion
An empty case before the last default case is useless.
```

docs/rules/no-useless-switch-case.md
```diff
@@ -0,0 +1,67 @@
+# Disallow useless case in switch statement
```
sindresorhus, line 1:
```suggestion
# Disallow useless case in switch statements
```

docs/rules/no-useless-switch-case.md
```diff
@@ -0,0 +1,67 @@
...
+		// Fallthrough
+	default:
+		handleDefaultCase();
+		break;
+}
+```
+
+```js
+switch (foo) {
+	// This is actually useless, but we only check cases where the last case is `default` case
```
sindresorhus, line 58:
```suggestion
	// This is actually useless, but we only check cases where the last case is the `default` case
```

rules/no-useless-switch-case.js
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const isEmptySwitchCase = node => node.consequent.every(node => isEmptyNode(node));
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	* 'SwitchStatement[cases.length>1]'(switchStatement) {
+		const {cases} = switchStatement;
+
+		// `typescript` allow multiple `default` cases
```
sindresorhus, line 19:
```suggestion
		// TypeScript allows multiple `default` cases
```

sindresorhus (changes requested):

test/no-useless-switch-case.mjs
```diff
@@ -0,0 +1,179 @@
...
+				}
+				default:
+					handleDefaultCase();
+					break;
+			}
+		`,
+		outdent`
+			switch (foo) {
+				case a:
+				case (( b ))         :
```
yakov116, line 113:
```suggestion
				case (( b )):
```
Curious were those spaces intentional?

fisker: Yes

fisker: For this function https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1779/files#diff-d14f0af39629703585e6c7779ca7b3c2a8207c63c330bc60f2849cb40a8e14f2R16, so people can't refactor to use something like `end + 1`

test/no-useless-switch-case.mjs
```diff
@@ -0,0 +1,179 @@
...
+// 					break;
+// 				case 1:
+// 				default:
+// 					handleDefaultCase2();
+// 					break;
+// 			}
+// 		`,
+// 	],
+// 	invalid: [],
+// });
```
sindresorhus, line 179: If it's intentional that this is commented out, it would be good with a short comment about why.

fisker: It's a mistake, thanks.

PR #1780, opened 2022-03-31
Add `prefer-modern-math-apis` rule

### LINKED ISSUE #1527: Rule proposal: `prefer-modern-math-apis`
I'm not good at math, but I found people use the complicated one, maybe because IE don't support it.

## Fail

```js
const foo = Math.log(bar) * Math.LOG10E;
```

```js
const foo = Math.LOG10E * Math.log(bar);
```

```js
const foo = Math.log(bar) / Math.LN10;
```

## Pass

```js
const foo = Math.log10(bar);
```

`Math.log2` seems the same, so maybe we should come up a better name.

fisker:
Fixes #1527


docs/rules/prefer-modern-math-apis.md
```diff
@@ -0,0 +1,64 @@
+# Prefer modern `Math` apis over legacy algorithm
```
sindresorhus, line 1:
```suggestion
# Prefer modern `Math` APIs over legacy patterns
```

sindresorhus (changes requested):

fisker:
Don't know why tests fail without a reason..

PR #1787, opened 2022-04-02
Deprecate `import-index` rule

fisker:


kachkaev:
@fisker WDYT of a rule that bans `index.{js,ts}{,x}` files? I am currently using [`canonical/filename-no-index`](https://github.com/gajus/eslint-plugin-canonical#eslint-plugin-canonical-rules-filename-no-index), but it’d be much cooler to have this rule in `unicorn` too!

I know that it's better to discuss such things in a new issue – this is just a quick sense-check.

fisker:
I don't know, I alway use index.js like this https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/selectors/index.js

fisker:
Maybe only allow re-export in index.js?

kachkaev:
I used to use `index.js` for re-exporting stuff for a while. Here was my typical dev flow:

1. Create `foo.js`
1. Add code
1. Realise that `foo.js` is too big
1. Replace the file with `foo/index.js` which re-exports `foo/bar.js`, `foo/baz.js`, etc.

The problem with this approach is that the last step is quite tedious. It’s not only necessary to delete the original file, but also to think about filenames in the subfolder. It is not always easy to come up with nice mini-groups to avoid actual code in `foo/index.js`.

Another problem with `foo/index.js` is that it can’t be imported the same way as `foo.js`, unless we rely on the Node.js resolution mechanism. Pure ESM modules don’t support `import { foo } from "./foo";` — it’s either `import { foo } from "./foo.js";` or `import { foo } from "./foo/index.js";`.

So I’ve decided to ban `index.js` completely and here is what I do now:

1. Create `foo.js`
1. Add code
1. Realise that `foo.js` is too big
1. Create `foo` folder and place `foo/bar.js` into it.
1. Add `import { bar } from "./foo/bar.js"` in `foo.js`. Thus, `foo.js` works as the stable ‘entry point’ for _foo_ and `foo` folder contains ‘private’ stuff that can only be imported from `foo.js`. Other files can only import from `./foo.js` (as if it’s a local mini-npm library), but they must never look into files like `./foo/bar.js`, because that’s like calling a private method of a class.

What’s cool is that an upgrade from `foo.js` into `foo.js` + `foo/*.*` is not a leap any more. The transition feels very natural and does not stand on my way.

This approach is inspired by https://github.com/benawad/destiny — see this [youtube talk](https://www.youtube.com/watch?v=rGYbrIf-y58). If you want to check out an example file structure, here are a couple of public projects I’ve got using it:

- https://github.com/kachkaev/njt (`pages`) 
- https://github.com/kachkaev/tooling-for-how-old-is-this-house (`pages`, `shared`)

PR #1792, opened 2022-04-09
Add `prefer-event-target` rule

### LINKED ISSUE #1740: Rule proposal: prefer EventTarget instead of EventEmitter
### Description

why? `EventTarget` exist in Deno and browser (unlike `EventEmitter` - which is a NodeJS only thing) and `EventTarget` too also has support for using the options `{ once, signal }` to cancel some event listener you can also know if something is "isTrusted" and emitted by nodejs itself. It also have other useful tool such stopPropagation, target, timestamp and other things. maybe we will expect to see more use of it when/if nodejs decides to add support for something like `ExtendableEvent.waitUntil`

using this rule could also potentially reduce the bundle size and make you think of other solutions that is more cross env friendlier. sometimes you might not always need a event handler at all and you can simply just pass options to a constructor or something else

### Fail

```js
import EventEmitter from 'node:event'

class Foo extends EventEmitter { ... }
```

```js
foo.on('ready', fn)
```

### Pass

```js
class Foo extends EventTarget { ... }
```
```js
// ✓ ok (most oftan you only need one listener)

class Foo {
  onmessage = null
  
  init() {
    this.ws.onmessage = ({data}) => {
      doSomethingWith(data)
      this.onmessage?.(data)
    }
  }
}
```
```js
foo.ready.then(fn)
```

jopemachine:
Thanks for this awesome lib!

Added `prefer-event-target` rule, Fixes #1740.


sami12000 (approved):

rules/prefer-event-target.js
```diff
@@ -0,0 +1,44 @@
...
+const MESSAGE_ID = 'prefer-event-target';
+const messages = {
+	[MESSAGE_ID]: 'Prefer `EventTarget` over `EventEmitter`.',
+};
+
+const eventEmitterClassSelector = [
+	':matches(ClassDeclaration, ClassExpression)',
+	'[superClass.name="EventEmitter"]',
+	'[body.type="ClassBody"]',
+].join('');
```
fisker, line 11:
```suggestion
const eventEmitterSuperClassSelector = [
	':matches(ClassDeclaration, ClassExpression)',
	'[body.type="ClassBody"]',
	' > ',
	'[superClass.name="EventEmitter"]',
].join('');
```

So we can report on the super class instead of the whole class.

rules/prefer-event-target.js
```diff
@@ -0,0 +1,44 @@
...
+
+const newEventEmitterSelector = [
+	'NewExpression',
+	'[callee.name="EventEmitter"]',
+].join('');
```
fisker, line 16: Same here.

rules/prefer-event-target.js
```diff
@@ -0,0 +1,44 @@
...
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = () => ({
+	[eventEmitterClassSelector](node) {
+		return {
+			node,
+			messageId: MESSAGE_ID,
+		};
+	},
+	[newEventEmitterSelector](node) {
```
fisker, line 26: If these two listeners have the same logic, you can `matches()` from `selectors/index.js` to combine these two selectors into one.

test/prefer-event-target.mjs
```diff
@@ -0,0 +1,32 @@
+import outdent from 'outdent';
+import {getTester} from './utils/test.mjs';
+
+const {test} = getTester(import.meta);
+
+test.snapshot({
+	valid: [
+		outdent`class Foo {}`,
+		outdent`class Foo extends OtherClass {}`,
+		outdent`class Foo extends EventTarget {}`,
```
fisker, line 10: It's single line, just use a normal string.

fisker (changes requested):

rules/prefer-event-target.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+const MESSAGE_ID = 'prefer-event-target';
+const messages = {
+	[MESSAGE_ID]: 'Prefer `EventTarget` over `EventEmitter`.',
+};
+
+const eventEmitterClassSelector = [
+	':matches(ClassDeclaration, ClassExpression)',
+	'[superClass.name="EventEmitter"]',
+	'[body.type="ClassBody"]',
```
fisker, line 10: You added this body type check, but not tests covers this, is it possible to have different body here? Maybe in TypeScript? Please test it.

jopemachine: > You added this body type check, but not tests covers this, is it possible to have different body here? Maybe in TypeScript? Please test it.

I'm sorry, but I think I actually don't understand this feedback.

Is it possible to handle typescript with this lib?
I tried to insert a simple typescript statement (`let a: number = 3;`) and tested, but it seems like parsing error occurs.

I think class declaration's with other body type is not possible, according to [this ECMA document](https://tc39.es/ecma262/#sec-class-definitions) in Javascript,
In case of Typescript, [this specification's 117 page](https://javascript.xgqfrms.xyz/pdfs/TypeScript%20Language%20Specification.pdf)

And FYI, when I added this selector, I referenced to `/rules/no-static-only-class.js`'s selector.

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,39 @@
+# Prefer EventTarget instead of EventEmitter
```
sindresorhus, line 1:
```suggestion
# Prefer `EventTarget` over `EventEmitter`
```

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,39 @@
...
+
+<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
+<!-- RULE_NOTICE -->
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+<!-- /RULE_NOTICE -->
+
+While `EventEmitter` could be used in only Node.js, `EventTarget` exist in *Deno* and *Browser* too.
+
+This rule could potentially reduce the bundle size, and make your code more cross-platform friendly.
```
sindresorhus, line 11:
```suggestion
This rule could potentially reduce the bundle size and make your code more cross-platform friendly.
```

sindresorhus: It wouldn't potentially reduce bundle size, it would.

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,39 @@
...
+
+You can check their [differences](https://nodejs.org/api/events.html#eventtarget-and-event-api) between `EventEmitter` and `EventTarget`.
+
+## Fail
+
+```js
+import EventEmitter from 'node:event'
```
sindresorhus, line 18:
```suggestion
import EventEmitter from 'node:event';
```

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,39 @@
...
+<!-- RULE_NOTICE -->
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+<!-- /RULE_NOTICE -->
+
+While `EventEmitter` could be used in only Node.js, `EventTarget` exist in *Deno* and *Browser* too.
+
+This rule could potentially reduce the bundle size, and make your code more cross-platform friendly.
+
+You can check their [differences](https://nodejs.org/api/events.html#eventtarget-and-event-api) between `EventEmitter` and `EventTarget`.
```
sindresorhus, line 13:
```suggestion
See the [differences](https://nodejs.org/api/events.html#eventtarget-and-event-api) between `EventEmitter` and `EventTarget`.
```

sindresorhus (changes requested):

rules/prefer-event-target.js
```diff
@@ -0,0 +1,49 @@
...
+	' > ',
+	'[name="EventEmitter"]',
+].join('');
+
+const newEventEmitterSelector = [
+	'NewExpression',
+	'[callee.name="EventEmitter"]',
+	' > ',
+	'[name="EventEmitter"]',
+].join('');
```
fisker, line 22:
```suggestion
const newEventEmitterSelector = [
	'NewExpression',
	' > ',
	'Identifier.callee',
	'[name="EventEmitter"]',
].join('');
```

rules/prefer-event-target.js
```diff
@@ -0,0 +1,49 @@
...
+	[MESSAGE_ID]: 'Prefer `EventTarget` over `EventEmitter`.',
+};
+
+const eventEmitterSuperClassSelector = [
+	':matches(ClassDeclaration, ClassExpression)',
+	'[superClass.name="EventEmitter"]',
+	'[body.type="ClassBody"]',
+	' > ',
+	'[name="EventEmitter"]',
+].join('');
```
fisker, line 15:
```suggestion
const eventEmitterSuperClassSelector = [
	matches(['ClassDeclaration', 'ClassExpression']),
	'[body.type="ClassBody"]',
	' > ',
	'Identifier.superClass',
	'[name="EventEmitter"]',
].join('');
```

test/prefer-event-target.mjs
```diff
@@ -0,0 +1,32 @@
...
+import {getTester} from './utils/test.mjs';
+
+const {test} = getTester(import.meta);
+
+test.snapshot({
+	valid: [
+		'class Foo {}',
+		'class Foo extends OtherClass {}',
+		'class Foo extends EventTarget {}',
+	],
```
fisker, line 11:
```suggestion
		'class EventEmitter extends Foo {}',
		'class Foo extends foo.EventEmitter {}',
],
```

test/prefer-event-target.mjs
```diff
@@ -0,0 +1,32 @@
...
+	invalid: [
+		'class Foo extends EventEmitter {}',
+		'class Foo extends EventEmitter { someMethod() {} }',
+		outdent`
+			class Foo extends EventEmitter {
+				addListener() {}
+				removeListener() {}
+			}`,
+	],
```
fisker, line 20: We didn't test `ClassExpression`.

```js
const Foo = class extends  EventEmitter {}
```

test/prefer-event-target.mjs
```diff
@@ -0,0 +1,32 @@
...
+});
+
+test.snapshot({
+	valid: [
+		'new EventTarget',
+		'const target = new EventTarget;',
+	],
```
fisker, line 27:
```suggestion
		'EventEmitter()',
		'new Foo(EventEmitter)',
		'new foo.EventEmitter()',
	],
```

fisker (changes requested):

rules/prefer-event-target.js
fisker: I don't know if it's possible to have different body in TS, I guess it's fine to keep this just in case.

If you find anything want test, you can use

```js
test.typescript();
// OR
test.snapshot({
	valid: [
		{
			code: '',
			parser: parsers.typescript,
		}
	],
});
```

There are examples in other rules.

test/prefer-event-target.mjs
fisker: This is not what I wrote, I mean they should be valid.

test/prefer-event-target.mjs
fisker: Same here.

test/prefer-event-target.mjs
jopemachine: > This is not what I wrote, I mean they should be valid.

I'm sorry for my misunderstanding, I think I missed 14 lines' array closed braket.

rules/prefer-event-target.js
```diff
@@ -0,0 +1,48 @@
...
+
+const eventEmitterSuperClassSelector = [
+	matches(['ClassDeclaration', 'ClassExpression']),
+	'[body.type="ClassBody"]',
+	' > ',
+	'Identifier.superClass[name="EventEmitter"]',
+].join('');
+
+const newEventEmitterSelector = [
+	matches(['NewExpression', 'CallExpression']),
```
fisker, line 17:
```suggestion
	'NewExpression',
```

test/prefer-event-target.mjs
```diff
@@ -0,0 +1,47 @@
...
+			}
+		`,
+	],
+});
+
+test.snapshot({
+	valid: [
+		'EventTarget()',
+		'new EventTarget',
+		'const target = new EventTarget;',
```
fisker, line 38:
```suggestion
		'const target = new EventTarget;',
		'const target = new Foo(EventEmitter);',
```

Make sure we only area about the callee.

fisker (approved):
Two comments, otherwise looks good to me. Thanks!

rules/prefer-event-target.js
jopemachine: I think probably we need the `CallExpression` selector here to match the below codes.

```js
// 1
EventEmitter()

// 2
const emitter = EventEmitter()
```

fisker: In theory, we should not allow `EventEmitter` at all, in following examples, `EventEmitter` can use in some other plcaes to construct.

```js
// pass `EventEmitter` to another function
foo(EventEmitter);
```

```js
foo.bar = EventEmitter;
```

```js
foo = {bar: EventEmitter};
```

but I assume you only want to check common cases, "super class" and "new expression".

```js
EventEmitter()
```

is not common.

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,38 @@
...
+See the [differences](https://nodejs.org/api/events.html#eventtarget-and-event-api) between `EventEmitter` and `EventTarget`.
+
+## Fail
+
+```js
+import EventEmitter from 'node:event';
+
+class Foo extends EventEmitter {
+
+}
```
fisker, line 21:
```suggestion
import {EventEmitter} from 'node:event';

class Foo extends EventEmitter {}
```

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+```js
+const emitter = new EventEmitter;
```
fisker, line 25:
```suggestion
const emitter = new EventEmitter();
```

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+## Pass
+
+```js
+class Foo extends EventTarget {
+
+}
```
fisker, line 33:
```suggestion
class Foo extends EventTarget {}
```

docs/rules/prefer-event-target.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+```js
+const target = new EventTarget;
```
fisker, line 37:
```suggestion
const target = new EventTarget();
```

rules/prefer-event-target.js
```diff
@@ -6,26 +6,16 @@ const messages = {
...
-	' > ',
-	'Identifier.superClass',
+const selector = [
+	'Identifier',
 	'[name="EventEmitter"]',
+	matches([
+		'ClassDeclaration > .superClass',
+		'ClassExpression > .superClass',
+		'NewExpression > .callee',
+	]),
```
fisker, line 16: I don't think this rule cases about `ClassBody`, so I just removed it. @jopemachine

fisker (approved):

fisker:
@jopemachine Can you fix the snapshot? `npx ava -u` should work.

jimmywarting:
Woho! thank you for this! 🎉

PR #1793, opened 2022-04-09
Add `prefer-error-cause` rule

### LINKED ISSUE #1342: Rule proposal: `prefer-error-cause`
When throwing a new error in `try/catch` or `Promise#catch`, should add the old error as new error's cause property.

- ["Error Cause" proposal](https://github.com/tc39/proposal-error-cause) 

## Fail

```js
try {} catch {
	throw new Error('oops');
}
```

```js
// (I'm not sure about this)
try {} catch (error) {
	error.message = 'oops';
	throw error;
}
```

```js
promise.catch(() => {
	throw new Error('oops');
})
```

```js
// Same as above
promise.then(null, () => {
	throw new Error('oops');
})
```

## Pass

```js
try {} catch (error) {
	throw new Error('oops', {cause: error});
}
```

```js
promise.catch((error) => {
	throw new Error('oops', {cause: error});
})
```

jopemachine:
Added `prefer-error-cause` rule, Fixes #1342.

I'm trying to create many tests as possible I can.

https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1342#issuecomment-1085527190 this rule is not implemented yet in this PR, but I made this PR because I thought it'd be good to get some feedback before going further.

And this PR includes some uncertain assumptions such that the object argument including the `cause` property should exist at the last argument of the expression.

Any kind of reviews would be highly appreciated.


rules/prefer-error-cause.js
```diff
@@ -0,0 +1,297 @@
...
+	if (!array || array.length === 0) {
+		return undefined;
+	}
+
+	return array[array.length - 1];
+};
+
+// Same logic with '/rules/fix/append-argument.js'
+// But the function does not support NewExpression.
+const appendArgument = (fixer, node, text, sourceCode) => {
```
fisker, line 39: In your case, you only have `NewExpression` with at least one argument, you can update that check logic in /rules/fix/append-argument.js to only throw when `if (node.type === "NewExpression" && node.arguments.length === 0)`.(I hope you know why that check exists). So we can use that here.

rules/fix/append-argument.js
```diff
@@ -2,10 +2,10 @@
 const {isCommaToken} = require('eslint-utils');
 
 function appendArgument(fixer, node, text, sourceCode) {
-	// This function should also work for `NewExpression`
-	// But parentheses of `NewExpression` could be omitted, add this check to prevent accident use on it
+	// Parentheses of `NewExpression` could be omitted,
+	// which means appendArgument should not be called when node.arguments.length is 0
 	/* c8 ignore next 3 */
-	if (node.type !== 'CallExpression') {
+	if (node.type === "NewExpression" && node.arguments.length === 0) {
```
sindresorhus, line 8: Always run lint locally before pushing.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,286 @@
...
+
+	return fixer.insertTextBefore(lastToken, text);
+};
+
+const generateCauseInspector = inspectProperty => objectExpression => {
+	if (!objectExpression || objectExpression.type !== 'ObjectExpression') {
+		return false;
+	}
+
+	return objectExpression.properties.some(prop =>
```
sindresorhus, line 51:
```suggestion
	return objectExpression.properties.some(property =>
```

sindresorhus:
Reviewing a lint rule in detail is a lot of work. Please look over your own diff and try to improve and simplify the code: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1793/files

sindresorhus (changes requested):

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
+/* eslint-disable capitalized-comments */
+/* eslint-disable brace-style */
```
fisker, line 2:
```suggestion
```

Please fix the problem instead of disable the rule

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
...
+	].join(''),
+]);
+
+const peekTop = array => {
+	if (_.isEmpty(array)) {
+		return undefined;
+	}
+
+	return array[array.length - 1];
+};
```
fisker, line 39: This function is meaningless. Use `array[array.length - 1]` directly. Or `const peekTop = array => array[array.length - 1];`

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
...
+	return fixer.insertTextBefore(lastToken, text);
+};
+
+const generateCauseInspector = inspectProperty => objectExpression => {
+	if (!objectExpression || objectExpression.type !== 'ObjectExpression') {
+		return false;
+	}
+
+	return objectExpression.properties.some(property =>
+		property.key.name === 'cause' && inspectProperty(property),
```
fisker, line 56:
```suggestion
		!property.computed && property.key.name === 'cause' && inspectProperty(property),
```

```js
const cause = 'not-cause'
new Error('', {[cause]: 'foo'});
```

We should check `...` element here.

```js
new Error('', {...foo});
```

We can't know if `cause` is in `foo`.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
...
+		':has(ThrowStatement)',
+	].join(''),
+]);
+
+const promiseCatchSelector = matches([
+	[
+		methodCallSelector({method: 'catch', argumentsLength: 1}),
+		':has(ThrowStatement)',
+	].join(''),
+]);
```
fisker, line 31: It seems you didn't cover the second argument in `.then()`.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/59218e3cd653fdf9ad7e7f5d2666e8c2fa775601/rules/catch-error-name.js#L24-L27

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
...
+	const nodesToFixCandidates = getAllNodesToFix({node, context, throwStatement});
+	if (_.isEmpty(nodesToFixCandidates)) {
+		return;
+	}
+
+	const nodesToFix = nodesToFixCandidates.filter(({node}) =>
+		!hasValidCauseProperty(node),
+	);
+
+	if (!_.isEmpty(nodesToFix)) {
```
fisker, line 318: Meaningless check.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
...
+						context,
+						statementToFix: nodeToFix.statementToFix,
+						errorConstructorLastArgument: nodeToFix.node,
+						errorArgumentIdentifier,
+						isCauseNameValid: hasInvalidCauseProperty(nodeToFix.node),
+					});
+
+					if (result) {
+						yield result;
+					}
```
fisker, line 341: return the result directly, the fix function don't has to be a generator function.

fisker (changes requested):

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,388 @@
...
+
+		result.push({
+			node: errorConstructorLastArgument,
+			statementToFix: throwStatement,
+		});
+	}
+	// Assume Error's constructor exists in other node of the block
+	// It could be VariableDeclarator or AssignmentExpression.
+	else {
+		const newExpressionSelector = esquery.parse(matches([
```
fisker, line 220: I think we should use scope information here.

fisker:
I'm really sorry. Super busy recently, I'll review ASAP.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+	[ERROR]: 'Prefer specifying an old error as `cause` option where rethrowing the error.',
+	[SUGGESTION]: 'Specify `cause` option to Error\'s constructor.',
+};
+
+const catchClauseSelector = matches([
+	[
+		'CatchClause',
+		':has(ThrowStatement)',
+	].join(''),
+]);
```
fisker, line 21:
```suggestion
const catchClauseSelector = 'CatchClause:has(ThrowStatement)';
```

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+
+const promiseThenCatchSelector = matches([
+	[
+		matches([
+			methodCallSelector({method: 'then', argumentsLength: 2}),
+			methodCallSelector({method: 'catch', argumentsLength: 1}),
+		]),
+		':has(ThrowStatement)',
+	].join(''),
+]);
```
fisker, line 31:
```suggestion
const promiseThenCatchSelector = [
	matches([
		methodCallSelector({method: 'then', argumentsLength: 2}),
		methodCallSelector({method: 'catch', argumentsLength: 1}),
	]),
	':has(ThrowStatement)',
].join('');
```

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+	insertCauseArgument({fixer, node, sourceCode, value}) {
+		return appendArgument(
+			fixer,
+			node,
+			`{cause: ${value}}`,
+			sourceCode,
+		);
+	},
+
+	insertCauseProperty({fixer, node, sourceCode, value}) {
```
fisker, line 71: Those two methods just makes the code less readable, why not use directly.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+			return;
+		}
+
+		errorArgumentIdentifier = 'error';
+		return fixerUtils.handleEmptyArgument({fixer, node, sourceCode, errorArgumentIdentifier});
+	}
+
+	if (errorConstructorLastArgument.type === 'ObjectExpression') {
+		if (isCauseNameValid) {
+			return fixerUtils.renameObjectProperty({
```
fisker, line 140: Should pass `Property` to `renameObjectProperty()`, not key/value.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+			`{cause: ${value}}`,
+			sourceCode,
+		);
+	},
+
+	insertCauseProperty({fixer, node, sourceCode, value}) {
+		return insertProperty(fixer, node, `cause: ${value}`, sourceCode);
+	},
+
+	renameObjectProperty({fixer, node, key, value}) {
```
fisker, line 75: Should accept `Property` , not key/value.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+		});
+
+		if (!identifier) {
+			throw new Error(`objectExpression doesn't have ${key} property`);
+		}
+
+		return replaceReferenceIdentifier(identifier.value, value, fixer);
+	},
+
+	handleEmptyArgument({fixer, node, sourceCode, errorArgumentIdentifier}) {
```
fisker, line 95: Why not split this into two functions, `insertErrorCatchClauseParameter` and `insertFunctionParameter`, if I didn't get this wrong.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+				statementToFix,
+			});
+		}
+	}
+
+	return result;
+};
+
+const handleCatchBlock = ({node, context, statements, parameter}) => {
+	const throwStatement = statements.find(statement => statement.type === 'ThrowStatement');
```
fisker, line 258: This is not right

```js
try {} catch (error) {
	if (error instanceof SynctaxError) {
		return ''
	} else {
		throw new Error('')
	}
}
```

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+	// Filter block having valid cause property
+	// `try {} catch (err) { throw new Error('oops', {cause: err})}`
+	// `promise.catch(err => { throw new Error('oops', {cause: err})`
+	if (errorArgumentIdentifier && errorArgumentIdentifier === throwStatementArgument.name) {
+		return;
+	}
+
+	// Report none identifier parameter
+	// `try {} catch ({error}) {}`
+	// `promise.catch({error} => {})`
```
fisker, line 279:
```suggestion
	// `promise.then(undefined, ({error}) => {})`
	// `promise.catch(({error}) => {})`
```

rules/prefer-error-cause.js
fisker: Feel hard to implement this, we should make sure the `throw` is in correct scope. Maybe check ancestors to get `function` and `try/catch`, make sure they are in current one?

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+		return;
+	}
+
+	const errorArgumentIdentifier = parameter?.name;
+	const throwStatementArgument = throwStatement.argument;
+
+	// Filter block having valid cause property
+	// `try {} catch (err) { throw new Error('oops', {cause: err})}`
+	// `promise.catch(err => { throw new Error('oops', {cause: err})`
+	if (errorArgumentIdentifier && errorArgumentIdentifier === throwStatementArgument.name) {
```
fisker, line 273: We should not check the name. 


```js
try {} catch (outerError) {
	try {} catch (innerError) {
		throw new Error('', {cause: outerError});
	}
}
```

I think this make sense, just make sure `cause` property exist.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,364 @@
...
+
+	const nodesToFix = nodesToFixCandidates.filter(({node}) =>
+		!hasValidCauseProperty(node),
+	);
+
+	for (const nodeToFix of nodesToFix) {
+		// Prevent repetitive fixing
+		if (hasValidCauseProperty(nodeToFix.node)) {
+			continue;
+		}
```
fisker, line 304: You just `.filter()` with the same condition above.

rules/prefer-error-cause.js
jopemachine: > Feel hard to implement this, we should make sure the `throw` is in correct scope. Maybe check ancestors to get `function` and `try/catch`, make sure they are in current one?

Thanks for pointing this out.
IMHO, we might need to use `esquery` here because throwStatement could emerge in any childnode.
If using esquery is too costly, how about trying to check closest ancestor using `getMatchingAncestorOfType` of `prefer-keyboard-event.key.js`?
I just updated the code using `esquery` for trying to correct this, if this is still incorrect or not in the right direction, please let me know.

fisker: That should work, there is another similar on https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c1cb41a67a5449e58bae60ceeb97279482b60354/rules/no-array-for-each.js#L49

fisker: > we might need to use esquery here because throwStatement could emerge in any childnode.

Maybe we don't need, we can listen on throw expression and find ancestor we want(try-catch or promise callback). Instead of get the ancestor first.

fisker: For example, in this case https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1793#discussion_r860563200, we don't need care about the outer one at all.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,341 @@
...
+		if (parameter && parameter.type !== 'Identifier') {
+			context.report({
+				node,
+				messageId: ERROR,
+			});
+			continue;
+		}
+
+		const nodesToFixCandidates = getAllNodesToFix({node, context, throwStatement});
+		if (_.isEmpty(nodesToFixCandidates)) {
```
fisker, line 274: Empty array can also filter and `for...of`, don't really see the point checking empty array here.

rules/prefer-error-cause.js
jopemachine: > Maybe we don't need, we can listen on throw expression and find ancestor we want(try-catch or promise callback). Instead of get the ancestor first.

I just updated the code by changing the selector to just `ThrowStatement`, and using `getMatchingAncestorOfType` to find out catchBlock.

fisker:
Can you fix the failing tests?

sindresorhus:
@jopemachine Bump :)

jopemachine:
I'm sorry for the late response.
I checked the integration test passed on my local mac.

docs/rules/prefer-error-cause.md
```diff
@@ -0,0 +1,48 @@
+# Prefer specifying an old error as `cause` option where rethrowing the error
```
sindresorhus, line 1:
```suggestion
# Prefer specifying an old error as `cause` option when rethrowing the error
```

sindresorhus: I think this could be better written. "old error" is not really the terminology I would use.

docs/rules/prefer-error-cause.md
```diff
@@ -0,0 +1,48 @@
...
+<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
+<!-- RULE_NOTICE -->
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+🔧 *This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems).*
+<!-- /RULE_NOTICE -->
+
+Prefer specifying old error as `cause` option where rethrowing the error.
+
+This rule also supports custom error classes in addition to built-in one.
```
sindresorhus, line 12:
```suggestion
This rule also supports custom error classes in addition to built-in ones.
```

docs/rules/prefer-error-cause.md
```diff
@@ -0,0 +1,48 @@
...
+
+In case of custom errors, this rule assumes the last argument of the constructor is object able to receive `cause` property.
```
sindresorhus, line 14: This should be part of the previous paragraph and simplified.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,319 @@
...
+const {appendArgument} = require('./fix/index.js');
+const getReferences = require('./utils/get-references.js');
+const getMatchingAncestorOfType = require('./utils/get-matching-ancestor-of-type.js');
+
+const ERROR = 'error';
+const SUGGESTION = 'suggestion';
+
+const messages = {
+	[ERROR]: 'Prefer specifying an old error as `cause` option where rethrowing the error.',
+	[SUGGESTION]: 'Specify `cause` option to Error\'s constructor.',
```
sindresorhus, line 13:
```suggestion
	[SUGGESTION]: 'Specify the `cause` option in the Error\'s constructor.',
```

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,319 @@
...
+		}
+
+		case 'ThrowStatement': {
+			nodeToInsertArgument = statementToFix.argument;
+			break;
+		}
+
+		default: {
+			throw new Error('statementToFix is not invalid type');
+		}
```
sindresorhus, line 122: Add a code coverage ignore comment here.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,319 @@
...
+			reportCannotBeFixed(catchBlock, context);
+			return [];
+		}
+
+		result.push({
+			node: errorConstructorLastArgument,
+			statementToFix: throwStatement,
+		});
+	} else {
+		// Assume Error's constructor exists in other node of the block
```
sindresorhus, line 155:
```suggestion
		// Assume Error's constructor exists in other node's of the block
```

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,319 @@
...
+			} else if (thrownErrorDeclarator.type === 'AssignmentExpression') {
+				targetArguments = thrownErrorDeclarator.right.arguments;
+			}
+
+			errorConstructorLastArgument = peekTop(targetArguments);
+
+			if (!errorConstructorLastArgument) {
+				reportCannotBeFixed(catchBlock, context);
+				return [];
+			}
```
sindresorhus, line 190: Needs a test for this for code coverage.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,319 @@
...
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	[throwStatementSelector](node) {
+		const catchBlock
+			= getMatchingAncestorOfType(node, 'CatchClause')
+			|| getMatchingAncestorOfType(node, 'CallExpression', parent => isThenMethod(parent) || isCatchMethod(parent));
+
+		if (!catchBlock) {
+			return;
+		}
```
sindresorhus, line 280: Needs a test for this for code coverage.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,319 @@
...
+				context,
+				parameter: catchBlock.param,
+				throwStatement: node,
+			});
+		} else {
+			const functionExpression = catchBlock.callee.property.name === 'then' ? catchBlock.arguments[1] : catchBlock.arguments[0];
+			const functionInnerBlockStatements = functionExpression.body?.body;
+			if (!functionInnerBlockStatements) {
+				return;
+			}
```
sindresorhus, line 294: Needs a test for this for code coverage.

test/prefer-error-cause.mjs
```diff
@@ -0,0 +1,429 @@
...
+	invalid: [
+		// ** Not sure #1342
+		// outdent`
+		// 	try {} catch (error) {
+		// 		error.message = 'oops';
+		// 		throw error;
+		// 	}
+		// `,
+
+		// cause not specified
```
sindresorhus, line 88:
```suggestion
		// `cause` not specified
```

test/prefer-error-cause.mjs
```diff
@@ -0,0 +1,429 @@
...
+				throw new CustomError('oops', {}, {});
+			}
+		`,
+	],
+});
+
+test.snapshot({
+	valid: [
+		// ** Not sure need to check inner blocks of the variable
+		// outdent`promise.catch(someCallback);`,
```
sindresorhus, line 225: This needs to be resolved.

test/prefer-error-cause.mjs
```diff
@@ -0,0 +1,429 @@
...
+		'promise.catch(oldError => { throw new Error(\'oops\', {cause:oldError}); });',
+		outdent`
+			promise.catch(oldError => {
+				const error = new Error('oops', {cause:oldError});
+				throw error;
+			});
+		`,
+		outdent`
+			promise.catch(function (oldError) {
+				const error = new Error('oops', {cause:oldError});
```
sindresorhus, line 244:
```suggestion
				const error = new Error('oops', {cause: oldError});
```

sindresorhus (changes requested):

sindresorhus:
Failing: https://github.com/sindresorhus/eslint-plugin-unicorn/runs/7158388862?check_suite_focus=true

docs/rules/prefer-error-cause.md
jopemachine: > I think this could be better written. "old error" is not really the terminology I would use.

How about `caught error`?

```
Prefer specifying a caught error as `cause` option when rethrowing the error
```

rules/prefer-error-cause.js
jopemachine: > Needs a test for this for code coverage.

I wonder if these statements are moot.
I think maybe `FunctionExpression` without 'body(`BlockStatement`).body' is not possible since it causes parsing error.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,320 @@
...
+};
+
+const reportCannotBeFixed = (node, context) => {
+	context.report({
+		node,
+		messageId: ERROR,
+	});
+};
+
+const fixerUtils = {
```
fisker, line 37: Let's define those functions directly, instead of putting on an object.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,320 @@
...
+		}
+
+		errorArgumentIdentifier = 'error';
+
+		// In case of Promise#catch
+		if (catchBlock.type === 'CallExpression') {
+			return fixerUtils.insertFunctionParameter({fixer, catchBlock, sourceCode, errorArgumentIdentifier});
+		}
+
+		return fixerUtils.insertErrorCatchClauseParameter({fixer, catchBlock, errorArgumentIdentifier});
```
fisker, line 90: Why don't we put this one line fix here directly, this just making the code hard to read.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,320 @@
...
+	&& callExpression?.callee.computed !== true
+	&& callExpression?.callee.property.type === 'Identifier'
+	&& callExpression?.callee.property.name === 'catch';
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	[throwStatementSelector](node) {
+		const catchBlock
+			= getMatchingAncestorOfType(node, 'CatchClause')
+			|| getMatchingAncestorOfType(node, 'CallExpression', parent => isThenMethod(parent) || isCatchMethod(parent));
```
fisker, line 277: I haven't read all the code, but I guess you are searching for the "old" error, but I don't think this is the correct way.

```js
try {} catch (error1) {
   return foo(bar).catch(error2 => {
       // This should be caused by `error2` not `error1`
    throw new Error();
  })
      
  function foo(bar) {
     // Not caused by `error1`
     if (bar < 0)
       throw new Error();
  }
}
```

---

I guess we should look for `CatchClause` or `:function`, 
1. not `CatchClause` first.
2. all function, not only argument of `Promise#{then, catch}()`, and check if the function is a `{then,catch}` argument.

jopemachine: I think this is resolved now if I understand correctly.
I replaced existing `getMatchingAncestorOfType` call with `getCatchBlock` for resolving (1).
And I used `esquery` in the `getCatchBlock` for implementing for checking if the function is `{then, catch}`'s argument (2).
And `getCatchBlock` return undefined if the throw statement is in the function to avoid existing wrong fix. (Fixing throw statement in the `function foo` in your example code)

jopemachine:
> Failing: sindresorhus/eslint-plugin-unicorn/runs/7158388862?check_suite_focus=true

I'm trying to resolve this issue, but I'm not sure how to handle this.
CI failed because `run-rules-on-codebase` test include `prefer-error-cause` rule when checking `intergration/test.mjs` file.
But when I added `eslint-disable` comment on the `intergration/test.mjs`'s failed lines, it cause lint error due to eslint can not find `prefer-error-cause` rule.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,392 @@
...
+			&& esquery.matches(expression.arguments[argumentOrder].body, esquery.parse(`[type="CallExpression"][callee.name="${identifier}"]`))
+		) {
+			return true;
+		}
+	}
+
+	return false;
+};
+
+const getCatchBlock = node => {
```
fisker, line 66: I have trouble to understand this function, can we do this.

1. Add a function to search ancestor that is function or `CatchClause`, whatever comes first.
2. Add a function to test that ancestor is `.then`/`.catch`.

The "CatchBlock" should always be a function or `CatchClause`.

Correct me if this logic is wrong.

rules/prefer-error-cause.js
```diff
@@ -0,0 +1,392 @@
...
+				});
+			},
+		});
+	}
+};
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	'ThrowStatement'(node) {
+		const catchBlock = getCatchBlock(node);
```
fisker, line 324: We should do `hasCauseProperty` check first.

test/snapshots/prefer-error-cause.mjs.md
```diff
@@ -0,0 +1,1907 @@
...
+Generated by [AVA](https://avajs.dev).
+
+## Invalid #1
+      1 | try {} catch (error) { throw new Error(); }
+
+> Error 1/1
+
+    `␊
+    > 1 | try {} catch (error) { throw new Error(); }␊
+        |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer specifying an old error as \`cause\` option when rethrowing the error.␊
```
fisker, line 14: Let's report on the new error.

test/snapshots/prefer-error-cause.mjs.md
```diff
@@ -0,0 +1,1907 @@
...
+## Invalid #24
+      1 | try {} catch (error) {
+      2 | 	throw new CustomError('oops', {}, {});
+      3 | }
+
+> Output
+
+    `␊
+      1 | try {} catch (error) {␊
+      2 | 	throw new CustomError('oops', {}, {cause: error});␊
```
fisker, line 611: Let's not auto-fix CustomError, we can't know how it uses `.cause`

rules/prefer-error-cause.js
fisker, line 324: Maybe easier if we listen on `NewExpression` instead of `ThrowStatement`? Do we don't need trace variables.

rules/prefer-error-cause.js
jopemachine: > 1. Add a function to search ancestor that is function or CatchClause, whatever comes first.

I think `getCatchBlock` is doing this although it additionally checks the file scope's callback function.

> 2. Add a function to test that ancestor is .then/.catch's argument.

I added `isThenOrCatchMethod` method for this thing and to simplify the code.

> The "CatchBlock" should always be a function or CatchClause, and it should not cross function or CatchClause.

I think `getCatckBlock` should return `function` or `CatchClause` if my implementation is not wrong.

rules/prefer-error-cause.js
jopemachine: How can we do `hasCauseProperty` check here?
I think maybe we can check `hasCauseProperty` after deciding `Catchblock` and `nodesToFixCandidates` in the block by calling `getAllNodesToFix`.

test/snapshots/prefer-error-cause.mjs.md
jopemachine: I think to implement reporting correctly `new Error` line when fix is possible, I need to extract line number of given Node (and also column start and end) for passing `loc` argument to `context.report`.
By any chance, are there any utility functions for this sort of purpose?

rules/prefer-error-cause.js
jopemachine: IMHO, when listen `NewExpression`, we need to find out if the line is throw statement and if not, find throw statements in the block and check each throw statement's argument.

So I think maybe the code would be complicated as much as now it is.

fisker:
```js
try {} catch (error) {
	throw new Error('message', {cause: foo})
}
```

There is already `cause`, why should we search for old error?

sindresorhus:
Bump

sindresorhus:
Closing for lack of activity.

PR #1803, opened 2022-04-29
Drop support for Node.js v12

fisker:
Enable `unicorn/prefer-top-level-await`

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,9 +10,9 @@ jobs:
       fail-fast: false
       matrix:
         node-version:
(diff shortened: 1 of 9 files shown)


bmish (approved):
Perhaps drop Node 12 and update the recommended rules in separate PRs so they show up separately in the changelog?

sindresorhus:
> Perhaps drop Node 12 and enable update the recommended rules in separate PRs so they show up separately in the changelog?

I agree, but not worth blocking this PR for. I'll make sure to note it in the release notes.

Issue #1751, opened 2022-03-19
Show exact solutions to incorrect code in the documentation


fregante:
The [documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-number-properties.md) shows good and bad examples in two large blocks, but this means that finding the _right_ way to do things takes longer than necessary. I'd suggest this format:

```js
❌ if (Object.is(foo, NaN)) {}
✅if (Object.is(foo, Number.NaN)) {}
```
```js
❌ isNan(1)
✅ Number.isNaN(1)
⚠️ These are not equivalent, you might want this instead:
  ✅ if (typeof numberOrIdk !== number || Number.isNaN(numberOrIdk)) {
```

sindresorhus:
We just borrowed ESLint's docs style, but I agree, it's been sometimes bothering me too.

fregante:
Oh yeah I mentioned that in the other issue and forgot it here.

> This is not Unicorn’s fault as ESLint’s docs are equally lacking

sindresorhus:
I think the code should be valid though.

How about this?

**Blocks**

```js
// ❌
if (Object.is(foo, NaN)) {}

// ✅
if (Object.is(foo, Number.NaN)) {}
```

**Single-line**

```js
var foo = '';   // ❌
const foo = ''; // ✅
```

fisker:
What do you think about live lint result like this? https://ota-meshi.github.io/eslint-plugin-regexp/rules/no-contradiction-with-assertion.html

fregante:
I think that's unrelated. Live rules or not, even with autofix you'd want to see what's the suggested solution.

PR #1811, opened 2022-05-11
Meta: Merge fail and pass examples

fregante:
- Follows https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1751

We can start from rule proposals. This is what new issues would look like:

## Examples

```js
❌ var replace = 'me';
✅ const replace = 'me';
```
```js
function foo() {
  var replace = 'me'; ❌
  return replace;
}

function foo() {
  return 'me'; ✅
}
```


acarfx:
`
❌ const replace = 'me';
✅ let replace = 'me';
`

PR #1813, opened 2022-05-11
`no-array-for-each`: Fix auto-fix causing SyntaxError

### LINKED ISSUE #1812: Auto-fixing 'unicorn/no-array-for-each' breaks the code in certain cases
Autofixing code with 'unicorn/no-array-for-each' rule creates a SyntaxError in cases where `.forEach()` callback has a default argument value.

Let's say I have this piece of code:
```js
items.forEach((item = 0) => {
    console.log(item)
});
```

This code works perfectly fine.

However, if I run `eslint --fix` command then 'unicorn/no-array-for-each' rule autofixes it to this:
```js
for (const item = 0 of items) {
    console.log(item)
}
```
This auto-fixed code has a syntax error in it: `SyntaxError: 'for-of' loop variable declaration may not have an initializer.`.

TimShilov:
Fixes #1812


rules/no-array-for-each.js
```diff
@@ -361,6 +361,7 @@ function isFixable(callExpression, {scope, functionInfo, allIdentifiers, context
 	const parameters = callback.params;
 	if (
 		!(parameters.length === 1 || parameters.length === 2)
+		|| (parameters.length === 1 && parameters[0].type === 'AssignmentPattern')
```
fisker, line 364:
```suggestion
		// `array.forEach((element = defaultValue) => {})`
		|| (parameters.length === 1 && parameters[0].type === 'AssignmentPattern')
```

fisker (approved):
Thank you!

Issue #652, opened 2020-03-31
Rule proposal: `prefer-await`


yvele:
The rule name may be named `unicorn/prefer-await`.

This only applies for ES6 code that have the [`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) keyword available.

#  Prefer `await` over `Promise#then()`, `Promise#catch()` and `Promise#finally()`

Prefer using [`await`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await) operator over [`Promise#then()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then), [`Promise#catch()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) and [`Promise#finally()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally), which is easier to write and to read afterwards.

~~This rule is fixable.~~

## Fail

```js
await doSomething().then(() => {
  // …
});
```

```js
doSomething().then(() => {
  // …
});
```

```js
await doSomething().catch((err) => {
  // …
});
```

```js
doSomething().catch((err) => {
  // …
});
```

## Pass

```js
 await doSomething();
```

```js
try {
  await doSomething();
} catch (err) {
  // …
}
```

fisker:
How about `no-promise-catch`.

yvele:
Yeah right `prefer-await` fits better and will forbid using `.then()`, `.catch()` and `.finally()` methods.

fisker:
There is one rule called [`prefer-await-to-then`](https://github.com/xjamundx/eslint-plugin-promise/blob/master/rules/prefer-await-to-then.js).

There is also proposal in eslint project, https://github.com/eslint/eslint/issues/9649 and also a PR https://github.com/eslint/eslint/pull/11962 :smile:

yvele:
> There is one rule called [`prefer-await-to-then`](https://github.com/xjamundx/eslint-plugin-promise/blob/master/rules/prefer-await-to-then.js) [](/xjamundx/eslint-plugin-promise/blob/HEAD@{2020-03-31T12:12:22Z}/rules/prefer-await-to-then.js).

Sure but the repo looks a bit abandoned.

[...]

PR #1816, opened 2022-05-21
Add `prefer-await` rule

fisker:
Part of #652

The original plan is to add fix for many cases, but it's hard.

The reason I start to implement this rule is that [`promise/eslint-plugin-promise`](https://github.com/xjamundx/eslint-plugin-promise/blob/development/docs/rules/prefer-await-to-then.md) can't check `.then` in `await` expression.

I started to add fix for simple `.then()` only, but this rule already get huge.

I think we should merge this rule first, and try to fix more cases in separate PRs.

In this PR, there is no autofix for any case, it just simply check `.then/catch/finnaly` and provide suggestion for one case, so it should be safe to enable for everyone.


docs/rules/prefer-await.md
```diff
@@ -0,0 +1,68 @@
+# Prefer using `await` operator over `Promise#{then,catch,finally}()`
```
sindresorhus, line 1:
```suggestion
# Prefer using the `await` operator over `Promise#{then,catch,finally}()`
```

rules/prefer-await.js
```diff
@@ -0,0 +1,210 @@
...
+const {getParenthesizedText, isParenthesized} = require('./utils/parentheses.js');
+const shouldAddParenthesesToCallExpressionCallee = require('./utils/should-add-parentheses-to-call-expression-callee.js');
+const getFunctionParameterVariables = require('./utils/get-function-parameter-variables.js');
+const needsSemicolon = require('./utils/needs-semicolon.js');
+
+const MESSAGE_ID_ERROR = 'prefer-await/error';
+const MESSAGE_ID_SUGGESTION = 'prefer-await/suggestion';
+const messages = {
+	[MESSAGE_ID_ERROR]: 'Do not use `Promise#{{method}}(…)`.',
+	[MESSAGE_ID_SUGGESTION]: 'Use `await` expression (UNSAFE).',
```
sindresorhus, line 13:
```suggestion
	[MESSAGE_ID_SUGGESTION]: 'Use `await` operator (UNSAFE FIX).',
```

To make it clear that it's the "fix" that is unsafe, not using "await".

rules/prefer-await.js
```diff
@@ -0,0 +1,210 @@
...
+};
+
+function isInsideTryStatement(node, stopNode) {
```
sindresorhus, line 16: I'm thinking we could make a more general utility for this. Something like `hasAncestor(node, {stopNode})`. It could potentially use another utility internally: `traverseUp()` which would be an Iterable that simply traverses parents.

If you agree, but don't have time to do it now, just add it as a TODO comment.

sindresorhus (changes requested):

fregante:
Would it be possible to add an option to allow `.catch`? I find this pattern pretty convenient over the alternative

```js
const isConnected = await ping().catch(() => false);
```

```js
const username = await getLoggedUser().catch(() => {}); // not logged in
```

If anything I’d probably almost always prefer .catch over a try/catch block, especially when I’m assigning a value. The alternative is suboptimal:

```js
let isConnected = false;
try {
  isConnected = await ping()
} catch {}
```

sindresorhus:
I'm personally ok with an option for that.

sindresorhus:
@fisker What should we do about this one?

sindresorhus:
@fisker Is this something you think you'll ever get to or should I close?

fisker:
Let's close for now.

PR #1818, opened 2022-05-23
Replace `lodash.get` with a native solution

XhmikosR:


sindresorhus (approved):

fisker:
There are unnecessary `?.`s.

XhmikosR:
Maybe, I just went with the safe approach which matches the lodash behavior. Feel free to push to this branch if you have any suggestions.

PR #1820, opened 2022-05-23
Replace `_.lowerFirst` and `_.upperFirst` with native solutions

XhmikosR:
Needs tests but I don't have the time right now to add them.

That being said, after #1818 and this PR there are only 5 lodash functions left:

- [ ] camelCase
- [ ] kebabCase
- [ ] snakeCase
- [ ] defaultsDeep
- [ ] pick

### DIFF
--- a/rules/catch-error-name.js
+++ b/rules/catch-error-name.js
@@ -3,6 +3,7 @@ const {findVariable} = require('eslint-utils');
 const avoidCapture = require('./utils/avoid-capture.js');
 const {renameVariable} = require('./fix/index.js');
 const {matches, methodCallSelector} = require('./selectors/index.js');
+const {upperFirst} = require('./utils/string-utils.js');
 
 const MESSAGE_ID = 'catch-error-name';
 const messages = {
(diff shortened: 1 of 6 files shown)


rules/catch-error-name.js
```diff
@@ -47,7 +48,7 @@ const create = context => {
 		name === expectedName
 		|| ignore.some(regexp => regexp.test(name))
 		|| name.endsWith(expectedName)
-		|| name.endsWith(expectedName.charAt(0).toUpperCase() + expectedName.slice(1));
+		|| name.endsWith(upperFirst(expectedName));
```
XhmikosR, line 51: Changed this place to use the new `upperFirst` function

sindresorhus:
I don't really see the point of removing Lodash functions if we need to manually implement them ourselves.

XhmikosR:
My plan is to finally be able to replace the remaining lodash functions with smaller packages. These two functions are pretty small and easy to implement hence the PR.

That being said, it's mostly a demonstration, feel free to pass.

To my eyes, getting rid of 1000+ files from node_modules will be a huge win for every consumer of the package.

FloEdelmann:
Isn't this PR following the exact strategy that you are linking to @fregante?

https://www.npmjs.com/package/eslint-plugin-you-dont-need-lodash-underscore#_upperFirst

XhmikosR:
My plan was to get rid of lodash and replace it with smaller packages as I mentioned above. I don't have the time to work on it further, but I opened the PR in case someone else wants to proceed with this.

IMHO it's totally worth dropping lodash, but like I said I don't have the time to finish the removal.

Closing.

Issue #1439, opened 2021-07-21
prefer-number-properties can't detect usages via global object


EvgenyOrekhov:
I think `prefer-number-properties` should detect usages via `global`/`window`/`self`:

```js
globalThis.isNaN(foo);
global.isNaN(foo);
window.isNaN(foo);
self.isNaN(foo);

globalThis.isFinite(foo);
global.isFinite(foo);
window.isFinite(foo);
self.isFinite(foo);
```

fisker:
Actually, I tried to use [this](https://eslint-utils.mysticatea.dev/api/scope-utils.html#referencetracker-class) to track global `isNaN` when I implement this rule, the reason I give up this solution is it can't report anything if `{globals: {isNaN: "off"}}` which I already did in my personal config. But maybe I was wrong? Welcome to give a shot.

EvgenyOrekhov:
Also added `globalThis` to examples.

PR #1829, opened 2022-05-26
`new-for-builtins`: Refactor to use `ReferenceTracker`

fisker:
Use [`ReferenceTracker`](https://eslint-utils.mysticatea.dev/api/scope-utils.html#tracker-iterateglobalreferences) instead of rely on the callee names.

As mentioned https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1439#issuecomment-885055275,

If we set `{globals: {Array: 'off'}}`, then

```js
Array()
```

will not be reported.

But this makes sense, because if user set `{globals: {Array: 'off'}}`, then the `Array` will be `undefined`, should already caught by [`no-undef`](https://eslint.org/docs/rules/no-undef), and if user uses `globalThis.Array`, we'll report it.

As a result of this improment, we can also check reassigned `const BuiltinArray = globalThis.Array`.

If we agree on this approach for trace global objects, I can fix https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1439 in the same way.

FYI: Lot's of ESLint builtin rules works this way.


sindresorhus:
> If we agree on this approach for trace global objects, I can fix https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1439 in the same way.

I agree. This is the best approach.

PR #1830, opened 2022-05-26
Add `prefer-logical-operator-over-ternary` rule

### LINKED ISSUE #1798: Rule proposal: `no-useless-ternary`
### Description

I was looking for a rule to forbid ternary that equivalent to a "logical or" expression.

The only thing I got is [`no-unneeded-ternary`](https://eslint.org/docs/rules/no-unneeded-ternary) rule, it checks `x ? x : 1` behind an option. [Even if I set `defaultAssignment: false`, this rule only check identifier](https://eslint.org/docs/rules/no-unneeded-ternary#defaultassignment)

### Fail

```js
foo ? foo : bar;
```

```js
foo.bar ? foo.bar : foo.baz
```

### Pass

```js
foo || bar;
```

```js
foo.bar || foo.baz;
```

### Additional Info

We should check all expressions instead of identifiers only.

This rule is safe to fix to logical or(`||`), but sometimes `??` operator maybe better.

So, we may want to use the suggestions API to draw attention?

fisker:
Fixes #1798

The rule name is not final. https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1798#issuecomment-1138289054


test/snapshots/prefer-string-starts-ends-with.mjs.md
```diff
@@ -541,7 +541,7 @@ Generated by [AVA](https://avajs.dev).
     ␊
     --------------------------------------------------------------------------------␊
     Suggestion 3/3: Use nullish coalescing \`(… ?? '').startsWith()\`.␊
-      1 | (a || b ?? '').startsWith('a')␊
+      1 | ((a || b) ?? '').startsWith('a')␊
```
fisker, line 544: This was bug 

```text
(a || b ?? '').startsWith('a')
VM576:2 Uncaught SyntaxError: Unexpected token '??'
﻿```

File: docs/rules/prefer-logical-operator-over-ternary.md [hunk 2]
```diff
@@ -0,0 +1,52 @@
+# Prefer using logical operator over ternary
```
sindresorhus, line 1:
```suggestion
# Prefer using a logical operator over a ternary
```

rules/prefer-logical-operator-over-ternary.js
```diff
@@ -0,0 +1,155 @@
+'use strict';
+const {isParenthesized, getParenthesizedText} = require('./utils/parentheses.js');
+const isSameReference = require('./utils/is-same-reference.js');
+const shouldAddParenthesesToLogicalExpressionChild = require('./utils/should-add-parentheses-to-logical-expression-child.js');
+const needsSemicolon = require('./utils/needs-semicolon.js');
+
+const MESSAGE_ID_ERROR = 'prefer-logical-operator-over-ternary/error';
+const MESSAGE_ID_SUGGESTION = 'prefer-logical-operator-over-ternary/suggestion';
+const messages = {
+	[MESSAGE_ID_ERROR]: 'Prefer using logical operator over ternary.',
```
sindresorhus, line 10:
```suggestion
	[MESSAGE_ID_ERROR]: 'Prefer using a logical operator over a ternary.',
```

sindresorhus (changes requested):

fisker:
Suggestions applied.

Are you fine with the rule name?

sindresorhus:
> Are you fine with the rule name?

👍

PR #1831, opened 2022-05-26
Verify fix and suggestion output

fisker:
Prevent bug https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1830#discussion_r882559871


sindresorhus:
Good idea 👍

PR #1843, opened 2022-06-19
Add `prefer-object-define-properties` rule

### LINKED ISSUE #1729: Rule proposal: `prefer-object-defineProperties`
### Description

I didn't know [`Object.defineProperties`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties) existed. Would have been nice for a rule to teach me as [I wrote consecutive `Object.defineProperty`](https://github.com/concept-not-found/module-linking-polyfill/blob/b7bc16056f4faf42173cc4fd334d4ef177df0121/packages/transformer/parser/grammar.js#L76-L83).

### Fail

```js
Object.defineProperty(matcher, 'logger', {
  value: () => {},
  writable: true,
})
Object.defineProperty(matcher, 'builder', {
  value: (value) => value.map(({ build }) => build()),
  writable: true,
})
```

### Pass

```js
Object.defineProperties(matcher, {
  logger: {
    value: () => {},
    writable: true,
  },
  builder: {
    value: (value) => value.map(({ build }) => build()),
    writable: true,
  }
})
```

Slowlife01:
Added `prefer-object-define-properties` rule, resolves #1729

Might need more test cases and better description
And unsure if fixer can be improved (formatting the output etc)


Slowlife01:
Should I check if they are in the same scope?

Slowlife01:
I tried to update the snapshot but it doesn't seems to work

fisker:
> Should I check if they are in the same scope?

Shouldn't they next to each other?

Slowlife01:
> > Should I check if they are in the same scope?
> 
> Shouldn't they next to each other?

Oh right

fisker:
This is how I did in other rule https://github.com/sindresorhus/eslint-plugin-unicorn/blob/a7f600796d53d67553959d6d1aa8b0993abd2633/rules/no-array-push-push.js#L22

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,122 @@
+'use strict';
+const isSameReference = require('./utils/is-same-reference.js');
+const {removeArgument} = require('./fix/index.js');
+
+const MESSAGE_ID = 'prefer-object-define-properties';
+const messages = {
+	[MESSAGE_ID]: 'Prefer `{{replacement}}` over multiple `{{value}}`.',
+};
+
+function getFirstExpression(node, sourceCode) {
```
fisker, line 10: This function only works for `ExpressionStatement`, you'll get unexpected result without checking node type

```js
foo(
	Object.defineProperty({}, "a", {}),
	Object.defineProperty({}, "b", {}),
);
```

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,122 @@
...
+				node: reference.identifier.parent.parent,
+				arguments: reference.identifier.parent.parent.arguments,
+			}));
+
+		const groups = [];
+		let currentGroup;
+
+		for (const reference of references) {
+			const firstExpression = getFirstExpression(reference.node.parent, sourceCode);
+			const newGroup = firstExpression?.expression?.callee.object
```
fisker, line 51: We should also check  `Object.defineProperties()` + `Object.defineProperty()`, not only `Object.defineProperty()` + `Object.defineProperty()`.

So we can also handle case that call `Object.defineProperty()` more than twice.

Slowlife01:
Putting this on draft for now

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,154 @@
...
+			}
+		}
+	}
+}
+
+function isObjectDefineProperty(node) {
+	return (
+		node
+		&& node.type === 'CallExpression'
+		&& node.callee.type === 'MemberExpression'
```
fisker, line 30:
```suggestion
		&& node.callee.type === 'MemberExpression'
		&& !node.computed
````

`Object[defineProperty](...)`

File: rules/prefer-object-define-properties.js [hunk 4]
```diff
@@ -0,0 +1,154 @@
...
+				scope: reference.from.block,
+				node: reference.identifier.parent.parent,
+				arguments: reference.identifier.parent.parent.arguments,
+			}));
+
+		const groups = [];
+		let currentGroup;
+
+		for (const reference of references) {
+			const firstExpression = getFirstExpression(reference.node.parent, sourceCode);
```
fisker, line 61: I didn't see `ExpressionStatement` check https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1843#discussion_r909226611, let's only check `ExpressionStatement`.

rules/prefer-object-define-properties.js
fisker: Please add test for it.

rules/prefer-object-define-properties.js
fisker: Need test, you can use https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1843#discussion_r909226611

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,178 @@
...
+					scope: reference.scope,
+					references: [reference],
+				};
+				groups.push(currentGroup);
+			} else {
+				currentGroup.references.push(reference);
+			}
+		}
+
+		for (const group of groups.filter(group => group.references.length > 1)) {
```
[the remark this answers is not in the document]
Slowlife01: How would I go about doing this? I'm a bit confused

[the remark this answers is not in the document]
fisker: We don't need care about the third one, ESLint will report on next round.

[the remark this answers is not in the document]
fisker: See https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/snapshots/no-array-push-push.mjs.md#invalid-15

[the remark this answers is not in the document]
Slowlife01: Not sure if this is how it's supposed to be. I can't get it to report like in the example, but the output seems to be fine

rules/prefer-object-define-properties.js
Slowlife01: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1843/files#diff-9303f48cf071a86d51ccd26a329a6dcb887acc1bb20a64298170dcd3e2c5d2a0R78-R83

I can only think of something like tihs

fisker:
@Slowlife01

I made some changes. I hope it's easier to understand the code. 

Mainly changes:
1. Instead of grouping calls. It only checks two sibling calls.
2. Relax the node type check, not only Identifier/Literal for property
3. Fix the fix logic for some untested node types. (eg, use `...descriptors`)

This not done yet.

1. Need more test for the stricter `isObjectDefinePropertyOrObjectDefineProperties` 
2. The "secondCallExpression" removal is not correct, it's tricky for ASI problem.

Maybe other isssues, I'll review again later.

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,179 @@
...
+		}
+		return `${propertyText}: ${sourceCode.getText(descriptor)}`;
+	}
+
+	const [, descriptors] = callExpression.arguments;
+	if (descriptors.type === 'ObjectExpression') {
+		return descriptors.properties.map(property => sourceCode.getText(property)).join(',\n');
+	}
+
+	return `...(${sourceCode.getText(descriptors)})`;
```
fisker, line 80: @Slowlife01 We are adding `()` to all descriptors, can you exclude some of them with this function https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/should-add-parentheses-to-spread-element-argument.js?

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,179 @@
...
+	return false;
+}
+
+function getDescriptorsText(callExpression, sourceCode) {
+	const method = callExpression.callee.property.name;
+
+	if (method === 'defineProperty') {
+		let [, property, descriptor] = callExpression.arguments;
+		let propertyText = sourceCode.getText(property);
+		if (property.type !== 'Literal') {
```
fisker, line 69: This is not right, if string is "a.b", the fix will cause syntax error. The value should be a indentifier.

rules/prefer-object-define-properties.js
```diff
@@ -0,0 +1,179 @@
+'use strict';
+const isSameReference = require('./utils/is-same-reference.js');
+const {removeArgument} = require('./fix/index.js');
+
+const MESSAGE_ID = 'prefer-object-define-properties';
+const messages = {
+	[MESSAGE_ID]: 'Prefer `{{replacement}}` over multiple `{{value}}`.',
```
fisker, line 7: This message may cause confusing, since the "secondCallExpression" can be `Object.defineProperties`, we need improve it.

rules/prefer-object-define-properties.js
fisker: Sorry, mistake, we didn't unquote the string, so the string won't cause problem.

rules/prefer-object-define-properties.js
fisker: Need add tests, and should do `isParenthesized` check first.

docs/rules/prefer-object-define-properties.md
```diff
@@ -0,0 +1,41 @@
+# Prefer using `Object.defineProperties` over multiple `Object.defineProperty` calls
+
+<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
+<!-- RULE_NOTICE -->
+✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+
+🔧 *This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems).*
+<!-- /RULE_NOTICE -->
+
+When defining more than one properties, [Object.defineProperties](https://mdn.io/Object.defineProperties) should be preferred.
```
XhmikosR, line 10:
```suggestion
When defining more than one property, [Object.defineProperties](https://mdn.io/Object.defineProperties) should be preferred.
```

Slowlife01:
I won't be able to finish this now

PR #1850, opened 2022-06-29
Replace `markdownlint-cli` with `markdownlint-cli2`

fisker:
[markdownlint-cli2](https://github.com/DavidAnson/markdownlint-cli2)


fisker:
I thought it's successor of `markdownlint-cli`, but seems not.

PR #1872, opened 2022-07-30
Make `template-indent` preserve trailing spaces

### LINKED ISSUE #1868: `template-indent` rule should have a setting to preserve trailing spaces
Sometimes you need to have a multiline string where trailing whitespaces are intentional

The plugin breaks the usecases of such strings.

`template-indent`

```js
const str = dedent`
  Line 1 with trailing spaces___
  We use underscores to visualize spaces

  Previous line is intentionally left empty
`;
```

The plugin converts it to

```js
const str = dedent`
  Line 1 with trailing spaces
  We use underscores to visualize spaces
  Previous line is intentionally left empty
`;
```

So the trailing spaces and blank lines are removed which breaks the usecase.

mnaoumov:
Fixes #1868

I realized that having it as a setting is not a right thing, because linter fixes should be essentially the same code, just restyled. Trimming trailing spaces essentially modifies the string and this should be considered as a bug.

Corresponding unit tests were added

### DIFF
--- a/rules/template-indent.js
+++ b/rules/template-indent.js
@@ -67,9 +67,11 @@ const create = context => {
 		}
 
 		const dedented = stripIndent(joined);
+		const trimmed = dedented.replace(new RegExp(`^${eol}|${eol}[ \t]*$`, 'g'), '');
+
 		const fixed
 			= eol
-			+ indentString(dedented.trim(), 1, {indent: parentMargin + indent})
+			+ indentString(trimmed, 1, {indent: parentMargin + indent})
 			+ eol
 			+ parentMargin;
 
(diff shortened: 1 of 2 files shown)


rules/template-indent.js
```diff
@@ -67,9 +67,11 @@ const create = context => {
 		}
 
 		const dedented = stripIndent(joined);
+		const trimmed = dedented.replace(new RegExp(`^${eol}|${eol}[ \t]*$`, 'g'), '');
```
fisker, line 70: I don't think this is correct.

It should only remove trailing empty new lines.

mnaoumov: @fisker This is the way it is already working and I was not sure if we should alter the existing behavior because it would be a breaking change

fisker: Somehow I missed space in the regexp, it's good.

fisker (approved):

PR #1874, opened 2022-08-08
Mention Node version support in `prefer-node-protocol` rule

### LINKED ISSUE #1867: `node:` imports aren't supported in Node.js v14
The [rule.js](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/rule.js#L2-L3) file uses `node:` imports which are supported since Node.js v16, but `eslint-plugin-unicorn` says it supports v14 and up.

The original error message:

```
Error: Failed to load plugin 'unicorn' declared in '.eslintrc.json': Cannot find module 'node:path'
Require stack:
- eslint-plugin-unicorn@43.0.2_eslint@8.20.0/node_modules/eslint-plugin-unicorn/rules/utils/rule.js
- eslint-plugin-unicorn@43.0.2_eslint@8.20.0/node_modules/eslint-plugin-unicorn/index.js
- @eslint+eslintrc@1.3.0/node_modules/@eslint/eslintrc/dist/eslintrc.cjs
```

The issue occurs after upgrading from 42.0.0 to 43.0.2.

It's a bit confusing that the tests are passing despite that they're running with v14, too.

bmish:
Document the supported Node versions for the node protocol import feature to reduce potential confusion or mistaken adoption of this lint rule.

While eslint-plugin-unicorn only supports Node 14.18 and above:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/20e959ffadeca007697534f1ad58c45d29cfdb04/package.json#L14

Some libraries that intend to support all versions of Node 14 might accidentally adopt this lint rule since it's very common to only test against the latest Node 14 and not necessarily against early versions of it.

Fixes #1867.

### DIFF
--- a/docs/rules/prefer-node-protocol.md
+++ b/docs/rules/prefer-node-protocol.md
@@ -9,6 +9,12 @@
 
 When importing builtin modules, it's better to use the [`node:` protocol](https://nodejs.org/api/esm.html#node-imports) as it makes it perfectly clear that the package is a Node.js builtin module.
 
+Note that Node.js support for this feature began in:
+
+> v16.0.0, v14.18.0 (`require()`)
+>
+> v14.13.1, v12.20.0 (`import`)
+
 ## Fail
 
 ```js


docs/rules/prefer-node-protocol.md
```diff
@@ -11,6 +11,12 @@ When importing builtin modules, it's better to use the [`node:` protocol](https:
 
 And don't forget to [upvote this issue](https://github.com/nodejs/node/issues/38343) if you agree.
 
+Note that this feature is not available in early versions of Node 12 and Node 14:
+
+> v16.0.0, v14.18.0 (ESM import and CommonJS require())
```
ddzz, line 16: Why not mention Node 18 here as well?

bmish: This text was copied directly from https://nodejs.org/api/esm.html#node-imports and is only supposed to mention the initial Node versions that supported this feature. Since Node 16 has full support for it, any subsequent versions can be assumed to fully support it as well.

sindresorhus, line 16: This is correct for `require` but not for ESM.

sindresorhus, line 16:
```suggestion
> v16.0.0, v14.18.0 (CommonJS `require()`)
```

Not clear to me why this is mentioning ESM.

docs/rules/prefer-node-protocol.md
```diff
@@ -11,6 +11,12 @@ When importing builtin modules, it's better to use the [`node:` protocol](https:
 
 And don't forget to [upvote this issue](https://github.com/nodejs/node/issues/38343) if you agree.
 
+Note that this feature is not available in early versions of Node 12 and Node 14:
```
sindresorhus, line 14: I don't think we need to mention Node.js 12 as it's not supported anymore.

sindresorhus:
Bump

docs/rules/prefer-node-protocol.md
bmish: Removed unnecessary mentions of ESM/CJS.

docs/rules/prefer-node-protocol.md
bmish: I removed "Node 12" from this sentence. However, I do think we should continue to mention it in the versions list below for completeness / to accurately satisfy the wording "Node.js support for this feature began in". Once all of eslint-plugin-unicorn's Node versions support this feature, I think we can just remove this section completely.

bmish:
Sorry for the delay. Should be ready if still interested.

PR #1886, opened 2022-09-10
Support ESLint new config system

### LINKED ISSUE #1885: Export new eslint config
Currently, configs/all.js and configs/recommended.js do not follow the esilnt's new config system. I think new modules should be created and exported along with the legacy.
Thanks :)

### LINKED ISSUE #2220: Add support for flat config (`eslint.config.js`)
From the next version 9, [ESLint will deprecate](https://eslint.org/blog/2023/10/flat-config-rollout-plans/) the current eslintrc format, and by default will use the [new flat config format](https://eslint.org/docs/latest/use/configure/configuration-files-new) (eslint.config.js). Already from [version 8.21.0](https://github.com/eslint/eslint/releases/tag/v8.21.0) you can use the flat config format. But your [README](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/readme.md) doesn't say anything about how to configure in flat config, so I'm guessing you don't support it.

Guide to migrating a plugin to a flat config: https://eslint.org/docs/latest/extend/plugin-migration-flat-config

jjangga0214:
Supporting eslint's new config system of eslint.

Note that legacy config system always has `require()`d plugins and sharable configs, while the system is ESM.

Thus [conditional export](https://nodejs.org/api/packages.html#conditional-exports) is great to keep compatibility.

Fixes #1885
Fixes #2220


package.json
```diff
@@ -13,6 +13,16 @@
...
 	},
+	"main": "index.js",
+	"exports": {
+		".": {
+			"import": "./index.mjs",
+			"require": "./index.js"
+		},
+		"./all": "./configs/all.mjs",
+		"./recommended": "./configs/recommended.mjs",
+		"./package.json": "./package.json"
```
sindresorhus, line 24: package.json should not be exported.

jjangga0214: I heard some _old_ static analysis tools read package.json by `require()` not `fs.readFile(Sync)`. 
If such a tool runs on modern nodejs (so `require()` can only read what explicitly `exports`ed), it would not work. 

I personally do not have a strong opinion on this issue. How do you think?

jjangga0214:
@sindresorhus 

If I try to create a "drop-in" replacement for the old config system, recommended.js would be like this.

```js
import unicorn from '../index.mjs'
import globals from 'globals'

export default {
	languageOptions: {
		globals: {
			...globals.node,
			...globals.es2021
		}
	},
	plugins: {
		unicorn,
	},
	rules: {}
}
```

However, in the new config system, there is actually no such "root" config.
Everything is kind of `overrides`, which was implemented comparatively recently on the old config system.
So the main intention of design is **chaining**.

`languageOptions` is a root-like options, likely to be configured by the end-users' preference.

If we include the default `languageOptions`, then the end-user might feel convenient when they use unicorn as a drop-in config, but have to manually override `languageOptions`.

Which one do you prefer?

jjangga0214:
@sindresorhus And I also want to ask how you think about including `files` options.

```js
export default {
	files: ['**/*.{js,cjs,mjs,jsx,ts,tsx}'],
	plugins: {
		unicorn,
	},
	rules: { /* ... */ }
}
```

package.json
fisker: I perfer `./*` toexport everything.

sindresorhus: I think it's better to be explicit so we don't accidentally export something we don't want to be a public export.

sindresorhus: We can keep the package.json export until we move to ESM fully.

sindresorhus:
> languageOptions is a root-like options. So I am not sure if it'd be a good convention for a sharable config to provide the default languageOptions in the new config system.

I would prefer to keep including them. We can reevaluate it when the new config system is more mature and has established conventions.

sindresorhus:
> And I also want to ask how you think about including files options.

If that's not the default in the new system, it makes sense to include it here for convenience.

sindresorhus:
You need to fix the lint issues.

jjangga0214:
> I would prefer to keep including them. We can reevaluate it when the new config system is more mature and has established conventions.

✅ Done.

> If that's not the default in the new system, it makes sense to include it here for convenience.

스크린샷 2022-09-11 오후 8 12 08

✅ Yes it's not default and I've configured it.

> You need to fix the lint issues.

✅ Right. Fixed :)

sindresorhus:
Linting is still failing

jjangga0214:
@sindresorhus Well.. fixed again.

package.json
fisker: If so, I don't understand why we need `./recommend` export, shouldn't ESLint load config via plugin entry?

jjangga0214: @fisker In the new system, eslint does not automatically load config from a plugin as well. 
Since `plugin:` protocol(e.g. `plugin:unicorn/recommended`) is no longer used in the new config system, it's not expected for a plugin to have `configs` property. That's why I created index.mjs differently from index.js.
Everything should be explicitly configured by nodejs's native module resolution.
Users should have a way to import sharable configs independently.

fisker: Thanks for the explanation.

jjangga0214:
I wrote a guide for the new config system and usage in readme.
Please check it out :)

readme.md
```diff
@@ -41,6 +41,58 @@ Use a [preset config](#preset-configs) or configure each rules in `package.json`
 }
 ```
 
+## Usage (new: `eslint.config.js`)
+
+From [`v8.21.0`](https://github.com/eslint/eslint/releases/tag/v8.21.0), eslint announced a new config system.
+In the new system, `.eslintrc*` is no longer used. `eslint.config.js` would be the default config file name.
+In eslint `v8`, the legacy system (`.eslintrc*`) would still be supported, while in eslint `v9`, only the new system would be supported.
```
AriPerkkio, line 48: > while in eslint `v9`, only the new system would be supported

ESLint has not made such decision. 

```suggestion
```

sindresorhus:
I think the docs are a bit to verbose. It's not our job to document how the ESLint config system works. We should just show a small example and then link to ESLint docs for more.

readme.md
jjangga0214: Thanks for fixing my confusion :)

jjangga0214:
> It's not our job to document how the ESLint config system works.

Yes, that's 100% right ***logically***.

jjangga0214:
> It's not our job to document how the ESLint config system works.

Yes, that's 100% right ***logically***.
However, ***psychologically***, a person can feel assured by our a little bit verbose docs.
Eslint's docs only describe how the new system works, not how the specific shareable config(unicorn's) should be configured.
Of course, people can surely infer how unicorn should be configured after reading docs, but having assurance is a different matter.
This feeling can be especially useful at the beginning of the new specification.
So we can make our docs concise later, when people get used to the new config system.

fisker:
After learning the new config system. I don't think adding those exports is the right thing.

My recommend config example:

### Preset

```js
import eslintPluginUnicorn from 'eslint-plugin-unicorn';

export default [
	eslintPluginUnicorn.config.recommend,
	// ...
];
```

### Manual

```js
import eslintPluginUnicorn from 'eslint-plugin-unicorn';

export default [
	{
		languageOptions: {
			// ...
		},
		plugins: {
			unicorn: eslintPluginUnicorn,
		},
		rules: {
			'unicorn/better-regex': 'error',
		},
	},
	// ...
];
```

And of course, `require()` can be used.

fisker:
So, I think we only need update `configs/recommended.js` to remove `env`, `parserOptions`, and `plugins`, add `languageOptions`.

The `index.js` file should like this:

```js
const eslintPluginUnicorn = {
	rules: {
		...loadRules(),
		...deprecatedRules,
	},
	configs: {
		recommended: recommendedConfig,
		all: allRulesEnabledConfig,
	},
};

plugin.configs.recommended.plugin = {unicorn: eslintPluginUnicorn};
plugin.configs.all.plugin = {unicorn: eslintPluginUnicorn};

module.exports = eslintPluginUnicorn;
```

jjangga0214:
@fisker 
Well, I disagree.

First, there's no `<plugin>.config` property in the spec of the new config system.
Second, creating a circular dependency is not good, IMHO.

jjangga0214:
> So, I think we only need update configs/recommended.js to remove env, parserOptions, and plugins, add languageOptions.
> The index.js file should like this:

If we patch index.js like this, it's a breaking change that no longer works in the legacy config system.
To avoid the breaking change, we should provide both the legacy and new shareable config.
[conditional export](https://nodejs.org/api/packages.html#conditional-exports) is great to do so.

fisker:
If breaking the legacy config is your concern, adding new entry is better. Anyway, we should not enforce people to use `import` to import the new entry `eslint-plugin-unicorn/recommended`. It should be a commonjs module.

fisker:
> there's no <plugin>.configs property

I don't think so.  `plugin.configs` will be kept for a long time.

jjangga0214:
> It should be a commonjs module.

Yes, it can be CJS as the new config system is available in both ESM and CJS.
But I know @sindresorhus wants the community to migrate to pure ESM eventually.
I think this is good timing to do so progressively.
No matter whether you agree or disagree, I'd respect the decision.
The decision is up to you, maintainers.

---
>  >  there's no `<plugin>.configs` property

> I don't think so. `plugin.configs` will be kept for a long time.

I don't understand what you meant by "I don't think so".
The statement 
___"there's no `<plugin>.configs` property in the spec of the new config system"___
is not my subjective opinion. The statement is an objective fact by itself.

fisker:
Turns out it's recommended  to use `plugins.configs` https://github.com/eslint/rfcs/tree/main/designs/2019-config-simplification#plugins-specifying-their-own-namespaces

fisker:
> "there's no <plugin>.configs property in the spec of the new config system"

I mean we'll keep `.configs` for a long time, since the old config system still functional.

jjangga0214:
> I mean we'll keep `.configs` for a long time, 

Ah, I see :)
Actually, the final decision is always up to you, and I always respect your opinion.
Especially for usage design.
But I just wanted to say my opinion is that `exports` is better than `.configs`.

> Turns out it's recommended to use plugins.configs in the RFC https://github.com/eslint/rfcs/tree/main/designs/2019-config-simplification#plugins-specifying-their-own-namespaces

Ah, actually, no. 🤔 
It's just an example and not necessarily recommended. Eslint does not announce the official design for it. Trust me at this point seriously. The fact is "`.config` and `.environment` is removed from the plugin". This is the only thing eslint officially speak out.

fisker:
Questions:

- Do we want the entry point to be `eslint-plugin-unicorn/configs/recommended` or `eslint-plugin-unicorn/recommended` (@jjangga0214 proposaled original)
- Do we still want to keep the equivalence of `{env: {es2024: true}}` (legacy), which adds `globals` and `ecmaVersion`(already set). To make it work as before I use `@eslint/eslintrc` to get the `globals`
- Anything else not addressed?

sindresorhus:
> Do we want the entry point to be eslint-plugin-unicorn/configs/recommended or eslint-plugin-unicorn/recommended (@jjangga0214 proposaled original)

I think we should stick to what ESLint recommends, just `unicorn.configs.recommended`.

https://eslint.org/docs/latest/use/configure/configuration-files-new#using-configurations-included-in-plugins

sindresorhus:
> Do we still want to keep the equivalence of {env: {es2024: true}} (legacy), which adds globals and ecmaVersion(already take care). To make it work as before I use @eslint/eslintrc to get the globals

No, I don't think that makes sense. We can rather document it and also include it in the recommended config.

sindresorhus:
> Anything else not addressed?

Just make sure it follows the recommendations in https://eslint.org/docs/latest/extend/plugin-migration-flat-config#backwards-compatibility

fisker:
>  just unicorn.configs.recommended.

We can't, we already have `unicorn.configs.recommended` as legacy config. So `unicorn.configs['flat/recommended']` for now?

sindresorhus:
> We can't, we already have unicorn.configs.recommended as legacy config. So unicorn.configs['flat/recommended'] for now?

Yes. Let's do it like that and we can switch it around later on when we fully move to flat config.

jjangga0214:
This would be better, IMHO.

```js
import unicorn from 'eslint-plugin-unicorn/flat'

export default [
        unicorn.configs.recommended,
]
```

fisker:
> This would be better then, IMHO.
> 
> ```js
> import unicorn from 'eslint-plugin-unicorn/flat'
> 
> export default [
>         unicorn.configs.recommended,
> ]
> ```

We can switch back to `unicorn.configs.recommended` in the future.

jjangga0214:
@fisker I know. But I mean `/flat` entry point. You may missed it?

fisker:
If we don't add `/flat` entry, we don't need define `exports`. 

If we add `/flat.js` file, people need use

```
import unicorn from 'eslint-plugin-unicorn/flat.js'
```

and

```
const unicorn = require('eslint-plugin-unicorn/flat')
// OR
const unicorn = require('eslint-plugin-unicorn/flat.js')
```

I think `unicorn.configs['flat/recommended']` is fine.

fisker:
I think it's ready for review.

PR #1895, opened 2022-09-17
`no-array-for-each`: Use `Reference#isWrite` to check parameter reassign

fisker:
Turns out `eslint-scope` has this reassign info, let's use it.


test/snapshots/no-array-for-each.mjs.md
```diff
@@ -3572,7 +3572,7 @@ Generated by [AVA](https://avajs.dev).
 > Output
 
     `␊
-      1 | for (const [index = indexDefaultValue, element = elementDefaultValue] of foo.entries()) {}␊
+      1 | for (let [index = indexDefaultValue, element = elementDefaultValue] of foo.entries()) {}␊
```
fisker, line 3575: It considers parameter default value is "write", but I don't think this is a big deal, not breaking code.

fisker: Or maybe because it's parameter, so it's writeable? https://github.com/eslint/eslint-scope/blob/4c005347cd556b4fa97ba0b626decdd0fce95962/lib/reference.js#L101, but shouldn't every parameter writeable (even without default value)? Not sure how it works.

fisker: There is another [`.isReadOnly()`](https://github.com/eslint/eslint-scope/blob/4c005347cd556b4fa97ba0b626decdd0fce95962/lib/reference.js#L123), but same result

fisker: Update: ESLint core uses  `reference.init === false && reference.isWrite()`, lt's do the same https://github.com/eslint/eslint/search?q=isWrite

PR #1904, opened 2022-09-19
Add `no-unnecessary-await` rule

### LINKED ISSUE #1558: Rule proposal: `no-unnecessary-await`
This prevents mistakes.

## Fail

```js
const foo = await await bar;
```

## Pass

```js
const foo = await bar;
```

---

Or maybe a more generic rule forbid `await` on non-promise objects?
So we can report `await 1`, `await true`, `await nonPromiseObject` etc.
This rule will cover the case above, because `await` result can't be a promise.

fisker:
Fixes #1558


rules/no-unnecessary-await.js
```diff
@@ -0,0 +1,103 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Disallow await non-promise value.',
```
sindresorhus, line 97:
```suggestion
			description: 'Disallow awaiting non-promise values.',
```

sindresorhus (changes requested):

PR #1906, opened 2022-09-20
New integration test

fisker:
1. Use ESLint API instead of CLI to run test
1. Improved log
1. Speed up by spliting into groups


fisker:
The ts parser is so damn slow, took 11min to lint 500 files https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/3089488168/jobs/4997163035#step:5:83

fisker:
Missing ruleId in message https://github.com/eslint/eslint/discussions/16333

fisker:
> Missing `ruleId` in message [eslint/eslint#16333](https://github.com/eslint/eslint/discussions/16333)

Mistake, there is no ruleId if it's a syntax error https://github.com/eslint/eslint/blob/2c152ff0fb709b99e62c19ecd2c95689efacbe4c/lib/linter/linter.js#L845, I guess ESLint can't trace which rule cause that invalid fix.

sindresorhus:
Very nice! 🎉

PR #1913, opened 2022-09-21
Use optional chaining

fisker:


rules/utils/rule.js
```diff
@@ -3,7 +3,7 @@ const path = require('node:path');
 const fs = require('node:fs');
 const getDocumentationUrl = require('./get-documentation-url.js');
 
-const isIterable = object => typeof object[Symbol.iterator] === 'function';
+const isIterable = object => object && typeof object[Symbol.iterator] === 'function';
```
sindresorhus, line 6:
```suggestion
const isIterable = object => typeof object?.[Symbol.iterator] === 'function';
```

fisker:
Double checked, should be good to go. This is only the part of code I checked, there should be more places can use it.

PR #1922, opened 2022-10-01
Add `no-unnecessary-negation` rule

### LINKED ISSUE #856: Rule proposal: simplify negation
## Description
This rule would enforce users to simplify negations: `!(a === b)` -> `a !== b`.
There might be other conditions like this that could be simplified.
I don't know about the rule's name though, `simplify-negation` might be too vague...

## Fail

```js
if (!(a > b)) {}
const foo = !(bar === baz);
const foo = !(typeof bar === 'undefined');
```

## Pass

```js
if (a <= b) {}
const foo = bar !== baz;
const foo = typeof bar !== 'undefined';
const foo = !Array.isArray(bar);
```

jguddas:
Resolves #856.

This does not cover swapping for if-statements and ternary expressions, nor does it cover binary algebra optimization.

### DIFF
--- a/docs/rules/no-unnecessary-negation.md
+++ b/docs/rules/no-unnecessary-negation.md
@@ -2,21 +2,35 @@
 
 <!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
 <!-- RULE_NOTICE -->
-✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
 
-🔧 *This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems).*
+✅ _This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config._
+
+🔧 _This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems)._
+
 <!-- /RULE_NOTICE -->
 
-<!-- Remove this comment, add more detailed description. -->
+Dissallow unnecessary negations when using `Boolean` or `!`.
 
 ## Fail
 
 ```js
-const foo = 'unicorn';
+!(a != b);
+!Boolean(a);
+!!!a;
+if (!!a) {
+}
 ```
 
 ## Pass
 
 ```js
-const foo = '🦄';
+!!a;
+Boolean(a);
+!a;
+
+a == b;
+!a;
+!a;
+if (a) {
+}
 ```
--- a/docs/rules/no-unnecessary-negation.md
+++ b/docs/rules/no-unnecessary-negation.md
@@ -1,0 +1,22 @@
+# This rule checks if conditions can be simpified
+
+<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
+<!-- RULE_NOTICE -->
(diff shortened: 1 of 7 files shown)


jguddas:
### All the places this rule found in integration group 1

- https://github.com/avajs/ava/blob/main/lib/assert.js#L115
- https://github.com/prettier/prettier/blob/main/src/language-js/comments.js#L521
- https://github.com/remix-run/react-router/blob/main/packages/react-router/lib/hooks.tsx#L240
- https://github.com/ariakit/ariakit/blob/main/packages/ariakit/src/menu/__examples__/menu-item-checkbox/index.tsx#L19
- https://github.com/ariakit/ariakit/blob/main/packages/ariakit/src/composite/composite-item.tsx#L386
- https://github.com/ariakit/ariakit/blob/main/packages/ariakit/src/combobox/__examples__/combobox-group/index.tsx#L41
- https://github.com/ariakit/ariakit/blob/main/packages/ariakit/src/menu/__examples__/menu-item-checkbox/index.tsx#L19
- https://github.com/ariakit/ariakit/blob/main/packages/ariakit/src/tab/tab.ts#L90

rules/utils/boolean.js
```diff
@@ -1,10 +1,11 @@
-'use strict';
```
XhmikosR, line 1: This should stay since the file is not ESM?

docs/rules/no-unnecessary-negation.md
```diff
@@ -0,0 +1,36 @@
...
+🔧 _This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems)._
+
+<!-- /RULE_NOTICE -->
+
+Dissallow unnecessary negations when using `Boolean` or `!`.
+
+## Fail
+
+```js
+!(a != b);
```
sindresorhus, line 17:
```suggestion
!(a !== b);
```

docs/rules/no-unnecessary-negation.md
```diff
@@ -0,0 +1,36 @@
...
+## Pass
+
+```js
+!!a;
+Boolean(a);
+!a;
+
+a == b;
+!a;
+!a;
```
sindresorhus, line 33: Duplicates

rules/no-unnecessary-negation.js
```diff
@@ -0,0 +1,118 @@
...
+const {
+	isLogicNot,
+	isBooleanCall,
+	getBooleanAncestor,
+	isSafelyBooleanCastable,
+} = require('./utils/boolean.js');
+
+const MESSAGE_ID = 'no-unnecessary-negation';
+const messages = {
+	[MESSAGE_ID]: 'Expression can be simpified.',
```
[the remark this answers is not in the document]
jguddas: What about: `Expression can use less negations.`?

The problem here is that there are also cases like `if (Boolean(a)) { … }` that will show this message.

docs/rules/no-unnecessary-negation.md
```diff
@@ -0,0 +1,36 @@
+# This rule checks if conditions can be simpified
```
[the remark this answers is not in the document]
jguddas: What about: `Use as few negations as possible`?

Or: `Use as few negations as possible and remove unnecessary Boolean casting`.

jguddas:
I just realized that there is [no-extra-boolean-cast](https://eslint.org/docs/latest/rules/no-extra-boolean-cast).
I will close this for now and if I have time see if there are improvements that can me made in `no-extra-boolean-cast`.

PR #1929, opened 2022-10-09
`prefer-at`: Support private field

### LINKED ISSUE #1928: `prefer-at` rule doesn't support private class members
[`prefer-at`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v44.0.2/docs/rules/prefer-at.md) rule doesn't support private class members.

### `index.js`

```JavaScript
class Foo {
    bar;
    #baz;

    qux() {
        console.log(this.bar[this.bar.length - 1]);
        console.log(this.#baz[this.#baz.length - 1]);
    }
}
```

### `package.json`

```JSON
{
    "name": "testcase",
    "version": "1.0.0",
    "dependencies": {
        "eslint": "8.25.0",
        "eslint-plugin-unicorn": "44.0.2"
    },
    "eslintConfig": {
        "parserOptions": {
            "ecmaVersion": "latest"
        },
        "plugins": ["unicorn"],
        "rules": {
            "unicorn/prefer-at": "error"
        }
    }
}
```

## To reproduce

```Shell
npm install
npx eslint index.js
```

## Results

```
/home/regseb/testcase/index.js
  6:30  error  Prefer `.at(…)` over `[….length - index]`  unicorn/prefer-at

  1 error and 0 warnings potentially fixable with the `--fix` option.
```

regseb:
Fix #1928

I added `case 'PrivateIdentifier':` in [`isSameReference()`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/dd92f06323ceccd26648723008b5e755fe51b712/rules/utils/is-same-reference.js#L131) (cf. https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1928#issuecomment-1272505234) but the private fields are still not supported.

```
prefer-at.mjs › Invalid #20
        1 | class Foo {#bar; baz() {return this.#bar[this.#bar.length - 1]}}

  Invalid case should have at least one error.

  Value is deeply equal:

  []

  › file://test/utils/snapshot-rule-tester.mjs:188:8

  ─

  1 test failed
```


fisker:
`MemberExpression` should update too.

regseb:
Tests passed with `"PrivateIdentifier"` in `getStaticPropertyName()`. [2a9e359#diff](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1929/commits/2a9e359ba55fde6eb3f64cb4b246f350a284df95#diff-3c0e4635c33c28fbf6f80411e3c4ecbd0949be7038871cac1285618a654e9aa2)

There isn't `"PrivateIdentifier"` in [ESLint](https://github.com/eslint/eslint/blob/94ba68d76a6940f68ff82eea7332c6505f93df76/lib/rules/utils/ast-utils.js#L283). There is a bug in ESLlint?

fisker:
I mean this part https://github.com/eslint/eslint/blob/94ba68d76a6940f68ff82eea7332c6505f93df76/lib/rules/utils/ast-utils.js#L420, your solution isn't correct, it will cause `this.#foo[this.foo.length - 1]` pass.

fisker (approved):

PR #1930, opened 2022-10-10
Automate docs with eslint-doc-generator

bmish:
I built this CLI tool [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) for automating the generation of the README rules list and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of our custom rules through better documentation and streamline the process of adding new rules).

This is a follow-up to the work I did to add the original notices in #1576. eslint-doc-generator is significantly more robust compared to the previous documentation generator script/tests which I have now removed. It has additional functionality (for example, the rules list legend is auto-generated, and the rules list will show additional columns of information when applicable) and 100% test coverage.

There are some discrepancies between the old docs and the new docs. Some of these are intentional and likely improvements. If you don't like any of the changes, let me know and I can tweak eslint-doc-generator or add a config option to support the desired behavior.


fisker:
I slightly prefer to keep codes in `scripts/`, so we can tweak them anytime we want.

The only missing part is the rule title, it's easy to support, I planned to do that, but haven't got time.

bmish:
> I slightly prefer to keep codes in scripts/, so we can tweak them anytime we want.

It is true you can more-easily tweak the current in-house scripts. I'm hoping the benefits I described in the PR description will outweigh that downside, in addition to these selling points:

* I plan to move [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) to @eslint-community sometime after reaching v1
* eslint-doc-generator will be less likely to break vs. maintaining in-house scripts since it is fully unit tested, written in TypeScript, and will be adopted by / tested on many of the top ESLint plugins
* eslint-doc-generator will be increasingly-configurable so many desired tweaks will be possible simply through a config option

> The only missing part is the rule title, it's easy to support, I planned to do that, but haven't got time.

What do you mean the "rule title" is missing?

fisker:
> What do you mean the "rule title" is missing?

I mean, currently, we are not auto-generating rule titles, but this PR did.

fisker:
I'll leave the decision to Sindre.

docs/rules/better-regex.md
```diff
@@ -1,11 +1,10 @@
 # Improve regexes by making them shorter, consistent, and safer
 
-<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
-<!-- RULE_NOTICE -->
-✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+✅ This rule is enabled in the `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs).
```
sindresorhus, line 3:
```suggestion
✅ This rule is enabled in the [`recommended` config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs).
```

THis feels more correct. The link is for the "recommended config".

docs/rules/better-regex.md
```diff
@@ -1,11 +1,10 @@
...
 
-🔧 *This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems).*
-<!-- /RULE_NOTICE -->
+🔧 This rule is automatically fixable by the `--fix` [CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
```
sindresorhus, line 5: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1930/files#r990917116
```suggestion
🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
```

docs/rules/consistent-destructuring.md
```diff
@@ -1,11 +1,12 @@
...
-<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
-<!-- RULE_NOTICE -->
-✅ *This rule is part of the [recommended](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config) config.*
+✅ This rule is enabled in the `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs).
 
-🔧💡 *This rule is [auto-fixable](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) and provides [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).*
-<!-- /RULE_NOTICE -->
+🔧 This rule is automatically fixable by the `--fix` [CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+💡 This rule is manually fixable by editor [suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
```
sindresorhus, line 7:
```suggestion
💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
```

docs/rules/better-regex.md
```diff
@@ -1,11 +1,10 @@
 # Improve regexes by making them shorter, consistent, and safer
 
-<!-- Do not manually modify RULE_NOTICE part. Run: `npm run generate-rule-notices` -->
```
sindresorhus, line 3: @bmish I think it may be smart to have a notice like this.

Maybe:

```
<!-- end rule header -->
<!-- DO NOT EDIT ANYTHING ABOVE HERE -->
```

sindresorhus:
I'm a big 👍 on this. I hope all ESLint plugins adopt this so we can get some universal consistency.

sindresorhus:
You need to ensure the rule names are not word wrapped. See the difference between our and yours:

- https://github.com/sindresorhus/eslint-plugin-unicorn#rules
- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/68a674c09162e8dac934be3c3e9bcfb018158143/readme.md#rules

sindresorhus:
Can you apply `--rule-doc-section-include` for the Fail and Pass sections?

sindresorhus:
Would be nice if you could define the config in package.json instead of as CLI flags.

https://github.com/sindresorhus/pkg-conf or a similar package.

sindresorhus:
The docs for `--rule-doc-title-format` should have an example for each style.

https://github.com/bmish/eslint-doc-generator#configuration-options

bmish:
Thanks for the feedback! Will work on making the tweaks and responding in the next few days.

docs/rules/better-regex.md
bmish: Agree with the other suggestions, but for this one, I currently only have a [`--url-configs` option](https://github.com/bmish/eslint-doc-generator#configuration-options) for specifying a link to general documentation about configurations in the plugin, and not the `recommended` config specifically. So that's why I put the link on the `config` word only. Do you still think it makes sense to include the config name in the link, even though the link isn't specifically pointing to that configuration?

sindresorhus: I missed that it was not pointing at a specific config. You're right. Doesn't make sense.

docs/rules/better-regex.md
bmish: Fixed.

docs/rules/consistent-destructuring.md
bmish: Fixed.

package.json
```diff
@@ -14,14 +14,14 @@
...
+		"create-rule": "node ./scripts/create-rule.mjs && npm run fix:eslint-docs",
 		"fix": "run-p --continue-on-error fix:*",
+		"fix:eslint-docs": "eslint-doc-generator --ignore-deprecated-rules --ignore-config all --rule-doc-title-format desc --url-configs \"https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs\"",
 		"fix:js": "npm run lint:js -- --fix",
 		"fix:md": "npm run lint:md -- --fix",
-		"generate-rule-notices": "node ./scripts/generate-rule-notices.mjs",
-		"generate-rules-table": "node ./scripts/generate-rules-table.mjs",
 		"integration": "node ./test/integration/test.mjs",
 		"lint": "run-p --continue-on-error lint:*",
+		"lint:eslint-docs": "npm-run-all \"fix:eslint-docs --check\"",
```
ddzz, line 24: Is `npm-run-all` needed here?

bmish: `npm-run-all` allows us to call another NPM script without having to specify either `npm` or `yarn` or whatever package manager the user might be using.

docs/rules/better-regex.md
bmish: I added those original warning comments back in. But they aren't managed by eslint-doc-generator as they are outside the managed area indicated by the `<!-- end rule header -->` comment.

bmish: Filed an issue to consider making this part of the standard comment marker: https://github.com/bmish/eslint-doc-generator/issues/150

package.json
fisker:
```suggestion
		"lint:eslint-docs": "npm run fix:eslint-docs -- --check",
```

Make sense to use npm run, like other scripts did.

fisker (approved):
One comment.

PR #1943, opened 2022-10-28
Update eslint-doc-generator to 0.17

bmish:
Follow-up to #1930. Minor version update to [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator).

Changes that affect us here:

* Marker comments make it more clear the content is auto-generated (https://github.com/bmish/eslint-doc-generator/pull/155)
* Rule doc notices now indicate when a config disables a rule (https://github.com/bmish/eslint-doc-generator/pull/156)
* Improved non-breaking spacing of rule names in rules list (https://github.com/bmish/eslint-doc-generator/pull/172)


sindresorhus:
Linting is failing

sindresorhus (approved):

readme.md
```diff
@@ -44,122 +44,122 @@ Use a [preset config](#preset-configs) or configure each rules in `package.json`
...
-
-<!-- end rules list -->
+| Name                                                                                             | Description                                                                                                                                                                                                       | ✅  | 🔧  | 💡  |
+| :----------------------------------------------------------------------------------------------- | :---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :-- | :-- | :-- |
+| [better-regex](docs/rules/better-regex.md)                                                       | Improve regexes by making them shorter, consistent, and safer.                                                                                                                                                    | ✅  | 🔧  |     |
```
fisker, line 55: Is there an option to not align this table? It's so hard to see diff.

bmish: Normally, the diff will be clearer because the columns will only resize if the longest item in them changes. Most of the time, only a single row will be affected when rule data changes. So hopefully, this won't be a problem most of the time.

The reason the entire table is affected this time is because of the change to the non-breaking spaces in the "Name" column header which helpfully reduced the width of that column substantially.

fisker: But still possible add a new longest rule name.

fisker: Is it aligned because other repo use Prettier?

bmish: The columns are currently aligned for two reasons:

1. We format the generated markdown with prettier if prettier is installed in the repo, and prettier aligns columns.
2. We use [markdown-table](https://www.npmjs.com/package/markdown-table) to generate the table which defaults to `alignDelimiters`.

Benefits of aligning the columns:

* The table markdown looks pretty and easy to read.
* The columns will only resize when the longest rule name or description changes (rare).
* Since this section is auto-generated and managed by the tool, having a minimal diff doesn't really matter.

Benefits of not aligning the columns:

* Always a minimal diff.

If you still prefer, I suppose I could add an `alignDelimiters` option to eslint-doc-generator to pass-thru to markdown-table. I just want to make sure there's a good argument for adding a fairly low-level option like this.

I think I may also try to remove the prettier formatting entirely and let the consumer run prettier on their own if they want to. This would reduce complexity of the tool and put the consumer in control.

fisker: Do whatever you decide in your repo. Don't have strong opinion.

fisker: BTW, the `removeTrailingNewline` you are using, isn't it a `.trimEnd()`? That `RegExp` is much slower.

bmish: Thanks, fixed: https://github.com/bmish/eslint-doc-generator/pull/186. Appreciate the feedback.

bmish: Removing prettier formatting in: https://github.com/bmish/eslint-doc-generator/pull/188

PR #1949, opened 2022-11-07
Add two projects to integration test

fisker:
These two projects are popular, and they use `eslint-plugin-unicorn`


fisker:
#1950 Should fix the test.

PR #1953, opened 2022-11-07
`prevent-abbreviations`: Improve fix for `retVal`

### LINKED ISSUE #1937: `prevent-abbreviation` suggests replacing `someRetVal` with `someReturnValueValue`
Rule `unicorn/prevent-abbreviation` suggests replacing `someRetVal` with `someReturnValueValue`.

Rule: `unicorn/prevent-abbreviation`

```js
const expectedRetVal = 'that should be ok';

// Expected:
const expectedReturnValue = 'that should be ok';

// Actual:
const expectedReturnValueValue = 'that should be ok';
```

fisker:
Fixes #1937


sindresorhus:
Would it not be better to make the check more general and ensure we don't repeat any words?

PR #1962, opened 2022-11-14
`prefer-string-replace-all`: Ignore regex with pipe

### LINKED ISSUE #1961: `prefer-string-replace-all` may suggest an incorrect alternative
`unicorn/prefer-string-replace-all` seems to be reporting incorrect alternatives for regular expressions with the global flag & with a `|`. Consider the following:

```TypeScript
declare const content: string;

content.replace(/\*|_/g, ""); // Prefer `String#replaceAll()` over `String#replace()`. eslint(unicorn/prefer-string-replace-all)
```

Upon applying the suggested fix, the code results to:

```TypeScript
declare const content: string;

content.replaceAll('*|_', "");
```

The regular expression has been lost and the code obviously does not have the original intentions.

It looks like the issue is that `|` is not included here:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/505a2039627350fe7ae0b990d48cd035d8184949/rules/prefer-string-replace-all.js#L20-L23

<hr>

- Node.js: 18.8.0
- @typescript-eslint/eslint-plugin: 5.42.1
- @typescript-eslint/parser: 5.42.1
- eslint: 8.27.0
- eslint-plugin-unicorn: 44.0.2
- typescript 4.8.4

hampuskraft:
Fixes #1961

### DIFF
--- a/test/prefer-string-replace-all.mjs
+++ b/test/prefer-string-replace-all.mjs
@@ -87,6 +87,11 @@ test({
 			output: 'foo.replaceAll(\'\\\\.\', bar)',
 			errors: [error],
 		},
+		{
+			code: 'foo.replace(/\\|/g, bar)',
+			output: 'foo.replaceAll(\'|\', bar)',
+			errors: [error],
+		},
 	],
 });
 
--- a/test/prefer-string-replace-all.mjs
+++ b/test/prefer-string-replace-all.mjs
@@ -14,6 +14,7 @@ test({
 		'foo.replace(/[a]/g, bar)',
 		'foo.replace(/a?/g, bar)',
 		'foo.replace(/.*/g, bar)',
+		'foo.replace(/a|b/g, bar)',
 		'foo.replace(/\\W/g, bar)',
 		'foo.replace(/\\u{61}/g, bar)',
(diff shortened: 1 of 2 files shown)


test/prefer-string-replace-all.mjs
```diff
@@ -14,6 +14,7 @@ test({
 		'foo.replace(/[a]/g, bar)',
 		'foo.replace(/a?/g, bar)',
 		'foo.replace(/.*/g, bar)',
+		'foo.replace(/a|b/g, bar)',
```
fisker, line 17: Can you add a test for escaped `|`?

hampuskraft: Should be resolved in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1962/commits/fbbccddcec24f9d03b2330ba6ea4138f0738f1bc, and the `unicorn/better-regex` error was solved in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1962/commits/f67a9f2ca208be1763d27a2af09b0dfcf39ba279.

fisker (approved):

PR #1964, opened 2022-11-16
`filename-case`: Add `includePath` option

### LINKED ISSUE #686: Enforce directory casing in `filename-case`
We'd like to enforce casing of both files and directories. Is that possible, or would it need some kind of "root" directory to avoid caring about directory name outside of the project? Easy enough to achieve with js config files I guess. There is `context.getCwd()`, could probably work as well

mrmckeb:
This PR aims to resolve #686 by allowing the rule to also lint paths.

I've added varied test cases, and have manually tested locally against a production project.

### DIFF
--- a/docs/rules/filename-case.md
+++ b/docs/rules/filename-case.md
@@ -105,3 +105,20 @@ Don't forget that you must escape special characters that you don't want to be i
 	}
 ]
 ```
+
+### includePath
+
+Type: `boolean`\
+Default: `false`
+
+Checks the path for case style, in addition to the filename.
+
+```js
+"unicorn/filename-case": [
+	"error",
+	{
+		"case": "kebabCase",
+		"includePath": true,
+	}
+]
+```
--- a/rules/filename-case.js
+++ b/rules/filename-case.js
@@ -1,6 +1,5 @@
 'use strict';
 const path = require('node:path');
-const process = require('node:process');
 const {camelCase, kebabCase, snakeCase, upperFirst} = require('lodash');
 const cartesianProductSamples = require('./utils/cartesian-product-samples.js');
 
@@ -157,7 +156,7 @@ const create = context => {
 			const extension = path.extname(filenameWithExtension);
 			const filename = path.basename(filenameWithExtension, extension);
 
-			const relativeFilenameWithExtension = path.relative(process.cwd(), filenameWithExtension);
+			const relativeFilenameWithExtension = path.relative(context.getCwd(), filenameWithExtension);
 			const dirname = path.dirname(relativeFilenameWithExtension);
 			const directoryAndFilename = dirname + path.sep + filename;
 
--- a/rules/filename-case.js
+++ b/rules/filename-case.js
@@ -162,7 +162,19 @@ const create = context => {
 
 			const base = filename + extension;
 
-			if (ignoredByDefault.has(base) || ignore.some(regexp => regexp.test(includePath ? relativeFilenameWithExtension : base))) {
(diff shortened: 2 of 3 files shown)


docs/rules/filename-case.md
```diff
@@ -105,3 +105,20 @@ Don't forget that you must escape special characters that you don't want to be i
...
+Default: `false`
+
+Checks the path for case style, in addition to the filename.
+
+```js
+"unicorn/filename-case": [
+	"error",
+	{
+		"case": "kebabCase",
+		"includePath": true,
```
fisker, line 121: I don't think we need an option to enable the check. Should always enabled.

rules/filename-case.js
```diff
@@ -154,13 +156,19 @@ const create = context => {
 		Program() {
 			const extension = path.extname(filenameWithExtension);
 			const filename = path.basename(filenameWithExtension, extension);
+
+			const relativeFilenameWithExtension = path.relative(process.cwd(), filenameWithExtension);
```
fisker, line 160: #686 Suggest to use `context.getCwd()`

docs/rules/filename-case.md
mrmckeb: That would make this a breaking change. Is that OK?

I agree that it makes more sense as a default, but perhaps it would then need an opt-out? Would we call that `excludePath` or `fileOnly`, or perhaps something else?

rules/filename-case.js
mrmckeb: Done, thanks!

fisker (changes requested):

rules/filename-case.js
```diff
@@ -154,13 +155,19 @@ const create = context => {
...
 			const filename = path.basename(filenameWithExtension, extension);
+
+			const relativeFilenameWithExtension = path.relative(context.getCwd(), filenameWithExtension);
+			const dirname = path.dirname(relativeFilenameWithExtension);
+			const directoryAndFilename = dirname + path.sep + filename;
+
 			const base = filename + extension;
 
-			if (ignoredByDefault.has(base) || ignore.some(regexp => regexp.test(base))) {
+			if (ignoredByDefault.has(base) || ignore.some(regexp => regexp.test(includePath ? relativeFilenameWithExtension : base))) {
```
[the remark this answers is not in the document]
mrmckeb: Agreed, I'll split here.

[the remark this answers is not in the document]
fisker: If file is not in CWD, we should also ignore directories.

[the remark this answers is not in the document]
mrmckeb: Done, thanks!

docs/rules/filename-case.md
mrmckeb: Can this be an option that we add later? We could roll this out as a feature release, and if there's demand, we can change the default and extend the rule again later?

I can see that some people may also want to go down the path of having different rules for different directories - like `PascalCase` for React components, but `kebab-case` for other directories. That becomes quite complex to build out.

fisker: We always do major release. https://github.com/sindresorhus/eslint-plugin-unicorn/releases

effinrich:
First, great library, best filename linter I've used by far. This feature would make it perfect. Is it dead in the water, or will it make it to release?  I see there were changes requested, but then nada for a couple of months.

mrmckeb:
I haven't had time to come back to this sorry @effinrich. If you'd like to pick it up, that would be great.

sindresorhus:
Closing for lack of activity.

PR #1965, opened 2022-11-16
`prefer-string-replace-all`: Report all `String#replace()` when the pattern has `g` flag

### LINKED ISSUE #1948: Enforce `.replaceAll()` over `.replace()` when the pattern has `g` flag
I believe they are the same, but use `.replaceAll()` has more clear intention

# Fail

```js
foo.replace(/\s/g, '')
```

# Pass

```js
foo.replaceAll(/\s/g, '')
```

```js
foo.replace(/\s/, '')
```

```js
foo.replace(" ", '')
```

fisker:
Fixes #1948


docs/rules/prefer-string-replace-all.md
```diff
@@ -8,11 +8,16 @@
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
 The [`String#replaceAll()`](https://github.com/tc39/proposal-string-replaceall) method is both faster and safer as you don't have to escape the regex if the string is not a literal.
+Even if regexp has to be used, `String#replaceAll()` has more clear intention.
```
sindresorhus, line 11:
```suggestion
The [`String#replaceAll()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) method is both faster and safer as you don't have to use a regex and remember to escape it if the string is not a literal. And when used with a regex, it makes the intent clearer.
```

PR #1966, opened 2022-11-17
Add `no-typeof-undefined` rule

### LINKED ISSUE #1492: Rule proposal: `no-typeof-undefined`

fisker:
Fixes #1492


rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,125 @@
...
+						&& !isParenthesized(typeofNode, sourceCode)
+					) {
+						yield * addParenthesizesToReturnOrThrowExpression(fixer, parent, sourceCode);
+						return;
+					}
+
+					const tokenBefore = sourceCode.getTokenBefore(binaryExpression);
+					if (needsSemicolon(tokenBefore, sourceCode, secondToken.value)) {
+						yield fixer.insertTextBefore(binaryExpression, ';');
+					}
```
fisker, line 93: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e4aaa42a780e2b155fa25fcba457162308dec36b/rules/no-negated-condition.js#L105, but I'm not going to make a reusable function in this PR, will do later.

docs/rules/no-typeof-undefined.md
```diff
@@ -0,0 +1,59 @@
+# Enforce compare with `undefined` directly
```
sindresorhus, line 1:
```suggestion
# Disallow comparing `undefined` using `typeof`
```

docs/rules/no-typeof-undefined.md
```diff
@@ -0,0 +1,59 @@
...
+
+✅ This rule is enabled in the `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs).
+
+🔧💡 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) and manually fixable by [editor suggestions](https://eslint.org/docs/developer-guide/working-with-rules#providing-suggestions).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+Enforce compare value with `undefined` directly instead of compare `typeof value` with `'undefined'`.
```
sindresorhus, line 10:
```suggestion
Checking if a value is `undefined` by using `typeof value === 'undefined'` is needlessly verbose. It's generally better to compare against `undefined` directly. The only time `typeof` is needed is when a global variable potentially does not exists, in which case, using `globalThis.value === undefined` may be better.
```

docs/rules/no-typeof-undefined.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## Options
+
+### checkGlobalVariables
+
+Type: `boolean`\
+Default: `false`
+
+This rule ignores variables not defined in file by default.
```
sindresorhus, line 47:
```suggestion
The rule ignores variables not defined in the file by default.
```

rules/no-typeof-undefined.js
```diff
@@ -0,0 +1,140 @@
...
+const isOnSameLine = require('./utils/is-on-same-line.js');
+const needsSemicolon = require('./utils/needs-semicolon.js');
+const {
+	isParenthesized,
+} = require('./utils/parentheses.js');
+
+const MESSAGE_ID_ERROR = 'no-typeof-undefined/error';
+const MESSAGE_ID_SUGGESTION = 'no-typeof-undefined/suggestion';
+const messages = {
+	[MESSAGE_ID_ERROR]: 'Compare with `undefined` directly instead of `typeof` check.',
```
sindresorhus, line 17:
```suggestion
	[MESSAGE_ID_ERROR]: 'Compare with `undefined` directly instead of using `typeof`.',
```

sindresorhus (changes requested):

PR #1971, opened 2022-11-18
`prefer-string-replace-all`: Improve RegExp to string fix

fisker:
Add `regjsparser` to better understand RegExp.

I'm not sure how to persevere escape.

Should be able to imporve `prefer-string-starts-ends-with` with this module too.


sindresorhus:

sindresorhus:
> I'm not sure how to persevere escape.

Maybe leave a TODO comment about it as a future improvement?

PR #1973, opened 2022-11-18
Run `prefer-export-from` test with ESpree

fisker:


sindresorhus:
Linting is failing

fisker:
Why am I sending from my main branch...

PR #1977, opened 2022-11-18
`explicit-length-check`: Ignore `.length || number`

### LINKED ISSUE #1933: `explicit-length-check` rule should not be triggered when length is part of a numeric expression
The `explicit-length-check` rule is getting triggered in the following context where the length is actually being used for it's numeric value rather than as a boolean.  Telling the user to convert it to a logical comparison expression is not appropriate.

CleanShot 2022-10-21 at 15 41 12@2x

`explicit-length-check`

```js
  const x = [1,2].length || 1;
```

CleanShot 2022-10-21 at 15 44 20@2x

fisker:
Fixes #1933

Ideally, we should only ignore when right side is a number, but not sure worth the effort.


sindresorhus:
> Ideally, we should only ignore when right side is a number, but not sure worth the effort.

Can't we just place a simple `isLiteralNumber(…) || !isLiteral(…)` check in there?

sindresorhus:
> Can't we just place a simple isLiteralNumber(…) || !isLiteral(…) check in there?

If this is too difficult, I'm happy to just open an issue for that as a future improvement.

yoyo837:
Hello, Why is this still in draft?

fisker:
I haven't got time to finish it. Will back to this soon.

fisker:
Added check, only ignores when `|| number`. Sorry for the long delay.

PR #1981, opened 2022-11-19
`prefer-string-replace-all`: Check pattern even it's already using `.replaceAll`

fisker:


rules/prefer-string-replace-all.js
```diff
@@ -82,13 +84,37 @@ const create = context => ({
...
+		if (methodName === 'replaceAll') {
+			if (!patternReplacement) {
+				return;
+			}
+
+			return {
+				node: pattern,
+				messageId: MESSAGE_ID_USE_STRING,
+				data: {
+					replacement: patternReplacement.length < 20 ? patternReplacement : 'literal',
```
sindresorhus, line 99: Can you add a code comment here? I don't really understand the purpose of the `< 20` check.

PR #1985, opened 2022-11-20
Update dependencies

fisker:


sindresorhus:
Can you fix the conflict?

fisker:
I'll be busy with other works from next week, I may won't able to contribute here in near future.

sindresorhus:
> I'll be busy with other works from next week, I may won't able to contribute here in near future.

Ok. Thanks for letting me know and no worries :)

readme.md
```diff
@@ -46,121 +46,123 @@ Use a [preset config](#preset-configs) or configure each rules in `package.json`
 <!-- Do not manually modify this list. Run: `npm run fix:eslint-docs` -->
 <!-- begin auto-generated rules list -->
 
-✅ Enabled in the `recommended` [configuration](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs).\
+💼 [Configurations](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs) enabled in.\
+🚫 [Configurations](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs) disabled in.\
```
yakov116, line 50: @bmish The word recommended is missing here

bmish: That's intentional. The `recommended` config is mentioned on the next line:

💼 [Configurations](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs) enabled in.\
🚫 [Configurations](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs) disabled in.\
✅ Set in the `recommended` [configuration](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs).

We have separate emojis for each config severity present (error, warn, off) and separate emojis for each config present (in this case, only `recommended`). This allows us to represent different configurations ("recommended config disables some rule", "foo config sets a rule to warn" etc).

yakov116: But the wording is very off

bmish: ?

Here are some more examples:

- https://github.com/import-js/eslint-plugin-import#rules
- https://github.com/bmish/eslint-plugin-jest/blob/eslint-doc-generator/README.md#rules

yakov116: I think the wording is off, `Configurations disabled in.` is not a valid English sentence

yakov116: So then it should be "Configuration is disabled" where does the word "in" have to do with this.

I think the word recommended should be restored

bmish: "Configuration is disabled" doesn't seem to make sense to me. Configurations are not disabled, rules are. And we use "Configurations" plural because multiple configurations could be involved. This column is not necessarily specific to the single/recommended config and could involve multiple configs as you can see in the examples I linked to earlier.

yakov116: I am not sure why you are so against adding the word back. I disagree with your logic.

yakov116: @sindresorhus I would like to hear your opinion

sindresorhus: I agree, the wording is off. I think the whole thing is a bit confusing even if it had better naming. Why even have two columns for enabled/disabled? Checkmark if it's enabled, nothing if it's not.

bmish: We need to indicate when a rule is explicitly disabled by a config. That's why we have a separate column for that, but it only shows up when there are configs present that explicitly disable rules.

sindresorhus: Why do you need to indicate that? Is there really a difference between not being included in the preset and being disabled in the preset?

bmish: I think the confusion is happening because eslint-plugin-unicorn sets all non-recommended rules to `off` in its recommended config, possibly just as a TODO reminder to enable them someday, but doesn't intend to impart any significant meaning from this. Example:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/23d03265089acce6138d734776645f5a031f22a0/configs/recommended.js#L76-L77

To solve this in eslint-plugin-unicorn, we could either:

1. Stop setting rules to `off` unless specifically necessary.
2. Or just use `--rule-list-columns` to hide the column indicating which configs rules are disabled in since this information may not be meaningful in the context of eslint-plugin-unicorn configs.

I do want to note that this setup of explicitly setting all non-recommended rules to "off" is unusual and not something I have seen commonly in other plugins. It's fine if you want to do this, but it can be a bit confusing to automated tooling which has to assume that any configuration is meaningful.

As for other plugins, indications of what configs enable, disable, or warn for rules can all be meaningful, important, and necessary for completeness to fully represent a config's behavior.

This might not seem as important for a simple plugin with just a recommended config, but it becomes apparent when a plugin exports multiple configs or uses various severity levels when configuring its rules.

For example, in the [rules list](https://github.com/jsx-eslint/eslint-plugin-react#list-of-supported-rules) for eslint-plugin-react, [react/jsx-uses-react](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-uses-react.md) is enabled in the `recommended` config, but the same rule is then disabled in the `jsx-runtime` config. I have seen a similar pattern for "prettier" configs which disable rules from a recommended config that conflict with prettier.

So the current information architecture went through several iterations to reach this point where it can accurately represent the configurations in both simple and complex plugins.

sindresorhus: I can appreciate how difficult it is to solve all the needs of different plugins.

---

> Or just use --rule-list-columns to hide the column indicating which configs rules are disabled in since this information may not be meaningful in the context of eslint-plugin-unicorn configs.

👍 I would prefer to just hide the column.

---

> For example, in the [rules list](https://github.com/jsx-eslint/eslint-plugin-react#list-of-supported-rules) for eslint-plugin-react, [react/jsx-uses-react](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-uses-react.md) is enabled in the recommended config, but the same rule is then disabled in the jsx-runtime config.

I honestly found that one confusing too. I think if we had multiple presets, I would have preferred simply a single emoji indicating each preset and only one column.

PR #1992, opened 2022-11-27
Add historical note to `no-typeof-undefined`

fregante:
This rule seemingly goes against the "best practice" of always using `typeof`. I'd like to explain that the practice is no longer "best", but I'm not sure if this is the best way to phrase it.


fisker:
I don't know what that mean. I don't think use `typeof` or not related to the strict mode.

fregante:
Before strict mode, `undefined` can be defined and `undefined === 1` was entirely possible. For this reason, direct comparison with `undefined` was frowned upon and is probably still inherently seen as a bad pattern today.

sindresorhus (approved):

fisker:
In strict mode `const undefined = 1` still valid, only `globalThis.undefined` is readonly. Correct me if I'm wrong.

fisker:
I believe the global `undefined` is readonly because, it changed in ES5 https://262.ecma-international.org/5.1/#sec-15.1.1.3 spec.

It has nothing to do with the strict mode.

```console
$ node -p "(function() {'use strict';const undefined = 1; return typeof undefined})()"
number

$ node -p "(function() {const undefined = 1; return typeof undefined})()"
number
```

fregante:
Gotcha, so the rule was related to pre-ES5 browsers, IE9 only partially supported ES5. I'll update the wording

docs/rules/no-typeof-undefined.md
```diff
@@ -9,6 +9,8 @@
 
 Checking if a value is `undefined` by using `typeof value === 'undefined'` is needlessly verbose. It's generally better to compare against `undefined` directly. The only time `typeof` is needed is when a global variable potentially does not exists, in which case, using `globalThis.value === undefined` may be better.
 
+Historical note: Comparing against `undefined` without `typeof` was frowned upon until ES5. This is no longer a problem since all engines currently in use no longer allow reassigning the `undefined` global.
```
fregante, line 12: Is this more correct?

fregante: Indeed broken on IE8 👍 

Screen Shot 10

fregante: Brokener on IE9 😂 

Screen Shot 11

fisker (approved):
Not sure about this, I use `typeof`(since long time ago, maybe before IE6) not because I'm worried about `undefined` reassigned, but worried about error thrown on undefined global variable. But I guess it's fine to add this to the doc. Won't block.

PR #1999, opened 2022-11-28
`prefer-number-properties`: fix MDN links

stephenwade:
This PR updates the "they have slightly different behavior" links in the prefer-number-properties docs to point to the correct anchors.


fisker (approved):
Thanks!

PR #2002, opened 2022-11-30
Add jsx and tsx to file check pattern for no empty file

### LINKED ISSUE #2001: No empty file check for tsx files
Hello, I don't know if a bug report is the right thing here but I wanted to use the no-empty-file rule for my react project.
However is does not check .tsx files.
Is it possible to add these to the files to check?

Rule: no-empty-file

empty files:

test.ts // es lint error (no empty file)
tester.tsx // no error

Felix-kln:
Adding support for jsx and tsx file pattern

Fixes #2001


rules/no-empty-file.js
```diff
@@ -19,7 +19,7 @@ const hasTripeSlashDirectives = comments =>
 const create = context => {
 	const filename = context.getPhysicalFilename().toLowerCase();
 
-	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename)) {
+	if (!/\.(?:js|mjs|cjs|ts|jsx|tsx|mts|cts)$/.test(filename)) {
```
fisker, line 22:
```suggestion
	if (!/\.(?:js|mjs|cjs|jsx|ts|mts|cts|tsx)$/.test(filename)) {
```

Let's make `js/ts*` in this consistent order.

rules/no-empty-file.js
```diff
@@ -17,9 +17,9 @@ const hasTripeSlashDirectives = comments =>
 
 /** @param {import('eslint').Rule.RuleContext} context */
 const create = context => {
-	const filename = context.getPhysicalFilename().toLowerCase();
+	const filename = context.getPhysicalFilename();
 
-	if (!/\.(?:js|mjs|cjs|ts|mts|cts)$/.test(filename)) {
+	if (!/\.(?:js|mjs|cjs|jsx|ts|mts|cts|tsx)$/i.test(filename)) {
```
silverwind, line 22: Could be generalized to `/\.[mc]?[jt]sx?/` which would also support hypothetical `mjsx` and more.

silverwind, line 22:
```suggestion
	if (!/\.(?:[mc]?[jt]sx?)$/i.test(filename)) {
```

Would support stuff like `mjsx` too.

fisker: I prefer current pattern, it's more readable and won't match `mjsx`.

silverwind: Both are equally readable to me, but whatever you prefer. I think `mjsx` is inevitable.

sindresorhus: I prefer the current pattern too.

Issue #1931, opened 2022-10-10
`punycode/` is misflagged as requiring the `node:` prefix


nbouvrette:
```js
import { encode as punycodeEncode } from 'punycode/'
```

Is reporting: 

```
Prefer `node:punycode/` over `punycode/`.eslint[unicorn/prefer-node-protocol](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v44.0.0/docs/rules/prefer-node-protocol.md)
```

When in fact, `node:punycode` is deprecated and `punycode/` is a npm package.

fisker:
https://github.com/sindresorhus/is-builtin-module/issues/7

silverwind:
I think to fix this, the rule could just ignore the import if the specifier ends in `/`, which in itself seems to be some kind of hack around specific to node's module loader (I would not recommend using that, as it's bound to cause issues in other loaders).

silverwind:
The import from `punycode/` does not work for me in node ESM, it only works in CJS, so I suggest to only support it for CJS, if at all. Better alternatives include either `punycode/punycode.js` or the `punycode.js` npm module.

```js
> await import("punycode/")
Uncaught:
Error [ERR_UNSUPPORTED_DIR_IMPORT]:

kachkaev:
Same [here](https://github.com/hashintel/hash/pull/1704) with `import { Buffer } from "buffer/"` being converted to `import { Buffer } from "node:buffer/"`.

fisker:
Fixed by #2060

PR #2016, opened 2022-12-15
`prefer-node-protocol`: Don't suggest `node:` prefix for punycode

silverwind:
The [`punycode` core module](https://nodejs.org/api/punycode.html) is deprecated since Node 7.0.0. There is a [same-name npm module](https://www.npmjs.com/package/punycode) which is the recommended alternative. I think this situation is one-of-a-kind for core modules, so I added a simple condition to ignore it.

Related: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1931

### DIFF
--- a/rules/prefer-node-protocol.js
+++ b/rules/prefer-node-protocol.js
@@ -26,6 +26,7 @@ const create = () => ({
 			typeof value !== 'string'
 			|| value.startsWith('node:')
 			|| !isBuiltinModule(value)
+			|| value === 'punycode'
 		) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


silverwind:
Note, I'm unsure if this fix is correct in terms of which module the user gets as I'm not sure how node would behave in case of such a module name conflict. Behaviour might be different between `require` and `import` too.

All I know is that I have the npm dependency `punycode` installed and the linter is incorrectly suggesting to add the `node:` prefix to it.

silverwind:
Actually there appears to be a non-ambiguous npm package named [`punycode.js`](https://www.npmjs.com/package/punycode.js), I'll be using that so the rule can stay as-is.

Issue #1871, opened 2022-07-28
Failed to load plugin in monorepo


dev-justtrade:
ESLint with plugin unicorn brakes in a NX monorepo when having packages on a deeper folder level.

It works fine in packages like `/libs/first-level` (path `@nx-monorepo/first-level`) but throws this error when used in packages like `/libs/first-level/second-level` (path `@nx-monorepo/first-level/second-level`):
```
Error: Failed to load plugin 'unicorn' declared in '.eslintrc.json » ../../../.eslintrc.json': Invalid name: "@nx-monorepo/first-level/second-level"

Error: Failed to load plugin 'unicorn' declared in '.eslintrc.json » ../../../.eslintrc.json': Invalid name: "@nx-monorepo/first-level/second-level"
    at ensureValidName (/home/xxx/nx-monorepo/node_modules/normalize-package-data/lib/fixer.js:336:15)
    at Object.fixNameField (/home/xxx/nx-monorepo/node_modules/normalize-package-data/lib/fixer.js:215:5)
    at /home/xxx/nx-monorepo/node_modules/normalize-package-data/lib/normalize.js:32:38
    at Array.forEach (<anonymous>)
    at normalize (/home/xxx/nx-monorepo/node_modules/normalize-package-data/lib/normalize.js:31:15)
    at module.exports.sync (/home/xxx/nx-monorepo/node_modules/read-pkg/index.js:37:36)
    at module.exports.sync (/home/xxx/nx-monorepo/node_modules/read-pkg-up/index.js:27:24)
    at Object.<anonymous> (/home/xxx/nx-monorepo/node_modules/eslint-plugin-unicorn/rules/expiring-todo-comments.js:50:33)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
```

I seems to be related to the `expiring-todo-comments` rule.

schmkr:
Also experiencing this issue, in an NX monorepo.

Would it be an idea, to add a configuration option for the location of the `package.json` to use? Similar to how `import/no-extraneous-dependencies` [does this](https://github.com/import-js/eslint-plugin-import/blob/main/src/rules/no-extraneous-dependencies.js)?

schmkr:
[...]

PR #2020, opened 2022-12-22
`expiring-todo-comments`: Do not normalize package.json (#1871)

schmkr:
As per discussion in #1871, this PR disables normalizing the package.json data, so invalid names (or other properties) in the packages.json will not cause EsLint to break when using this plugin.

### DIFF
--- a/rules/expiring-todo-comments.js
+++ b/rules/expiring-todo-comments.js
@@ -47,7 +47,10 @@ const messages = {
 		'Unexpected \'{{matchedTerm}}\' comment without any conditions: \'{{comment}}\'.',
 };
 
-const packageResult = readPkgUp.sync();
+// We don't need to normalize the package.json data, because we are only using 2 properties and those 2 properties
+// aren't validated by the normalization. But when this plugin is used in a monorepo, the name field in the
+// package.json is invalid and would make this plugin throw an error. See also #1871
+const packageResult = readPkgUp.sync({normalize: false});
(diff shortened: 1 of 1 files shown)


fisker (approved):

rules/expiring-todo-comments.js
```diff
@@ -47,7 +47,7 @@ const messages = {
 		'Unexpected \'{{matchedTerm}}\' comment without any conditions: \'{{comment}}\'.',
 };
 
-const packageResult = readPkgUp.sync();
+const packageResult = readPkgUp.sync({normalize: false});
```
fisker, line 50: Can you add a comment to explain why we are doing this here?

schmkr: Ah, yes of course. Will do.

schmkr: Done.

fisker (approved):

schmkr:
@fisker thanks for the approval. I see a [check is failing](https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/3758656596/jobs/6387257436#step:5:89), but not too familiar with this project to identify if this is indeed a result of the changes in this PR. Do you know?

fisker:
> if this is indeed a result of the changes in this PR

Not related.

schmkr:
Thanks for that confirmation.

What is now the next step, what is still needed to get this PR merged?

Issue #2029, opened 2023-02-01
Audit all rules for "Change Array by copy"


fisker:
https://github.com/tc39/proposal-change-array-by-copy#overview

- `Array.prototype.toReversed() -> Array`
- `Array.prototype.toSorted(compareFn) -> Array`
- `Array.prototype.toSpliced(start, deleteCount, ...items) -> Array`
- `Array.prototype.with(index, value) -> Array`

----

- `TypedArray.prototype.toReversed() -> TypedArray`
- `TypedArray.prototype.toSorted(compareFn) -> TypedArray`
- `TypedArray.prototype.with(index, value) -> TypedArray`

----

- [x] `no-useless-spread` #2030
- [x] `prefer-negative-index` #2031
- [x] `prefer-set-has` #2032
- [x] `prefer-spread` #2034

fisker:
Done.

fisker:
In `prefer-spread` we forbid `array.slice()` , `array.slice(0)`, and `array.concat()` to clone array.

> In theory, array.toSpliced()/array.toSpliced(0?, 0?) should be able to used to clone array.


I don't think anyone doing this for now. Let's add it if someone request.

sindresorhus:
> I don't think anyone doing this for now. Let's add it if someone request.

I think we should add it. But can be a new issue.

fisker:
I need to confirm how `toSpliced` works, to confirm just `array.toSpliced()`, or `array.toSpliced(anything)` , `array.toSpliced(anything, 0)` should be included.

No browser support it yet. will test/read spec later.

> But can be a new issue.

We can keep this open.

fisker:
All PR merged. Closing

PR #2030, opened 2023-02-01
`no-useless-spread`: Check `Array#{toReversed,toSorted,toSpliced,with}`

fisker:
Ref: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2029


boris-petrov:
@fisker this PR is great and handling `[...foo.toSorted()]` is awesome, however it would be nice to also handle `[...foo].toSorted()` **and** `[...foo].sort()`, both of which can be replaced with `foo.toSorted()`.

sindresorhus:
@boris-petrov Open a new issue please.

PR #2032, opened 2023-02-01
`prefer-set-has`: Support `Array#{toReversed,toSorted,toSpliced,with}`

fisker:
Ref: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2029


fisker:
Missed `.with`

PR #2065, opened 2023-04-07
Add `prefer-blob-reading-methods` rule

### LINKED ISSUE #1269: Rule proposal: `prefer-blob-reading-methods`
[`FileReader`](https://developer.mozilla.org/en-US/docs/Web/API/FileReader) is kind of old, and it doesn't use promise. A better way to read file should be ["blob reading methods"](https://github.com/w3c/FileAPI/pull/117)

## Fail

```js
const arrayBuffer = await new Promise((resolve, reject) => {
	const fileReader = new FileRead();
	fileReader.onload = () => resolve(fileReader.result)
	fileReader.onerror = () => reject(fileReader.error)
	fileReader.readAsArrayBuffer(blob);
});
```

```js
const arrayBuffer = await new Response(blob).arrayBuffer();
```

## Pass

```js
const arrayBuffer = await blob.arrayBuffer();
```

NotWoods:
Fixes #1269

I opted to allow `new Response(blob).arrayBuffer()` because it's difficult to flag without type-checking (`blob` might be a typed array, stream, etc) and because it has better browser support than the blob methods.


fisker:
> because it's difficult to flag without type-checking (blob might be a typed array, stream, etc) 

make sense 👍

fisker:
Can you fix the linting problem?

fisker (approved):

silverwind:
It won't trigger on `reader.readAsDataURL`, right?

jimmywarting:
On the + side. if you use `` then it could maybe even possible be read lazily on demand from the disc when the image are in viewport.

and it's so much better than reading a hole video as base64 which would be huge.

silverwind:
I see `URL.createObjectURL` does not actually create Data URIs, but `/<uuid>` style pseudo-URLs that are only valid for current browsing session.

So depending on use case, it's not a replacement for `readAsDataURL`, like for example when one needs a portable string that obeys normal JS garbage collection rules. Think UI with a button to copy an image as data uri to clipboard.

Issue #1752, opened 2022-03-19
Include reason to use rule in documentation


fregante:
Very often I find rules that have no explicit reason to exist and one is supposed to just _get it_. This is not Unicorn’s fault as ESLint’s docs are equally lacking (and including text like: "When not to use this rule: Don't use it when you don't need it")

Especially when collaborating with others using the same set of rules in a project, I think this is a crucial piece of information together with:

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1751

The lack of it can frustrate developers and lead to the deactivation of the rule.

sindresorhus:
Do you have any specific Unicorn rules we should prioritize for this?

fregante:
Not sure, it appears that actually many/most rules actually do have such paragraph, like, random picks:

- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-add-event-listener.md
- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-module.md

Maybe this one is an _exception_ (…) more than the _rule_ (😏):

- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-number-properties.md

I think `no-array-reduce` could also have further details since it's controversial:

- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-reduce.md

fregante:
No reason included:

- https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/explicit-length-check.md
- Reason kinda found here: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/12

PR #2070, opened 2023-04-15
Extend explanations for some rules

fregante:
- Closes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1752

### DIFF
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Enforce explicitly checking the length of an object and enforce the comparison style.
+This rule is only meant to enforce a specific style and make comparisons more clear.
 
 This rule is fixable, unless it's [unsafe to fix](#unsafe-to-fix-case).
 
--- a/docs/rules/no-array-callback-reference.md
+++ b/docs/rules/no-array-callback-reference.md
@@ -7,6 +7,8 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
+Passing functions to iterator methods can cause issues when the function is changed without realizing that the iterator passes 2 more parameters to it. **This also applies when using TypeScript,** albeit only if the function accepts the same parameter type used by the iterator method.
+
 Suppose you have a `unicorn` module:
 
 ```js
@@ -35,6 +37,15 @@ const unicorn = require('unicorn');
 
 [1, 2, 3].map(unicorn);
 //=> [2, 3, 5]
+```
+
+This rule helps safely call the function with the expected number of parameters:
+
+```js
+const unicorn = require('unicorn');
+
+[1, 2, 3].map(x => unicorn(x));
+//=> [2, 3, 4]
 ```
 
 ## Fail
--- a/docs/rules/no-array-reduce.md
+++ b/docs/rules/no-array-reduce.md
@@ -9,7 +9,7 @@
 
 It's only somewhat useful in the rare case of summing up numbers, which is allowed by default.
 
-Use `eslint-disable` comment if you really need to use it.
+Use `eslint-disable` comment if you really need to use it or disable the rule entirely if you prefer functional programming.
 
 This rule is not fixable.
 
--- a/docs/rules/no-nested-ternary.md
+++ b/docs/rules/no-nested-ternary.md
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Improved version of the [`no-nested-ternary`](https://eslint.org/docs/latest/rules/no-nested-ternary) ESLint rule, which allows cases where the nested ternary is only one level and wrapped in parens.
+Improved version of the [`no-nested-ternary`](https://eslint.org/docs/latest/rules/no-nested-ternary) ESLint rule. This rule allows cases where the nested ternary is only one level and wrapped in parens.
 
 ## Fail
 
--- a/docs/rules/no-null.md
+++ b/docs/rules/no-null.md
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Disallow the use of the `null` literal, to encourage using `undefined` instead. You can find out more about this in <https://github.com/sindresorhus/meta/discussions/7>
+Disallow the use of the `null` literal, to encourage using `undefined` instead. You can learn why in <https://github.com/sindresorhus/meta/discussions/7>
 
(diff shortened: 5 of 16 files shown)


docs/rules/explicit-length-check.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Enforce explicitly checking the length of an object and enforce the comparison style.
+This rule is only meant to enforce a specific style and make comparisons more clear.
```
fregante, line 10: The previous description was nearly identical to the title

docs/rules/no-array-callback-reference.md
```diff
@@ -7,6 +7,8 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
+Passing functions to iterator methods can cause issues when the function is changed without realizing that the iterator passes 2 more parameters to it. **This also applies when using TypeScript,** albeit only if the function accepts the same parameter type used by the iterator method.
```
fregante, line 10: This rule like `no-array-reduce` is very "controversial", so the clearer the better.

docs/rules/no-array-reduce.md
```diff
@@ -9,7 +9,7 @@
 
 It's only somewhat useful in the rare case of summing up numbers, which is allowed by default.
 
-Use `eslint-disable` comment if you really need to use it.
+Use `eslint-disable` comment if you really need to use it or disable the rule entirely if you prefer functional programming.
```
fregante, line 12: You can't change the mind of functional programmers, so it's best to make it clear this is not for them.

docs/rules/no-nested-ternary.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Improved version of the [`no-nested-ternary`](https://eslint.org/docs/latest/rules/no-nested-ternary) ESLint rule, which allows cases where the nested ternary is only one level and wrapped in parens.
+Improved version of the [`no-nested-ternary`](https://eslint.org/docs/latest/rules/no-nested-ternary) ESLint rule. This rule allows cases where the nested ternary is only one level and wrapped in parens.
```
fregante, line 10: "Which" was ambiguous

docs/rules/no-null.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Disallow the use of the `null` literal, to encourage using `undefined` instead.
+Disallow the use of the `null` literal, to encourage using `undefined` instead. You can find out more about this in: https://github.com/sindresorhus/meta/discussions/7
```
fregante, line 10: Controversial rule. Best to link to the most comprehensive explanation.

docs/rules/no-useless-undefined.md
```diff
@@ -7,6 +7,10 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
+`undefined` is the default value for new variables, parameters, return statements, etc… so specifying it doesn't make any difference. 
+
+The only case where passing `undefined` is required is due to bad TypeScript types in functions, in which case you can use `checkArguments: false` option.
```
fregante, line 12: Is this true? Are there more cases?

docs/rules/no-zero-fractions.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-There is no difference in JavaScript between, for example, `1`, `1.0` and `1.`, so prefer the former for consistency.
+There is no difference in JavaScript between, for example, `1`, `1.0` and `1.`, so prefer the former for consistency and brevity.
```
fregante, line 10: What's a word for "completely useless and redundant"? 😅

docs/rules/prefer-module.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Prefer using the [JavaScript module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) format over the legacy CommonJS module format.
+Prefer using the [JavaScript module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) format over the legacy CommonJS module format. Together with other changes, this helps the ecosystem migrate to a single, native module format.
```
fregante, line 10: This is "controversial" and difficult rule to fix. I think we should add even more than this reasoning without starting CJS vs ESM flames.

docs/rules/prefer-number-properties.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Enforces the use of:
+ECMAScript 2015 moved globals onto the `Number` constructor for consistency and to slightly improve them. This rule enforces their usage to limit the usage of globals:
```
fregante, line 10: I couldn't find why they did this, this is my assumption.

docs/rules/prefer-query-selector.md
```diff
@@ -7,7 +7,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-It's better to use the same method to query DOM elements.
+It's better to use the same method to query DOM elements. This helps keep consistency and it lends itself to future improvements (e.g. more specific selectors).
```
fregante, line 10: Best I could do. The rule is more about style than utility (`getElementsByTagName` looks obsolete).

docs/rules/no-zero-fractions.md
fisker: `1..toString()`is not.useless.

docs/rules/prefer-prototype-methods.md
```diff
@@ -31,6 +31,7 @@ const array = Array.prototype.slice.apply(bar);
 
 ```js
 const hasProperty = Object.prototype.hasOwnProperty.call(foo, 'property');
+// Or better yet Object.hasOwn(foo, 'property')
```
fisker, line 34: We can use `Object.prototype.toString` as example.

docs/rules/no-zero-fractions.md
fregante: The rule doesn't affect that, so why are you mentioning it?

fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/f14433068bbf92d56b3aaa457c5a1aa0ec19e7ab/test/snapshots/no-zero-fractions.mjs.md?plain=1#L2630

PR #2072, opened 2023-04-23
`prefer-top-level-await`: Ignore `.cjs` files

fisker:


rules/prefer-top-level-await.js
```diff
@@ -59,6 +59,12 @@ const isAwaitArgument = node => {
 
 /** @param {import('eslint').Rule.RuleContext} context */
 function create(context) {
+	const filename = context.getFilename();
+
+	if (filename.toLowerCase().endsWith('.cjs')) {
```
sindresorhus, line 64:
```suggestion
	if (context.getFilename().toLowerCase().endsWith('.cjs')) {
```

Since it's not used anywhere else.

PR #2084, opened 2023-05-09
`prefer-dom-node-remove`: Fix incorrect auto-fix

### LINKED ISSUE #2082: `prefer-dom-node-remove`: incorrect autofix with optional chaining
```js
node?.parentNode.removeChild(node);
```

gets incorrectly autofixed to

```js
node.remove();
```

it should be

```js
node?.remove();
```

fisker:
Fixes #2082


fisker:
Left some TODOs #2085

PR #2097, opened 2023-05-14
`prefer-array-some`: ignore filter calls with literal argument

thorn0:


fisker:
There is a function  `isNodeValueNotFunction` https://github.com/sindresorhus/eslint-plugin-unicorn/blob/7895e5d71a1e377820a969dfbf93a864fc0e43bb/rules/utils/index.js#L4

Let's use it.

PR #2111, opened 2023-05-16
Refactor `simpleArraySearchRule`

fisker:
- Use simple `CallExpression` selector.

Since `prefer-array-index-of` listens two different cases, I don't want to merge listeners manually, so I added a `.on` method to `context`. https://github.com/sindresorhus/eslint-plugin-unicorn/commit/ecf02e2dfaa99cd39ba9a1e101e631290a90be28

Feel it can be useful, we can

```js
module.exports = {
	create(context) {
		context.on('CallExpression', node => {
			checkOneCaseWithCallExpression();
		})
		context.on('CallExpression', node => {
			checkAnotherCaseWithCallExpression();
		})
	}
}
```

instead of 

```js
module.exports = {
	create(context) {
		return {
			CallExpression(node) {
				checkOneCaseWithCallExpression();
				checkAnotherCaseWithCallExpression();
			}
		}
	}
}
```

Existing rule 

https://github.com/fisker/eslint-plugin-unicorn/blob/c3a485ddb9521b66961c38709a31b9af92042b4f/rules/prefer-date-now.js#L35-L72

Can be separate into two listeners, we don't have to put two total different logic in one function

### DIFF
--- a/rules/prefer-array-index-of.js
+++ b/rules/prefer-array-index-of.js
@@ -13,10 +13,10 @@ const lastIndexOfOverFindLastIndexRule = simpleArraySearchRule({
 
 /** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
-	create: context => ({
-		...indexOfOverFindIndexRule.createListeners(context),
-		...lastIndexOfOverFindLastIndexRule.createListeners(context),
-	}),
+	create(context) {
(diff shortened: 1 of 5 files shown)


sindresorhus:
Good idea

sindresorhus:
Can you fix the conflict?

PR #2116, opened 2023-05-16
Refactor to use simple selectors

fisker:


rules/prefer-dom-node-dataset.js
```diff
@@ -1,34 +1,40 @@
...
+				isMethodCall({
+					method: 'setAttribute',
+					argumentsLength: 2,
+					optionalCall: false,
+					optionalMember: false,
+				})
+				|| isMethodCall({
+					methods: ['getAttribute', 'removeAttribute', 'hasAttribute'],
+					argumentsLength: 1
+					optionalCall: false,
```
yakov116, line 27: Just curious, why not just make this the default?

fisker: What if we want ignore `optionalCall` check

```js
isMethodCall({
	optionalCall: undefined
})
```

? Odd to me.

fisker: Maybe introduce another flag?

```js
isMethodCall(node, {
	checkOptionalCall: false
})
```

Issue #2156, opened 2023-06-20
unicorn/prefer-dom-node-dataset - Undetected exception if optional chaining


gcunsolo:
Hi Unicorn, first things first: kudos for your priceless work, big fan of yours 🦄 

Got a problem tho with `unicorn/prefer-dom-node-dataset` and optional chaining

The following code should raise an exception, and yet it doesn't:
```js
document.querySelector('body')?.getAttribute('data-some-dataset');
```
Removing the optional chaining "?." raise an error as expected ;)

Thanks in advance for your time, wish you joy and rainbows

sindresorhus:
If anyone wants to work on this, see the initial attempt in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2158.

PR #2158, opened 2023-06-22
`prefer-dom-node-dataset` - Add support to optional chaining

gcunsolo:
999d3df1089b83b5cc187be263cbf1a84ca08c24 failing tests on #2156 
81705555530a87b9a1c807ef737d79d3349ce728 fix
...
3da90d7493b200126e5bde667fedec0f9a7f9020 snapshots update


test/snapshots/prefer-dom-node-dataset.mjs.md
```diff
@@ -269,6 +269,22 @@ Generated by [AVA](https://avajs.dev).
         | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Prefer \`.dataset\` over \`setAttribute(…)\`.␊
     `
 
+## Invalid #17
+      1 | element?.setAttribute("data-optional-chaining", "foo");
+
+> Output
+
+    `␊
+      1 | element.dataset.optionalChaining = "foo";␊
```
fisker, line 278: We can't fix to this, it will break code.

rules/prefer-dom-node-dataset.js
```diff
@@ -27,6 +27,18 @@ const create = context => ({
...
+					method: 'setAttribute',
+					argumentsLength: 2,
+					optionalCall: false,
+					optionalMember: true,
+				})
+				|| isMethodCall(node, {
+					methods: ['getAttribute', 'removeAttribute', 'hasAttribute'],
+					argumentsLength: 1,
+					optionalCall: false,
+					optionalMember: true,
```
fisker, line 40: We can just remove `optionalMember: false,` above.

gcunsolo: u're right!
`done();`

test/snapshots/prefer-dom-node-dataset.mjs.md
gcunsolo: Hello Fisk, thanks!
How exactly can I help?

fisker: We should either remove the fix for optional members or add a safe fix.

gcunsolo: I'm sorry @fisker, it took me a while to understand what was wrong, I think I got it now:

We should expect the autofix feature to keep the optional chaining, is that it?

Like 

 ```suggestion
       1 | element?.dataset.optionalChaining = "foo";␊
```
Did I get it right?

Thanks again for your review, cheers

fisker: I don't think so, optional chaining can't be LHS.

fisker: 1. `element?.setAttribute('data-foo', 'bar')`

Since `foo?.dataset.foo = 'bar'` is invalid, and we don't want auto fix to wrap it in a `if`, so we should just remove the autofix.

2. `element?.getAttribute('data-foo')`

Should be safe to fix to `element?.dataset.foo`

3. `element?.removeAttribute('data-foo')`

Should be safe to fix to `delete element?.dataset.foo`

4. `element?.hasAttribute('data-foo')`

Since `Object.hasOwn(element?.dataset, 'foo')` can cause runtime error, so I guess we should remove the autofix too.

fisker: I'm adding `hasOptionalChainElement` utility in #2169, you can use it after we merge.

sindresorhus:
@gcunsolo Bump :)

sindresorhus:
Bump

sindresorhus:
Closing for lack of activity.

PR #2169, opened 2023-07-10
`prefer-dom-node-dataset`: Remove broken fix for `element.setAttribute`

fisker:
Previously, we fix

```js
optional?.element.setAttribute("data-unicorn", "🦄");
```

to

```js
optional?.element.dataset.unicorn = "🦄";
```

This is invalid code.


rules/prefer-dom-node-dataset.js
```diff
@@ -10,50 +10,29 @@ const messages = {
...
-			return;
-		}
-
-		const method = node.callee.property.name;
-		const name = dashToCamelCase(attributeName.slice(5));
+	// `foo?.bar = ''` is invalid
+	// TODO: Remove this restriction if https://github.com/nicolo-ribaudo/ecma262/pull/4 get merged
+	if (method === 'setAttribute' && hasOptionalChainElement(callExpression.callee)) {
+		return;
+	}
```
fisker, line 20: This is the actual change.

PR #2173, opened 2023-07-11
Remove use of deprecated methods

fisker:
Ref https://github.com/eslint/eslint/issues/16999


scrungrth:
This seems like a breaking change, some of these params were only added recently, requiring a new higher minimum eslint version.

fisker:
Already required working eslint verison https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1b6757e80cdb7a120d90f2df5fef087a8b0f1cc0/package.json#L96, released as major https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v48.0.0 what else do you except?

PR #2180, opened 2023-07-23
Fix directive check in no-empty-file for TypeScript parser

### LINKED ISSUE #2175: no-empty-file always false positives in @typescript-eslint/parser@6
Right now, `no-empty-file` assumes that if an AST node includes a `directive` property, then that's enough to indicate it's a directive:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/ab7732e1de0f431adc7316595dd5636781d33806/rules/no-empty-file.js#L9

That assumption no longer holds true in `@typescript-eslint/parser`'s v6.0.0 due to https://github.com/typescript-eslint/typescript-eslint/pull/6274. We changed the AST shape roughly like:

```diff
- directive?: string;
+ directive: string | undefined;
```

https://github.com/typescript-eslint/typescript-eslint/blob/df131e258c93e5714c88c0373cfeb2e1e75afc75/packages/ast-spec/src/statement/ExpressionStatement/spec.ts#L8

The PR explains why we did that (tl;dr: normalizing AST node shapes). So `eslint-plugin-unicorn` will need to update with a change like:

```diff
- const isDirective = node => node.type === 'ExpressionStatement' && 'directive' in node;
+ const isDirective = node => node.type === 'ExpressionStatement' && node.directive;
```

Right now, `no-empty-line` is flagging totally valid files such as https://github.com/jakebailey/pyright-action/pull/19/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1. Thanks @jakebailey for reporting!

mcous:
Fixes #2175

To ensure the fix is valid, I updated the `@typescript-eslint/parser` dev dependency and added a few simple `test.typescript` cases to `no-empty-file` tests. Please let me know if there's any additional testing I can / should add to this PR!


rules/no-empty-file.js
```diff
@@ -6,7 +6,7 @@ const messages = {
 	[MESSAGE_ID]: 'Empty files are not allowed.',
 };
 
-const isDirective = node => node.type === 'ExpressionStatement' && 'directive' in node;
+const isDirective = node => node.type === 'ExpressionStatement' && node.directive !== undefined;
```
fisker, line 9:
```suggestion
const isDirective = node => node.type === 'ExpressionStatement' && typeof node.directive === 'string';
```

Just in case parser change it to `null` in future.

test/no-empty-file.mjs
```diff
@@ -77,3 +77,13 @@ test.snapshot({
...
 	],
 });
+
+// Test for https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2175
+test.typescript({
+	valid: [
+		{code: '(() => {})();', filename: 'example.ts'},
+	],
+	invalid: [
+		{code: '"";', filename: 'example.ts', errors: 1},
```
fisker, line 87:
```suggestion
		{code: '"";', filename: 'example.ts', errors: 1},
		{code: '"use strict";', filename: 'example.ts', errors: 1},
```

fisker (approved):

PR #2182, opened 2023-08-03
add unopinionated preset

### LINKED ISSUE #896: Add `unopinionated` preset
I propose turning off the following rules by default, so that the `recommended` config could act as a baseline which may be extended explicitly:

- `unicorn/filename-case` – This goes against the principle set by one of the most popular linting packages, [`eslint-config-airbnb`](https://www.npmjs.com/package/eslint-config-airbnb). See this excerpt from their style guide:

  > [23.6](https://github.com/airbnb/javascript#naming--filename-matches-export) A base filename should exactly match the name of its default export.

- `unicorn/no-nested-ternary` – Doesn't work properly when [used along with Prettier](https://github.com/prettier/eslint-config-prettier), as formatting removes parentheses used in ternary expressions.

- `unicorn/no-null` – This is very opinionated and external libraries like React and Prisma depend on `null`. Also, `something == null` checks for both `null` and `undefined` at once, so it's less prone to errors in comparison to `something === undefined || something === null`.

- `unicorn/no-useless-undefined` – Conflicts with ESLint's built-in [`consistent-return`](https://eslint.org/docs/rules/consistent-return) rule, which is [enforced by Airbnb's config](https://github.com/airbnb/javascript/blob/5b462a04cdfebba712a3df18926edfcb4103d2a9/packages/eslint-config-airbnb-base/rules/best-practices.js#L22-L23). Causes all of the following code to be invalid:

  ```js
  function fetchModel() {
    if (exists) return { something: "some data" };
    return undefined; // As embraced by `unicorn/no-null`
  }

  function fetchModel() {
    if (exists) return { something: "some data" };
    return; // Violates `consistent-return`, even when omitted 
  }
  ```

- `unicorn/prefer-query-selector` – The alternatives, namely `getElementById` and `getElementsByClassName`, offer better performance. Selector-based queries should be avoided in general, even in CSS, as they get complex pretty fast. Using IDs and class names for locating elements is a well-tested practice, while complex selectors are questionable.

- `unicorn/prevent-abbreviations` – While the intent is great, way too many false positives may occur. For instance, React uses the term `props` to identify attributes passed to a JSX element. They aren't widely called `properties` in the documentation and tutorials. Same goes for the `req` and `res` parameters of an API handler function, as popularized [by Express](https://expressjs.com/en/starter/hello-world.html).

As an alternative, a `recommended-loose` or `recommended-unopinionated` config may also be added to avoid the issues outline above. My goal is to set up ESLint with a strict set of rules as simple as below:

```json
{
  "root": true,
  "parserOptions": { "project": "./tsconfig.json" },
  "extends": [
    "airbnb-typescript",
    "airbnb/hooks",
    "plugin:@typescript-eslint/recommended",
    "plugin:unicorn/recommended",
    "prettier",
    "prettier/@typescript-eslint",
    "prettier/react",
    "prettier/unicorn"
  ]
}
```

RicardoValdovinos:
Hello,

I'd like to work on and resolve #896

Currently, I have disabled four rules

```
no-array-for-each
no-array-reduce
no-null
prefer-switch
```

I was also considering disabling 

```
switch-case-braces
empty-brace-spaces
```

as these are more related to style and should probably be handled by a formatter.

Lastly, as referenced in #1636, I'd like to suggest modifying `prevent-abbreviations` default setting to allow standard abbreviations such as `arg`, `args`, `db`, and `env`. There may be others but these are the immediate ones that came to mind.

### DIFF
--- a/configs/unopinionated.js
+++ b/configs/unopinionated.js
@@ -1,0 +1,15 @@
+'use strict';
+const {rules, ...baseConfigs} = require('./recommended.js');
+
+module.exports = {
+	...baseConfigs,
+	rules: {
+		...Object.fromEntries(Object.entries(rules)
+		.filter(([ruleId, severity]) => severity === 'error')
+		),
+		'unicorn/no-array-for-each': 'off',
+		'unicorn/no-array-reduce': 'off',
+		'unicorn/no-null': 'off',
+		'unicorn/prefer-switch': 'off'
+	},
+};
--- a/configs/unopinionated.js
+++ b/configs/unopinionated.js
@@ -4,9 +4,7 @@ const {rules, ...baseConfigs} = require('./recommended.js');
 module.exports = {
 	...baseConfigs,
 	rules: {
-		...Object.fromEntries(Object.entries(rules)
-			.filter(([_ruleId, severity]) => severity === 'error'),
-		),
+		...Object.fromEntries(Object.entries(rules)),
 		'unicorn/no-array-for-each': 'off',
 		'unicorn/no-array-reduce': 'off',
 		'unicorn/no-null': 'off',
--- a/configs/unopinionated.js
+++ b/configs/unopinionated.js
@@ -5,11 +5,11 @@ module.exports = {
 	...baseConfigs,
 	rules: {
 		...Object.fromEntries(Object.entries(rules)
-		.filter(([ruleId, severity]) => severity === 'error')
+			.filter(([_ruleId, severity]) => severity === 'error'),
 		),
 		'unicorn/no-array-for-each': 'off',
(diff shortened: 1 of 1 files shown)


devunt:
An other option is to make `prettier` preset separately for rules that conflicting with what prettier do.
Following rules can be disabled with `prettier` preset for now.

```
empty-brace-spaces
number-literal-case
```

Zamiell:
@devunt Isn't that already covered by `eslint-config-prettier`? I don't think it needs to be in a separate config.

sindresorhus:
See: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/896#issuecomment-1958814311

Closing for now.

PR #2183, opened 2023-08-04
`no-array-for-each`: Add more reasoning to the docs

Zamiell:
n/a

### DIFF
--- a/docs/rules/no-array-for-each.md
+++ b/docs/rules/no-array-for-each.md
@@ -13,6 +13,8 @@ Benefits of [`for…of` statement](https://developer.mozilla.org/en-US/docs/Web/
 - Better readability
 - Ability to exit early with `break` or `return`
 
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):

Zamiell:
how long do PRs usually sit after being approved but before being merged?

fisker (approved):

cmdcolin:
random quip:  it would be cool to provide "examples" of what this text means e.g. a little code snippet showing the improved type narrowing, etc. as it's a little advanced (i'm just an onlooker of this repo)

Zamiell:
The type narrowing difference is illustrated by the following code:

```ts
declare const foo: Foo | undefined;

if (foo === undefined) {
  return;
}

for (const element of array) {
  // type of `foo` is `Foo`
}
```

```ts
declare const foo: Foo | undefined;

if (foo === undefined) {
  return;
}

array.forEach((element) => {
  // type of `foo` is `Foo | undefined`
});
```

The variable mutation difference is illustrated by the following code:

```ts
function countArray(array: number[]) {
  let count = 0;

  for (const element of array) {
    count += element;
  }

  return count;
}
```

```ts
function countArray(array: number[]) {
  let count = 0;

  array.forEach((element) => {
    count += element;
  });

  return count; // lint error: count is never modified
}
```

cmdcolin:
excellent, thanks for the demo!

PR #2195, opened 2023-09-22
Support RegExp `v` flag

fisker:


fisker:
Blocked by #2196

fisker:
@novemberborn Sorry for bothering you. 

As you can see test is failing in this PR. We use a old version of ava https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1629ebeb9d967f23245546aa2e0c2c182611ebb4/package.json#L71, we use it weird, we delete snapshots, and force ava to regenerate snapshots, to make sure snapshots won't change. https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1629ebeb9d967f23245546aa2e0c2c182611ebb4/.github/workflows/main.yml#L40-L45

It works well until this PR. I already regenerate snapshots locally, but it still changes on CI. Maybe you known what possibly go wrong?

novemberborn:
Looks like a test failed, it didn’t get to the diffing of the snapshots. And only in Node 20. Wouldn’t that be a real bug?

> we use it weird, we delete snapshots, and force ava to regenerate snapshots, to make sure snapshots won't change

That sounds like what AVA checks for in CI, what am I missing? Is this to handle differences between Node versions?

fisker:
> Looks like a test failed, it didn’t get to the diffing of the snapshots. 

You are right.. I thought that's the "lint-test" job, which does the snapshot diff check, but it's not, really sorry.

> Wouldn’t that be a real bug?

Run test on Node.js v20 locally, tests all passed.
Run out of ideas...

fisker:
Should be something wrong here https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1629ebeb9d967f23245546aa2e0c2c182611ebb4/rules/prefer-regexp-test.js#L79-L92

I'll try to fix it later.

fisker:
@novemberborn You are right, there is a bug indeed, thanks for your help.

Issue #2208, opened 2023-10-27
`prefer-module` should not report `module` name in types


ST-DDT:
The `prefer-module` rule incorrectly marks the `module` name as forbidden in custom types/interfaces.

````txt
  18:4   error  Do not use "module"     unicorn/prefer-module
````

```js
// This should be allowed
export interface ModuleRegistry {
  [module: string]: Entry;
}
```

fisker:
Would you like to fix it?

Add `TSIndexSignature` here

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/61e66017f0c582c8a2537ca23b2e858e1fad8908/rules/ast/is-reference-identifier.js#L4

should fix that.

ST-DDT:
Created a PR to fix that: 

- https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2209

PR #2209, opened 2023-10-27
`prefer-module`: allow `module` as TSIndexSignature names or TSTypeAliasDeclaration ids

ST-DDT:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2208

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2208

I tried adding a test for that, but was unable to do that (maybe because it requires ts support):

````txt
  message: `␊
    > 1 | type ModuleRegistry = { [module: string]: string };␊
        |      ^ Parsing error: Unexpected token ModuleRegistry`,                                                                                                            
  }                                                                                                                                                                          

  » |      ^ Parsing error: Unexpected token ModuleRegistry
  » verify (file:///test/utils/snapshot-rule-tester.mjs:138:9)
  » file:///test/utils/snapshot-rule-tester.mjs:169:23
````

So I used https://github.com/faker-js/faker/pull/2510/files#diff-7170563e0d86e93e8997b4acb856e5e494809a0ef41b9f4386f26e14bc961908R19 to test that.

### DIFF
--- a/rules/ast/is-reference-identifier.js
+++ b/rules/ast/is-reference-identifier.js
@@ -120,6 +120,11 @@ function isNotReference(node) {
 			return parent.parameters.includes(node);
 		}
 
+		// `type Identifier = Foo`
+		case 'TSTypeAliasDeclaration': {
+			return parent.id === node;
+		}
+
 		case 'TSPropertySignature': {
 			return parent.key === node;
 		}
--- a/rules/ast/is-reference-identifier.js
+++ b/rules/ast/is-reference-identifier.js
@@ -115,6 +115,7 @@ function isNotReference(node) {
 			return parent.id === node;
 		}
 
+		// `type Foo = { [Identifier: string]: string }`
 		case 'TSIndexSignature': {
 			return parent.parameters.includes(node);
 		}
--- a/rules/ast/is-reference-identifier.js
+++ b/rules/ast/is-reference-identifier.js
@@ -115,6 +115,10 @@ function isNotReference(node) {
 			return parent.id === node;
 		}
 
+		case 'TSIndexSignature': {
+			return parent.parameters.includes(node);
+		}
+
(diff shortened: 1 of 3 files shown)


fisker:
Take this as example https://github.com/sindresorhus/eslint-plugin-unicorn/blob/61e66017f0c582c8a2537ca23b2e858e1fad8908/test/no-array-for-each.mjs#L571

Put your code in `valid` part.

ST-DDT:
> Take this as example

Thanks for the help. 
Is one test sufficient or do we need more?

fisker (approved):

fisker:
Added more, but feel free to add more.

Thanks!

sindresorhus:
Tests are failing

ST-DDT:
fisker added an additional case where the type name is module.
The additional (failing) test, is also an (expectedly) valid and similar case, but it is different from the issue this PR originally tried to fix.

I'm not sure how I can fix that. Is there a way to run only a specific test? Maybe then I can try to debug it.

fisker:
I'll take a look.

fisker (approved):

ST-DDT:
It looks like we missed a case:

````ts
type Data = {
  [module in 'foo']: number;
};
````

e.g.

````ts
type Data = {
  [module in keyof string]: number;
};
````

fisker:
> I need to know how I can run a specific test(case) only.

Currently we don't support it yet. But it should be possible if we upgrade `eslint-ava-rule-tester` https://github.com/jfmengels/eslint-ava-rule-tester/releases/tag/v4.1.0

ST-DDT:
I used a workaround, aka deleting all tests except for the one I wanted and created a PR with the fix:

- https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2213

PR #2219, opened 2023-11-12
fix: typo "hasSuggestion" in no-for-loop rule, should be plural form instead

virtuallyunknown:
As defined in here:

https://github.com/eslint/eslint/blob/6d470d2e74535761bd56dcb1c021b463ef9e8a9c/lib/shared/types.js#L153


sindresorhus:
Linting is failing.

virtuallyunknown:
Yep, sorry about that. The entire commit was literally one single character, but it seems it triggered some other errors. I think it should be fine all now, I will come check back later 👍

PR #2221, opened 2023-11-16
Additional cases to consider handling in your rule

gechdejenb:
1. Multiple Update Expressions: Ensure that there is only one update expression in the for loop, and it is an increment by one. If there are multiple update expressions or the update is not by one, consider flagging it.

2.Non-Integer Step in For-Of Replacement:
Ensure that the step in the for loop is always 1. If it's not, consider flagging it.

3.Array Length Check:
Check if there's a condition checking the array's length in the for loop. If the loop should run until the end of the array, you may not need such a check.

4.Declaration Type Check:
Ensure that the variable declared in the loop is a let if it's being modified within the loop. If it's not being modified, consider recommending const.

5.Array Modification Check:
Check if there are any array modifications within the loop. If the array is modified, it might affect the behavior of a for-of loop.

6.Function Call within Loop Header:
Check if there's a function call within the loop header. If there is, consider flagging it, as it might have side effects that affect the loop behavior.

7.Use of break or continue Statements:
Check if there's the use of break or continue statements within the loop body. If there is, it might complicate the translation to a for-of loop.

### DIFF
--- a/rules/no-for-loop.js
+++ b/rules/no-for-loop.js
@@ -406,6 +406,81 @@ const create = context => {
 					}
 				};
 			}
+			// 1. Multiple Update Expressions
+            if (node.update && node.update.type === 'BinaryExpression' && node.update.operator !== '+') {
+                const start = node.update.range[0];
+                const end = node.update.range[1];
+                return {
+                    loc: toLocation([start, end], sourceCode),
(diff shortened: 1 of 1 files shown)


sindresorhus:
We don't accept ChatGPT generated contributions. You are also missing tests.

PR #2228, opened 2023-12-10
Add regex global flag to replaceAll in docs

Logicer16:
I've realised a mistake in one of the passing examples. When using `String#replaceAll()` with a `RegExp`, the regex global flag (`g`) must be used, as was mentioned in the [issue](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1948#issuecomment-1321301671) which originally prompted this passing case to be added to the documentation. Without the global flag, a `TypeError` is thrown at runtime.


fisker (approved):
Thanks!

PR #2232, opened 2023-12-19
`no-useless-undefined`: Add `checkArrowFunctionBody` option

### LINKED ISSUE #2191: Rule change proposal: `no-useless-undefined` add option to skip checking arrow function body
Sometimes it's more clear to use `() => undefined` than `() => {}`,

https://github.com/babel/babel/pull/15547/files#diff-27084e8878ce0b72bf5cd31ec6a0a719096f981bcfa32ca71a57a0f35fb84676R64

```js
const {foo = () => {}} = {};
const {foo = () => undefined} = {};
```

```js
const {
  foo = () => {},
  bar = () => ({})
} = {};
const {
  foo = () => undefined,
  bar = () => ({})
} = {};
```

Suggest add option `checkArrowFunctionBody: false`.

fisker:
Fixes #2191

### DIFF
--- a/docs/rules/no-useless-undefined.md
+++ b/docs/rules/no-useless-undefined.md
@@ -9,7 +9,9 @@
 
 `undefined` is the default value for new variables, parameters, return statements, etc… so specifying it doesn't make any difference.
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
CI is failing

PR #2233, opened 2023-12-19
Update dependencies

fisker:


fisker:
@novemberborn The snapshot files are different on Windows and Linux, do you know what causes?

novemberborn:
@fisker I'm not aware of any differences between the OSes. It'll be different with Node 21 though because they changed the gzip library.

Would need the raw bytes to compare.

fisker:
You are right, it about the Node.js version.
As I tested, I found that the result of  `zlib.gzipSync` is different between Node.js v20.9.0 and v20.10.0 on Windows.

PR #2235, opened 2023-12-19
Update ava to v6

fisker:
This seems never going to get fixed, https://github.com/avajs/ava/discussions/2928

I guess we'll have to live with it.


fisker:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2233#issuecomment-1862177908 Still an issue.

PR #2236, opened 2023-12-19
Add `test.only` helper to test one case

fisker:
Can be used as

```js
test({
  valid: [
    test.only`code`,
    test.only('code'),
    test.only({code: 'code'}),
    {code: 'code', only: true},
  ],
})
```


scripts/internal-rules/no-test-only.js
```diff
@@ -0,0 +1,74 @@
...
+				}
+
+				context.report(problem);
+			},
+		};
+	},
+	meta: {
+		fixable: 'code',
+		messages: {
+			[messageId]: '`test.only` cannot be used.',
```
sindresorhus, line 71: Not immediately clear from the message why it cannot be used?

fisker, line 71:
```suggestion
			[messageId]: '`test.only` should only be used for debugging purposes, please remove it before commit.',
```

Better?

sindresorhus:
@fisker Do you plan any more changes this week? If not, I'm going to do a new release.

fisker:
Unfortunately, I have very limited time for open source projects this year. Don't have plan for other changes now, you can release if you want.

PR #2237, opened 2023-12-19
`prefer-negative-index`: Check `TypedArray#subarray()`

fisker:


fisker:
I didn't use TypedArray much, is `.byteLength` always equal to `.length`?

sindresorhus:
Only for Uint8Array

PR #2242, opened 2023-12-21
Fix `filename-case`

### LINKED ISSUE #2240: In v50.0.0 , lodash dependency is missing
`const {camelCase} = require('lodash');` is used in https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v50.0.0/rules/no-unnecessary-polyfills.js

But, `"lodash"` isn't available in https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v50.0.0/package.json

So, when we consume this package (in another ESLint config), it is causing an issue when we try to load the latest version:
`Failed to load plugin 'eslint-plugin-unicorn' declared in '.eslintrc.js ...`

fisker:
Fixes #2240

### DIFF
--- a/rules/no-unnecessary-polyfills.js
+++ b/rules/no-unnecessary-polyfills.js
@@ -2,7 +2,7 @@
 const path = require('node:path');
 const readPkgUp = require('read-pkg-up');
 const coreJsCompat = require('core-js-compat');
-const {camelCase} = require('lodash');
+const {camelCase} = require('./utils/lodash.js');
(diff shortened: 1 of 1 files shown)


fisker:
I'll try to release a version, just test that if I can do it 😄 . @sindresorhus

fisker:
It works https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v50.0.1

PR #2247, opened 2024-01-02
`prefer-prototype-methods`: Fix argument of `isMethodCall`

haocheng6:
This PR fixes the option used by the `isMethodCall` function in the `prefer-prototype-methods` rule.


fisker (approved):
Thank you!

PR #2249, opened 2024-01-04
Use `FlatESLint` in package test

fisker:


test/package.mjs
```diff
@@ -98,68 +97,6 @@ test('validate configuration', async t => {
...
-		};
-		const runEslint = async options => {
-			const [{messages}] = await new ESLint({...baseOptions, ...options}).lintText(text);
-			return messages;
-		};
-
-		const [{message}] = await runEslint();
-		t.is(message, 'Parsing error: The keyword \'import\' is reserved');
-		t.deepEqual(await runEslint({baseConfig: eslintPluginUnicorn.configs.recommended}), []);
-	}
```
fisker, line 162: `FlatESLint` uses the latest syntax by default now, I guess we should just remove these.

Issue #1230, opened 2021-04-28
Rule proposal: `prefer-destructured-variables`


fisker:
I understand we already have [`consistent-destructuring`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-destructuring.md), and it already covers cases I'm going to post, but that rule is buggy, and I don't like destructuring all used properties in one, I prefer destructuring when needed, for example:

```js
const {a} = foo;

if (a) {
  // ... do something not related to `foo.b`
}

const {b} = foo;

// ... do something with `b`
```

## Fail

```js
const {a} = foo;
console.log(foo.a); 
```

## Pass

```js
const {a} = foo;
console.log(a, foo.b); 
```

brettz9:
This looks like a good solution for your use case and a good compromise solution for those of us who do want to require destructuring where possible and practical, but not tyrannically so (as long as it is implemented to meet [my case](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1005#issuecomment-758325847) as it sounds it would be and [GuoSirius'](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1005#issuecomment-758353250) (since that is technically a variable but just not yet available)). In any case, some false negatives are not concealing technically bad code.

sindresorhus:
I agree, I thought `consistent-restructuring` would be a good idea, but in practice, it's just been annoying. I think we should also turn `consistent-destructuring` off by default.

sindresorhus:
This is now accepted.

sindresorhus:
This is now covered by the simplified `consistent-destructuring` behavior. The rule only reports when the same direct property already has a corresponding destructured variable, and it does not expand existing destructuring patterns.

The reassignment caveat mentioned above is also covered: reads after writes like `foo.a = …`, `foo.a++`, `delete foo.a`, destructuring assignment writes, and loop writes are intentionally ignored.

Closing as completed by #3115.

PR #2260, opened 2024-01-17
Remove `consistent-destructuring` from recommend preset config

vossmalte:
See https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1230#issuecomment-828366222
!1230


package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "eslint-plugin-unicorn",
-	"version": "50.0.1",
+	"version": "51.0.0",
```
sindresorhus, line 3: You should never bump a version in a pull request. It's up to the maintainers to decide when and what version to use.

Issue #2266, opened 2024-01-29
Consider remove autofixing unicorn/prevent-abbreviations


jpolo:
When using prevent-abbreviations with VSCode fix on save, some variable is automagically renamed.
This could look at first sight but when the code was not written by yourself this is problematic...

The autofixer also takes a risk of breaking business logic "just because of formatting issue", which IMHO does not worth the tradeoff.

unicorn/prevent-abbreviations

I did some digging in github and it seems that many people (including me) are disabling this rule so it indicates this rule has some usage issue at the moment...

sindresorhus:
> When using prevent-abbreviations with VSCode fix on save, some variable is automagically renamed.
This could look at first sight but when the code was not written by yourself this is problematic...

The problem here is reformatting other people's code with your own rules, not this rule in particular. You should only have "format on save" enabled for your own projects.

sindresorhus:
I find the rule very useful, and it would be a chore without the auto-fixing, so we are unlikely to disable that.

It also does not make sense for every rule to have a setting to turn off auto-fixing, but you could maybe try to get ESLint to add a way to disable auto-fixing for certain rules.

jpolo:
Sorry my explanation was confusing. I meant by "reformatting the code other people wrote", that you are in VSCode, you save => it is possible that a code that you don't see (maybe scrolling down) is fixed by the rule.
 
Like you, I am absolutely convinced that this rule is very important :), that is why I am giving the feedback : I wish I could turn it on again.

I know it would be a chore without auto-fixing, but I found that this rule can break so many things, that it is still a chore to validate the modifications made...
Reformulated : either you have few changes and it is not so hard to type refactor the changes -OR- there are many changes and the potential breakage are so important that it takes a lot of time.
In the mean time, the "refactor variable" feature  in VSCode is much safer for renaming purpose.

fregante:
I'm convinced autofix-on-save is not a safe behavior, so I'm 👎 on this basis. 

While _formatting_ on save is fine, "fixing bad code" isn't, because my code is at various stages of _bad_ until I commit it. It doesn't make sense to attempt to fix incomplete code because lint rules (and their autofixes) assume they're not working on WIP code.

I get red underlines while I write code, but the editor should not automatically alter my code while I'm still working on it (unless again it's just formatting)

jpolo:
You might be right on the "fix on-save" behavior and I am opened to change my dev practices.

[...]

Issue #2181, opened 2023-07-28
Failed to warn empty file


adoin:
![image](https://github.com/sindresorhus/eslint-plugin-unicorn/assets/20517172/ce112c1f-0d8d-474e-a66a-9bdea47fa949)
But 
![image](https://github.com/sindresorhus/eslint-plugin-unicorn/assets/20517172/b0cc4f62-ec70-4f07-9131-16979ced001c)

And commit lint didn't stop commit through eslint.

fregante:
Support is explicitly limited to JS/TS files:


https://github.com/sindresorhus/eslint-plugin-unicorn/blob/61912993131b359d3c5789a2d1b2567b144e2f56/rules/no-empty-file.js#L22-L24

Igorkowalski94:
As an alternative you can use https://github.com/Igorkowalski94/eslint-plugin-project-structure has the same capabilities + force only folders/files, has a built-in case for name, regex support, all extensions support, ability to inherit the name from the parent (the child inherits the name of the folder in which it is located), folder recursion (You can nest a given folder structure recursively).

Issue #2085, opened 2023-05-09
Improve `prefer-dom-node-remove`


fisker:
- [ ] Use snapshot test
- [ ] Improve fix, not use the original text but manipulate tokens
- [ ] Improve fix for `node?.parentNode.removeChild(node)` (node can be anything, the object and the argument should be same reference, but can be different AST, eg: `a?.b?.parentNode.remove(a.b)`)

PR #2279, opened 2024-02-10
Add issue template for a rule change

fregante:
It looks like most open issues are to request changes or options to existing rules, so I thought I'd create a template.


.github/ISSUE_TEMPLATE/rule_change.yml
```diff
@@ -0,0 +1,56 @@
+name: Rule change
+description: As rule could be improved or extended?
+title: "`rule-name` should do xyz"
+labels:
+  - enhacement
```
fregante, line 5:
```suggestion
  - enhancement
```

Alternatively, it could use a `change request` label like we have in Refined GitHub

.github/ISSUE_TEMPLATE/rule_change.yml
```diff
@@ -0,0 +1,56 @@
...
+      value: |
+        Before opening a new issue:
+        - Look for existing [open or closed rule proposals](https://github.com/sindresorhus/eslint-plugin-unicorn/issues?q=is:issue+label%25enhancement
+        - [Report a bug](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/new?assignees=&labels=&projects=&template=bug_report.md) instead if it's a false positive
+  - type: textarea
+    validations:
+      required: true
+    attributes:
+      label: Description
+      description: Explain here what change or improvement you'd like.
```
sindresorhus, line 19:
```suggestion
      description: Explain the improvement you would like to see.
```

.github/ISSUE_TEMPLATE/rule_change.yml
```diff
@@ -0,0 +1,56 @@
...
+      value: |
+        ```js
+        const replace = 'me';
+        ```
+
+        ```js
+        function foo() {
+          return 'me';
+        }
+        ```
```
sindresorhus, line 52: Is this really necessary for this type of issue?

sindresorhus (changes requested):

.github/ISSUE_TEMPLATE/rule_change.yml
fregante: The "good code" isn't needed if the suggestion is:

- _"Also detect this other way to do the same thing"_
	-  like https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2181

but it's definitely needed if it's:

- _"Also autofix this other similar but unrelated method"_
	- Like https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2085

.github/ISSUE_TEMPLATE/rule_change.yml
```diff
@@ -0,0 +1,56 @@
+name: Rule change
+description: As rule could be improved or extended?
```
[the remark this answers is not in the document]
fregante: "A rule"

This follows the "Bug report" format

Screenshot 3

.github/ISSUE_TEMPLATE/rule_change.yml
fregante: `improvement` is a synonym of `enhancement`

The choice would be between "change" and "improve". Technically every change is an improvement, but in this case I'm seeing "change" as `I don't agree with this rule, please alter it`, like

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2266

sindresorhus: Let's go with "change request" then. Doesn't really matter that much.

fregante:
```suggestion
  - change request
```

sindresorhus:
You also need to resolve the other inline feedback.

.github/ISSUE_TEMPLATE/rule_change.yml
fregante, line 2:
```suggestion
description: A rule could be improved or extended?
```

.github/ISSUE_TEMPLATE/rule_change.yml
```diff
@@ -0,0 +1,56 @@
+name: Rule change
+description: A rule could be improved or extended?
+title: "`rule-name` should do xyz"
```
fisker, line 3:
```suggestion
title: "`rule-name`: "
```

PR #2290, opened 2024-02-24
Do not ignore logical expressions when checking functions

fisker:


fisker:
Thought it's used in other rules. will remove it in #2289

Issue #2294, opened 2024-02-28
`unicorn/no-array-callback-reference` does not report when optional chaining


czizzy:
this rule is not reported when optional chaining

`unicorn/no-array-callback-reference`

```js
function isString(arg) {
  return typeof arg === 'string';
}

const obj = { a: [], b: [] };

obj.a.filter(isString);
obj.b?.filter(isString);
```

only the first filter report an error.

fisker:
Remove this line should fix this

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/231529aee793251f30416b93dc3d5f9bfa27ee47/rules/no-array-callback-reference.js#L238


PR welcome.

PR #2296, opened 2024-02-29
`no-array-callback-reference`: check optional memeber 

czizzy:
fix https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2294


fisker:
Please add tests to verify your change https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/no-array-callback-reference.mjs

sindresorhus:
@czizzy Bump

PR #2297, opened 2024-03-14
`import-style`: Support `node:` protocol

### LINKED ISSUE #1200: Support `node:` module name prefix in the `import-style` rule
It doesn't handle `import util from 'node:util';`.

https://nodejs.org/api/esm.html#esm_node_imports

zanminkian:
When enabling `unicorn/import-style` and `unicorn/prefer-node-protocol` at the same time, `unicorn/import-style` will not work for `path` and `util` modules.

```js
{
  rules: {
    'unicorn/import-style': 'error',
    'unicorn/prefer-node-protocol': 'error',
  },
}
```

```js
import {join} from 'path';
import * as util from 'util';

// After running `eslint . --fix`, codes will be auto-fixed like this without errors.
import {join} from 'node:path';
import * as util from 'node:util';

```

Closes #1200.


GeoffreyBooth:
I think the docs for `import-style` need to be updated with the example changed from `"path": { "named": true }` to `"node:path": { "named": true }`

I had this code:

```js
import { join } from 'node:path'
```

And I wanted to keep it as written, but couldn’t until I added the `node:` prefix to the override:

```js
"unicorn/import-style": ["error", {"styles": {"node:path": {"named": true}}}],
```

Issue #2298, opened 2024-03-18
`prefer-spread` should not report on optional chaining


JounQin:
```js
scopeManager?.scopes.slice()
```

fisker:
Note: Add test for `(scopeManager?.scopes).slice()` when work on it.

MichaelBlm:
Can I attempt to fix this issue?

PR #2304, opened 2024-04-01
fix: [2298] `prefer-spread`: should not report on optional chaining

MichaelBlm:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2298


MichaelBlm:
main branch fails `npm run integration -- --group 3 `

fisker:
> @sindresorhus main branch fails `npm run integration -- --group 3 `

I'll fix that.

test/prefer-spread.mjs
```diff
@@ -312,6 +312,8 @@ test.snapshot({
 		'buffer.slice()',
 		'file.slice()',
 		'class A {foo() {this.slice()}}',
+		'scopeManager?.scopes.slice()',
+		'(scopeManager?.scopes).slice()',
```
fisker, line 316: I think we should report `(scopeManager?.scopes).slice()` since it's not an optional call.

MichaelBlm: Ah it wasnt clear in the issue if the (scopeManager?.scopes).slice() was valid or invalid. Let me update the PR.

Issue #2292, opened 2024-02-25
`prevent-abbreviations` - non-ASCII characters ignored in filenames


Kristinita:
### 1. Summary

If I use [**eslint-plugin-html**](https://github.com/BenoitZugmeyer/eslint-plugin-html) and check JavaScript inside HTML by using Unicorn, I get errors like this:

```text
8:3  error  The filename `Мiръ.html` should be named `Мindexръ.html`. A more descriptive name will do too  unicorn/prevent-abbreviations

✖ 1 problem (1 error, 0 warnings)
```

#### 2.1. Files

`.eslintrc.yml`:

```yaml
extends:
- plugin:unicorn/all

plugins:
- html
- unicorn

```

`Мiръ.html` ([**“мiръ” meaning “world”**](https://russianlife.com/the-russia-file/war-and-peace-7-fun-facts/)):

```html
<!DOCTYPE html>
<html lang="en">
<head>
	<meta charset="UTF-8">
	<meta name="viewport" content="width=device-width, initial-scale=1.0">
	<title>Document</title>
	<script>
		console.log("Kira Goddess!");
	</script>
</head>
<body>
	<p>Kira Goddess!</p>
</body>
</html>

```

#### 2.2. Steps to reproduce

```text
npm install --save-dev eslint eslint-plugin-html eslint-plugin-unicorn

npx eslint Мiръ.html
```

##### 2.3.1. Desired

No errors.

##### 2.3.2. Current

```text
8:3  error  The filename `Мiръ.html` should be named `Мindexръ.html`. A more descriptive name will do too  unicorn/prevent-abbreviations

✖ 1 problem (1 error, 0 warnings)
```

### 3. Possible implementations of the desired behavior

1. Make it possible to ignore HTML files via the `ignore` option or by introducing a new option such as `checkFilenamesExtensions`. For example:

	```yaml
	unicorn/prevent-abbreviations:
	- error
	- checkFilenamesExtensions:
	  - ".js"
	```
	
	In this case, Unicorn will solely check names of files with the `.js` extension, but not with other extensions like `.html` or `.md` (if the user is using [**eslint-plugin-markdown**](https://github.com/eslint/eslint-plugin-markdown)).

1. Make it possible to skip filenames like `Мiръ` containing non-ASCII symbols. For example, via the option `checkNonASCIIFilenames: false`.

### 4. Not helped

[...]

PR #2308, opened 2024-04-04
`prevent-abbreviations` - non-ASCII characters ignored in filenames

MichaelBlm:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2292

### DIFF
--- a/rules/prevent-abbreviations.js
+++ b/rules/prevent-abbreviations.js
@@ -116,7 +116,7 @@ const getNameReplacements = (name, options, limit = 3) => {
 	}
 
 	// Split words
-	const words = name.split(/(?=[^a-z])|(?<=[^A-Za-z])/).filter(Boolean);
+	const words = name.split(/(?=\P{Lowercase_Letter})|(?<=\P{Letter})/u).filter(Boolean);
 
 	let hasReplacements = false;
 	const combinations = words.map(word => {
--- a/test/prevent-abbreviations.mjs
+++ b/test/prevent-abbreviations.mjs
@@ -1931,6 +1931,10 @@ test({
 			code: 'foo();',
 			filename: 'err/http-error.js',
 		},
+		{
+			code: 'foo();',
+			filename: 'Мiръ.html',
+		},
 		// `ignore` option
 		{
 			code: outdent`


MichaelBlm:
Here is an explanation of the updated regular expression:

1. `(?=\P{Ll})`: This is a positive lookahead assertion. It matches a position where the next character is not a Unicode lowercase letter. `\P{Ll}` is a Unicode character class that negates the `Ll` (lowercase letter) class.

2. `(?<=\P{L})`: This is a positive lookbehind assertion. It matches a position where the previous character is not a Unicode letter (uppercase or lowercase). `\P{L}` negates the `L` (letter) Unicode character class.

fisker:
The solution looks tricky to me, I prefer just ignore if non-ascii characters included.

rules/prevent-abbreviations.js
```diff
@@ -116,7 +116,7 @@ const getNameReplacements = (name, options, limit = 3) => {
 	}
 
 	// Split words
-	const words = name.split(/(?=[^a-z])|(?<=[^A-Za-z])/).filter(Boolean);
+	const words = name.split(/(?=\P{Ll})|(?<=\P{L})/u).filter(Boolean);
```
sindresorhus, line 119: Use the verbose names (`Lowercase_Letter`), for readability.

sindresorhus:
@fisker Tricky how? The solution looks fine to me.

rules/prevent-abbreviations.js
MichaelBlm: Updated to include the verbose names

MichaelBlm:
My explanation of the regex wasn't that clear. \P{Letter} is equivalent to [^a-zA-Z] but it includes non-ASCII characters

fisker:
> @fisker Tricky how? The solution looks fine to me.

In different languages, the latin letters can be used with other characters together to be a word, so the non-letter characters not a word boundary.

For example, in Chinese, "B超" "T恤" are words. Similar in Japanese, I think.

rules/prevent-abbreviations.js
fisker: We should make a rule for it 😄

fisker: https://github.com/ota-meshi/eslint-plugin-regexp/issues/720

MichaelBlm: I'll work on it actually

MichaelBlm:
> > @fisker Tricky how? The solution looks fine to me.
> 
> In different languages, the latin letters can be used with other characters together to be a word, so the non-letter characters not a word boundary.
> 
> For example, in Chinese, "B超" "T恤" are words. Similar in Japanese, I think.

Can you please provide me a valid/invalid test case to see if this fix doesn't already account for that?

rules/prevent-abbreviations.js
fisker: Note: the proposal is on another repo.

fisker:
I don't really have a case now.

fisker (approved):

fregante:
@fisker you're saying that `龖.js` will be counted as "one-letter long", but in Chinese it's a fully-formed "word", not an abbreviation. Hence "ignore if non-ascii characters included"

fisker:
I don't think Chinese characters are "letter"s, but I can be wrong.

fregante:
That's right, it's why quoted "one-letter word". The rule counts characters as "letters" but not all writing systems have letters. 

https://en.wikipedia.org/wiki/Writing_system#Basic_terminology

fregante:
Now that I think about it, we don't need a regex to split words: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter

MichaelBlm:
I'll explore a solution using the segmenter. Looks promising

MichaelBlm:
I added this code to the valid tests and it is passing with the current implementation? Is that the correct behavior?
`		{
			code: 'foo();',
			filename: '龖.js',
		},`

PR #2315, opened 2024-04-08
`prefer-number-properties`: Add `checkNaN` option

gurgunday:
I think the `NaN` check should also be toggleable, because I use it like I use `null` (absence of an object), to represent an absence of a number; and since `null` is global, this setting lets me be more consistent

Thanks for maintaining this amazing plugin!


gurgunday:
CI error not related I believe

sindresorhus:
// @Richienb

fisker:
Should the option be `checkNaN` or `checkNan` ?

Richienb:
Could we maybe set it to false by default? (breaking)
I also think that `NaN` is better than `Number.NaN`.

sindresorhus:
> Could we maybe set it to false by default? (breaking)

👍

docs/rules/prefer-number-properties.md
```diff
@@ -93,6 +93,13 @@ Default: `false`
 
 Pass `checkInfinity: true` to enable check on `Infinity`.
 
+### checkNaN
+
+Type: `boolean`\
+Default: `true`
+
+Pass `checkNaN: false` to disable check on `NaN`.
```
Richienb, line 101:
```suggestion
Default: `false`

Pass `checkNaN: true` to enable check on `NaN`.
```

Richienb (changes requested):

docs/rules/prefer-number-properties.md
gurgunday: Yes, there are also a number of tests that need to be modified (ran with checkNaN: true), will do them

docs/rules/prefer-number-properties.md
```diff
@@ -93,6 +93,13 @@ Default: `false`
...
 Pass `checkInfinity: true` to enable check on `Infinity`.
 
+### checkNaN
+
+Type: `boolean`\
+Default: `false`
+
+Pass `checkNaN: true` to enable check on `NaN`.
+
 #### Fail
```
silverwind, line 103: Your docs change is in the wrong place, the `Fail` section below is for `checkInfinity`, so you should at it below and add your own `Fail` examples.

gurgunday:
Yeah, sorry y'all — when I turn it off by default, a bunch of tests are failing and even though I modify them to be like Infinity checks, there are errors

I do not have the time to look into it, so let's focus on adding this feature in this PR

See f544b63a4221e51fd5dc30fbf8e1c2027bfb502e to view my latest try

Richienb (changes requested):
I would like some examples in the docs

Richienb:
> I do not have the time to look into it, so let's focus on adding this feature in this PR

That's ok, we can change it in the next PR.

Added to the tracking issue

gurgunday:
I added one more, what else would you like to see?

docs/rules/prefer-number-properties.md
```diff
@@ -126,3 +126,39 @@ const isPositiveZero = value => value === 0 && 1 / value === Number.POSITIVE_INF
...
+const foo = NaN;
+```
+
+```js
+// eslint unicorn/prefer-number-properties: ["error", {"checkNaN": false}]
+const foo = -NaN;
+```
+
+```js
+// eslint unicorn/prefer-number-properties: ["error", {"checkNaN": true}]
```
Richienb, line 162: Add cases that specifically show that `Number.NaN` and `-Number.NaN` pass
I don't think this example is necessary.

Richienb (changes requested):

docs/rules/prefer-number-properties.md
gurgunday: Oh sorry; I was confused because Inifinity doesn't have examples like that either

Richienb (approved):

sindresorhus (approved):

PR #2316, opened 2024-04-09
Fix codecov config

fisker:


fisker:
@sindresorhus Can you add `CODECOV_TOKEN`?

sindresorhus:
> Can you add CODECOV_TOKEN?

Done

fisker:
There is a bug in v4, https://github.com/codecov/feedback/issues/265
I'll see if v3 can work with token.

PR #2335, opened 2024-05-07
Add `no-magic-array-flat-depth` rule

### LINKED ISSUE #1920: Rule proposal: `no-magic-array-flat-depth`
### Description

A special case of [`no-magic-numbers`](https://eslint.org/docs/latest/rules/no-magic-numbers) rule.

When calling `Array#flat(depth)` with depth. Normally the depth should be `1`/`Number.POSITIVE_INFINITY`. Otherwise it should be reported, but I don't want this rule to get annoying, so

- If it's not a number literal, we assume user know the max depth.
- If there is a comment before the number, we assume the number is explained. 

### Fail

```js
array.flat(2);
```

```js
array.flat(99);
```

### Pass

```js
array.flat();
```

```js
array.flat(Number.POSITIVE_INFINITY);
```

```js
array.flat(depth);
```

```js
array.flat(/* The depth is always 2 */ 2);
```

### Additional Info

This rule should only be useful for users don't want to enable [`no-magic-numbers`](https://eslint.org/docs/latest/rules/no-magic-numbers), otherwise the failed cases shold already covered.

fisker:
Fixes #1920


docs/rules/no-magic-array-flat-depth.md
```diff
@@ -0,0 +1,40 @@
+# Disallow magic number as the `depth` argument in `Array#flat(…).`
```
sindresorhus, line 1:
```suggestion
# Disallow a magic number as the `depth` argument in `Array#flat(…).`
```

rules/no-magic-array-flat-depth.js
```diff
@@ -0,0 +1,51 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Disallow magic number as the `depth` argument in `Array#flat(…).`',
```
sindresorhus, line 47:
```suggestion
			description: 'Disallow a magic number as the `depth` argument in `Array#flat(…).`',
```

sindresorhus (changes requested):

PR #2336, opened 2024-05-07
Update `eslint-plugin-eslint-plugin` to v6

fisker:
Fixes new rule `require-meta-docs-recommended` https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/require-meta-docs-recommended.md


sindresorhus:
Could we generate the `recommended` preset based on the property? So we don't have two sources of truth.

fisker:
> Could we generate the `recommended` preset based on the property? So we don't have two sources of truth.

Sure, let's do it in another PR.

PR #2339, opened 2024-05-08
Add `prefer-string-raw` rule

### LINKED ISSUE #2048: Rule proposal: `prefer-string-raw`
### Description

To avoid escape `\`.

### Fail

```js
console.log('"\\" found.');
```

```js
const directory = "C:\\windows\\style\\path";
```

```js
new RegExp('foo\\.bar')
```

### Pass

```js
console.log(String.raw`"\" found.`);
```

```js
const directory = String.raw`C:\windows\style\path`;
```

```js
new RegExp(String.raw`foo\.bar`)
```

```js
const foo = '\\ `';
// Use `String.raw` will need escape `
const foo = '\\ ${foo}';
// Use `String.raw` will need escape $ or `{`
```

```js
// Places can't use a tagged template

"A directive\\"

import foo from "./foo\\bar.js";

const foo = {
	'\\': 'a non-computed key'
}
```

---

This rule can also check when building a tagged function like [this](https://github.com/fregante/code-tag/commit/b20054ce644edb85f17a93ec364371bc3943455f), but it's hard to detect such pattern.

fisker:
Fixes #2048


fisker:
Blocked by #2342

fisker:
Once this gets reviewed, we can revert https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2339/commits/9bc7b933cbb1a34be79f4163cca1e75bb271b68e before merge.

docs/rules/prefer-string-raw.md
```diff
@@ -0,0 +1,22 @@
+# Prefer using `String.raw` tag to avoid escaping `\`
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+<!-- Remove this comment, add more detailed description. -->
```
sindresorhus, line 10: Should link to String.raw on MDN here.

docs/rules/prefer-string-raw.md
```diff
@@ -0,0 +1,22 @@
+# Prefer using `String.raw` tag to avoid escaping `\`
```
sindresorhus, line 1:
```suggestion
# Prefer using the `String.raw` tag to avoid escaping `\`
```

docs/rules/prefer-string-raw.md
```diff
@@ -0,0 +1,22 @@
...
+
+```js
+const foo = 'unicorn';
+```
+
+## Pass
+
+```js
+const foo = '🦄';
+```
```
sindresorhus, line 22: ^

sindresorhus (changes requested):

rules/prefer-string-raw.js
```diff
@@ -0,0 +1,92 @@
+'use strict';
+const {isStringLiteral, isDirective} = require('./ast/index.js');
+const {fixSpaceAroundKeyword} = require('./fix/index.js');
+
+const MESSAGE_ID = 'prefer-string-raw';
+const messages = {
+	[MESSAGE_ID]: '`String.raw` should be used to avoid escaping `\\`.',
+};
+
+const BACKSLASH = '\\';
```
sindresorhus, line 10: Shouldn't these use `String.raw`?

fisker: ``String.raw`\` `` is an error and ``String.raw`\\` `` products two backslash, so it's impossible to create a a single backslash with `String.raw`.

We can't do ``String.raw`foo\` `` either. That's why we ignore string ends with backslash  https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2339/files#diff-def49face997bfe9729f748b6720c2847611a4ad3b6a9714a96f0ef408259d75R65

fisker: The one in `messages` can't use `String.raw` either, because it contains `` ` ``, `` String.raw`foo`bar` `` this is invalid code. Ignored in code too https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2339/files#diff-def49face997bfe9729f748b6720c2847611a4ad3b6a9714a96f0ef408259d75R57

sindresorhus: Ah ok.

sindresorhus:
This looks good to me 👍

CI needs to be fixed though.

PR #2349, opened 2024-05-09
Add `consistent-empty-array-spread` rule

### LINKED ISSUE #2074: Rule proposal: `consistent-empty-array-spread`
### Description

When spreading a ternary in array, we can use both

```js
[
	element1,
	...(a ? [b, c] : []),
]
```

and

```js
[
	element1,
	...(a ? [b, c] : ''),
]
```

but it's bad to have different style in one project, should enforce consistent style.

Personally, I slightly prefer to use empty array when the consequent is array, but empty string should be cheaper. I hope this rule default to enforce use the same type for "consequent" and "alternate".

### Fail

```js
[
	element1,
	...(a ? [b, c] : ''), // use `[]` since "consequent" is array
]
```

```js
[
	element1,
	...(a ? 'bc' : []), // should use `""` since "consequent" is string
]
```

### Pass

```js
[
	element1,
	...(a ? [b, c] : []), // use `[]` since "consequent" is array
]
```

```js
[
	element1,
	...(a ? 'bc' : ''), // should use `""` since "consequent" is string
]
```

fisker:
Fixes #2074


rules/consistent-empty-array-spread.js
```diff
@@ -0,0 +1,126 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Prefer consistent type when spreading a ternary in array literal.',
```
sindresorhus, line 119:
```suggestion
			description: 'Prefer consistent types when spreading a ternary in an array literal.',
```

sindresorhus (changes requested):

PR #2353, opened 2024-05-11
Add `no-negation-in-equality-check` rule

### LINKED ISSUE #1421: Rule proposal: Forbid `!foo === bar`
I'm not sure what name it should be, this is a common mistake, espeically when doing quick debugging. Personally I've done this several times. 

We have this syntax forbidden in Prettier codebase, with [`no-restricted-syntax` rule](https://github.com/prettier/prettier/blob/c2599d47145f7b4bd310f66edd69d9ca14b2c60d/.eslintrc.js#L38-L39), because I made [a mistake](https://github.com/prettier/prettier/commit/c7e354e38cfd051d3f64b56aeda35e9a0a669de6#r47946398) once there.

## Fail

```js
if (!foo === bar);
```

```js
if (!foo !== bar);
```

```js
// Don't have to be in `if`
const isEqual = !foo === bar;
```

## Pass

```js
if (!(foo === bar));
```

```js
// I guess we should allow this, it's type-casting
if (!!foo === bar);
```

fisker:
Fixes #1421


francoismassart:
Hi @fisker, 

Great plugin 👍  
I tried to enable this rule (`no-negation-in-equality-check`) and I'm getting this error:
`Definition for rule 'unicorn/no-negation-in-equality-check' was not found  unicorn/no-negation-in-equality-check`

When I take a look inside the `node_modules/eslint-plugin-unicorn/` I do not see the rule file:
image

The installed version is `53.0.0`

fisker:
It's released in v54 https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v54.0.0

PR #2355, opened 2024-05-11
Remove useless "recommended" properties

DEVTomatoCake:
This removes the `recommended` property from rules with no apparent use, this just seems like an oversight from a Find and Replace to add the property after all `description` properties, however without ensuring that only those within the `docs` object are affected.

`recommended` itself is not used within these files, and it doesn't seem like they're used externally.


fisker (approved):

fisker:
Thanks. It's on me. I used a regular expression to search `description`, and add `recommended` in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2336.

Issue #2278, opened 2024-02-10
Switch to flat config


fregante:
I'm opening this tracking issue to collect related tickets.

This change blocks:

- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2151
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2163

silverwind:
Does it have to be a "switch"? There are a number of plugins that support both configs at the same time and I think it would ease the migration to flat config.

fregante:
I think the flat config is already supported:

- https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1886

So this issue is to _drop support_ for the old config, essentially. It's nothing something that will happen anytime soon, I suppose.

philsherry:
[ESLint v8.x end of life was October 5, 2024](https://eslint.org/blog/2024/09/eslint-v8-eol-version-support/), so [Flat config](https://eslint.org/docs/latest/use/configure/configuration-files) is the only way forward now.

sindresorhus:
@fisker What do you think about doing this now? (Asking for your opinion, not for you to do the work)

fisker:
I'm fine with it, but should we wait for XO to get ready?

sindresorhus:
No, I don't want this blocked on XO.

sindresorhus:
I'm working on this now.

silverwind:
> [Flat config](https://eslint.org/docs/latest/use/configure/configuration-files) is the only way forward now.

FWIW, eslint v9 still supports the deprecated config format. It will be removed only in v10.

sindresorhus:
CI is failing because snapshot test binaries differ. I suspect it's a difference since I'm on macOS and tests run on Linux. Same Node.js version though. Anyone on Linux and Windows and can try updating the snapshots with Node.js 23.6.0 locally?

fregante:
Codespace might be an option

PR #2371, opened 2024-05-27
Upgrade read-pkg-up

erezarnon:
Currently we depend through it on semver@5 which has a security vulnerability: CVE-2022-25883


sindresorhus:
We cannot update this until we move to ESM, which requires moving completely to ESLint flat config.

https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2278

Issue #2324, opened 2024-05-01
missing TypeScript declaration file


Zamiell:
Now that ESLint 9 is released and everyone is migrating to flat config, importing the plugin into a flat config by following the official documentation actually results in a red squiggly line, because no declaration file exists (when using `checkJs` and/or `@ts-check`).

I suspect many/most users will be using TypeScript to validate that their new flat configs do not have typos in them. Thus, shipping a declaration file for this plugin would really be appreciated!

TheElegantCoding:
same here 
![image](https://github.com/sindresorhus/eslint-plugin-unicorn/assets/70770072/a44525b9-4266-409f-b525-fc3a15423da0)

fisker:
Types already available https://github.com/eslint-types/define-config-plugin-types/tree/main/types/unicorn

Maybe @Shinigami92 can help us move to this repo?

silverwind:
Here's a workaround, but it's not much use because it does not validate `rules`. In fact I'm not sure how `rules` validation could even be made to work in typescript.

```ts
declare module "eslint-plugin-unicorn" {
  import type {Eslint} from "eslint";
  const plugin: Eslint.Plugin;
  export = plugin;
}
```

PR #2381, opened 2024-06-15
Improve type inference for configs in index.js

luizffgv:
Currently, using a unicorn config in a type-checked eslint.config.js gives me this:

image

This happens because `createConfig`'s return type has `name` inferred as `boolean` and `plugins` as the union of `typeof {unicorn}` and `typeof ['unicorn']`.

I separated `createConfig` into two versions so TypeScript will infer more specific types for each configuration, fixing the above error.

### DIFF
--- a/index.js
+++ b/index.js
@@ -47,7 +47,7 @@ const allRules = Object.fromEntries(
 	]),
 );
 
-const createLegacyConfig = (rules) => ({
+const createLegacyConfig = rules => ({
 	...legacyConfigBase,
 	plugins: ['unicorn'],
 	rules: {...externalRules, ...rules},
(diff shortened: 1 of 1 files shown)


sindresorhus:
It would be better to do https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2324 to get proper types.

luizffgv:
> It would be better to do #2324 to get proper types.

Definitely, that's much better. I think my PR can still be a good change along the declaration file because the inferred types in index.js will be compatible with the declared types, and my change removes the red squiggly lines from index.js when passing a string to `flatConfigName`.

PR #2382, opened 2024-06-15
Add TypeScript declaration file

### LINKED ISSUE #2324: missing TypeScript declaration file

hyoban:
fixes #2324 
closes #2381


sindresorhus (approved):

dsvgl:
When is this going to be released? 🙏🏻

PR #2384, opened 2024-06-16
prefer-node-protocol: ignore bun modules

### LINKED ISSUE #2383: Bun modules still incorrectly flagged as requiring the `node:` prefix when running ESLint on Bun
everything is the same as it was described in #2226, but happens only when ESLint is run on Bun: `bun --bun eslint .`

happens because bun adds its own modules to the node's [builtinModules](https://nodejs.org/api/module.html#modulebuiltinmodules), which is used by the `builtin-modules` v3 package, which is used by the `is-builtin-module` v3 package, which is used by this plugin

SunsetTechuila:
fixes #2383

### DIFF
--- a/rules/prefer-node-protocol.js
+++ b/rules/prefer-node-protocol.js
@@ -32,7 +32,7 @@ const create = () => ({
 		if (
 			typeof value !== 'string'
 			|| value.startsWith('node:')
-			|| value.match(/^bun(:\w+|$)/)
+			|| /^bun(:\w+|$)/.test(value)
 			|| !isBuiltinModule(value)
 		) {
 			return;
--- a/rules/prefer-node-protocol.js
+++ b/rules/prefer-node-protocol.js
@@ -32,7 +32,7 @@ const create = () => ({
 		if (
 			typeof value !== 'string'
 			|| value.startsWith('node:')
-			|| /^bun(:\w+|$)/.test(value)
+			|| /^bun(?::|$)/.test(value)
 			|| !isBuiltinModule(value)
 		) {
 			return;
--- a/rules/prefer-node-protocol.js
+++ b/rules/prefer-node-protocol.js
@@ -32,8 +32,8 @@ const create = () => ({
 		if (
 			typeof value !== 'string'
 			|| value.startsWith('node:')
+			|| value.match(/^bun(:\w+|$)/)
 			|| !isBuiltinModule(value)
-			|| !isBuiltinModule(`node:${value}`)
 		) {
 			return;
 		}


sindresorhus:
This is already how the latest version of the existing package works. It requires ESM, so we would need to either inline it or wait for https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2278. I'm not interested in switching to `is-core-module`.

SunsetTechuila:
> wait for https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2278

it may take a long time

> I'm not interested in switching to is-core-module.

then, will it be okay if i do this for now?

```js
|| value.match(/^bun(:|$)/)
```

sindresorhus:
> will it be okay if i do this for now?

👍

SunsetTechuila:
@sindresorhus done

rules/prefer-node-protocol.js
```diff
@@ -32,8 +32,8 @@ const create = () => ({
 		if (
 			typeof value !== 'string'
 			|| value.startsWith('node:')
+			|| /^bun(:\w+|$)/.test(value)
 			|| !isBuiltinModule(value)
-			|| !isBuiltinModule(`node:${value}`)
```
fisker, line 36: Why are we removing this line?

SunsetTechuila: no longer needed, as it was only added to detect bun's modules

fisker: I don't understand what's the problem then, are you saying both `bun:foo` and `node:bun:foo` passes `isBuiltinModule()` check?

SunsetTechuila: `isBuiltinModule(bun:foo)` returns `true` when eslint is run on bun - https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2383#issue-2355755215

SunsetTechuila: <details>

<summary>Screenshot</summary>

image

</details>

SunsetTechuila: https://github.com/sindresorhus/is-builtin-module/blob/2b22980d890cd656419106e2d2cca5696453bd7f/index.js#L12-L14

fisker (approved):

rules/prefer-node-protocol.js
```diff
@@ -32,8 +32,8 @@ const create = () => ({
 		if (
 			typeof value !== 'string'
 			|| value.startsWith('node:')
+			|| /^bun(:\w+|$)/.test(value)
```
fisker, line 35:
```suggestion
			|| /^bun(?::|$)/.test(value)
```

PR #2390, opened 2024-06-24
Replace deprecated read-pkg-up with read-package-up

benmccann:
https://www.npmjs.com/package/read-pkg-up says it's been deprecated


benmccann:
oops. looks like this isn't as simple as just switching it as three tests are failing now. I'll close this

PR #2391, opened 2024-06-25
`prevent-abbreviations`:  `db` is also the abbreviation for `decibel`

yukulele:
db is also the abbreviation for decibel


sindresorhus:
The abbreviation for decibel is `dB`. Not sure if `dB` is also caught be the `db` rule.

sindresorhus:
This will need a test to make sure it's caught by the `db` rule, and if not, it needs to be under `dB`.

PR #2392, opened 2024-06-26
docs: `require('fs')` should fail `prefer-node-protocol`

benmccann:
I was a bit confused by the docs and am not sure they're correct. @fisker do you think you accidentally forgot to mark this example as invalid as part of https://github.com/sindresorhus/eslint-plugin-unicorn/pull/1827?

### DIFF
--- a/docs/rules/prefer-node-protocol.md
+++ b/docs/rules/prefer-node-protocol.md
@@ -30,6 +30,10 @@ import fs from 'fs/promises';
 ```
 
 ```js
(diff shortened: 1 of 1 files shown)


fisker (approved):
Thank you, it's in `Pass` section because we ignores `require()` at first. https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1198#issuecomment-825470155

benmccann:
Thanks for taking the time to confirm!

PR #2398, opened 2024-07-05
prefer-query-selector: Add getElementsByName 

### LINKED ISSUE #2374: `prefer-query-selector`: Add `getElementsByName`
### Description

[`getElementsByName`](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByName) is a lesser known DOM query method, it should be included in the list of flagged methods for [`prefer-query-selector`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-query-selector.md).

### Fail

```js
getElementsByName('foo');
```

### Pass

```js
querySelectorAll(`[name="foo"]`);
```

axetroy:
fixed #2374


rules/prefer-query-selector.js
```diff
@@ -11,17 +11,37 @@ const disallowedIdentifierNames = new Map([
 	['getElementById', 'querySelector'],
 	['getElementsByClassName', 'querySelectorAll'],
 	['getElementsByTagName', 'querySelectorAll'],
+	['getElementsByName', 'querySelectorAll'],
 ]);
 
 const getReplacementForId = value => `#${value}`;
 const getReplacementForClass = value => value.match(/\S+/g).map(className => `.${className}`).join('');
+const getReplacementForName = (value, originQuote) => value.match(/\S+/g).map(name => `[name=${wrapQuoted(name, originQuote)}]`).join('');
```
fisker, line 19:
```suggestion
const getReplacementForName = (value, originalQuote) => `[name=${wrapQuoted(value, originalQuote)}]`;
```

I don't think `getElementsByName` accepts mutiple names,

rules/prefer-query-selector.js
```diff
@@ -11,17 +11,37 @@ const disallowedIdentifierNames = new Map([
...
+			return `'${value}'`;
+		}
+
+		case '`': {
+			return `'${value}'`;
+		}
+
+		// No default
+	}
+};
```
fisker, line 43: I don't think this is safe, but I guess no one put quotes or other strange stuff in the name?

A better fix should direct update the raw string

```diff
- ('foo')
+ ('[name="foo"]')
```

```diff
- ("foo")
+ ("[name=\"foo\"]")  // Personally, I like double quoted attribute selector.
```

```diff
- (`foo`)
+ (`[name="foo"]`)
```

Anyway, if it's hard we can leave it as is for now.

rules/prefer-query-selector.js
axetroy: You are right, and I have added the test for multiple names.

fregante:
It's a little complicated and maybe too rare to be worth it, but the correct replacement for `form` elements would be `form.elements.namedItem('foo')`, not `form.querySelector('[name="foo"]')

https://stackoverflow.com/a/76206952/288906

silverwind:
> It's a little complicated and maybe too rare to be worth it, but the correct replacement for `form` elements would be `form.elements.namedItem('foo')`, not `form.querySelector('[name="foo"]')`
> 
> https://stackoverflow.com/a/76206952/288906

If it's to complicated to autofix, I suggest not having a autofixer, just the detection only.

fregante:
I think it's fine as is for now, `querySelector` is not a wrong fix for this rule. Perhaps the selections of form items would have to be its own rule because it's kinda complicated

fregante:
> `querySelector` is not a wrong fix for this rule

I said it's _not_ wrong here, but `namedItem()` would still be the better API to select form items.

silverwind:
Ah I see what you mean now. If you have multiple forms on the page with duplicate `name` fields in them, then yes, only `form.elements.namedItem` will work.

fisker (approved):

PR #2410, opened 2024-07-26
Add `prefer-global-this` rule

### LINKED ISSUE #978: Rule proposal: `prefer-global-this`
[`globalThis`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) is the unified way to access the "global" object.

This has the benefit of making code more cross-platform and more consistent.

`globalThis` requires Node.js 12, so we cannot add it to the recommended preset until April 2021 (date of Node.js 10 obsoletion).

## Fail

```js
window.foo; // Browser

self.foo; // Web Worker

global.foo; // Node.js
```

## Pass

```js
globalThis.foo;
```

axetroy:
Close #978


axetroy:
The current implementation uses the Identifier node for judgment.

However, this node appears in many cases, so it may cause misjudgment. Maybe this isn't a good way to do it.

fisker:
> The current implementation uses the Identifier node for judgment.
> 
> However, this node appears in many cases, so it may cause misjudgment. Maybe this isn't a good way to do it.

We can use [GlobalReferenceTracker](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/utils/global-reference-tracker.js) or use `ReferenceTracker` directly from  `@eslint-community/eslint-utils`

axetroy:
I have reimplemented it and I have handled almost all Expression.

```ts
export interface ExpressionMap {
    ArrayExpression: ArrayExpression;
    ArrowFunctionExpression: ArrowFunctionExpression;
    AssignmentExpression: AssignmentExpression;
    AwaitExpression: AwaitExpression;
    BinaryExpression: BinaryExpression;
    CallExpression: CallExpression;
    ChainExpression: ChainExpression;
    ClassExpression: ClassExpression;
    ConditionalExpression: ConditionalExpression;
    FunctionExpression: FunctionExpression;
    Identifier: Identifier;
    ImportExpression: ImportExpression;
    Literal: Literal;
    LogicalExpression: LogicalExpression;
    MemberExpression: MemberExpression;
    MetaProperty: MetaProperty;
    NewExpression: NewExpression;
    ObjectExpression: ObjectExpression;
    SequenceExpression: SequenceExpression;
    TaggedTemplateExpression: TaggedTemplateExpression;
    TemplateLiteral: TemplateLiteral;
    ThisExpression: ThisExpression;
    UnaryExpression: UnaryExpression;
    UpdateExpression: UpdateExpression;
    YieldExpression: YieldExpression;
}
```

It can match more accurately.

fregante (changes requested):
The rule should accept and pass all known `window` properties, so it likely needs to import a list from npm.

### Pass


```js
window.innerHeight
window.addEventListener()
window.name = 'Future'
```

https://developer.mozilla.org/en-US/docs/Web/API/Window

fregante:
> These codes will still work after changing to `globalThis`

No, those are strictly part of `window`. `addEventListener` must be called on `window` to work, not on `globalThis`.

sindresorhus:
I agree, actual `window` properties should stay being `window.` since it's weird to get, for example, the `innerHeight` from `globalThis`.

sindresorhus:
@fregante I don't think it can be an automatic list as https://developer.mozilla.org/en-US/docs/Web/API/Window lists a lot of non-window-specific properties too. For example, `window.crypto` should be `globalThis.crypto`.

fregante:
> These codes will still work after changing to `globalThis`

Interestingly, that appears to be true. I was under the impression that `globalThis` was not _exactly_ `window` in browsers. Firefox for example returns `false` for `'chrome' in window` in extensions (or at least it used to)




> I don't think it can be an automatic list

As "automatic" as a list manually written and published by someone else, following the same meaning, that's what I meant 😃 

But yeah the list isn't huge so a copy-paste from that page and manual exclusion isn't a huge problem, at least to get this started.

axetroy:
So, how should I distinguish which properties/objects are more reasonable to access on window? This should be the DOM/BOM API.

Do I need to maintain this list in plugin, or a separate npm package?

axetroy:
I added several common APIs, basically DOM and BOM interfaces, welcome to add the missing parts

docs/rules/prefer-global-this.md
```diff
@@ -0,0 +1,48 @@
+# Prefer `globalThis` instead of `window`, `self` and `global`
```
sindresorhus, line 1:
```suggestion
# Prefer `globalThis` over `window`, `self`, and `global`
```

rules/prefer-global-this.js
```diff
@@ -0,0 +1,494 @@
...
+	'navigator',
+	'clientInformation',
+	'originAgentCluster',
+	'alert',
+	'confirm',
+	'prompt',
+	'print',
+	'postMessage',
+
+	// global events
```
sindresorhus, line 72: Not sure these should be here? They don't seem window-specific.

test/prefer-global-this.mjs
```diff
@@ -0,0 +1,158 @@
...
+		'import window, {foo} from "xxx"',
+		'export { window }  from "xxx"',
+		'export * as window from "xxx";',
+		outdent`
+			try {
+
+			} catch (window) {}
+		`,
+
+		// Use window specify apis
```
sindresorhus, line 44:
```suggestion
		// Use window specific APIs
```

docs/rules/prefer-global-this.md
```diff
@@ -0,0 +1,48 @@
...
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+🔧💡 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) and manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+This rule will enforce the use of `globalThis` over `window`, `self`, and `global`.
+
+However, some window-specific APIs are still allowed (e.g. `window.innerWidth`, `window.innerHeight`, `window.navigator`). You can find the list of APIs in the source code of this rule.
```
sindresorhus, line 12: Link to the file.

rules/prefer-global-this.js
```diff
@@ -0,0 +1,491 @@
...
+	'onstorage',
+	'onunhandledrejection',
+	'onunload',
+
+	// https://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-flow
+	'addEventListener',
+	'removeEventListener',
+	'dispatchEvent',
+
+	// https://dom.spec.whatwg.org/#idl-index
```
sindresorhus, line 174: These are no window-specific either.

rules/prefer-global-this.js
```diff
@@ -0,0 +1,491 @@
...
+
+	'self',
+	'location',
+	'navigator',
+	'onerror',
+	'onlanguagechange',
+	'onoffline',
+	'ononline',
+	'onrejectionhandled',
+	'onunhandledrejection',
```
sindresorhus, line 257: While these are technically on `window`, they don't seem window-specific.

sindresorhus (changes requested):

fregante:
I can help filter out some of these. 

From a first look, it varies:

- some are definitely not window properties, like the `Screen` constructor
- some are generally used without `window.` so they could be with globalThis or with nothing at all
- some are events that aren’t strictly related to the window, like `onpaste` (`globalThis.onpaste` might work just as well)

For the first two categories, they should either use globalThis or nothing at all (unless they’re used as `globalThis.location?.href`). So I wonder if instead of “prefer globalThis” this rule should just be “don’t use window”

docs/rules/prefer-global-this.md
```diff
@@ -0,0 +1,48 @@
...
+## Fail
+
+```js
+window
+window.foo
+window[foo]
+window.foo()
+global
+global.foo
+global[foo]
```
fregante, line 23: I suppose this can’t be covered anymore since we don’t know whether foo is an accepted property.

sindresorhus:
> So I wonder if instead of “prefer globalThis” this rule should just be “don’t use window”

But it's "don't use window, or self, or global". I think "prefer `globalThis`" is easier to say.

fregante:
What I was suggesting is that you should still prefer nothing at all over `globalThis` in some cases (unless you're using `?.`), like you wouldn't use `new globalThis.Image()` or `globalThis.Array.from()` or `globalThis.globalThis`

rules/prefer-global-this.js
```diff
@@ -0,0 +1,491 @@
...
+	'screenY',
+	'screenTop',
+	'screenWidth',
+	'screenHeight',
+	'devicePixelRatio',
+	'MediaQueryList',
+	'MediaQueryListEvent',
+	'Screen',
+	'CaretPosition',
+	'VisualViewport',
```
fregante, line 240: Note that this list is somewhat arbitrary. My logic for things that belong here:

- it's directly related to window (✅ `window.close()`)
- it does NOT work well as `globalThis.x` or `x` (✅ `window.frames`, `window.top`)

More easily, I removed:

- all events that bubble up (❌ `window.onplay`)
- all events that works just as well on a "global" (✅ `globalThis.online`)
- all constructors

Note that some constructors are occasionally related to window (like Element) but works just as well as globals, unless you're accessing `iframe.contentWindow.Element`, which is not touched.

```suggestion
	// Properties and methods
	// https://html.spec.whatwg.org/multipage/nav-history-apis.html#the-window-object
	'name',
	'locationbar',
	'menubar',
	'personalbar',
	'scrollbars',
	'statusbar',
	'toolbar',
	'status',
	'close',
	'closed',
	'stop',
	'focus',
	'blur',
	'frames',
	'length',
	'top',
	'opener',
	'parent',
	'frameElement',
	'open',
	'originAgentCluster',
	'postMessage',

	// Events commonly associated with "window"
	'onresize',
	'onblur',
	'onfocus',
	'onload',
	'onscroll',
	'onscrollend',
	'onwheel',
	'onbeforeunload', // Browsers might have specific behaviors on exactly `window.onbeforeunload =`
	'onmessage',
	'onmessageerror',
	'onpagehide',
	'onpagereveal',
	'onpageshow',
	'onpageswap',
	'onunload',

	// To add/remove/dispatch events that are commonly associated with "window"
	// https://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-flow
	'addEventListener',
	'removeEventListener',
	'dispatchEvent',

	// https://dom.spec.whatwg.org/#idl-index
	'event', // Deprecated and quirky, best left untouched

	// https://drafts.csswg.org/cssom-view/#idl-index
	'screen',
	'visualViewport',
	'moveTo',
	'moveBy',
	'resizeTo',
	'resizeBy',
	'innerWidth',
	'innerHeight',
	'scrollX',
	'pageXOffset',
	'scrollY',
	'pageYOffset',
	'scroll',
	'scrollTo',
	'scrollBy',
	'screenX',
	'screenLeft',
	'screenY',
	'screenTop',
	'screenWidth',
	'screenHeight',
	'devicePixelRatio',
```

rules/prefer-global-this.js
```diff
@@ -0,0 +1,491 @@
...
+	'onpopstate',
+	'onrejectionhandled',
+	'onstorage',
+	'onunhandledrejection',
+	'onunload',
+
+	// https://www.w3.org/TR/DOM-Level-2-Events/events.html#Events-flow
+	'addEventListener',
+	'removeEventListener',
+	'dispatchEvent',
```
fregante, line 172: In a future PR, not now, it would be cool to warn on _some_ events.

```js
window.addEventListener('resize', fn); // ✅ 
window.addEventListener('play', fn); // ❌ 
globalThis.addEventListener('play', fn); // ✅ 
```

fregante (changes requested):

rules/prefer-global-this.js
axetroy: Implemented and add test.

rules/prefer-global-this.js
axetroy: It is difficult to determine which properties or methods are window-specific.

But Federico had updated it. I think is have been resolved.

rules/prefer-global-this.js
fregante: Can you turn this message into a comment so we know why some properties are missing and when some should not be added? Because I expect this list to receive some updates and we want to make sure that they make sense.

test/prefer-global-this.mjs
```diff
@@ -0,0 +1,160 @@
...
+				window.foo();
+			}
+		`,
+		'foo.window',
+		'foo.global',
+		'import window from "xxx"',
+		'import * as window from "xxx"',
+		'import window, {foo} from "xxx"',
+		'export { window }  from "xxx"',
+		'export * as window from "xxx";',
```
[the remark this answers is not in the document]
axetroy: test added

rules/prefer-global-this.js
axetroy: Add to comment a496c6f3ab6fadf7dae952a4bdaa4227e6cf2068

fregante (approved):

rules/prefer-global-this.js
```diff
@@ -0,0 +1,380 @@
...
+					if (node.parent && node.parent.type === 'CallExpression') {
+						if (node.parent.arguments.length > 0 && node.parent.arguments[0].type === 'Literal' && windowSpecificEvents.has(node.parent.arguments[0].value)) {
+							return;
+						}
+					} else {
+						return;
+					}
+				} else {
+					return;
+				}
```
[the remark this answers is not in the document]
axetroy: I have simplified in e8412fb1da3e1125da1e8de5a1913bbf698b02a7

docs/rules/prefer-global-this.md
```diff
@@ -0,0 +1,49 @@
...
+// Worker specific APIs
+self.postMessage('Hello')
+self.onmessage = () => {}
+
+globalThis
+globalThis.foo
+globalThis[foo]
+globalThis.foo()
+const {foo} = globalThis
+```
```
sindresorhus, line 49: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2414#discussion_r1696783852

rules/prefer-global-this.js
```diff
@@ -0,0 +1,375 @@
...
+	// ===== Declarations =====
+
+	/** @param {import('estree').VariableDeclarator} node */
+	VariableDeclarator(node) {
+		handleNodes(context, node.init);
+	},
+	/** @param {import('estree').AssignmentPattern} node */
+	AssignmentPattern(node) {
+		handleNodes(context, node.right);
+	},
```
fisker, line 359: We can't write rule like this.

Have you tried `GlobalReferenceTracker` metioned in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2410#issuecomment-2252217275?

If it doesn't fit your needs, we have [`isReferenceIdentifier`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/77f32e5a6b2df542cf50dfbd371054f2cd8ce2d6/rules/ast/is-reference-identifier.js#L150C10-L150C31) may help too.

fisker (changes requested):
Please try to reuse existing utilities.

rules/prefer-global-this.js
axetroy: Okay, I'll take a look at how to use this.

axetroy: The code has been Refactored with GlobalReferenceTracker without change to any test case. 817d8aa36b27588c350ca077664f6afd3d5b7433

docs/rules/prefer-global-this.md
axetroy: Updated at ff4346535b77e051c6d5c14fcaca147d673c3aea

Here is the [Rendered Preview](https://github.com/axetroy/eslint-plugin-unicorn/blob/prefer-global-this/docs/rules/prefer-global-this.md)

rules/prefer-global-this.js
```diff
@@ -0,0 +1,266 @@
...
+const isWebWorkerSpecificAPI = node => node.type === 'MemberExpression' && node.object.name === 'self' && node.property.type === 'Identifier' && webWorkerSpecificAPIs.has(node.property.name);
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => {
+	const tracker = new GlobalReferenceTracker({
+		objects: globalIdentifier,
+		handle(reference) {
+			const {node} = reference;
+
+			switch (node.type) {
```
sindresorhus, line 223: No point in doing a switch over a single case.

axetroy:
> Please try to reuse existing utilities.

@fisker  Can you review it again?

sindresorhus (approved):

rules/prefer-global-this.js
```diff
@@ -0,0 +1,259 @@
...
+
+@param {import('estree').MemberExpression} node
+@returns {boolean}
+*/
+const isWindowSpecificAPI = node => {
+	if (node.type !== 'MemberExpression') {
+		return false;
+	}
+
+	if (node.object.name !== 'window' || node.property.type !== 'Identifier') {
```
fisker, line 192: Need add `!node.computed` here, don't forgot to test.

rules/prefer-global-this.js
```diff
@@ -0,0 +1,259 @@
...
+		return false;
+	}
+
+	if (windowSpecificAPIs.has(node.property.name)) {
+		if (['addEventListener', 'removeEventListener', 'dispatchEvent'].includes(node.property.name) && node.parent && node.parent.type === 'CallExpression') {
```
fisker, line 197: `node.parent &&` also unnecessary.

We need test `node === node.parent.callee`.

rules/prefer-global-this.js
```diff
@@ -0,0 +1,259 @@
...
+	});
+
+	return {
+		...tracker.createListeners(context),
+		/** @param {import('estree').AssignmentExpression} node */
+		AssignmentExpression(node) {
+			handleNodes(context, [node.left]);
+		},
+		/** @param {import('estree').UpdateExpression} node */
+		UpdateExpression(node) {
```
fisker, line 240: According to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/77f32e5a6b2df542cf50dfbd371054f2cd8ce2d6/rules/utils/is-left-hand-side.js#L3, 
 Looks like we are missing `AssignmentPattern`
, `ArrayPattern`, `ObjectPattern` and `UnaryExpression`(`delete window.xxx`, but maybe no need to care.)

rules/prefer-global-this.js
```diff
@@ -0,0 +1,259 @@
+'use strict';
+
+const {GlobalReferenceTracker} = require('./utils/global-reference-tracker.js');
+
+const MESSAGE_ID_ERROR = 'prefer-global-this/error';
+const MESSAGE_ID_SUGGESTION = 'prefer-global-this/suggestion';
+const messages = {
+	[MESSAGE_ID_ERROR]: 'Prefer `{{replacement}}` over `{{value}}`.',
```
fisker, line 8: The `replacement` is always `globalThis`, no need use template.

rules/prefer-global-this.js
```diff
@@ -0,0 +1,259 @@
...
+@param {string} value
+*/
+function report(context, node, value) {
+	const fix = (/** @type {import('eslint').Rule.RuleFixer} fixer */ fixer) => fixer.replaceText(node, 'globalThis');
+	context.report({
+		node,
+		messageId: MESSAGE_ID_ERROR,
+		data: {value, replacement: 'globalThis'},
+		fix,
+		suggest: [{messageId: MESSAGE_ID_SUGGESTION, data: {value, replacement: 'globalThis'}, fix}],
```
fisker, line 156: Are we providing suggestion and autofix at the same time??


(We need a tool to prevent this from happening.)

rules/prefer-global-this.js
```diff
@@ -0,0 +1,259 @@
...
+@param {import('estree').Node | Array<import('estree').Node>} nodes
+*/
+function handleNodes(context, nodes) {
+	if (!Array.isArray(nodes)) {
+		nodes = [nodes];
+	}
+
+	for (const node of nodes) {
+		if (node && node.type === 'Identifier' && globalIdentifier.has(node.name)) {
+			const variable = findVariableInScope(context.sourceCode.getScope(node), node.name);
```
fisker, line 173: I'm not quite sure about this line, are are checking `window`/`self` is global not local, right?

If I'm correct, we have `isShadowed` for this purpose.(Not sure what we'll miss, but that utility exists for a while).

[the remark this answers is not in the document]
fisker: This also makes me thinking about that, do we need check the replacement(`globalThis`) is available in this scope? (Not sure if worth the effort)

rules/prefer-global-this.js
axetroy: resolved. 461014aefa976b59a6b8c86397248f711600f415

rules/prefer-global-this.js
axetroy: Resolved. 7fe5672e672732fe13ad835791f547c3352bd003

rules/prefer-global-this.js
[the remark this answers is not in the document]
axetroy: Yes, this is to check if this `window` is a shadowed variable
```js
function test (window) {

}
```

rules/prefer-global-this.js
axetroy: Resolved. c7887b6ddc3f26c8050fd529798ada51cc7ecc0a

rules/prefer-global-this.js
axetroy: I think `!node.computed` should not be added here, or am I missing some case?.

Regardless of whether it is a computed property or not, this rule should be applied.

There is the test that we already have

```js
window['title'] // node.computed === true
window.title    // node.computed === false
```

```md
## invalid(4): window.foo

> Input

    `␊
      1 | window.foo␊
    `

> Output

    `␊
      1 | globalThis.foo␊
    `

> Error 1/1

    `␊
    > 1 | window.foo␊
        | ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `

## invalid(5): window[foo]

> Input

    `␊
      1 | window[foo]␊
    `

> Output

    `␊
      1 | globalThis[foo]␊
    `

> Error 1/1

    `␊
    > 1 | window[foo]␊
        | ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `

## invalid(6): window["foo"]

> Input

    `␊
      1 | window["foo"]␊
    `

> Output

    `␊
      1 | globalThis["foo"]␊
    `

> Error 1/1

    `␊
    > 1 | window["foo"]␊
        | ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `
```

rules/prefer-global-this.js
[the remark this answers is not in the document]
axetroy: Resvoled. 99c7031f3ae56b6ffa1e8b92d689a05516abb9f5

rules/prefer-global-this.js
axetroy: `delete window.foo` is a `MemberExpression` that we already tested.

```md
## invalid(27): delete window.foo

> Input

    `␊
      1 | delete window.foo␊
    `

> Output

    `␊
      1 | delete globalThis.foo␊
    `

> Error 1/1

    `␊
    > 1 | delete window.foo␊
        |        ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `
```

axetroy: `ObjectPattern` and `ArrayPattern` are necessary to test.
```js
// ObjectPattern
const { window } = obj

// ArrayPattern
const [window] = obj
```

### AssignmentPattern

```markdown
## invalid(48): function greet({ name = window.foo }) {}

> Input

    `␊
      1 | function greet({ name = window.foo }) {}␊
    `

> Output

    `␊
      1 | function greet({ name = globalThis.foo }) {}␊
    `

> Error 1/1

    `␊
    > 1 | function greet({ name = window.foo }) {}␊
        |                         ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `

```

fisker: Can you add test for 

`({ foo = window.foo } =  {});`

and
`[window.foo] = []`;

axetroy: These are invalid syntax

```
unknown: Unexpected token, expected "," (1:28)

> 1 | function greet({ foo: window.foo } =  {}) {}
    |                             ^
```

```
unknown: Unexpected token, expected "," (1:22)

> 1 | function greet([window.foo] = []) {}
    |                       ^
```

axetroy:
```markdown

## invalid(49): ({ foo: window.foo } = {})

> Input

    `␊
      1 | ({ foo: window.foo } =  {})␊
    `

> Output

    `␊
      1 | ({ foo: globalThis.foo } =  {})␊
    `

> Error 1/1

    `␊
    > 1 | ({ foo: window.foo } =  {})␊
        |         ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `

## invalid(50): [window.foo] = []

> Input

    `␊
      1 | [window.foo] = []␊
    `

> Output

    `␊
      1 | [globalThis.foo] = []␊
    `

> Error 1/1

    `␊
    > 1 | [window.foo] = []␊
        |  ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
    `

```

axetroy:
CI failed, But it doesn't seem to be caused by this PR

Same as https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/10377594904/job/28732103760

```
  no-for-loop › invalid(1): for (let i = 0; i < positions.length; i++) { \u0009let last: vscode.Position | vscode.Range = positions[i]; \u0009let selectionRange = allProviderRanges[i]; }

  Assertion failed: 

  A fatal parsing error occurred in autofix.
  Error: Parsing error: The variable declaration of a 'for...of' statement cannot have a type annotation.
  Autofix output:
  for (let [i, last]: [number, vscode.Position | vscode.Range] of positions.entries()) {
  	let selectionRange = allProviderRanges[i];
  }

  AssertionError [ERR_ASSERTION]: A fatal parsing error occurred in autofix.
  Error: Parsing error: The variable declaration of a 'for...of' statement cannot have a type annotation.
  Autofix output:
  for (let [i, last]: [number, vscode.Position | vscode.Range] of positions.entries()) {
  	let selectionRange = allProviderRanges[i];
  }
      at runRuleForItem (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:828:17)
      at testInvalidTemplate (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:967:28)
      at /home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:1274:33
      at file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint-ava-rule-tester/index.js:27:7
      at Test.callFn (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/test.js:525:26)
      at Test.run (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/test.js:534:33)
      at Runner.runSingle (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/runner.js:281:33)
      at Runner.runTest (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/runner.js:363:30)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Promise.all (index 92)



  no-for-loop › invalid(2): for (let i = 0; i < positions.length; i++) { \u0009const last /* comment */ : /* comment */ Position = positions[i]; \u0009console.log(i); }

  Assertion failed: 

  A fatal parsing error occurred in autofix.
  Error: Parsing error: The variable declaration of a 'for...of' statement cannot have a type annotation.
  Autofix output:
  for (const [i, last   /* comment */]: [number, /* comment */ Position] of positions.entries()) {
  	console.log(i);
  }

  AssertionError [ERR_ASSERTION]: A fatal parsing error occurred in autofix.
  Error: Parsing error: The variable declaration of a 'for...of' statement cannot have a type annotation.
  Autofix output:
  for (const [i, last   /* comment */]: [number, /* comment */ Position] of positions.entries()) {
  	console.log(i);
  }
      at runRuleForItem (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:828:17)
      at testInvalidTemplate (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:967:28)
      at /home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:1274:33
      at file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint-ava-rule-tester/index.js:27:7
      at Test.callFn (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/test.js:525:26)
      at Test.run (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/test.js:534:33)
      at Runner.runSingle (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/runner.js:281:33)
      at Runner.runTest (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/runner.js:363:30)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Promise.all (index 93)



  no-for-loop › invalid(3): for (let i = 0; i < positions.length; i++) { \u0009let last: vscode.Position | vscode.Range = positions[i]; }

  Assertion failed: 

  A fatal parsing error occurred in autofix.
  Error: Parsing error: The variable declaration of a 'for...of' statement cannot have a type annotation.
  Autofix output:
  for (let last: vscode.Position | vscode.Range of positions) {
  }

  AssertionError [ERR_ASSERTION]: A fatal parsing error occurred in autofix.
  Error: Parsing error: The variable declaration of a 'for...of' statement cannot have a type annotation.
  Autofix output:
  for (let last: vscode.Position | vscode.Range of positions) {
  }
      at runRuleForItem (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:828:17)
      at testInvalidTemplate (/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:967:28)
      at /home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint/lib/rule-tester/rule-tester.js:1274:33
      at file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/eslint-ava-rule-tester/index.js:27:7
      at Test.callFn (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/test.js:525:26)
      at Test.run (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/test.js:534:33)
      at Runner.runSingle (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/runner.js:281:33)
      at Runner.runTest (file:///home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/node_modules/ava/lib/runner.js:363:30)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Promise.all (index 94)

  ─

  3 tests failed
```

rules/prefer-global-this.js
fisker: `window[title]`

axetroy: There is no difference between `window[title]` and `window[foo]`

Both `title` and `foo` are variable accesses. test added.

```md
## invalid(6): window[title]

> Input

    `␊
      1 | window[title]␊
    `

> Output

    `␊
      1 | globalThis[title]␊
    `

> Error 1/1

    `␊
    > 1 | window[title]␊
        | ^^^^^^ Prefer \`globalThis\` over \`window\`.␊
```

huhuanming:
will there be any new progress? 😳

axetroy:
> will there be any new progress? 😳

This PR is almost ready to be merged. Block by #2424

rules/prefer-global-this.js
fisker: `window[name]` doesn't mean it's accessing `window.name`, since `name` can be other values.

fisker: We are not checking `window` is `MemberExpression.object` either, `foo[window]`

fisker: Please don't click `resolve` button, I can't find my comment.

fisker:
Not sure about the strategy for unknown properties here.

```
window[foo];
const foo = 'name'; window[foo];
const foo = 'not-name'; window[foo];
```

I think we should ignore them, or simply ignore all computed properties.

axetroy:
> Not sure about the strategy for unknown properties here.
> 
> ```
> window[foo];
> const foo = 'name'; window[foo];
> const foo = 'not-name'; window[foo];
> ```
> 
> I think we should ignore them, or simply ignore all computed properties.

I prefer to ignore all computed properties. We don't know whether the variable of the calculated property is in the whitelist.

sindresorhus:
I agree with ignoring them. It's an edge-case anyway.

fisker:
> I prefer to ignore all computed properties.

Agree, let's ignore computed properties then.

rules/prefer-global-this.js
axetroy: > We are not checking `window` is `MemberExpression.object` either, `foo[window]`

The test added in 4be46990636a294b5fa45064a478a4ecbdddeda8

test/prefer-global-this.mjs
```diff
@@ -0,0 +1,173 @@
...
+		'[window.name] = []',
+	],
+	invalid: [
+		'global',
+		'self',
+		'window',
+		'window.foo',
+		'window[foo]',
+		'window[title]',
+		'window["foo"]',
```
fisker, line 67: These should be valid, since they are computed.

axetroy: Update in e3e48c070298b61520b659c195c69ce1e86bb8a3

.github/workflows/title-formatter.yml
```diff
@@ -1,23 +0,0 @@
...
-      issues: write
-    steps:
-      - uses: actions/checkout@v4
-      - name: Auto-format rule names in titles
-        uses: fregante/title-replacer-action@v1
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          pattern-path: docs/rules
-          replacement: '`$0`'
-          trim-punctuation: true
```
fisker, line 23: File deleted by accident?

axetroy: Yeah. Reverted

sindresorhus:
Let us know when this is ready for another review. Make sure https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2410/files#r1710020285 is resolved first.

axetroy:
> Let us know when this is ready for another review. Make sure [#2410 (files)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2410/files#r1710020285) is resolved first.

This is already resolved. see https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2410/files#diff-560654447874168910d7716e1da47d1f8f03eebfc3f5eaa14f7977133d1d1882R57-R64

huhuanming (approved):

axetroy:
@fisker Could you please review it again and let me know if there is anything that needs to be modified

axetroy:
> I did not reviewed this PR carefully. Inspired by the implementation of eslint built-in rule [no-restricted-globals](https://github.com/eslint/eslint/blob/main/lib/rules/no-restricted-globals.js?rgh-link-date=2024-09-10T07%3A14%3A02Z). I provide a simple demo to implement it.
> 
> ```js
> {
>   create: (context) => ({
>     Program: (node) => {
>       const banned = ["global", "self"];
>       // Report variables declared elsewhere
>       const scope = context.sourceCode.getScope(node);
>       scope.variables.forEach((v) => {
>         if (banned.includes(v.name)) {
>           v.references.forEach((ref) => {
>             context.report({node:ref.identifier, message: "xxxxxx"});
>           });
>         }
>       });
>       // Report variables not declared at all
>       scope.through.forEach((ref) => {
>         if (banned.includes(ref.identifier.name)) {
>            context.report({node:ref.identifier, message: "xxxxxx"});
>         }
>       });
>     },
>   }),
> }
> ```

Looks good! I will try

rules/prefer-global-this.js
```diff
@@ -0,0 +1,205 @@
...
+@param {import('estree').MemberExpression} node
+@returns {boolean}
+*/
+const isWebWorkerSpecificAPI = node => node.type === 'MemberExpression' && node.object.name === 'self' && node.property.type === 'Identifier' && webWorkerSpecificAPIs.has(node.property.name);
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	Program(programNode) {
+		const scope = context.sourceCode.getScope(programNode);
+		for (const variable of scope.variables) {
```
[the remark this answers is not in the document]
axetroy: Thanks for report, now it fixed and added test cases that are not declared in `globals` options

rules/prefer-global-this.js
```diff
@@ -0,0 +1,220 @@
...
+@returns {boolean}
+*/
+const isWebWorkerSpecificAPI = node => node.type === 'MemberExpression' && node.object.name === 'self' && node.property.type === 'Identifier' && webWorkerSpecificAPIs.has(node.property.name);
+
+/**
+@param {import('eslint').Rule.RuleContext} context
+@param {import('estree').Identifier} node
+*/
+function reportProblem(context, node) {
+	if (isWindowSpecificAPI(node.parent) || isWebWorkerSpecificAPI(node.parent) || isComputedMemberExpression(node.parent, node)) {
```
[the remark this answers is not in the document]
fisker: No, we decide not to report computed properties.

rules/prefer-global-this.js
```diff
@@ -0,0 +1,220 @@
...
+@param {import('estree').MemberExpression} node
+@returns {boolean}
+*/
+const isWebWorkerSpecificAPI = node => node.type === 'MemberExpression' && node.object.name === 'self' && node.property.type === 'Identifier' && webWorkerSpecificAPIs.has(node.property.name);
+
+/**
+@param {import('eslint').Rule.RuleContext} context
+@param {import('estree').Identifier} node
+*/
+function reportProblem(context, node) {
```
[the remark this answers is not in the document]
fisker: Worker APIs should use `self`

fisker (approved):
Great work, and sorry for the delay.

rules/prefer-global-this.js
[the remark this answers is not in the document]
fregante: Should they? I think `self` was chosen because there's no `window`, but `globalThis` should still be preferred there I think.

- Service workers introduction: 2014
- `globalThis` introduction: 2019

I think using `self` in service workers is more of a historical artifact rather than a necessity. We can discuss/change this later though, I don't want to hold this PR any longer 🙏

sindresorhus:
Nice work, @axetroy 👍

rules/prefer-global-this.js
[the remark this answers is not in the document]
sindresorhus: @fregante Can you open an issue?

charles4221:
Just upgraded my version of the plugin and started getting lint errors for this rule on code like this:

```ts
export function detectPrefersDarkMode() {
  return (
    typeof window !== 'undefined' &&
    window.matchMedia('(prefers-color-scheme: dark)').matches
  );
}
```

Shouldn't this type of usage be excluded from this rule since `globalThis.matchMedia` will crash with a `TypeError: globalThis.matchMedia is not a function`?

fregante:
It works for me. That function does exist in your browser. Are you sure that's not just your test env crashing?

perrin4869:
So this rule is failing for me on this line:

https://github.com/dotcore64/react-stay-scrolled/blob/master/src%2Findex.js#L11

Since `globalThis` is defined in all environments, it can't be used to detect a browser environment or a different environment.

It would be cool if these edge cases could be handled

fregante:
@perrin4869 see #2468

rules/prefer-global-this.js
```diff
@@ -0,0 +1,210 @@
...
+]);
+
+const webWorkerSpecificAPIs = new Set([
+	// https://html.spec.whatwg.org/multipage/workers.html#the-workerglobalscope-common-interface
+	'addEventListener',
+	'removeEventListener',
+	'dispatchEvent',
+
+	'self',
+	'location',
```
stof, line 108: why allowing `self.location` in workers but not `window.location` ?

fregante: See #2458

PR #2413, opened 2024-07-30
[WIP] Add `prefer-math-abs` rule

### LINKED ISSUE #1249: Rule proposal: `prefer-math-abs`
## Fail

```js
const a = b < 0 ? -b : b;
const a = b <= 0 ? -b : b;
const a = b < 0 ? 0 - b : b;
const a = b <= 0 ? 0 - b : b;
```

```js
const a = b > 0 ? b : -b;
const a = b >= 0 ? b : -b;
```

```js
const a = b < 0 ? b * -1 : b;
```

```js
if (number > POSITIVE_CONSTANT || number < -POSITIVE_CONSTANT) {}
```

## Pass

```js
const a = Math.abs(b);
```

```js
if (Math.abs(number) > POSITIVE_CONSTANT) {}
```

axetroy:
Close #1249

Note: The current implementation may be a bit complicated and needs to be simplified


rules/utils/is-nodes-equal.js
```diff
@@ -0,0 +1,44 @@
...
+
+const ignoreKeys = new Set(['loc', 'start', 'end', 'parent', 'range']);
+
+/**
+ * Compare two AST nodes for equality by structure and value.
+ * @param {import('estree').Node} node1 - The first AST node.
+ * @param {import('estree').Node} node2 - The second AST node.
+ * @returns {boolean} - True if the nodes are structurally and value-wise equal, false otherwise.
+ */
+function isNodesEqual(node1, node2) {
```
fisker, line 11: How is this different than [`isSameReference`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/77f32e5a6b2df542cf50dfbd371054f2cd8ce2d6/rules/utils/is-same-reference.js#L111C10-L111C25)?

axetroy: `isSameReference` will treat two identical nodes as different because they have different references.

But `isNodesEqual` will treat them as the same.

For example:

```js
isSameReference({ foo: 1 }, { foo: 1 }) // false
isNodesEqual({ foo: 1 }, { foo: 1 }) // true
```

PR #2414, opened 2024-07-30
Add `prefer-map-has` rule

### LINKED ISSUE #1329: Rule proposal: `prefer-map-has`
When checking existence, should prefer `Map#has()` over `Map#get()`.

## Fail

```js
if (map.get(foo));
```

## Pass

```js
if (map.has(foo));
```

I think better ship this with [`prefer-map`](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1285) together, so user can notice this problem when we fix `object.foo` to `map.get(foo)` in `prefer-map`.

axetroy:
Close #1329


rules/prefer-add-event-listener.js
```diff
@@ -44,7 +44,7 @@ const shouldFixBeforeUnload = (assignedExpression, nodeReturnsSomething) => {
 		return false;
 	}
 
-	return !nodeReturnsSomething.get(assignedExpression);
+	return !Reflect.apply(nodeReturnsSomething.get, nodeReturnsSomething, [assignedExpression]);
```
axetroy, line 47: This is not an accidental modification, because if this is not changed, the CI test will not pass.

The test case still passes, nothing change.

docs/rules/prefer-map-has.md
```diff
@@ -0,0 +1,52 @@
+# Prefer `Map#has` over `Map#get`
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+<!-- Remove this comment, add more detailed description. -->
```
sindresorhus, line 10: ⬆️

docs/rules/prefer-map-has.md
```diff
@@ -0,0 +1,52 @@
...
+map.has(foo);
+
+while (map.has(foo)) {
+  // do staff
+}
+
+do {} while (map.has(foo));
+
+do {} while (map.has(foo) || otherCondition);
+```
```
sindresorhus, line 52: This is a good example, on why we should start doing this: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1751

Please format it like in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1751#issuecomment-1254913032, with one code block per example.

rules/prefer-map-has.js
```diff
@@ -0,0 +1,161 @@
...
+const messages = {
+	[MESSAGE_ID_ERROR]: 'Prefer `{{replacement}}` over `{{value}}`.',
+	[MESSAGE_ID_SUGGESTION]: 'Replace `{{value}}` with `{{replacement}}`.',
+};
+
+/**
+ * @param {import('eslint').Rule.RuleContext} context
+ * @param {import('estree').SimpleCallExpression} node
+ * @param {import('estree').Node} [replacement]
+ */
```
sindresorhus, line 14:
```suggestion
/**
@param {import('eslint').Rule.RuleContext} context
@param {import('estree').SimpleCallExpression} node
@param {import('estree').Node} [replacement]
*/
```

rules/prefer-map-has.js
```diff
@@ -0,0 +1,161 @@
...
+function processNode(context, node, replacement = node) {
+	if (node.callee.type === 'MemberExpression' && node.callee.object.type === 'Identifier'
```
sindresorhus, line 16: To reduce nesting, do an early-return.

sindresorhus (changes requested):

rules/prefer-map-has.js
axetroy: This doesn't seem to be the recommended style for `jsdoc`

sindresorhus: It's the style we use.

docs/rules/prefer-map-has.md
```diff
@@ -0,0 +1,56 @@
+# Prefer `Map#has` over `Map#get`
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+In some places where Boolean value judgment is required, it is more reasonable to prefer `Map#has` over `Math#get`.
```
sindresorhus, line 10: Needs a better description. Feel free to use ChatGPT to improve it.

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2414#discussion_r1696783852 was not resolved.

docs/rules/prefer-map-has.md
axetroy: I have already been using it, but I haven't found a more suitable description. Could you help me with this?

docs/rules/prefer-map-has.md
axetroy: Maybe I misunderstood. According to this format, there are no longer two code blocks `Fail` and `Pass`.

Has it become two code blocks: `Blocks` and `Single-line`?

axetroy: I did update a version of this

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2414/commits/4e05b9b2b97fd739743ef404d2f0454534a74564 ?

docs/rules/prefer-map-has.md
```diff
@@ -0,0 +1,56 @@
+# Prefer `Map#has` over `Map#get`
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+💡 This rule is manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+In Boolean expressions, prefer `Map#has` over `Map#get`. For example, in IfStatements, WhileStatements, DoWhileStatements, ForStatements, and ConditionalExpressions.
```
sindresorhus, line 10: I find it hard to believe this is the best you can do even with the help of AI.

This is what I got back, which is much better:

> Use `Map#has` instead of `Map#get` to check for the presence of a key in a `Map`. `Map#has` is more semantically correct for existence checks and avoids unnecessary retrieval of values.

sindresorhus:
This is not related to this pull request and should be done in a separate pull request: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2414/files#r1696698356

docs/rules/prefer-map-has.md
axetroy: My bad, I didn't find the keywords

axetroy:
> [4e05b9b](https://github.com/sindresorhus/eslint-plugin-unicorn/commit/4e05b9b2b97fd739743ef404d2f0454534a74564) ?

I did update but CI failed

```log
  package › Every rule has a doc with the appropriate content

  test/package.mjs:134

   133:     if (!RULES_WITHOUT_PASS_FAIL_SECTIONS.has(ruleName)) {              
   134:       t.true(documentContents.includes('## Pass'), `${ruleName} include…
   135:       t.true(documentContents.includes('## Fail'), `${ruleName} include…

  prefer-map-has includes '## Pass' examples section

  Value is not `true`:

  false

  › file://test/package.mjs:134:10
```

axetroy:
Close since it will bring unexpected false positives.

PR #2423, opened 2024-08-13
`prefer-default-parameters` - Document falsy values

### LINKED ISSUE #2422: `prefer-default-parameters` does not properly consider nulls
The rule `prefer-default-parameters` wants this:

```js
function doSomething(couldBeNull) {
  const normalized = couldBeNull ?? ''
  // Do something with normalized
}
```

to look like what exactly?
```js
function doSomething(couldBeNull = '') {
  // Do something with couldBeNull, but how to handle null without violating rule?
}
```

This is a problem because:
```js
doSomething(someExternalLibraryIHaveNoControlOverThatCouldReturnNull())
```

Of course I could fix all the call sites since I own `doSomething`:
```js
doSomething(someExternalLibraryIHaveNoControlOverThatCouldReturnNull() ?? '')
```

But, no thank you.

fregante:
- Close #2422


fregante:
Failing CI is not due to this change, right? 🤔

PR #2425, opened 2024-08-19
Add `consistent-existence-index-check` rule

### LINKED ISSUE #1024: Rule proposal: `consistent-non-exists-index-check`
Similar to [`explicit-length-check`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/explicit-length-check.md), enforce consistent style when check `index === -1` or `index !== -1`

## Fail

```js
const index = foo.indexOf('bar');

if (index < 0) {}
```

```js
const index = foo.indexOf('bar');

if (index >= 0) {}
```

## Pass

```js

const index = foo.indexOf('bar');

if (index === -1) {}
```

```js

const index = foo.indexOf('bar');

if (index !== -1) {}
```

But this is hard to detect, since we already [enforce `includes`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/prefer-includes.md) if `indexOf()` is in a position as `boolean`.

axetroy:
Close #1024


sindresorhus:
I think the rule should be called `consistent-indexof-check`.

docs/rules/consistent-non-exists-index-check.md
```diff
@@ -0,0 +1,36 @@
+# Enforce consistent styling when checking for element existence using `indexOf()`
```
sindresorhus, line 1:
```suggestion
# Enforce consistent style for element existence checks with `indexOf()`
```

sindresorhus (changes requested):

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2425#discussion_r1721190952

axetroy:
> [#2425 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2425#discussion_r1721190952)

I do remember I updated it 😢

axetroy:
Perhaps, it should be renamed to `consistent-existence-index-check`? since this rule check for `indexOf()`, `lastIndexOf()`, `findIndex()`, and `findLastIndex()`

rules/consistent-indexof-check.js
```diff
@@ -0,0 +1,116 @@
...
+	'>=': '<=',
+};
+
+/**
+Check if the node is a call expression of `indexOf` or `lastIndexOf` method.
+
+@param {import('estree').Node} node
+@returns {node is import('estree').CallExpression}
+*/
+function isIndexOfCallExpression(node) {
```
fisker, line 23: Please switch to use `isMethodCall`.

rules/consistent-indexof-check.js
```diff
@@ -0,0 +1,116 @@
...
+			return;
+		}
+
+		const replacement = getReplacement(operator, literal, identifier);
+
+		if (!replacement) {
+			return;
+		}
+
+		const variableFound = resolveVariableName(identifier, context.sourceCode.getScope(node));
```
fisker, line 71: What do you think listen on assignment `const index = ...indexOf(..)` then look for comparison instead?

rules/consistent-indexof-check.js
axetroy: b24f0c24f74d114dce607572e504653d334515a7

rules/consistent-indexof-check.js
axetroy: This implementation may get a better performance. (Because it searches for `const index = ...indexOf(..)` statement first, and then searches for `BinaryExpression`)

But it is also more complicated, There are countless possibilities for `BinaryExpression` to appear, and traverse almost the entire AST, I think it is not worth it.

for example:

```js
const index = "foo".indexOf("o");

function test() {
	class Man {
		method() {
			return {
				nest () {
					if (index > -1) { // ❌ You have to traverse from the top level to here
					
					}
				},
				noop () {
					// ❌ Even if there is no node here, you must traverse here to check
				}
			}
		}
	}
}
```

fisker: It's easy to find them, since `index` variable already include all the references.

axetroy: All right, I will try it later

axetroy: Done in 732ae0560cd55526c8e77c45ef3a7a479bae4d9a

axetroy: Should we catch the case of `AssignmentExpression` ?

```
let index = 0;

// do something

index = foo.indexOf("xx")

// ...

if (index > -1) {

}
```

axetroy: > Should we catch the case of `AssignmentExpression` ?
> 
> ```js
> let index = 0;
> 
> // do something
> 
> index = foo.indexOf("xx")
> 
> // ...
> 
> if (index > -1) {
> 
> }
> ```

Implemented

fisker: > Should we catch the case of AssignmentExpression ?

No, should only check  work for declaration with `const`

rules/consistent-indexof-check.js
```diff
@@ -0,0 +1,133 @@
...
+
+/**
+Find and process references to a given identifier.
+
+@param {import('eslint').Rule.RuleContext} context
+@param {import('estree').Identifier} identifierNode
+@returns
+*/
+function findAndProcessReferences(context, identifierNode) {
+	const variable = resolveVariableName(identifierNode.name, context.sourceCode.getScope(identifierNode));
```
fisker, line 58: `sourceCode.getDeclaredVariables` should work here.

rules/consistent-indexof-check.js
axetroy: > No, should only check declaration with const

Is there any reason for this?

fisker: We can't know how it will be used, we don't want false positives.

rules/consistent-indexof-check.js
axetroy: Update in 4d88f256840842b5946add5b903e0f92f52f5bbf

rules/consistent-indexof-check.js
```diff
@@ -0,0 +1,130 @@
...
+				operator,
+				literal,
+				identifier.name,
+			);
+
+			if (!replacement) {
+				continue;
+			}
+
+			context.report({
```
axetroy, line 90: This `context.report()` can not replace with a return statement

Otherwise, It can not report multiple issues when it returns.

fisker:
@axetroy I simplified the rule only check `index < 0`, `index >= 0`, `index > -1`. I think it's enough, what do you think?

axetroy:
> @axetroy I simplified the rule only check `index < 0`, `index >= 0`, `index > -1`. I think it's enough, what do you think?

I think we should support as many edge cases as possible, for example

Even if `index < -100` and `index < 0` are no different, it would be better if it can be fixed.

Of course, if it is not easy to do, then I am fine with it.

fisker (approved):

sindresorhus:
> Perhaps, it should be renamed to consistent-existence-index-check?

👍

axetroy:
> > Perhaps, it should be renamed to consistent-existence-index-check?
> 
> 👍

@sindresorhus Renamed

PR #2427, opened 2024-08-19
`prefer-string-slice`: remove unsafe autofix for `String#substr()`

### LINKED ISSUE #2291: `prefer-string-slice` bad autofix from `.substr()`
```js
let string = "::";

// Prefer `String#slice()` over `String#substr()`.
// eslint(unicorn/prefer-string-slice)
let eslintError = string.substr(-2, 2);
                  ^^^^^^^^^^^^^^^^^^^^
// After applying fix:
let fixed = string.slice(-2, -2 + 2);

eslintError // "::"
fixed // ""

// Uncaught AssertionError [ERR_ASSERTION]: '::' == ''
assert.equal(eslintError, fixed)
```

axetroy:
Fix #2291

When the second argument `length` is equal to the length of the string,

it will autofix to `.slice(start, 0)`

Now we make to `.slice(start, 0 || undefined)` to align the behavior of `.substr(start, length)`


rules/prefer-string-slice.js
```diff
@@ -67,7 +67,8 @@ function * fixSubstrArguments({node, fixer, context, abort}) {
 	if (argumentNodes.every(node => isNumber(node, scope))) {
 		const firstArgumentText = getParenthesizedText(firstArgument, sourceCode);
 
-		yield fixer.insertTextBeforeRange(secondArgumentRange, `${firstArgumentText} + `);
+		yield fixer.insertTextBeforeRange(secondArgumentRange, `(${firstArgumentText} + `);
+		yield fixer.insertTextAfterRange(secondArgumentRange, ') || undefined');
```
fisker, line 71: The original `Math.min()` already ugly, let's turn this fix into suggestion, I don't think people will use the fixed code directly.

axetroy: I admit, this is ugly, but I can't think of a more elegant fix

fisker: > let's turn this fix into suggestion

Use suggestion instead.

axetroy: The source code is too complicated, and I don't have the confidence to refactor it. Close PR

fisker:
Let's just remove the autofix not much value to provide a suggestion.

fisker (approved):

PR #2428, opened 2024-08-19
[WIP] Add `prefer-change-array-by-copy` rule

### LINKED ISSUE #1514: Rule proposal: `prefer-change-array-by-copy`
["Change Array by copy" proposal](https://github.com/tc39/proposal-change-array-by-copy#change-array-by-copy)

## Fail

```js
array.reverse();
```

```js
array.sort();
```

```js
const array = [...foo];
array.splice(0, 1);
```

```js
const array = [...foo];
array[1] = 'changed';
```

## Pass

```js
array = array.toReversed();
```

```js
array = array.toSorted();
```

```js
const array = foo.toSpliced(0, 1);
```

```js
const array = foo.with(1, 'changed');
```

### LINKED ISSUE #1743: Rule proposal: Prevent returning/assigning mutating array methods
### Description

Prevent returning or assigning mutating array methods, to improve readability and also protect against accidentally mutating an array in-place when you shouldn't. `Array#sort()`, `Array#fill()`, `Array#reverse()`.

### Fail

```js
console.log(foo.sort());
```

```js
const sorted = foo.sort();
```

### Pass

```js
foo.sort();
console.log(foo);
```

```js
foo.sort();
```

I guess we should also allow the case where there's an intermediate array?

```js
const sorted = foo.map(…).sort();
```

axetroy:
Close #1514
Close #1743

If it is autofixable, It may lead to unexpected behavior and bugs

```diff
const array = [1,2,3]
- array.reverse()
+ array.toReversed()

for (const item of array) {
    console.log(item)   // print `3,2,1` before fix
                        // print `1,2,3` after fix
}
```


docs/rules/prefer-change-array-by-copy.md
```diff
@@ -0,0 +1,35 @@
+# Prefer immutable array methods to generate new arrays over modifying original array objects
```
sindresorhus, line 1:
```suggestion
# Prefer immutable array methods over modifying original arrays
```

docs/rules/prefer-change-array-by-copy.md
```diff
@@ -0,0 +1,35 @@
...
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+`Array#sort()`, `Array#reverse()`, `Array#splice()`, and `Array#[index]` modify the original array, which can lead to unexpected behavior and bugs. To avoid these issues, prefer using immutable methods like `Array#toSorted()`, `Array#toReversed()`, `Array#toSpliced()`, and `Array#with()` to generate new arrays without altering the original one.
+
+## Examples
+
+```js
+const newArray = [3,2,1].sort(); // ❌
+const newArray = [3,2,1].toSorted(); // ✅
+
```
sindresorhus, line 17: Separate code blocks for each example.

docs/rules/prefer-change-array-by-copy.md
```diff
@@ -0,0 +1,35 @@
...
+Methods like `Array#sort()`, `Array#reverse()`, `Array#splice()`, and direct assignment (`Array#[index]`) modify the original array, potentially causing unexpected behavior. To avoid these issues, prefer using immutable methods like `Array#toSorted()`, `Array#toReversed()`, `Array#toSpliced()`, and `Array#with()` to generate new arrays without altering the original.
+
+## Examples
+
+```js
+const newArray = [3,2,1].sort(); // ❌
+const newArray = [3,2,1].toSorted(); // ✅
+
+const newArray = new Array({file: 10}).splice(0, 2); // ❌
+const newArray = new Array({file: 10}).toSpliced(0, 2); // ✅
```
sindresorhus, line 19: This is invalid code.

docs/rules/prefer-change-array-by-copy.md
```diff
@@ -0,0 +1,35 @@
...
+const newArray = array.toSpliced(start, deleteCount); // ✅
+
+array[0] = 'changed'; // ❌
+const newArray = array.with(0, 'changed'); // ✅
+
+// When the number of parameters does not match, these methods will be treated as user-defined methods instead of array methods.
+// For example, the following are not reported by this rule
+array.sort('There', 'is', 'more', 'than', 'one', 'parameter'); // ✅
+array.reverse('This should not have parameter'); // ✅
+array.splice(0, 1, "This have more then three parameter"); // ✅
```
sindresorhus, line 34: Not sure there is any point in documenting this. It's a given that we should try the hardest not to have false-positives.

sindresorhus (changes requested):

sindresorhus:
How about an opt-in option to enforce always using immutable methods, regardless of the context? This could be useful for those who prefer a functional programming style.

docs/rules/prefer-change-array-by-copy.md
axetroy: Indeed, users should not pay attention to these details

axetroy:
CI fail on linting current codebase, Should I fix them in this PR, or in another separate?

```
 /home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/rules/prefer-string-slice.js
Error:   108:4  error  Prefer `Array#toReversed()` over `Array#reverse()`  unicorn/prefer-change-array-by-copy

/home/runner/work/eslint-plugin-unicorn/eslint-plugin-unicorn/rules/prevent-abbreviations.js
Error:    97:38  error  Prefer `Array#toSorted()` over `Array#sort()`     unicorn/prefer-change-array-by-copy
Error:   148:5   error  Prefer `Array#toSpliced()` over `Array#splice()`  unicorn/prefer-change-array-by-copy

✖ 3 problems (3 errors, 0 warnings)

Error: Process completed with exit code 1.
```

axetroy:
Block, Since node 18.x is not support `Change Array By Copy` APIs. But this version still needs to be tested in CI.

`npm run run-rules-on-codebase` is always fail.

fregante:
Node 18 EOL is on 30 Apr 2025, does this PR have to wait 8 months? I think it's easier to disable this specific rule in `eslint.dogfooding.config.mjs`

axetroy:
> Node 18 EOL is on 30 Apr 2025, does this PR have to wait 8 months? I think it's easier to disable this specific rule in `eslint.dogfooding.config.mjs`

It sounds good, and it is not an auto fix, but a suggestion.

However, users need to identify whether the current environment supports `Change Array By Copy`

fregante:
> If it is autofixable

It can be autofixable in a very small subset of cases. I don’t think it's safe to even try unless usage can be easily determined. 

The only truly safe autofix is when there's an array literal in the current expression, like `return [...someOtherArray].reverse()`.

In case of `return someOtherArray.reverse()`, you'd have to safely determine that `someOtherArray` is not an argument:

```js
function x(someOtherArray) {
   return someOtherArray.reverse(); // Can't autofix
}
```

It's not used in callbacks:

```js
function x() {
   const someOtherArray = []
   setTimeout(() => {
       console.log(someOtherArray)
   }, 1)
   return someOtherArray.reverse(); // Can't autofix
}
```


It's not used after the `.reverse()` call:

```js
function x() {
   const someOtherArray = []
   try {
      return someOtherArray.reverse(); // Can't autofix
   } finally {
      console.log(someOtherArray)
   }
}
```

It's not used in the outer scope:

```js
const someOtherArray = []
function x() {
   return someOtherArray.reverse(); // Can't autofix
}
```


etc. So it's likely easier to just _not_ autofix here.

axetroy:
I prefer not to auto-repair at all. After all, code like this (`[...someOtherArray].reverse()`) is rare.

However, due to compatibility issues, should we set this rule as `recommended: false`?

fisker:
I hope we have a separate rule to forbid `.reverse()`, but allow `ExpressionStatement`.

```js
// Only exception
array.reverse(); // ✅

[...array].reverse(); // ❌
array.reverse()[0];  // ❌
```

Same for `.sort()` but two rules.

fregante:
What was missing here?

punkpeye:
Has anyone found an alternative?

PR #2430, opened 2024-08-20
Add `no-identical-alternatives-in-ternary` rule

### LINKED ISSUE #2153: Rule proposal:  `no-identical-alternate`
### Description

Enfore clean code.

### Fail

```js
a ? b ? c : 1 : 1
```

### Pass

```js
a && b ? c  : 1
```

### Additional Info

_No response_

axetroy:
Close #2153


docs/rules/no-identical-alternate.md
```diff
@@ -0,0 +1,22 @@
+# Disallows nested ternary expressions with repeated alternatives
```
sindresorhus, line 1:
```suggestion
# Disallow nested ternary expressions with repeated alternatives
```

sindresorhus:
I think the rule name should include `ternary`. So maybe `no-identical-alternatives-in-ternary`?

sindresorhus (changes requested):

rules/no-identical-alternate.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const {isSameReference} = require('./utils/index.js');
+
+const MESSAGE_ID = 'no-identical-alternate';
+const messages = {
+	[MESSAGE_ID]: 'Replace nested ternary expressions with a logical expression.',
```
sindresorhus, line 7: Should be more specific about what kind of ternary expression it targets.

axetroy: How about `Replace repeated alternatives of ternary expressions with a logical expression.` or `Simplify repeated alternatives of ternary expressions.` ?

axetroy:
> I think the rule name should include `ternary`. So maybe `no-identical-alternatives-in-ternary`?

I'm not very good with rule naming, but I think this is a better name

rules/no-identical-alternatives-in-ternary.js
```diff
@@ -0,0 +1,47 @@
...
+const messages = {
+	[MESSAGE_ID]: 'Replace repeated alternatives of ternary expressions with a logical expression.',
+};
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	/** @param {import('estree').ConditionalExpression} node */
+	ConditionalExpression(node) {
+		// Check if the alternate is a ConditionalExpression
+		if (node.alternate && node.consequent.type === 'ConditionalExpression') {
```
fisker, line 15: I think we should go deeper, 

```js
a ? (b ? 1 : 2) : (d ? 1 : 3);
```

```js
a ? (b ? 1 : 2) : (d ? 3 : 1)
```

should also be reported. I'm not sure about the equivalent... but there must be a better way to write?

axetroy: This is too complicated, I don't think it can be done easily.

axetroy: I think we should ignore this situation, it is almost impossible to happen in the real world.

Or create another rule that names `no-complex ternary` to disallow it.

axetroy:
If it needs to keep the comments, needs a test:

```js
a /** comment */ ? /** comment */ b : /** comment */ c ? 1 : 1
```

rules/no-identical-alternatives-in-ternary.js
```diff
@@ -0,0 +1,47 @@
...
+			const {test: innerTest, consequent: innerConsequent, alternate: innerAlternate} = node.consequent;
+
+			const {sourceCode} = context;
+
+			if (isSameReference(outerAlternate, innerAlternate)) {
+				context.report({
+					node,
+					messageId: MESSAGE_ID,
+					fix(fixer) {
+						const newExpression = `${sourceCode.getText(outerTest)} && ${sourceCode.getText(innerTest)} ? ${sourceCode.getText(innerConsequent)} : ${sourceCode.getText(outerAlternate)}`;
```
axetroy, line 26: TODO: Here will drop the comment

We should change to another replacement implementation

axetroy:
I have reimplemented it. It is supports multiple lines and complex comments retention.

Now it's ready for review @fisker

rules/utils/conditional-expression.js
```diff
@@ -0,0 +1,36 @@
...
+const {getParenthesizedRange} = require('./parentheses.js');
+
+/**
+Find the index of the question mark in a ConditionalExpression node
+
+@param {import('estree').ConditionalExpression} node
+@param {import('eslint').SourceCode} sourceCode
+@returns {number}
+*/
+function findIndexOfQuestionMarkInConditionalExpression(node, sourceCode) {
```
fisker, line 12:
```js
sourceCode.getTokenAfter(node.test, () => ...isQuestionMarkToken)
```

should enough

test/no-identical-alternatives-in-ternary.mjs
```diff
@@ -0,0 +1,75 @@
...
+
+const {test} = getTester(import.meta);
+
+test.snapshot({
+	valid: [
+		'a ? b ? c : 1 : 2',
+		'a && b ? c : 1',
+	],
+	invalid: [
+		'a?b?c:1:1',
```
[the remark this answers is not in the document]
axetroy: `a ? b ? 1: c :1` fix to `a && b ? 1 : c`

`a ? 1: b ? c : 1` fix to `a || !b ? 1 : c`

`a ? 1 : (b ? 1 : c)` fix to `(a || b) ? 1 : c`

This implementation is too complicated

axetroy:
Close because it is too complex to implement edge cases perfectly.

PR #2432, opened 2024-08-21
Add `prefer-math-min-max`

### LINKED ISSUE #2373: Rule proposal: `prefer-math-min-max`
Could there be a rule for fixing the use of ternaries instead of using `Math.min` and `Math.max` where appropriate?

Example of **_incorrect_** usage
```ts
const height = 30;
const value = height > 50 ? 50 : height;
```

Example of **_correct_** usage, using the Math API
```ts
const height = 30;
const value = Math.min(height, 50);
```

same example for `Math.max` with flipped operator sign

axetroy:
Close #2373


docs/rules/prefer-math-min-max.md
```diff
@@ -0,0 +1,56 @@
+# Prefer `Math.min()` and `Math.max()` over ternary expressions for simple comparisons
```
sindresorhus, line 1:
```suggestion
# Prefer `Math.min()` and `Math.max()` over ternaries for simple comparisons
```

docs/rules/prefer-math-min-max.md
```diff
@@ -0,0 +1,56 @@
...
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+Enforce the use of `Math.min()` and `Math.max()` instead of ternary expressions for simple comparisons. This makes the code more readable.
```
sindresorhus, line 10: This is repeating the title.

sindresorhus (changes requested):

rules/prefer-math-min-max.js
```diff
@@ -0,0 +1,95 @@
...
+
+		const {sourceCode} = context;
+		const {operator, left, right} = test;
+
+		const leftCode = sourceCode.getText(left);
+		const rightCode = sourceCode.getText(right);
+		const alternateCode = sourceCode.getText(alternate);
+		const consequentCode = sourceCode.getText(consequent);
+
+		switch (operator) {
```
fisker, line 45: I believe this can be simplify to just two branch, can you try?

fisker:
Two edge cases will be broken

1. need parens

```js
(0,foo) > 10 ? 10 : (0,foo)
```

2. need insert space before

```js
function a() {
  return+foo > 10 ? 10 : +foo
}
```

rules/prefer-math-min-max.js
```diff
@@ -0,0 +1,95 @@
...
+*/
+function reportPreferMathMinOrMax(context, node, left, right, method) {
+	const {sourceCode} = context;
+
+	context.report({
+		node,
+		messageId: MESSAGE_ID,
+		data: {
+			replacement: `${method}()`,
+			value: sourceCode.getText(node),
```
fisker, line 21: There is no `value` placeholder in message.

rules/prefer-math-min-max.js
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+@param {import('eslint').Rule.RuleContext} context
+@param {import('estree').ConditionalExpression} node
+@param {string} method
+*/
+function reportPreferMathMinOrMax(context, node, left, right, method) {
+	const {sourceCode} = context;
+
+	context.report({
```
fisker, line 16: Please use `return` or `yield` to avoid `content.report` in all your PRs.

axetroy:
> Two edge cases will be broken
> 
>     1. need parens
> 
> 
> ```js
> (0,foo) > 10 ? 10 : (0,foo)
> ```
> 
>     2. need insert space before
> 
> 
> ```js
> function a() {
>   return+foo > 10 ? 10 : +foo
> }
> ```

Good catch

rules/prefer-math-min-max.js
axetroy: I'm not sure why return a report is better than `context.report({})`, since I see it used elsewhere.

And function calls have intelligence with LSP, which is less error-prone than returning a value.

Anyway, update in 7cd3d39d86adde4186f5a2ae2a14fc222fbec36e

rules/prefer-math-min-max.js
axetroy: 7948804fc2d6d905ce92427ea8f2a3983d70ea1a

axetroy:
> Two edge cases will be broken
> 
>     1. need parens
> 
> 
> ```js
> (0,foo) > 10 ? 10 : (0,foo)
> ```
> 
>     2. need insert space before
> 
> 
> ```js
> function a() {
>   return+foo > 10 ? 10 : +foo
> }
> ```

Resolve the edge case in 87815e94088d0667e8faa1bf7d655a4d5baa8188 and 48de8a4ce8815b353b0f8fdf6b42886702d98937

rules/prefer-math-min-max.js
axetroy: dc5aedcb0e35ad6345541fd7ccbb7f19d173bb97

fisker: There are still 4 branches, the if condition can be merged.

rules/prefer-math-min-max.js
```diff
@@ -0,0 +1,109 @@
...
+		},
+		* fix(fixer) {
+			/**
+			 * ```js
+			 * function a() {
+			 *   return+foo > 10 ? 10 : +foo
+			 * }
+			 * ```
+			 */
+			if (node.parent.type === 'ReturnStatement' && node.parent.argument === node && node.parent.start + 'return'.length === node.start) {
```
fisker, line 32: Use this instead https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1e367bbe6f9e5a135fd7476b18df714e26928c4a/rules/fix/fix-space-around-keywords.js#L12

axetroy: f5afe1e12a478305043c09b3e6fcb473aeec070e

rules/prefer-math-min-max.js
axetroy: These four branches are definitely necessary and cannot be eliminated

axetroy:
Code coverage 100%, reached

rules/prefer-math-min-max.js
axetroy: Updated in 9bdef777c5997c9bedf92e38fe139434adf9b57d

rules/prefer-math-min-max.js
```diff
@@ -0,0 +1,90 @@
...
+		node,
+		messageId: MESSAGE_ID,
+		data: {replacement: `${method}()`},
+		/** @param {import('eslint').Rule.RuleFixer} fixer */
+		* fix(fixer) {
+			/** @type {{parent: import('estree'.Node)}} */
+			const {parent} = node;
+			if (
+				// Catch edge case: `return+foo > 10 ? 10 : +foo`
+				(parent.type === 'ReturnStatement' && parent.argument === node && parent.start + 'return'.length === node.start)
```
fisker, line 32: We should remove the AST check here, `fixSpaceAroundKeyword` handles more cases.

fisker (approved):
Good job

yoyo837:
`prefer-math-min-max` does not consider Implicit Type Conversion.

```js
col.width > 0 ? col.width : 0
```
conver to
```
Math.max(col.width, 0)
```

It broke the logic.

sindresorhus:
It's impossible to account for that without types. And I don't think it should account for that regardless. Implicit type conversion is an anti-pattern. If you want to do it, I suggest disabling the rule.

Issue #1063, opened 2021-01-24
Lack of clear alternative for reduceRight in unicorn/no-array-reduce


harrysarson:
I do not wish to re-litigate #623 but I am here as the rule triggered for `Array#reduceRight` after updating xo. 

For `Array#reduce` there is a clear alternative which is to use a `for (const ... of array) { ... }` loop.

However, there is no clean way to iterate over an Array backwards, the best I can come up with is `for (let i = array.length - 1; i >= 0; i--) { const ... = array[i]; ... }` but in my (admittedly subjective) opinion having to use indicies is worse than the complexity introduced by reverseRight.

Rule: unicorn/no-array-reduce

The code in question looks a bit like:

```js
const listEmpty = null;

const listCons = (element, list) => ({
  head: element,
  tail: list,
});

const listFromArray = (array) =>
  array.reduceRight((list, element) => listCons(element, list), listEmpty);
```

And I think that (the best alternative that I can think of)

```js
const listFromArray = (array) => {
    let list = listEmpty;
    for (let i = array.length - 1; i >= 0; i--) {
        const element = array[i];
        list = listCons(element, list);
    }
    return list;
}
```

is worse because of how easy it is to make an off-by-one error with the indexing logic.

---

I am happy to (and have) used a splash of `eslint-disable-next-line` so happy for this to be closed but I thought it is worth sharing.

Of course it is possible that there is a much nicer way to iterated backwards over an array that I have missed. If there is it would be great to add it to the docs for unicorn/no-array-reduce which currently do not show an alternative for `reduceRight`.

papb:
What about:

```js
const listFromArray = (array) => {
  let list = listEmpty;
  for (const element of array.reverse()) {
    list = listCons(element, list);
  }
  return list;
}
```

What do you think?

harrysarson:
The reverse method is nice but I don't have the motivation to benchmark it (even v8 isn't that smart is it?).

I think its easier to add the eslint-disable than the helper function in my case (my case is unusually function programming style for js so is probably uniquely suited to a reduce) but I agree it is probably the best of both worlds solution. Would a PR adding it to the docs be welcome?

fisker:
👎 for `.reverse()`, it reverse the array, you don't even know the current order if you plan to reuse the array.

[...]

PR #2457, opened 2024-09-29
Document `reduceRight` alternative for `no-array-reduce`

fregante:
- For https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1063


fisker (approved):

rules/no-array-reduce.js
```diff
@@ -4,7 +4,7 @@ const {isNodeValueNotFunction, isArrayPrototypeProperty} = require('./utils/inde
 
 const MESSAGE_ID = 'no-reduce';
 const messages = {
-	[MESSAGE_ID]: '`Array#{{method}}()` is not allowed',
+	[MESSAGE_ID]: '`Array#{{method}}()` is not allowed. Prefer other types of loop for readability.',
```
fregante, line 7: I’ll alter this message to specifically mention .toReversed() for reduceRight. https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2428 isn’t out yet so it’s possible that they will use `.reverse()` instead.

rules/no-array-reduce.js
```diff
@@ -104,8 +106,7 @@ const create = context => {
 				const methodNode = getMethodNode(callExpression);
 				yield {
 					node: methodNode,
-					messageId: MESSAGE_ID,
-					data: {method: methodNode.name},
+					messageId: methodNode.name,
```
fregante, line 109: Pattern seen in https://github.com/sindresorhus/eslint-plugin-unicorn/blob/1558cbea81bf4476ee27d80724155d991433ce49/rules/prefer-string-slice.js#L141

sindresorhus (approved):

PR #2250, opened 2024-01-04
Update ESLint to v9, drop support for Node.js v16

fisker:

PR #2461, opened 2024-09-30
Fix: missing ESLint v9 peer dependency

rtritto:
Continue of #2250


rtritto:
`>=8.56.0` includes `^9.0.0`

PR #2462, opened 2024-09-30
Fix: use explicit versions of eslint peer dependency

rtritto:
`>=8.56.0` allows version 9 and future versions of `eslint` (10, 11 etc) and it doesn't advise with warnings of the package managers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
 		"yaml": "^2.5.1"
 	},
 	"peerDependencies": {
-		"eslint": ">=8.56.0"
+		"eslint": "^8.56.0 || ^9.0.0"
 	},
 	"ava": {
 		"files": [


fisker:
> it doesn't advise with warnings of the package managers.

Which one?

rtritto:
> > it doesn't advise with warnings of the package managers.
> 
> Which one?

With `>=8.56.0`.

E.g. with yarn berry and the next version of `eslint`:

`yarn add -D eslint@10.0.0 eslint-plugin-unicorn`
OR
`yarn add -D eslint@10.0.0`
`yarn add -D eslint-plugin-unicorn`
OR
`yarn add -D eslint-plugin-unicorn`
`yarn add -D eslint@10.0.0`

Results:
- With `>=8.56.0`, no warning on logs
- With `^8.56.0 || ^9.0.0`, you get a warning on logs:
  ```console
  ➤ YN0060: │ eslint is listed by your project with version 10.0.0 (<DEP_HASH>), which doesn't satisfy what eslint-plugin-unicorn and other dependencies request (^8.56.0 || ^9.0.0).
  ```

fisker:
Oh, you mean it won't warn if ESLint release v10?

rtritto:
Correct, it doesn't warn if `eslint-plugin-unicorn` isn't compatible with `eslint` v10

fisker:
Sorry that I misunderstood. I'm not sure what's the best practice, but I think it's not a bad idea to assume it will work for the future versions.

rtritto:
All ESLint plugins use the explicit logic for the eslint version.
It's a bad practice hide unsupported plugin version to ESLint (peer dependency check).

sindresorhus:
We will test when ESLint 10 is out and if it's not compatible, we'll add this then.

rtritto:
FYI @fregante

fregante:
Other than a [2023 article](https://eslint.org/blog/2023/10/flat-config-rollout-plans/#eslintrc-removed-in-eslint-v10.0.0), I can't find any references to eslint v10 even on the repo.

Presuming that this plugin has already been updated for v9 and stopped using deprecated features, I think it won't have issues with v10. I think v10 will just remove any APIs that v9 deprecated (e.g. eslintrc support)

rtritto:
ESLint v10 is only an example.
With `>=8.56.0` implies that `eslint-plugin-unicorn` can support future version of 8 (9, 10, 11 etc).
So, after upgrading Eslint with a major version, no alert is showed on package manager logs (mismatch in peer dependencies). This behavior can break ESLint.

With `^8.56.0 || ^9.0.0` you can assert that `eslint-plugin-unicorn` supports only >=8.56.0 and <10.0.0 versions.

It's a best practise that is used everywhere (e.g. plugins) for peer dependencies.

fregante:
But that brings zero advantages and all disadvantages. 

If you install v10 and it doesn’t work, then it doesn’t work. This change won’t change anything. 

If you install v10 and it works, then this change means that you *cannot* install v10 until this plugin is updated, which means you have to wait possibly days for that.

PR #2467, opened 2024-10-04
`prefer-math-min-max`: Ignore BigInt

### LINKED ISSUE #2465: Bug: `prefer-math-min-max` reports false-positive for bigint
the new rule `prefer-math-min-max` should ignore bigints

```js
const code = 420n > 69n ? 420n : 69n; // should be ok
```

chirokas:
Close #2465


fisker:
Maybe we should only report when there is a number, since strings, `Date`s can also be compared.

fisker:
@axetroy What do you think?

axetroy:
> Maybe we should only report when there is a number, since strings, `Date`s can also be compared.

Make sense.

> Ensure that everything, be it an Identifier, Literal, or any other expression, is a Number.

But I think we should only need to determine whether one of the left or right is a number (Identifier/Literal/Expression).

```js
const isNumber = require("./utils/is-number.js");

// ...

const { sourceCode } = context;
const scope = sourceCode.getScope(conditionalExpression);
const isValidExpression = [left, right].some((node) => isNumber(node, scope));

if (!isValidExpression) {
  return;
}


```

fisker:
The logic is good, but it makes this rule almost useless.. since it's rare that both left and right matches `isNumber`.

fisker:
Maybe we can skip this check with `strictTypes: false` (default: `true`) option? So this rule can be useful sometimes.

[`TypeTracker` from `eslint-plugin-regexp`](https://github.com/ota-meshi/eslint-plugin-regexp/blob/aa8857f0cda5afb35081f2c2a64cc36174018e46/lib/utils/type-tracker/index.ts) can be useful, but it's quiet big.

@sindresorhus Not sure what we should do here.

sindresorhus:
> The logic is good, but it makes this rule almost useless.. since it's rare that both left and right matches isNumber.

Not sure I understand. We only need to check if one of the sides is a number, right?

sindresorhus:
> TypeTracker from eslint-plugin-regexp can be useful, but it's quit big.

If we were to do that, it would be better to use TypeScrip as its type inference is much more reliable.

axetroy:
> But I think we should only need to determine whether one of the left or right is a number
> We only need to check if one of the sides is a number, right?

Just check one of them to make sure it is a comparison of numbers, not strings.

sindresorhus:
> Just check one of them to make sure it is a comparison of numbers, not strings.

Yes

fisker:
> We only need to check if one of the sides is a number, right?

I thought the same, https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2467#issuecomment-2394841085

Turns out I was wrong


```
10n > 0 ? 10n : 0
10n
Math.max(0, 10n)
VM762:1 Uncaught TypeError: Cannot convert a BigInt value to a number

new Date > 0 ? new Date : 0
Sat Oct 12 2024 18:11:59 GMT+0800 (中国标准时间)
Math.max(0, new Date)
1728727926364

'100' > 0 ? '100' : 0
'100'
Math.max(0, '100')
100
```

sindresorhus:
> Turns out I was wrong

That is awful code and not something we should care about. We should be pragmatic.

fisker (approved):

PR #2472, opened 2024-10-07
Update package.json - added homapage url

piwysocki:
Added missing homepage url.

with this property we could see url in VSCode :

obraz


sindresorhus:
npm infers this from the repo already.

PR #2475, opened 2024-10-10
Refactor: remove `is-builtin-module` dep

zanminkian:

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,6 @@
 		"esquery": "^1.6.0",
 		"globals": "^15.9.0",
 		"indent-string": "^4.0.0",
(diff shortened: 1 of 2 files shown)


fregante (approved):
👍 CI is failing for external issues (vs code syntax; lack of token)

sindresorhus:
The built-in method matches on a list of APIs that depends on which Node.js version the linter is running on. This can created inconsistent results, meaning it can pass locally, but fail in CI, etc. I think it's better to just wait and then update to https://github.com/sindresorhus/is-builtin-module v4 when we can require ESM, as it uses a static list.

PR #2476, opened 2024-10-11
`prefer-at`: Remove unsafe autofix for `.slice` with 1 argument

k-yle:
It's not safe to change `array.slice(-n)[0]` to `array.at(-n)`. This code is not equivalent for arrays with only 1 item. It's only equivalent for `.slice(-1)`

 Proof:
```js
testCases = [ 'a', 'ab', 'abc', 'abcd' ];
testCases.map(array => ({
    original: array.slice(-2)[0],
    autofixed: array.at(-2),
}));

// output:
[ 
  {original: 'a', autofixed: undefined}, // 🔴 this is different
  {original: 'a', autofixed: 'a'},
  {original: 'b', autofixed: 'b'},
  {original: 'c', autofixed: 'c'},
]
```


rules/prefer-at.js
```diff
@@ -96,9 +96,12 @@ function checkSliceCall(node) {
...
 		if (
-			firstElementGetMethod === 'zero-index'
-			|| firstElementGetMethod === 'shift'
-			|| (startIndex === -1 && firstElementGetMethod === 'pop')
+			startIndexNode.argument.value === 1
+			&& (
+				firstElementGetMethod === 'zero-index'
+				|| firstElementGetMethod === 'shift'
+				|| (startIndex === -1 && firstElementGetMethod === 'pop')
+			)
```
fisker, line 104: This is incorrect, the value of `startIndexNode` doesn't matter. It only result differently when the index is out of the boundary of string. And sting don't have `.shift`/`.pop`, so it doesn't matter.

k-yle: @fisker we're checking `startIndexNode` because `.slice(-1)[0]` can still be safely converted to `.at(-1)`, but only for `-1`, not any other number. For any other number the behaviour is different if the index is out of bounds

fisker: You are right.

But i think we should still report `array.slice(-9)[0]`, just remove the autofix.

sindresorhus:
Bump :)

rules/prefer-at.js
k-yle: > … we should still report `array.slice(-9)[0]` …

Are you sure we should intentionally raise a false positive error? If someone blindly accepted this suggestion they could introduce a bug which is quite hard to discover....

fisker: Okay, let's skip, it's a rare case I guess.

fisker:
@k-yle Can you  fix conflicts?

fisker (approved):

k-yle:
@fisker done

PR #2485, opened 2024-10-20
Update snapshots

fregante:
Following https://github.com/sindresorhus/eslint-plugin-unicorn/commit/fe4e7bcfa6d427f4abc4e94446a553f0a68b1fa1


package.json
```diff
@@ -22,6 +22,7 @@
 		"fix": "run-p --continue-on-error fix:*",
 		"fix:eslint-docs": "eslint-doc-generator",
 		"fix:js": "npm run lint:js -- --fix",
+		"fix:snapshots": "ava --update-snapshots",
```
fregante, line 25: I'm also including the snapshots in the `fix` command, which I thought was enough initially but it did not update the snapshots

fregante:
@fisker vscode's integration test has been failing for a few days, but the affected haven't changed in years.

https://github.com/microsoft/vscode/blob/main/src/vs/workbench/services/search/test/node/fixtures/examples/NullPoinderException.js#L10

fisker (approved):

fisker:
> but the affected haven't changed in years.

I'm not sure how the parser changed, but the typescript parser is super forgiving, it allow many kind of invalid syntax, it's been problem for years in Prettier to handle all  unexpected AST, until  last year they starts to add extra checks for invalid syntaxes. [My  PRs](https://github.com/typescript-eslint/typescript-eslint/commits?author=fisker)

fisker:
Ah, I believe the parser is correctly treat is as module now, it enbles strict mode, which also request by me.. https://github.com/typescript-eslint/typescript-eslint/issues/9101

PR #2495, opened 2024-11-15
`no-array-for-each`: Add `ignoreNamedIdentifier` option

azat-io:
The `no-array-for-each` rule is good for making code easier to understand and shorter. However, in some cases I find it easier to use `forEach` when I have a function that is applied via `forEach` in multiple places.

This PR adds the ability to ignore the rule if only the function name is passed in the arguments to `forEach`.

Example:
https://github.com/azat-io/eslint-plugin-perfectionist/blob/main/rules/sort-classes.ts#L341-L374

```ts
if ('elements' in nodeValue) {
  nodeValue.elements
    .filter(currentNode => currentNode !== null)
    .forEach(traverseNode)
}

if ('argument' in nodeValue && nodeValue.argument) {
  traverseNode(nodeValue.argument)
}

if ('arguments' in nodeValue) {
  nodeValue.arguments.forEach(traverseNode)
}

if ('declarations' in nodeValue) {
  nodeValue.declarations.forEach(traverseNode)
}

if ('properties' in nodeValue) {
  nodeValue.properties.forEach(traverseNode)
}

if ('expressions' in nodeValue) {
  nodeValue.expressions.forEach(traverseNode)
}
```

### DIFF
--- a/docs/rules/no-array-for-each.md
+++ b/docs/rules/no-array-for-each.md
@@ -61,3 +61,17 @@ for (const [index, element] of array.entries()) {
 	bar(element, index, array);
 }
 ```
+
+## Options
+
+### ignoreNamedIdentifier
+
+Type: `boolean`\
+Default: `false`
+
+Ignores the rule when the callback is function identifier.
+
+```js
+// Pass
+array.forEach(bar);
+```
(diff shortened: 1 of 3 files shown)


sindresorhus:
Thanks for the PR, but I don't think this is something we want. Passing a function directly into a callback is generally an anti-pattern. We even have a rule against it: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-callback-reference.md

azat-io:
Sad, thank you, I hear you!

PR #2501, opened 2024-11-21
`prefer-global-this`: Fix auto-fix for `typeof window`

### LINKED ISSUE #2468: `prefer-global-this` should ignore `typeof window === 'undefined'`
Hey thanks for the plugin, it's so useful !

I recently updated to v56 and was surprised about the autofix

 It's very common to have

```js
const isServer = typeof window === 'undefined';
const isBrowser = typeof window !== 'undefined'
```

It fixes to

```js
const isServer = typeof globalThis === 'undefined';
const isBrowser = typeof globalThis !== 'undefined';
```

Which isn't valid

Interested to know your advices.

PS: Thanks to @fregant comment below: `window' in globalThis`

axetroy:
fix #2468


sindresorhus:
I think it should emit an error, just not auto-fix. Or the check could auto-fix to `'window' in globalThis`.

axetroy:
Auto fix to `'window' in globalThis` is not safe. Maybe we should fix to `typeof globalThis.window`?

sindresorhus:
How is it not safe?

axetroy:
> How is it not safe?

As #2468 describes:

### There's a BREAKING CHANGE with auto-fix

```diff
- const isBrowser = typeof window !== 'undefined'
+ const isBrowser = typeof globalThis !== 'undefined'
```

It breaks the original behavior.

Now, this patch fixes it to

```diff
- const isBrowser = typeof window !== 'undefined'
+ const isBrowser = typeof globalThis.window !== 'undefined'
```

It is consistent with the original behavior

sindresorhus:
Lets go with `1`.

axetroy:
> Lets go with `1`.

This is exactly what this PR does.

@fisker Could you review it?

TrevorBurnham:
Using `typeof globalThis.window` instead of `typeof window` affects bundle sizes in Next.js, and likely other popular JS frameworks. I've spun up a new ticket for discussion: #2605

PR #2512, opened 2024-12-16
Add `prefer-class-fields` rule

### LINKED ISSUE #314: Rule proposal: `prefer-class-fields`
[Issuehunt badges](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/314)

There is a superior pattern for class field declarations that can declutter constructors (often making them totally unnecessary.) The feature is still considered experimental but it works out of the box with babel.

Fail:

```js
class Foo {
  constructor() {
    this.foo = 'foo';
  }
}
```

Pass:

```js
class Foo {
  foo = 'foo';
}
```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $30.00)

- [fregante fregante](https://oss.issuehunt.io/u/fregante) ($30.00)

### Submitted pull Requests
- [#2512 Add `prefer-class-fields` rule](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/pull/2512)
---

### Tips

- Checkout the [Issuehunt explorer](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://oss.issuehunt.io/r/new) on IssueHunt to raise funds.
</details>

FRSgit:
Adds `prefer-class-fields` rule which turns:

```js
class Foo {
  constructor() {
    this.foo = 'foo';
  }
}
```

into:

```js
class Foo {
  foo = 'foo';
}
```

Fixes #314

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#314: Rule proposal: `prefer-class-fields`](https://oss.issuehunt.io/repos/55832243/issues/314)
---
</details>


rules/prefer-class-fields.js
```diff
@@ -0,0 +1,116 @@
...
+};
+
+/**
+ * @param {import('eslint').Rule.Node} node
+ * @returns {node is import('estree').ExpressionStatement & {expression: import('estree').AssignmentExpression & {left: import('estree').MemberExpression & {object: import('estree').ThisExpression}}}}
+ */
+const isThisAssignmentExpression = node => {
+	if (
+		node.type !== 'ExpressionStatement'
+		|| node.expression.type !== 'AssignmentExpression'
```
fisker, line 17: The operator didn't checked

```js
			class Foo {
				constructor() {
					this.foo += 'foo';
				}
			}
```

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,116 @@
...
+	) {
+		return false;
+	}
+
+	const lhs = node.expression.left;
+
+	if (!lhs.object || lhs.object.type !== 'ThisExpression') {
```
fisker, line 24: The left property didn't checked

```js
			class Foo {
				constructor() {
					this[foo.bar] = 'foo';
				}
			}
```

fisker:
What should we do if the assignment is unreachable?

```js
			class Foo {
				constructor(x) {
					if (x) {return}
					this.foo = 'foo';
				}
			}
```

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,116 @@
...
+						node,
+						messageId: MESSAGE_ID,
+
+						/** @param {import('eslint').Rule.RuleFixer} fixer */
+						* fix(fixer) {
+							yield removeThisFieldAssignment(node, sourceCode, fixer);
+							yield fixer.insertTextAfterRange(
+								classBodyStartRange,
+								`\n${indent}${node.expression.left.property.name} = ${node.expression.right.raw};`,
+							);
```
fisker, line 93: The fix also changes code if the property already exists.

```js
			class Foo {
				foo = 'bar';
				constructor() {
					this.foo = 'foo';
				}
			}
```

FRSgit: True. IMO in this case the output should be:
```
class Foo {
    foo = 'foo';
    constructor() {
    }
}
```

docs/rules/prefer-class-fields.md
```diff
@@ -0,0 +1,41 @@
...
+
+This rule will enforce the use of field declarations over `this` assignments in constructors for static values.
+
+> To avoid leaving empty constructors after autofix use [`no-useless-constructor` rule](https://eslint.org/docs/latest/rules/no-useless-constructor).
+
+## Fail
+
+```js
+class Foo {
+  constructor() {
```
sindresorhus, line 18: tab-indentation

docs/rules/prefer-class-fields.md
```diff
@@ -0,0 +1,41 @@
+# Prefer class field declarations over assigning static values in constructor using `this`
```
sindresorhus, line 1:
```suggestion
# Prefer class field declarations over `this` assignments in constructor
```

sindresorhus (changes requested):

fregante:
> What should we do if the assignment is unreachable?

It's dynamically set, so it cannot be statically defined. It should not be reported by this rule.

rules/prefer-class-fields.js
FRSgit: You're right, it should be ignored. Fixed here: 21e3cf1a172d1834bb67bbeabe39215cd0d3830e

rules/prefer-class-fields.js
FRSgit: I think in this case we should ignore any kind of reporting - `foo.bar` might be dynamic and checking if it is or not is a rabbit hole with potentially lots of nested jumps.

docs/rules/prefer-class-fields.md
FRSgit: 378f58bf45ea56b930cba9b48f7fbac71e4a2131

rules/prefer-class-fields.js
FRSgit: Added a valid test case to cover the dynamic case: 633829c7daf29bd071a2ed371bbee6de5f28cd9a

FRSgit: And here: 0e6afb3a27ad2de8ccb5dc8cc7dcb2e89261863c added support for simple strings & template strings.

FRSgit:
Hey @fregante!
Thanks for your input about the linting - is there a ruleset somewhere which I could use? The `lint` task in this repo is not doing this much (at least for me) and it removes some of your changes 😄 It seems that the part `/** @type {import('eslint').Rule.RuleModule} */` is required by the `lint` task. After your change it was failing with:
image
So, I've reverted this one change.

Or maybe we can just agree that once the work in this PR is done you'll maybe just run your linter on it once?

rules/prefer-class-fields.js
fisker: I didn't mean to ask support for strings, please test

```js
			class Foo {
				constructor() {
					this[foo] = 'foo';
				}
			}
```

rules/prefer-class-fields.js
FRSgit: done: a7edb371f78b157570e5c96d6d901c6a653dc8b0

rules/prefer-class-fields.js
FRSgit: Don't worry, I didn't understood it well then :)
Done: f9caf1b8fdf8f85b87c2036023b6bf2e7972e476

FRSgit:
Handled stopping of the static analysis on unsupported case: 3376845a24b906cc6d5fec99524b99570397136c

FRSgit:
Okay, I fixed all test/linting errors connected with my rule - should I fix other as well? They seems to be connected with some default options missing in already existing rules?
cc: @sindresorhus @fregante

FRSgit:
Hey! I've merged last changes from main to the branch of this PR - now it's ready to be reviewed again

docs/rules/prefer-class-fields.md
FRSgit: Applied this change to readme.md and rules' markdown file - it seems that `prefer-class-fields.md` was getting updated using content of the `readme.md` by `fix` npm script.
Now it's okay.

docs/rules/prefer-class-fields.md
```diff
@@ -0,0 +1,41 @@
...
+
+```js
+class Foo {
+	foo = 'foo';
+}
+
+class MyError extends Error {
+	name = 'MyError'
+}
+```
```
sindresorhus, line 41: Can you use this style for the examples?

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/4e539b4b084ce06165a0fdbc02d309ce7cf75c82/docs/rules/consistent-existence-index-check.md?plain=1#L16-L36

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,182 @@
...
+};
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description:
+				'Prefer class field declarations over `this` assignments in constructor',
```
sindresorhus, line 175:
```suggestion
			description: 'Prefer class field declarations over `this` assignments in constructors.',
```

sindresorhus:
Change the test file extension from `.mjs` to `.js`.

docs/rules/prefer-class-fields.md
FRSgit: Sure, sorry for that - previously I've looked to [prefer-switch docs](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-switch.md) for inspiration.
Change done in: e839988df76908e9935e782884401bf9d7b3503e

FRSgit:
@sindresorhus done: c7632d9c5e23595882c7ed286bea6eaca0af4312

sindresorhus (approved):

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,182 @@
...
+					return {
+						node,
+						messageId: MESSAGE_ID,
+
+						/**
+						@param {import('eslint').Rule.RuleFixer} fixer
+						*/
+						* fix(fixer) {
+							yield removeFieldAssignment(node, sourceCode, fixer);
+							yield addOrReplaceClassFieldDeclaration(
```
fisker, line 151: Let's not remove code, especially there can be side effects. We can use suggestions or simply just report error, if property already exists.

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,182 @@
...
+
+			const constructorBody = constructor.value.body?.body;
+
+			if (!constructorBody) {
+				return;
+			}
+
+			const firstInvalidProperty = constructorBody.findIndex(
+				node => node.type !== 'ExpressionStatement',
+			);
```
fisker, line 123: Please help me understand this, are you assuming assignments before non-`ExpressionStatement`s are reachable? And it's addressing https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2512#issuecomment-2547309562 ?

rules/prefer-class-fields.js
FRSgit: My thinking was: property in the constructor always takes priority over class field declaration, so there are no side-effects. IMO it's a safe assumption.
WDYT?

fisker:
```js
const getFoo = () => console.log('called')

new class {
  foo = getFoo();  <-- this function call is removed.
  constructor() {
     this.foo = 1
  }
}
```

rules/prefer-class-fields.js
fisker: I'm not sure about this, but I don't have a better idea. 

Anyway, let's allow `EmptyStatement` too. since in no-semi style, it's common to put leading `;` at start.

rules/prefer-class-fields.js
FRSgit: You're right, I've rewritten it as a suggestion 3153a8922da7aa79d28680302dfc186b560bd6e0

rules/prefer-class-fields.js
FRSgit: done 7863af915754a4ccf6fc5689ad365c5409eea3ed

FRSgit:
ready to be re-reviewed @fisker

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,204 @@
...
+const create = context => {
+	const {sourceCode} = context;
+
+	return {
+		ClassBody(classBody) {
+			const constructor = classBody.body.find(x => x.kind === 'constructor');
+
+			if (!constructor || constructor.type !== 'MethodDefinition') {
+				return;
+			}
```
fisker, line 106:
```js
			const constructor = classBody.body.find(x => x.kind === 'constructor' && constructor.type !== 'MethodDefinition');

			if (!constructor) {
				return;
			}
```

Does TypeScript allow multiple `constructor`? If so, it will find the correct `constructor`

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,204 @@
...
+
+			const firstInvalidProperty = constructorBody.findIndex(
+				node => !WHITELIST_NODES_PRECEDING_THIS_ASSIGNMENT.has(node.type),
+			);
+			const validConstructorProperties
+				= firstInvalidProperty === -1
+					? constructorBody
+					: constructorBody.slice(0, firstInvalidProperty);
```
fisker, line 120: We already use a loop, we can use the index directly, no need slice here.

rules/prefer-class-fields.js
```diff
@@ -0,0 +1,204 @@
...
+							messageId: MESSAGE_ID_SUGGESTION,
+							data: {
+								propertyName,
+								// Class expression does not have name, e.g. const a = class {}
+								className: classBody.parent?.id?.name ?? '',
+							},
+							/**
+							@param {import('eslint').Rule.RuleFixer} fixer
+							*/
+							* suggest(fixer) {
```
fisker, line 154: `suggest` should be an `array`.

rules/prefer-class-fields.js
FRSgit: Typescript allows multiple constructor declarations (overloads) but only one definition (implementation). [See an example here](https://stackoverflow.com/a/12702786/8805801)

But I don't think I understand what do you mean here - can you elaborate a bit more?

fisker: Example from your link

```js
class Box {
    public x: number;
    public y: number;
    public height: number;
    public width: number;

    constructor();                   // <- Your version will give up when find this
    constructor(obj: IBox); 
    constructor(obj?: IBox) {        // <- My version can locate this
        this.x = obj?.x ?? 0;
        this.y = obj?.y ?? 0;
        this.height = obj?.height ?? 0;
        this.width = obj?.width ?? 0;
    }   
}
```


Sorry, my version has a typo

```diff
-			const constructor = classBody.body.find(node => node.kind === 'constructor' && node.type !== 'MethodDefinition');
+			const constructor = classBody.body.find(node => node.kind === 'constructor' && node.type === 'MethodDefinition');

			if (!constructor) {
				return;
			}

```

sindresorhus:
@FRSgit Bump :)

sindresorhus:
Closing for lack of activity.

FRSgit:
@sindresorhus Can you reopen this PR please? I've just pushed changes

fisker:
@FRSgit I pushed some changes

- Make code easier to understand
- Strictly check `computed`, `static` for finding existing property (previously, it incorrectly considered a static/computed property as existing)
- Insert property at end instead of before `constructor`, safer.
- Play safer by only checking the first expression (currently only ignores `EmptyStatement`)

  Take the existing example

  ```
	class MyError extends Error {
		constructor(message: string) {
			super(message);
			this.name = 'MyError';
		}
	}
  ```

  We can't know what `super(message)` do, it can change `name`, the original will override what `super()` do, but if we use a property, it will override by super class, this is rarely but still possible. 

  Not sure how to improve this part.

  Potential issue if we decide to ignore more expressions, the current `removeFieldAssignment` need deal with the ASI problem.

  for example, if we decide to ignore `super()` call.

  ```js
  class Foo {
    constructor() {
      super()  
      this.bar = 1
      ;[a]
    }
  }


---

@FRSgit This rule should work for private fields, do you think you can continue?

fisker:
> It looks like both ways: defining value in constructor and directly as a class field override what is set by parent constructor.

Sorry, you are right about this.

Anyway, it's just an wrong example, I mean other nodes can change the result.

```
class A {
	constructor() {
		call(this);
		this.foo = 1;
	}
}
```

is not safe to fix to

```
class A {
	constructor() {
		call(this);
	}
	foo = 1
}
```

Am I right?

fisker:
Since it's not easy to say what can be skipped, I'm not sure what we can do about these totally unrelated expressions like this:

```js
class A {
	constructor() {
		const x = 2; // Can skip, but can't skip `const x = call()`
		this.foo = 1;
	}
}
```

FRSgit:
@fisker Yes, you're right with the example using function call - I bet there are many other cases that might potentially break unless rules' autofix is pretty restrictive.

IMO to not over complicate, we should allow only EmptyExpressions and other `this.foo = ` declarations to precede lines rule will autofix. In any other case - we print a suggestion and user must resolve by themselves 🤷 

Ofc, there are probably other cases where autofix could be applied just like that, but that's something for next iterations. When somebody comes with a proposition for improvement of this algo in the future.

fisker:
Can you make it work with private fields?

FRSgit:
@fisker Right, forgot about that!
Here you go: a6648b45a8f39c98e7fdb2e380d88b52e459d292

PR #2523, opened 2025-01-10
Add `no-instanceof-builtin-object` rule

### LINKED ISSUE #723: Rule proposal: `prefer-error-iserror`
[Disallow `instanceof Error` checks](https://github.com/sindresorhus/execa/issues/422#issuecomment-625750262).

## Fail

```js
const isError = foo instanceof Error;
```

## Pass

```js
const isError = Object.prototype.toString.call(foo) === '[object Error]'
```

### LINKED ISSUE #2157: Rule proposal: `no-instanceof-function`
### Description

similar to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-instanceof-array.md

### Fail

```js
foo instanceof Function
```

### Pass

```js
typeof foo === 'function'
```

### Additional Info

_No response_

### LINKED ISSUE #2452: Rule proposal: `no-instanceof-builtin-object`
### Description

`instanceof` operator in JavaScript has some [limitations](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms).

To detect the type of a variable, we should encourage user to use: 
- `typeof` operator
```js
typeof foo === 'string' // detect whether foo is string
```
- `node:util/types` module (for Node)
```js
import {isNumberObject, isMap} from 'node:util/types';
isNumberObject(new Number(123)) // true
isNumberObject(123) // false
isMap(new Map()) // true
```
- `@sindresorhus/is` module (for browser)
```js
import is from '@sindresorhus/is';
is(new Map()) === 'Map' // true
```

Therefore, builtin objects like `Map`/`Promise`/`Error` can be detected by `typeof`, `node:util/types` and `@sindresorhus/is`. Should not use `instanceof`.

Rule `no-instanceof-array` can be removed if this rule is implemented.

### Fail

```js
foo instanceof Array
foo instanceof Map
foo instanceof Promise
```

### Pass

```js
import types from 'node:util/types'
Array.isArray(foo)
types.isMap(foo)
types.isPromise(foo)
```

```js
import is from '@sindresorhus/is'
is(foo) === 'Array'
is(foo) === 'Map'
is(foo) === 'Promise'
```

### Proposed rule name

no-instanceof-builtin-object

### Additional Info

_No response_

axetroy:
Close #2452
Close #2157 
Close #723

This PR includes the content of `no-instanceof-array`

`no-instanceof-array` should be retired in another PR.

------

The fix is not 100% safe, e.g.

```js
// The input
"abc" instanceof String // false
// Fixed to
typeof "abc" === "string" // true
```


rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,123 @@
...
+						if (name === 'Array') {
+							return fixer.replaceText(
+								node,
+								`Array.isArray(${sourceCode.getText(node.left)})`,
+							);
+						}
+
+						return fixer.replaceText(
+							node,
+							// TODO: Because quotes are used here, this may cause exceptions in some templates, such as Vue.
```
axetroy, line 100: @zanminkian Can you offer some advice?

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,123 @@
...
+					messageId: MESSAGE_ID,
+					fix(fixer) {
+						const instanceofToken = sourceCode.getTokenAfter(node.left);
+
+						const text = sourceCode.getText(node);
+
+						const instanceofBeforeText = text.slice(node.left.end - node.start, instanceofToken.start - node.start);
+						const instanceofAfterText = text.slice(instanceofToken.end - node.start, node.right.start - node.start);
+
+						return fixer.replaceText(
```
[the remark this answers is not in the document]
axetroy: >  So, I personally do not think we should fix it. Just fix it for some cases like foo instanceof Array

This make sense.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,123 @@
+'use strict';
+const {checkVueTemplate} = require('./utils/rule.js');
+
+const MESSAGE_ID = 'no-instanceof-builtin-object';
+const messages = {
+	[MESSAGE_ID]: 'Unsafe instanceof should not be used to check type.',
+};
+
+const primitiveTypes = new Set(['String', 'Number', 'Boolean', 'BigInt', 'Symbol']);
+const referenceTypes = new Set([
```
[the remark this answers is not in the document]
axetroy: > `Object.keys(globals.builtin).filter(key => /^[A-Z]/.test(key))` ?

This doesn't seem like a good approach, for example, `NaN`, `JSON` are not constructors

axetroy:
> Currently, there are 3 cases we can fix in this PR.
> 
>     * foo instanceof Array -> Array.isArray(foo)
> 
>     * foo instanceof Function -> typeof foo === 'function'
> 
>     * foo instanceof Object -> type foo === 'object' && foo !== null

It's not safe

```diff
<template>
-  <div v-if="foo instanceof Function">bar</div>
+  <div v-if="typeof Function === "function"">bar</div>
</template>
```

rules/no-instanceof-builtin-object.js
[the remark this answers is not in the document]
fisker: I suggest hard code here, relay on runtime behavior is not a good idea.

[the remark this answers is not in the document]
fisker: Anyway, some data already exits, for example shared/typed-array.js

rules/catch-error-name.js
```diff
@@ -41,7 +41,7 @@ const create = context => {
 	};
 	const {name: expectedName} = options;
 	const ignore = options.ignore.map(
-		pattern => pattern instanceof RegExp ? pattern : new RegExp(pattern, 'u'),
+		pattern => Object.prototype.toString.call(pattern) === '[object RegExp]' ? pattern : new RegExp(pattern, 'u'),
```
fisker, line 44: I don't think this is improvement here.

test/snapshots/no-instanceof-builtin-object.mjs.md
```diff
@@ -0,0 +1,719 @@
...
+> Input
+
+    `␊
+      1 | function foo(){return[]instanceof Array}␊
+    `
+
+> Output
+
+    `␊
+      1 | function foo(){returnArray.isArray([])}␊
```
fisker, line 471: Broken

test/snapshots/no-instanceof-builtin-object.mjs.md
```diff
@@ -0,0 +1,719 @@
...
+      33 |␊
+      34 | 	// comment␊
+    `
+
+> Output
+
+    `␊
+      1 | Array.isArray(foo)␊
+      2 |␊
+      3 | 	// comment␊
```
fisker, line 527: Expected https://github.com/sindresorhus/eslint-plugin-unicorn/blob/e42118dc2739a48cdc6a0a245c5a3d6b85450bcd/test/snapshots/no-instanceof-array.mjs.md?plain=1#L216

fisker:
Need a options to allow fix to use [`Error.isError`](https://github.com/tc39/proposal-is-error?tab=readme-ov-file#erroriserror)

axetroy:
> Need an options to allow fix to use [`Error.isError`](https://github.com/tc39/proposal-is-error?tab=readme-ov-file&rgh-link-date=2025-01-15T15%3A55%3A17Z#erroriserror)

Should we wait until it reaches stage 4 before modifying the rules?

After all, no browser has implemented it yet (Firefox/Chrome)

rules/catch-error-name.js
axetroy: So, should we ignore the rule in this line?

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,102 @@
...
+		}
+
+		const {name} = node.right;
+
+		if (builtinConstructors.has(name)) {
+			if (name === 'Array') {
+				context.report({
+					node,
+					messageId: MESSAGE_ID,
+					fix: fixer => fixer.replaceText(node, `Array.isArray(${context.sourceCode.getText(node.left)})`),
```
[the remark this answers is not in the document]
axetroy: Now aligned with the `no-instanceof-array` rule

test/snapshots/no-instanceof-builtin-object.mjs.md
axetroy: > Now aligned with the `no-instanceof-array` rule

Fixed

test/snapshots/no-instanceof-builtin-object.mjs.md
```diff
@@ -0,0 +1,719 @@
...
+> Input
+
+    `␊
+      1 | (0, array) instanceof Array␊
+    `
+
+> Output
+
+    `␊
+      1 | Array.isArray(0, array)␊
```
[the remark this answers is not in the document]
axetroy: > Now aligned with the `no-instanceof-array` rule

Fixed

test/snapshots/no-instanceof-builtin-object.mjs.md
axetroy: > Now aligned with the `no-instanceof-array` rule

Fixed

axetroy:
> Need an options to allow fix to use [`Error.isError`](https://github.com/tc39/proposal-is-error?tab=readme-ov-file&rgh-link-date=2025-01-15T15%3A55%3A17Z#erroriserror)

Already implemented in f5d907edcfb8ee9cb2ccb205812c4b0a84d6f5e9

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,174 @@
...
+			}
+		},
+	};
+};
+
+const schema = [
+	{
+		type: 'object',
+		properties: {
+			shippedProposals: {
```
fisker, line 152: Something like `useErrorIsError: true` is easier to understand.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,174 @@
...
+			const {left, right} = node;
+
+			let tokenStore = sourceCode;
+			let instanceofToken = tokenStore.getTokenAfter(left, isInstanceofToken);
+			if (!instanceofToken && sourceCode.parserServices.getTemplateBodyTokenStore) {
+				tokenStore = sourceCode.parserServices.getTemplateBodyTokenStore();
+				instanceofToken = tokenStore.getTokenAfter(left, isInstanceofToken);
+			}
+
+			if (primitiveConstructors.has(right.name) || right.name === 'Function') {
```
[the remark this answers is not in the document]
axetroy: I am not a fan of this, but anyway, updated.

rules/no-instanceof-builtin-object.js
axetroy: This is based on [Babel's parameter](https://babeljs.io/docs/babel-preset-env#shippedproposals) design

If others believe this parameter is unsuitable, we can consider using an alternative instead.

fisker: > once Error.isError is officially incorporated into ECMAScript, the option useErrorIsError: true will no longer be necessary.

We did this a lot.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,187 @@
...
+	'BigUint64Array',
+	'Float32Array',
+	'Float64Array',
+	'Int16Array',
+	'Int32Array',
+	'Int8Array',
+	'Uint16Array',
+	'Uint32Array',
+	'Uint8Array',
+	'Uint8ClampedArray',
```
fisker, line 46: Please import these from shared/typed-array.js

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,187 @@
...
+const referenceConstructors = new Set([
+	// Error types
+	'AggregateError',
+	'Error',
+	'EvalError',
+	'RangeError',
+	'ReferenceError',
+	'SyntaxError',
+	'TypeError',
+	'URIError',
```
fisker, line 24: Please extract them and share with https://github.com/sindresorhus/eslint-plugin-unicorn/blob/4e539b4b084ce06165a0fdbc02d309ce7cf75c82/rules/error-message.js#L15

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,187 @@
...
+
+						yield * replaceNodeOrTokenAndSpacesBefore(instanceofToken, '', fixer, sourceCode, tokenStore);
+						yield * replaceNodeOrTokenAndSpacesBefore(right, '', fixer, sourceCode, tokenStore);
+					},
+				});
+
+				return;
+			}
+
+			if (right.name === 'Error' && shippedProposals) {
```
fisker, line 137: Please merge this branch with the `Array` one.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,187 @@
...
+
+						yield * replaceNodeOrTokenAndSpacesBefore(instanceofToken, '', fixer, sourceCode, tokenStore);
+						yield * replaceNodeOrTokenAndSpacesBefore(right, '', fixer, sourceCode, tokenStore);
+					},
+				});
+
+				return;
+			}
+
+			context.report({node, messageId: MESSAGE_ID});
```
fisker, line 156: Normally, we use this pattern.

```
const problem = ...

if (...) {
  problem.fix = ...
}

return problem
```

fisker:
To be honest, I don't think it's a good idea to forbidden all those instanceof check, especially in most cases we don't really need care about care about across realms.

But every one should agree that these are bad

- `instanceof Array` (`Error` too, once it's widely supportted)
- `instanceof Function`
- `instanceof <primitive wrapper>`

I don't want people end up has to disable the rule completely.

I suggest,

option 1) make two rules, one for the cases above, one for the aggressive one。
option 2）one rule, but only check cases above and configable to check more cases.

WTDY? @sindresorhus

docs/rules/no-instanceof-builtin-object.md
```diff
@@ -0,0 +1,58 @@
+# Disallow instanceof on built-in objects
```
sindresorhus, line 1:
```suggestion
# Disallow `instanceof` on built-in objects
```

rules/no-instanceof-builtin-object.js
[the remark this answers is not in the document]
sindresorhus: Guard makes it more readable.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,154 @@
...
+	},
+];
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create: checkVueTemplate(create),
+	meta: {
+		type: 'problem',
+		docs: {
+			description: 'Disallow instanceof on built-in objects',
```
sindresorhus, line 146:
```suggestion
			description: 'Disallow `instanceof` on built-in objects.',
```

sindresorhus (changes requested):

sindresorhus:
> option 2）one rule, but only check cases above and configable to check more cases.

This 👍

axetroy:
> option 2）one rule, but only check cases above and configable to check more cases.

Implemented by add some options:

#### `strategy`: 'loose' | 'strict'

loose: check `instanceof <primitive wrapper>` and `typeof Function` and `typeof Array`
strict: check all built-in constructor

#### include: Array<string>

specify the constructor to include

#### exclude: Array<string>

specify the constructor to exclude

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,201 @@
...
+	},
+];
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create: checkVueTemplate(create),
+	meta: {
+		type: 'problem',
+		docs: {
+			description: 'Disallow `instanceof` on built-in objects',
```
sindresorhus, line 205:
```suggestion
			description: 'Disallow `instanceof` with built-in objects.',
```

docs/rules/no-instanceof-builtin-object.md
```diff
@@ -0,0 +1,134 @@
+# Disallow `instanceof` on built-in objects
```
sindresorhus, line 1:
```suggestion
# Disallow `instanceof` with built-in objects
```

sindresorhus (changes requested):

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,219 @@
+const {checkVueTemplate} = require('./utils/rule.js');
+const {getParenthesizedRange} = require('./utils/parentheses.js');
+const {replaceNodeOrTokenAndSpacesBefore, fixSpaceAroundKeyword} = require('./fix/index.js');
+const {builtinErrors} = require('./error-message.js');
+const typedArray = require('./shared/typed-array.js');
```
sindresorhus, line 5: You can use `import` now.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,219 @@
...
+import {getParenthesizedRange} from './utils/parentheses.js';
+import {replaceNodeOrTokenAndSpacesBefore, fixSpaceAroundKeyword} from './fix/index.js';
+import {builtinErrors} from './error-message.js';
+import typedArray from './shared/typed-array.js';
+
+const isInstanceofToken = token => token.value === 'instanceof' && token.type === 'Keyword';
+
+const MESSAGE_ID = 'no-instanceof-builtin-object';
+const messages = {
+	[MESSAGE_ID]: 'Unsafe instanceof should not be used to check type.',
```
sindresorhus, line 11:
```suggestion
	[MESSAGE_ID]: 'Avoid using `instanceof` for type checking as it can lead to unreliable results.',
```

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,219 @@
...
+	'Proxy',
+
+	// Other
+	'DataView',
+	'Date',
+	'SharedArrayBuffer',
+	'FinalizationRegistry',
+]);
+
+const fixByReplaceMethod = function * ({fixer, node, sourceCode, tokenStore, instanceofToken}, method) {
```
fisker, line 56: `replaceWithFunctionCall` should be a better name.

docs/rules/no-instanceof-builtin-object.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+### useErrorIsError
+
+Type: `boolean`\
+Default: `false`
+
+Specifies using [`Error.isError()`](https://github.com/tc39/proposal-is-error) to determine whether it is an error object.
+
+This option requires `strategy: 'strict'`.
```
fisker, line 129: I don't think it should be under `strategy: 'strict'`.

rules/no-instanceof-builtin-object.js
```diff
@@ -0,0 +1,219 @@
...
+
+	const range = getParenthesizedRange(left, tokenStore);
+	yield fixer.insertTextBeforeRange(range, functionCall + '(');
+	yield fixer.insertTextAfterRange(range, ')');
+
+	yield * replaceNodeOrTokenAndSpacesBefore(instanceofToken, '', fixer, sourceCode, tokenStore);
+	yield * replaceNodeOrTokenAndSpacesBefore(right, '', fixer, sourceCode, tokenStore);
+};
+
+const replaceWithTypeOfStatement = function * ({fixer, node, sourceCode, tokenStore, instanceofToken}) {
```
fisker, line 69: `typeof` is not a statement, `replaceWithTypeOfExpression` or just `replaceWithTypeOf`

fisker (approved):
Good job!

fisker:
Follow up PR to remove `no-instanceof-array` #2534

PR #2525, opened 2025-01-15
Add `no-accessor-recursion` rule

### LINKED ISSUE #1052: Rule proposal: `no-accessor-recursion`
Accessing a property inside an accessor with the same name will cause infinite recursion.

## Fail

```js
get foo() {
    return this.foo;
}

set foo(v) {
    this.foo = v;
}
```

## Pass

```js
get foo() {
    return this._foo;
}

set foo(v) {
    this._foo = v;
}

get bar() {
    return that.bar;
}

set bar(v) {
    that.bar = v;
}
```

axetroy:
Close #1052


docs/rules/no-accessor-recursion.md
```diff
@@ -0,0 +1,74 @@
+# Disallow recursive access to this within getter and setter
```
sindresorhus, line 1:
```suggestion
# Disallow recursive access to `this` within getters and setters
```

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,101 @@
...
+	};
+};
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+	create,
+	meta: {
+		type: 'problem',
+		docs: {
+			description: 'Disallow recursive access to this within getter and setter',
```
sindresorhus, line 95:
```suggestion
			description: 'Disallow recursive access to `this` within getters and setters.',
```

sindresorhus:
You need to regenerate the snapshots `npx ava -u`.

axetroy:
@fisker 
CI is failing, and it doesn't seem to be related to the newly added rule, am I missing something?
I have updated the snapshot with `npx ava -u test/no-accessor-recursion.js`

sindresorhus (approved):

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,99 @@
...
+				while (scope.type !== 'function' || isArrowFunctionScope(scope)) {
+					scope = scope.upper;
+				}
+
+				// Check if `this` is in the current function expression scope
+				if (scope.block === property.value) {
+					/** @type {import('estree').MemberExpression} */
+					const {parent} = node;
+
+					const isThisAccessed = () => parent.type === 'MemberExpression' && parent.property.type === 'Identifier' && parent.property.name === property.key.name;
```
fisker, line 58: `computed` didn't checked.

axetroy: Good catch 👍

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,99 @@
...
+/** @param {import('eslint').Scope.Scope} scope */
+const isArrowFunctionScope = scope => scope.type === 'function' && scope.block.type === 'ArrowFunctionExpression';
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => {
+	const functionExpressionStack = [];
+
+	return {
+		// Getter for object literal
+		'Property[value.type="FunctionExpression"], [kind="get"]'(node) {
```
fisker, line 15: Did you mean `Property[value.type="FunctionExpression"][kind="get"]`?

axetroy: Yeah, But this should also be a legal selector syntax, just with a different style

```
attribute conditions: [attr!="foo"], [attr>2], [attr<3], [attr>=2], or [attr<=3]
```

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,99 @@
...
+		},
+		// Setter for class
+		'MethodDefinition[kind="set"]'(node) {
+			functionExpressionStack.push(node);
+		},
+		'MethodDefinition[kind="set"]:exit'() {
+			functionExpressionStack.pop();
+		},
+		ThisExpression(node) {
+			if (functionExpressionStack.length > 0) {
```
fisker, line 43:
```js
const property = functionExpressionStack.at(-1);

if (!property) {
  return;
}
```

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,99 @@
...
+				while (scope.type !== 'function' || isArrowFunctionScope(scope)) {
+					scope = scope.upper;
+				}
+
+				// Check if `this` is in the current function expression scope
+				if (scope.block === property.value) {
+					/** @type {import('estree').MemberExpression} */
+					const {parent} = node;
+
+					const isThisAccessed = () => parent.type === 'MemberExpression' && !parent.computed && parent.property.type === 'Identifier' && parent.property.name === property.key.name;
```
fisker, line 58: Let's move `parent.type === 'MemberExpression' && !parent.computed && parent.property.type === 'Identifier'` check to the top for perf.

rules/no-accessor-recursion.js
axetroy: > If so, shouldn't Property[kind="get"] enough?

Make sense.

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,99 @@
...
+						case 'get': {
+							if (isThisAccessed()) {
+								return {node: parent, messageId: MESSAGE_ID_ERROR};
+							}
+
+							break;
+						}
+
+						case 'set': {
+							if (isThisAccessed() && parent.parent.type === 'AssignmentExpression') {
```
fisker, line 70: Need make sure the memberExpression is on left

```
a = this.foo
``

File: rules/no-accessor-recursion.js (hunk 4 above)
Comment (fisker) [reply rules/no-accessor-recursion.js:15]: > Yeah, But this should also be a legal selector syntax, just with a different style
> 
> ```
> attribute conditions: [attr!="foo"], [attr>2], [attr<3], [attr>=2], or [attr<=3]
> ```

No, it's not. 

```
Property[value.type="FunctionExpression"], [kind="get"]
```

means


```
Property[value.type="FunctionExpression"],
*[kind="get"]
```

File: rules/no-accessor-recursion.js [hunk 8]
```diff
@@ -0,0 +1,97 @@
...
+		/** @param {import('estree').ThisExpression} node */
+		ThisExpression(node) {
+			/** @type {import('estree').Property | import('estree').MethodDefinition} */
+			const property = functionExpressionStack.at(-1);
+
+			if (!property) {
+				return;
+			}
+
+			const scope = getClosestFunctionScope(sourceCode, node);
```
fisker, line 57: Nice refactoring.

Now we have implemented this functionality, do we still need the functionExpressionStack? we can check `scope.block.parent` directly instead?

axetroy: Yes, this is indeed better.

axetroy:
There is an edge case, Private fields are not handled properly

```js
class Foo{
	get bar() {
		return this.#bar;
	}

	get #bar() {
		return this.#bar
	}
}
```

fisker:
Missed `computed` check  in property, fixed in [`818e194` (#2525)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2525/commits/818e194736383fc74763e3e1a900d64af8cdd79b)

fisker:
> ```js
> class Foo{
> 	get bar() {
> 		return this.#bar;
> 	}
> 
> 	get #bar() {
> 		return this.#bar
> 	}
> }
> ```

Yeah, should check `PrivateIdentifier`. Good catch.

fisker:
Class static accessors seems don't need special logic, but please add tests

```js
class Foo{
	static get bar() {
		return this.bar;
	} 
}
```

fisker:
There are missing cases for setters, please give a shoot, if it's hard to fix we can ignore.


```js
class Foo{
	set bar() {
		[this.bar] = array;
	}
}
```

```js
class Foo{
	set bar() {
		({propery: this.bar} = array);
	}
}
```

```js
class Foo{
	set bar() {
		Object.assign(this, {bar: 1})
	}
}
```

But this case for getters is common, we should report

```js
class Foo{
	get bar() {
		const {bar} = this;
	}
}
```

test/snapshots/no-accessor-recursion.js.md
```diff
@@ -226,3 +226,49 @@ Generated by [AVA](https://avajs.dev).
...
+      9 | }␊
+    `
+
+> Error 1/2
+
+    `␊
+      1 | class Foo{␊
+      2 | 	get bar() {␊
+    > 3 | 		return this.#bar;␊
+        | 		       ^^^^^^^^^ Disallow recursive access to \`this\` within getters and setters.␊
```
fisker, line 252: False positive, since you didn't compare node.type.

test/no-accessor-recursion.js
```diff
@@ -109,6 +109,14 @@ test.snapshot({
...
 			};
 		`,
+		// Static getter
+		outdent`
+			class Foo{
+				static get bar() {
+					return this.bar;
+				} 
+			}
+		`,
```
fisker, line 119: Shouldn't this invalid? I only asked to test, didn't mean to ignore.

axetroy: > Shouldn't this invalid? I only asked to test, didn't mean to ignore.

Well, I thought this didn't have infinite recursion until I tried it

axetroy:
> There are missing cases for setters, please give a shoot, if it's hard to fix we can ignore.
> 
> ```js
> class Foo{
> 	set bar() {
> 		[this.bar] = array;
> 	}
> }
> ```
> 
> ```js
> class Foo{
> 	set bar() {
> 		({propery: this.bar} = object);
> 	}
> }
> ```
> 
> ```js
> class Foo{
> 	set bar() {
> 		Object.assign(this, {bar: 1})
> 	}
> }
> ```
> 
> But this case for getters is common, we should report
> 
> ```js
> class Foo{
> 	get bar() {
> 		const {bar} = this;
> 	}
> }
> ```

```js
class Foo{
	get bar() {
		const {bar} = this;
	}
}
```

The last case is already supported and tested, the other cases I don't think anyone will write that way.

And they are difficult to detect, I think we should ignore it

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,109 @@
...
+*/
+const isRecursiveMemberAccess = (parent, property) => isDotNotationAccess(parent) && isSameKey(parent.property, property.key);
+
+/**
+Check if `this` is accessed recursively within a destructuring assignment.
+
+@param {import('estree').VariableDeclarator} parent
+@param {import('estree').Property | import('estree').MethodDefinition} property
+*/
+const isRecursiveDestructuringAccess = (parent, property) => parent.type === 'VariableDeclarator' && parent.id.type === 'ObjectPattern' && parent.id.properties.some(declaratorProperty => declaratorProperty.type === 'Property' && isSameKey(declaratorProperty.key, property.key));
```
fisker, line 55: Not checking `shorthand`, and `computed`.

rules/no-accessor-recursion.js
```diff
@@ -0,0 +1,112 @@
...
+	['Property', 'MethodDefinition'].includes(property.type) && !property.computed && ['set', 'get'].includes(property.kind)
+	&& (property.key.type === 'Identifier' || property.key.type === 'PrivateIdentifier');
+
+/**
+Check if two property keys are the same.
+
+@param {import('estree').Property['key']} KeyLeft
+@param {import('estree').Property['key']} KeyRight
+*/
+const isSameKey = (KeyLeft, KeyRight) => ['type', 'name'].every(key => KeyLeft[key] === KeyRight[key]);
```
fisker, line 39: The parameter should be in camelCase.

fisker:
@axetroy Handled more cases [`a17cef9` (#2525)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2525/commits/a17cef94434b6ce62bd3473fa21e99659f2df711), please take a look.

test/no-accessor-recursion.js
```diff
@@ -240,5 +240,20 @@ test.snapshot({
 				}
 			}
 		`,
+		//
+		...[
+			'++ this.bar;',
```
axetroy, line 245: 👍

fisker (approved):

fisker (changes requested):
Do you know how `this` works in [class static initialization blocks](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Static_initialization_blocks)? Can you check if we need logic adjustments?

fisker:
Alright, I found 

> The `this` inside a static block refers to the constructor object of the class. 

So, I think we need some updates to `getClosestFunctionScope`.


```js
const foo = {
	get bar() {
		class Foo {
			static {
				this.bar // <- This should not reported.
			}
		}
	}
}
```

fisker (approved):
Good job!

PR #2526, opened 2025-01-16
`prefer-includes`: Support Vue template

### LINKED ISSUE #2508: `prefer-includes` not working in vue templates
```vue
<template>
  <div v-if="['vue', 'pinia'].indexOf(pkg) > -1" />
</template>

<script setup lang="ts">
const pkg = 'vue'

console.log(['vue', 'pinia'].indexOf(pkg) > -1)
</script>
```

I added the rule `'unicorn/prefer-includes': 'error'` to my project, which checks for and replaces `.indexOf()` with `.includes()`. While this works properly within the `<script>` block of a Vue SFC, it fails to detect issues within the `<template>` block.

I attempted to add `checkVueTemplate` to the `prefer-includes.js` rule to catch errors within the `<template>` block. However, I encountered issues with the automatic fixes not working as expected. Is there a way to resolve this problem?

axetroy:
Close #2508


fisker (approved):
Good job!

PR #2528, opened 2025-01-17
Enhance `import-style` rule with auto-fix capability

### LINKED ISSUE #963: Auto-fix for `unicorn/import-style`?
```js
import {dirname} from 'path';

const foo = dirname(bar)
```

->

```js
import path from 'path';

const foo = path.dirname(bar)
```

I think it's doable.

CasperEngl:
- Updated the `import-style` rule to include an `autoFix` option, allowing automatic fixing of import styles.
- Modified the rule's documentation to reflect the new `autoFix` feature and its default behavior.
- Added test cases to verify the functionality of the new auto-fix feature for various import scenarios.

This change improves usability by enabling automatic corrections for import style violations.

This fixes #963


rules/import-style.js
```diff
@@ -138,6 +138,7 @@ const create = context => {
 			checkDynamicImport = true,
 			checkExportFrom = false,
 			checkRequire = true,
+			autoFix = false,
```
fisker, line 141: We don't need option for this, it should always do autofix.

rules/import-style.js
```diff
@@ -183,6 +184,25 @@ const create = context => {
...
+				if (node.type === 'ImportDeclaration' && allowedImportStyles.has('namespace')) {
+					const importedNames = node.specifiers
+						.filter(s => s.type === 'ImportSpecifier' || s.type === 'ImportDefaultSpecifier')
+						.map(s => s.local.name);
+
+					if (importedNames.length > 0) {
+						const namespaceIdentifier = moduleName === 'react' ? 'React' : importedNames[0];
+						return fixer.replaceText(
+							node,
+							`import * as ${namespaceIdentifier} from "${moduleName}"`,
```
fisker, line 201: We need find all references and replace them with `namespace.x`

rules/import-style.js
CasperEngl: Got it. When I ran tests there was an error that told me autofixes were unexpected. This is my first time submitting a PR here, so I can't quite remember what caused that to happen, but I can backtrack and fix it.

rules/import-style.js
CasperEngl: I'll look into this. If you know the solution, please feel free to share it with me.

fisker: You need

1) give a name for namespace import, it should be based on package name or filename
2) find all references, and check that variable name for the module are available in their scope.[this](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/4e539b4b084ce06165a0fdbc02d309ce7cf75c82/rules/utils/avoid-capture.js#L132) utility maybe helpful.
3) Replace all references.

CasperEngl: Fixed 👍

rules/import-style.js
```diff
@@ -127,244 +128,382 @@ const defaultStyles = {
...
-			type: 'object',
-			additionalProperties: {
-				type: 'boolean',
 			},
 		},
-	},
-};
-
-/** @type {import('eslint').Rule.RuleModule} */
-const config = {
```
sindresorhus, line 359: The `create` method should not be inlined in the `config` object, for readability.

https://github.com/CasperEngl/eslint-plugin-unicorn/blob/2c6b0f7690e5a77d61f397d0d08b2cab7e6f690b/rules/catch-error-name.js#L124-L126

rules/import-style.js
```diff
@@ -127,244 +128,382 @@ const defaultStyles = {
...
+					const isValidNode = isImportDeclaration || isVariableDeclaratorWithRequireCall;
+
+					if (!isValidNode) {
+						return;
+					}
+
+					let importedNames = [];
+					if (node.type === 'ImportDeclaration') {
+						importedNames = node.specifiers
+							.filter(s => s.type === 'ImportSpecifier' || s.type === 'ImportDefaultSpecifier')
```
fisker, line 208: `s` -> `specifier`

rules/import-style.js
```diff
@@ -127,244 +128,382 @@ const defaultStyles = {
...
+							.map(s => ({
+								localName: s.local.name,
+								importedName: s.type === 'ImportDefaultSpecifier' ? 'default' : s.imported.name,
+							}));
+					} else if (node.type === 'VariableDeclarator' && node.id.type === 'ObjectPattern') {
+						importedNames = node.id.properties
+							.map(p => {
```
fisker, line 215: `p` -> `property`

rules/import-style.js
```diff
@@ -127,244 +128,382 @@ const defaultStyles = {
...
+									importedName: p.key.name,
+								};
+							});
+					}
+
+					if (importedNames.length === 0) {
+						return;
+					}
+
+					const specialCases = {
```
fisker, line 234: This should defined in outer scope.

rules/import-style.js
```diff
@@ -127,244 +128,382 @@ const defaultStyles = {
...
+						'date-fns': 'dateFns',
+						ramda: 'R',
+						rxjs: 'Rx',
+						vue: 'Vue',
+						angular: 'Angular',
+					};
+
+					const scope = sourceCode.getScope(node);
+					const namespaceIdentifier = specialCases[moduleName]
+						|| moduleName.replaceAll(/-./g, x => x[1].toUpperCase());
```
fisker, line 257: This is not enough, `moduleName` can contain characters which is not allowed in identifier, eg: `.`, `#`. We need replace them first.

And it should only infer from the last part `from "./foo/bar.js"` should be `bar`

CasperEngl: I've added some more tests to check these use cases and updated the implementation to use `getNamespaceIdentifier` which handles creation of the namespace indentifier according to these rules.

rules/import-style.js
```diff
@@ -127,7 +128,48 @@ const defaultStyles = {
...
+	const lastPart = moduleName.split('/').pop().split('.')[0];
+
+	// For scoped packages, we want the package name, not the scope
+	// @scope/package -> package
+	// @scope/foo.bar -> foo
+	const packageName = lastPart.startsWith('@') ? moduleName.split('/').pop().split('.')[0] : lastPart;
+
+	// Replace invalid identifier characters and convert to camelCase
+	return packageName
+		.replaceAll(/[^\dA-Za-z-]/g, '-') // Replace invalid chars with hyphen
```
fisker, line 168: Will not be valid identifier if filename starts With number.

CasperEngl: Nice catch. I've updated the implementation and added test cases.

CasperEngl:
I reverted a lot of changes (by using new commits), and now the diff looks much nicer. Let me know what you think!

rules/import-style.js
```diff
@@ -127,6 +128,175 @@ const defaultStyles = {
...
+	'prop-types': 'PropTypes',
+	lodash: '_',
+	'lodash-es': '_',
+	jquery: '$',
+	'styled-components': 'styled',
+	redux: 'Redux',
+	'react-redux': 'ReactRedux',
+	axios: 'Axios',
+	moment: 'moment',
+	'date-fns': 'dateFns',
```
fisker, line 145: These two seems unnessary

rules/import-style.js
```diff
@@ -127,6 +128,175 @@ const defaultStyles = {
...
+};
+
+const getNamespaceIdentifier = moduleName => {
+	// Handle special cases first
+	if (specialCases[moduleName]) {
+		return specialCases[moduleName];
+	}
+
+	// Get the last part of the path and remove extension
+	const lastPart = moduleName.split('/').pop().split('.')[0];
```
fisker, line 159: Module name can be `foo/`

rules/import-style.js
```diff
@@ -127,6 +128,175 @@ const defaultStyles = {
...
+
+	// For scoped packages, we want the package name, not the scope
+	// @scope/package -> package
+	// @scope/foo.bar -> foo
+	const packageName = lastPart.startsWith('@') ? moduleName.split('/').pop().split('.')[0] : lastPart;
```
fisker, line 164: `lastPart` already strip `@scope` part.

rules/import-style.js
```diff
@@ -127,6 +128,175 @@ const defaultStyles = {
...
+ @param {import('estree').Node} options.node - The AST node representing the import/require statement.
+ @param {import('eslint').SourceCode} options.sourceCode - The ESLint source code object.
+ @param {string} options.moduleName - The name of the module being imported.
+ @param {Set<string>} options.allowedImportStyles - Set of allowed import styles for the module.
+ @returns {(fixer: import('eslint').Rule.RuleFixer) => Array<import('eslint').Rule.Fix>|void} A function that takes a fixer and returns an array of fixes or void.
+ */
+const createFix = ({node, sourceCode, moduleName, allowedImportStyles}) => fixer => {
+	if (!allowedImportStyles.has('namespace') || allowedImportStyles.has('named')) {
+		return;
+	}
```
fisker, line 191: Some check should be outside the function.

We don't want to add an empty `fix` function.

rules/import-style.js
```diff
@@ -127,6 +128,175 @@ const defaultStyles = {
...
+			return references;
+		};
+
+		const references = getAllReferences(programScope);
+
+		return references.map(reference => fixer.replaceText(reference.identifier, `${uniqueNamespaceIdentifier}.${importedName}`),
+		);
+	});
+
+	return [importFix, ...referenceFixes];
```
fisker, line 297: Please use a generator function instead return all at the end.

rules/import-style.js
CasperEngl: Removed.

rules/import-style.js
CasperEngl: Stripped trailing slashses

rules/import-style.js
CasperEngl: Using lastPart directly now.

rules/import-style.js
CasperEngl: Done 👍

rules/import-style.js
```diff
@@ -127,6 +128,169 @@ const defaultStyles = {
...
+
+		if (hasMatchingNamedImport) {
+			return;
+		}
+
+		for (const {localName, importedName} of importedNames) {
+			// Skip rest patterns since they should be kept as is
+			if (importedName === undefined) {
+				continue;
+			}
```
fisker, line 265: Please help me understand this. Since you turn the `VariableDeclarator` into an identifier, so the `RestElement` just removed?

rules/import-style.js
```diff
@@ -127,6 +128,169 @@ const defaultStyles = {
...
+					references = [...references, ...getAllReferences(childScope)];
+				}
+
+				return references;
+			};
+
+			const references = getAllReferences(programScope);
+
+			for (const reference of references) {
+				yield fixer.replaceText(reference.identifier, `${uniqueNamespaceIdentifier}.${importedName}`);
```
fisker, line 288: This is not the current fix.

```js
import {bar} from 'foo'

a = {bar}
```

will fix to 

```js
import foo from 'foo'

a = {foo.bar} // Syntax error
```

rules/import-style.js
```diff
@@ -127,6 +128,169 @@ const defaultStyles = {
...
+
+		for (const {localName, importedName} of importedNames) {
+			// Skip rest patterns since they should be kept as is
+			if (importedName === undefined) {
+				continue;
+			}
+
+			const programScope = sourceCode.getScope(sourceCode.ast);
+
+			const getAllReferences = scope => {
```
fisker, line 269: This part should use `sourceCode.getDeclaredVariables()` to get variable from the `ImportDeclaration` and then we have all the references.

rules/import-style.js
fisker: You can use the existing `renameVariable` utility.

rules/import-style.js
```diff
@@ -127,6 +128,169 @@ const defaultStyles = {
...
+
+	let importedNames = [];
+	if (node.type === 'ImportDeclaration') {
+		importedNames = node.specifiers
+			.filter(specifier => specifier.type === 'ImportSpecifier' || specifier.type === 'ImportDefaultSpecifier')
+			.map(specifier => ({
+				localName: specifier.local.name,
+				importedName: specifier.type === 'ImportDefaultSpecifier' ? 'default' : specifier.imported.name,
+			}));
+	} else if (node.type === 'VariableDeclarator' && node.id.type === 'ObjectPattern') {
```
fisker, line 205: I think we can focus on `import`s, leave `require()` alone, ESM should be preferred.

rules/import-style.js
fisker: `consistent-assert` has similar logic, you can take a look. https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2535/files

rules/import-style.js
sindresorhus: Yes, we no longer plan to handle anything CommonJS.

sindresorhus:
@CasperEngl Are you still interested in finishing this? If not, I'm gonna close it.

PR #2531, opened 2025-01-20
Update snapshots

fisker:
https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2278#issuecomment-2600853204


sindresorhus:
Thanks for fixing this 🙏🙏

PR #2536, opened 2025-01-22
Add jsdoc `type` annotation for create function

axetroy:


sindresorhus:
If accepted, it would be great to update existing rules too.

sindresorhus (approved):

fisker:
Actually our `create` is a little different than the ESLint one, maybe better define a new one?

1. Listeners  allow return/yield problem(s)
2. context has addtional `.on` and `.onExit` methods

sindresorhus:
:arrow_up: @axetroy

axetroy:
I think we should close since it is not a standard eslint's type definition

> Listeners allow return/yield problem(s)
> context has addtional .on and .onExit methods

This type is not easy to define

PR #2537, opened 2025-01-22
`no-instanceof-builtin-object`: Rename to `no-instanceof-builtins` and use suggestions for unsafe cases

fisker:
It's not safe to fix

```
new Number instanceof Number
true
typeof new Number === 'number'
false
```

Sorry that I missed this in #2523, cc @axetroy


axetroy:
I reminded it in that PR, I thought you knew, after all, there are very few such cases.

And the package `@sindresorhus/is` is also check with `typeof` and did not consider the object

https://github.com/sindresorhus/is/blob/e0976457e04ba5df210ca0c976844b580b62f741/source/index.ts#L610-L612

axetroy:
BTW, All of primitive wrappers did not have a safe fix.

We should disable them all

#### String

```js
// ❌
'abc' instanceof String; // false
typeof 'abc' === 'string'; // true
```

#### Number

```js
// ❌
123 instanceof Number; // false
typeof 123 === 'number'; // true
```

#### Boolean

```js
// ❌
true instanceof Boolean; // false
typeof true === 'boolean'; // true
```

#### BigInt

```js
// ❌
123n instanceof BigInt; // false
typeof 123n === 'bigint'; // true
```

#### Symbol

```js
// ❌
Symbol('[[test]]') instanceof Symbol; // false
typeof Symbol('[[test]]') === 'symbol'; // true
```

fisker:
> All of primitive wrappers did not have a safe fix.

I forgot that `123 instanceof Number; // false`

axetroy (approved):

fisker:
We already have a rule named `new-for-builtins`, should we rename this rule to `no-instanceof-builtins`?

axetroy:
> We already have a rule named `new-for-builtins`, should we rename this rule to `no-instanceof-builtins`?

The name seems more consistent

sindresorhus (approved):

sindresorhus:
Merge conflict

PR #2538, opened 2025-01-23
Add `no-named-default` rule

### LINKED ISSUE #2087: Rule proposal: `no-named-default`
### Description

During refactoring, eg:

```js
// We don't need `foo` at first.
doThingsWithoutFoo();

export {default as foo} from 'foo';
```

```js
import {default as foo} from 'foo'; // Didn't mean to import like this, but changed from previous version

// We need `foo` now
doThingsWithFoo();
export {foo}
```

### Fail

```js
import {default as foo} from 'foo';
```

```js
import {default as foo, bar} from 'foo';
```

### Pass

```js
import foo from 'foo';
```

```js
import foo, {bar} from 'foo';
```

### Additional Info

Should we check export?

#### Fail

```js
export {foo as default};
```

#### Pass

```js
export default foo;
```

fisker:
Fixes #2087


test/snapshots/no-named-default.js.md
```diff
@@ -0,0 +1,497 @@
...
+
+    `␊
+      1 | import defaultExport, {default as named} from "foo";␊
+    `
+
+> Output
+
+    `␊
+      1 | import named from "foo";␊
+      2 | import defaultExport from "foo";␊
```
fisker, line 103: I choose to duplicate `import` instead of 

```js
import defaultExport from "foo";
const named = defaultExport;
```

As mentioned in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2087#issuecomment-1543530438 and https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2087#issuecomment-1543546434, because theoretically it can be a `let`, live binding is safer.

Anyway, it's just an edge case.

docs/rules/no-named-default.md
```diff
@@ -0,0 +1,70 @@
+# Disallow default export as named
```
sindresorhus, line 1:
```suggestion
# Disallow named usage of default import and export
```

docs/rules/no-named-default.md
```diff
@@ -0,0 +1,71 @@
...
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#preset-configs-eslintconfigjs).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+<!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
+
+The [`default import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`default export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#using_the_default_export) syntax exists for the default exports.
```
sindresorhus, line 10:
```suggestion
Enforces the use of the [`default import`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and [`default export`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#using_the_default_export) syntax instead of named syntax.
```

sindresorhus (changes requested):

PR #2543, opened 2025-01-24
`new-for-builtins`: Correct fix for `Date()`

fisker:
```
typeof new Date()
'object'
typeof Date()
'string'
```


test/snapshots/new-for-builtins.js.md
```diff
@@ -1388,3 +1367,129 @@ Generated by [AVA](https://avajs.dev).
...
+> Input
+
+    `␊
+      1 | const foo = Date(/*comment*/);␊
+    `
+
+> Error 1/1
+
+    `␊
+    > 1 | const foo = Date(/*comment*/);␊
```
axetroy, line 1451: Why no fix this to

```js
const foo = String(new Date(/*comment*/));
```

It should be safe

fisker: It's suggestion, see L1455

fisker: Sorry, I misunderstood, we should try not remove comments.

fisker: The correct fix seems like

```js
		function * fix(fixer) {
			yield * fixSpaceAroundKeyword(fixer, node, sourceCode);
			yield fixer.insertTextBefore(node, 'String(');
			yield fixer.insertTextAfter(node, ')');
			yield * switchCallExpressionToNewExpression(node, sourceCode, fixer, /* fixSpaces */ false);
			yield * removeArguments();
		}
```

Just not worth the effort

fisker: Maybe you wonder why `switchCallExpressionToNewExpression` is necessary, something like

```js
Array(globalThis)[0].Date(/*comment*/);
````

is not easy to fix correctly, but `switchCallExpressionToNewExpression` already considered this.

PR #2544, opened 2025-01-24
Add `consistent-date-clone` rule

fisker:


docs/rules/consistent-date-clone.md
```diff
@@ -0,0 +1,22 @@
+# Prefer pass `Date` directly to the constructor when cloning a `Date`
```
sindresorhus, line 1:
```suggestion
# Prefer passing `⁠Date` objects to the constructor when cloning
```

rules/consistent-date-clone.js
```diff
@@ -0,0 +1,50 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+const config = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Prefer pass `Date` directly to the constructor when cloning a `Date`.',
```
sindresorhus, line 42:
```suggestion
			description: 'Prefer passing `Date` directly to the constructor when cloning.',
```

fisker: You add different suggestions for the same description.

PR #2548, opened 2025-01-28
`prefer-includes`: Drop duplicated line in docs

anna-liepina:
i noticed there is duplicated line I think it is typo :)


fisker (approved):
Thank you!

PR #2549, opened 2025-02-01
`prefer-global-this`: Also exclude `outerWidth` and `outerHeight`

fregante:
I forgot to add these two.

Also it would be great to add a `main` branch protection o this repo because I almost clicked "commit to main". Admins can be exempted.

### DIFF
--- a/rules/prefer-global-this.js
+++ b/rules/prefer-global-this.js
@@ -84,6 +84,8 @@ const windowSpecificAPIs = new Set([
 	'resizeBy',
 	'innerWidth',
 	'innerHeight',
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):

sindresorhus:
> Also it would be great to add a main branch protection o this repo because I almost clicked "commit to main". Admins can be exempted.

Rules are not advanced enough to create exceptions for certain users.

fregante:
They've been for quite a while. Also admins also have a checkbox to bypass. Refined GitHub also has limitations but I can still push directly to it

PR #2554, opened 2025-02-13
Refactor`no-instanceof-builtins`: optimize implementation

zanminkian:
Improve code readability.


fisker (changes requested):
Thanks for the contribution, but I found the original implementation is pretty readable.

BTW, in this implementation, mutation on `strategyConstructors` is a bug.

sindresorhus:
I also prefer the original code.

PR #2555, opened 2025-02-14
[WIP] Add `no-this-outside-of-class` rule

### LINKED ISSUE #2450: Rule proposal: `no-this-outside-of-class`
### Description

In modern JavaScript, `this` should not appear outside of an class.

### Fail

```js
function foo () {
  this.bar;
}
function Foo(name) {
  this.name = name;
}
class Foo {
  foo() {
    function bar() {
      this.baz;
    }
  }
}
```

### Pass

```js
class Foo {
  foo() {
    this.name;
  }
}

class Foo {
  foo() {
    const bar = () => {
      this.name;
    }
  }
}
```

### Proposed rule name

no-this-outside-of-class

### Additional Info

_No response_

axetroy:
Close #2450

It rules may have many false positives in some edge case.

So before that, we need to consider all the situations


axetroy:
Since this rule is unclear and there are a lot of false positives, let's close it.

If there is a clearer rule, please let me know and I'll re-open it

Issue #2530, opened 2025-01-18
Update style for rule fail/pass examples


sindresorhus:
See https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2529

I don't want lots of pull requests, so if you want to contribute, you have to do them all.

---

Enable this again when this is done: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/41a7644260e42e03a0b3ea61bd53528e5cf2e8dd/test/package.js#L145-L147

fregante:
> you have to do them all.

With 100+ rules that would be prohibitive even just to review. Let's say at least 20/30 rules

sindresorhus:
Thanks for working on this, @stephenwade 🙏

stephenwade:
Glad I can help @sindresorhus! FYI, this issue can be un-pinned now.

PR #2559, opened 2025-02-17
Add options to `escape-case` and `number-literal-case`

### LINKED ISSUE #2078: Rule proposal: Require escape sequences to use lowercase values
### Description

The `escape-case` and `number-literal-case` rules supported uppercase values now, however is it possible to add a **lowercase option** for it?

### Fail

```js
// unicorn/escape-case
const foo = '\xA9';
const foo = '\uD834';
const foo = '\u{1D306}';
const foo = '\cA';

// number-literal-case
const foo = 0xA9;
const foo = 0xFF;
```

### Pass

```js
// unicorn/escape-case
const foo = '\xa9';
const foo = '\ud834';
const foo = '\u{1d306}';
const foo = '\ca';

// number-literal-case
const foo = 0xa9;
const foo = 0xff;
```

### Additional Info

The config can be like this:
```json
{
    "rules": {
        "unicorn/escape-case": ["error", "lowercase"],
	"unicorn/number-literal-case": ["error", "lowercase"],
    }
}
```

otomad:
Close #2078 .

- `escape-case` add an option to accept lowercase escape value.
- `number-literal-case` add options to accept lowercase hexadecimal value<wbr>~~, uppercase radix identifier and exponential notation~~.


docs/rules/escape-case.md
```diff
@@ -26,3 +26,29 @@ const foo = '\uD834';
...
+const foo = '\uD834';
+const foo = '\u{1D306}';
+const foo = '\cA';
+
+// Pass
+const foo = '\xa9';
+const foo = '\ud834';
+const foo = '\u{1d306}';
+const foo = '\ca';
+```
```
sindresorhus, line 54: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2530

docs/rules/number-literal-case.md
```diff
@@ -64,3 +64,96 @@ const foo = 0xFFn;
...
+const foo = 0xFFn;
+
+// Pass
+const foo = 0Xff;
+const foo = 0xff;
+const foo = 0Xffn;
+const foo = 0xffn;
+```
+
+### radixIdentifier
```
sindresorhus, line 111: Do we really need to expose this and `exponentialNotation` as options? I can see the use-case for `hexadecimalValue`, but this one just makes for unreadable code. At minimum, it should have a big disclaimer.

otomad: I'm not sure, I just thought maybe it would give the user fine-grained control over every option.

Although I personally will only use the `hexadecimalValue` option.

docs/rules/number-literal-case.md
```diff
@@ -17,50 +17,148 @@ Differentiating the casing of the identifier and value clearly separates them an
...
+const foo = 2e5;
 ```
 
-## Pass
+## Options
 
-```js
-const foo = 0xFF;
-```
+Type: `object`
```
fisker, line 71: I think we can also allow `'lowercase'`/`'uppercase'` here. `'lowercase'` means everything should be lowercased.

fisker: Make sense.

sindresorhus:
I prefer dropping the `radixIdentifier` and `exponentialNotation` option. Every added option adds an additional maintenance burden and I don't think we should add things *just because*. I doubt anyone will actually use them, and every user has to read, consider, and understand each option.

sindresorhus:
Yes

otomad:
Done.

sindresorhus:
A lint issue: https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/13649042331/job/38153296589

sindresorhus (approved):

fisker (approved):

PR #2570, opened 2025-02-22
`prefer-set-has`: Fix to ignore String.prototype.slice()

### LINKED ISSUE #2216: Bug: prefer-set-has false positive for strings in JS
`prefer-set-has` seems to be giving false positives in cases where a string variable has an `.includes()` method called twice:

```js
const text = "abc".slice();
//    ~~~~
// `text` should be a `Set`, and use `text.has()` to check existence or non-existence.

text.includes("a") || text.includes("b");
```

Seen in real world code: [eslint/eslint@`59845bd` (#17515)](https://github.com/eslint/eslint/pull/17515/commits/59845bdb89eaf23ec028b2309fb6ef1dc1d895da#diff-c4eca4aa73ca33a81fc6bfbeb64ebfd27e369dfbeea7e36a73fff08daa46c1b9R252-R253)

Tsuyoshi84:
fix #2216

### DIFF
--- a/rules/prefer-set-has.js
+++ b/rules/prefer-set-has.js
@@ -103,11 +103,14 @@ const create = context => ({
 					optionalMember: false,
 				})
 				// Array methods that return an array
-				|| isMethodCall(parent.init, {
-					methods: arrayMethodsReturnsArray,
-					optionalCall: false,
-					optionalMember: false,
-				})
+				|| (
+					isMethodCall(parent.init, {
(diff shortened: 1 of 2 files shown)


fisker:
I don't think it's the right solution, we'll need type information to fix it.

Tsuyoshi84:
@fisker Thank you for your feedback. Then, I'll close this PR.

PR #2579, opened 2025-03-01
Fix: import warning regression in v57

### LINKED ISSUE #2561: [v57.0.0] Json import warning on Node 18.20
Warnings showing after upgrade to v57.0.0

```js
NODE_OPTIONS='--trace-warnings' eslint .

(node:159194) ExperimentalWarning: Importing JSON modules is an experimental feature and might change at any time
    at emitExperimentalWarning (node:internal/util:274:11)
    at ModuleLoader.jsonStrategy (node:internal/modules/esm/translators:308:3)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:299:14)
```

Caused by:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c44ea2641b55a6f244f648010b372db359ff4707/index.js#L4

Can we use this?

```js
import { createRequire } from "module";
const packageJson = createRequire(import.meta.url)("./package.json");
```

Zamiell:
Fixes #2561.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,9 @@
 import createDeprecatedRules from './rules/utils/create-deprecated-rules.js';
 import flatConfigBase from './configs/flat-config-base.js';
 import rules from './rules/index.js';
-import packageJson from './package.json' with {type: 'json'};
+import {createRequire} from 'node:module';
+
+const packageJson = createRequire(import.meta.url)('./package.json');
 
 const deprecatedRules = createDeprecatedRules({
 	// {ruleId: {message: string, replacedBy: string[]}}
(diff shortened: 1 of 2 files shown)


fisker:
Sorry, we are not going to accept this, you can fiz engines field instead. Because there are dependencies use import attributes too.

Zamiell:
> Sorry, we are not going to accept this, you can fix engines field instead.

My PR includes the fix to engines field. Do you want me to just edit the PR to remove the changes in "index.js"?

> Because there are dependencies use import attributes too.

Oh, that's annoying. :( Do you know which dependencies? I can submit a PR to them first?

PR #2581, opened 2025-03-02
Update `engines.node` to match real compatibility

fregante:
Addresses https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2561#issuecomment-2676673692

Particularly, the current config suggests that v57 is compatible with Node 19, but it's not.

It doesn't address the warnings issue, which can indeed be resolved by bumping Node (I'm not seeing the warning on v20.18.3)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
 	},
 	"sideEffects": false,
 	"engines": {
-		"node": ">=18.18"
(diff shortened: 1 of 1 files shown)


sindresorhus (approved):

fregante:
Since v57 technically already had this compatibility, I think it can be released as a patch update

sindresorhus:
The next version will b v58.

PR #2586, opened 2025-03-03
Update deprecated docs

### LINKED ISSUE #2585: The import-index rule is marked as deprecated but seems to have been removed
The `import-index` is marked as deprecated in the [documentation](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/deprecated-rules.md) but it seems it has been removed in version 57.0.0:

> This rule is outdated. JavaScript modules (ESM) do not support importing a directory.

Attempting to use it will produce an error:

```
rules: {
  "unicorn/import-index": "error"
}
```

warpdesign:
The `import-index` rule has been removed between `56.0.0` and `57.0.0`: this PR clearly marks it as **removed** in the docs, and not simply _deprecated_.

I think the release where it was removed should also mention it.

Fixes #2585

### DIFF
--- a/docs/deleted-and-deprecated-rules.md
+++ b/docs/deleted-and-deprecated-rules.md
@@ -1,73 +1,79 @@
 # Deprecated/Removed Rules
 
-## import-index
+## Deprecated rules
 
-This rule was removed. JavaScript modules (ESM) do not support importing a directory.
-
-## no-array-instanceof
+### no-instanceof-array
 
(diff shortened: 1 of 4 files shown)


fregante (approved):

fisker:
https://github.com/warpdesign/eslint-plugin-unicorn/blob/update-deprecated-rules/docs/deleted-and-deprecated-rules.md#deleted-rules

@fregante @warpdesign 

Looks good?

Issue #2547, opened 2025-01-26
BUG: `text-encoding-identifier-case` fixes `<meta charset="UTF-8" />` to an invalid value


taep96:
The `text-encoding-identifier-case` rule fixes `<meta charset="UTF-8" />` to an invalid value ("utf8"). As per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#charset):
> `charset`
>
> This attribute declares the document's character encoding. If the attribute is present, its value must be an ASCII case-insensitive match for the string "utf-8", because UTF-8 is the only valid encoding for HTML5 documents. <meta> elements which declare a character encoding must be located entirely within the first 1024 bytes of the document.

The only valid values would be `UTF-8` and `utf-8`, not `utf8`

fregante:
Duplicate of https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1926

Issue #1926, opened 2022-10-05
`text-encoding-identifier-case` error in JSX; `utf8` is not valid HTML


jp-knj:
According to [WhatWG](https://html.spec.whatwg.org/multipage/forms.html#attr-form-accept-charset), the value must be an [ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive) match for "UTF-8"

please text-encoding-identifier-case of rule check
Thanks for always helping us out

**example**
This is causing conflicts in our React code (in .tsx files that are linted by eslint):
```js
<form
  id="search_keyword"
  method="get"
  action={action}
  acceptCharset="UTF-8" // get eslint error
>
  <input
    type="search"
    name="q"
    id="q"
  />
        ....
</form>
```

Pyrolistical:
Similarly `new TextDecoder("utf-8")` is another false positive as according to the [spec](https://encoding.spec.whatwg.org/#names-and-labels):
> Authors must use the [UTF-8](https://encoding.spec.whatwg.org/#utf-8) [encoding](https://encoding.spec.whatwg.org/#encoding) and must use its ([ASCII case-insensitive](https://infra.spec.whatwg.org/#ascii-case-insensitive)) "utf-8" [label](https://encoding.spec.whatwg.org/#label) to identify it.

`utf-8` is cannon.

fisker:
It's hard to check places that only allows `utf-8`, but `utf-8` works everywhere (I think), how about we switch from `utf8` to use `utf-8`?

Mentioned this problem in orie propor, https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1690#issuecomment-1033685083

timdavish:
> It's hard to check places that only allows `utf-8`, but `utf-8` works everywhere (I think), how about we switch from `utf8` to use `utf-8`?
> 
> Mentioned this problem in original proposal, [#1690 (comment)](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1690#issuecomment-1033685083)

Does it make sense to make this configurable, at a minimum? I know you originally proposed configurability [here](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1690#issue-1098573807)

fisker:
Make sense to me.

carlosjeurissen:
Situations exist in which utf8 is the default, and utf-8 is an alias like in nodeJS:
[...]

PR #2600, opened 2025-03-19
rule text-encoding-identifier-case: Change utf8 canonical to utf-8

carlosjeurissen:
Addresses https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1926 and https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2547


fisker:
Please add an option to allow both.

sindresorhus:
The convention in Node.js (and as used in most Node.js docs) is `utf8`, and I don't see any reason to break that. I'm not against adding an option to support `utf-8`, but the default should continue to be `utf8`.

PR #2607, opened 2025-03-26
Add `prefer-import-meta-properties` rule

### LINKED ISSUE #2255: Support `import.meta.dirname` and `import.meta.filename` in `prefer-module` rule
- Replace `__dirname` and `__filename` with them.
- Replace `const __dirname = path.dirname(fileURLToPath(import.meta.url));` with them.

ota-meshi:
This PR adds `prefer-import-meta-properties` rule that reports the following code:

```js
import path from "path";
import { fileURLToPath } from "url";
const dirname1 = path.dirname(fileURLToPath(import.meta.url));
const dirname2 = path.dirname(import.meta.filename);
const dirname3 = fileURLToPath(new URL(".", import.meta.url));
const filename1 = fileURLToPath(import.meta.url);
const filename2 = fileURLToPath(new URL(import.meta.url));
```

close #2255


ota-meshi:
I realized that `new URL(".", import.meta.url).pathname` shouldn't be reported because it can be used in the browser.
I'll change this PR.

fisker:
> I realized that `new URL(".", import.meta.url).pathname` shouldn't be reported because it can be used in the browser. I'll change this PR.

`new URL("./", import.meta.url).pathname` should be the same logic.

ota-meshi:
Thank you!
I changed it to check `new URL("./", ...)` as well.

test/prefer-module.js
```diff
@@ -307,3 +307,84 @@ test.snapshot({
...
+	invalid: [
+		outdent`
+			import path from "path";
+			import { fileURLToPath } from "url";
+			const dirname1 = path.dirname(fileURLToPath(import.meta.url));
+			const dirname2 = path.dirname(import.meta.filename);
+			const dirname3 = path.dirname(new URL(import.meta.url).pathname);
+			const dirname4 = fileURLToPath(new URL(".", import.meta.url));
+			const dirname5 = fileURLToPath(new URL("./", import.meta.url));
+		`,
```
fisker, line 334: Can we split these tests? Little hard to review.

Something like

```js
...[
  'dirname(fileURLToPath(import.meta.url))'
].flatMap(code => [
  `
	  import {dirname} from "path";
	  import {fileURLToPath} from "url";
	  ${code}
  `,
  `
	  import path from "path";
	  import {fileURLToPath} from "path";
	  ${code.replace('dirname', 'path.dirname')}
  `,
  `
	  import {dirname} from "path";
	  import url from "url";
	  ${code.replace('fileURLToPath', 'url.fileURLToPath')}
  `,
  `
	  import path from "path";
	  import url from "path";
	  ${code.replace('dirname', 'path.dirname').replace('fileURLToPath', 'url.fileURLToPath')}
  `,
])
```

rules/prefer-module.js
```diff
@@ -358,6 +457,152 @@ function create(context) {
...
+		function buildProblem(node, name) {
+			const problem = {
+				node,
+				messageId: name === 'dirname' ? ERROR_CALC_DIRNAME : ERROR_CALC_FILENAME,
+				data: {name},
+			};
+			const fix = fixer =>
+				fixer.replaceText(node, `import.meta.${name}`);
+
+			if (filename.endsWith('.mjs')) {
```
fisker, line 591: Should `import.meta` already ensure these files are in ESM? Why this check needed?

fisker: Should we give the community some time to drop Node.js 18, and add an option instead?

ota-meshi: Oh, you're right, no need to check for mjs 😓 
It's definitely an ESM because it already uses `import.meta`.
I'll change that.

rules/prefer-module.js
```diff
@@ -198,6 +209,94 @@ const isTopLevelReturnStatement = node => {
...
+	if (callee.type === 'Identifier') {
+		// Check for propertyName(...);
+		const specifier = getImportSpecifier(callee);
+
+		return specifier?.type === 'ImportSpecifier' && specifier.imported.name === propertyName;
+	}
+
+	return false;
+
+	function getImportSpecifier(node) {
```
fisker, line 266: I understand this is a safe play, and you already added, it's good.

But there is another way to get these builtin modules now `process.getBuiltinModule`, maybe we can just simply check the callee is

```
fileURLToPath
url.fileURLToPath
``` 

?

Unless you want add support for `const url = process.getBuiltinModule('node:url')` and `const {fileURLToPath} = process.getBuiltinModule('node:url')`

ota-meshi: I think `fileURLToPath()` is mostly fine, but I think it's a bit dangerous to check `path.dirname()` only for identifiers.
`dirname()` seems like a common word, so I think it might be a user-created function.

I'll try changing it to check `process.getBuiltinModule()`.

---
By the way, this is the first time I've heard of the existence of `process.getBuiltinModule()` 😅.
Thanks for letting me know ❤️

fisker: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2611 😄

rules/prefer-module.js
ota-meshi: > Should we give the community some time to drop Node.js 18, and add an option instead?

With Node v18 reaching EOL next month, I'm not sure if it should be an option 😅

ota-meshi:
Thank you for your review!
I have made some changes to this PR.

rules/prefer-module.js
```diff
@@ -358,6 +497,138 @@ function create(context) {
...
+
+		/**
+		 @param { import('estree').Node} node
+		 @param {'dirname' | 'filename'} name
+		 */
+		function buildProblem(node, name) {
+			return {
+				node,
+				messageId: name === 'dirname' ? ERROR_CALC_DIRNAME : ERROR_CALC_FILENAME,
+				data: {name},
```
fisker, line 626: I didn't see placeholders in these two messages.

rules/prefer-module.js
```diff
@@ -358,6 +497,138 @@ function create(context) {
...
+					yield buildProblem(parent, 'dirname');
+				}
+			}
+		}
+
+		/**
+		 @param { import('estree').Node} node
+		 @param {'dirname' | 'filename'} name
+		 */
+		function buildProblem(node, name) {
```
fisker, line 622: We usually name it as `getProblem`.

rules/prefer-module.js
```diff
@@ -198,6 +205,138 @@ const isTopLevelReturnStatement = node => {
...
+	return node.value === '.' || node.value === './';
+};
+
+const isImportMeta = node =>
+	node.type === 'MetaProperty'
+	&& node.meta.name === 'import'
+	&& node.property.name === 'meta';
+
+/** @returns {node is import('estree').NewExpression} */
+const isNewURL = node =>
```
fisker, line 222: There is a existing `isCallExpression`

rules/prefer-module.js
```diff
@@ -198,6 +205,138 @@ const isTopLevelReturnStatement = node => {
...
+	node.type === 'NewExpression'
+	&& node.callee.type === 'Identifier'
+	&& node.callee.name === 'URL';
+
+/** @returns {node is import('estree').MemberExpression} */
+const isAccessPathname = node =>
```
fisker, line 228: `isMemeberExpression`

rules/prefer-module.js
```diff
@@ -198,6 +205,138 @@ const isTopLevelReturnStatement = node => {
...
+		}
+
+		if (node.type === 'CallExpression') {
+			if (checkKind !== 'module') {
+				return false;
+			}
+
+			// Check process.getBuiltinModule('x')
+			return (
+				isMemberExpression(node.callee, {property: 'getBuiltinModule', object: 'process'})
```
fisker, line 258:
```suggestion
				isMemberCall(node, {object: 'process', method: 'getBuiltinModule', argumentsLength: 1, optionalMember: false, optionalCall: false})
```

rules/prefer-module.js
```diff
@@ -198,6 +205,138 @@ const isTopLevelReturnStatement = node => {
...
+
+		return define.type === 'Variable' && checkPattern(define.name, checkKind);
+	}
+
+	/** @param {import('estree').Identifier | import('estree').ObjectPattern} node */
+	function checkPattern(node, checkKind) {
+		/** @type {{parent?: import('estree').Node}} */
+		const {parent} = node;
+		if (parent.type === 'VariableDeclarator') {
+			if (!parent.init || parent.id !== node) {
```
fisker, line 302: Let's make sure it's a `const`

rules/prefer-module.js
```diff
@@ -198,6 +205,138 @@ const isTopLevelReturnStatement = node => {
...
+	function checkDefinition(define, checkKind) {
+		if (define.type === 'ImportBinding') {
+			if (!isModuleLiteral(define.parent.source)) {
+				return false;
+			}
+
+			const specifier = define.node;
+			return checkKind === 'module'
+				? (specifier?.type === 'ImportDefaultSpecifier' || specifier?.type === 'ImportNamespaceSpecifier')
+				: (specifier?.type === 'ImportSpecifier' && specifier.imported.name === propertyName);
```
fisker, line 291: Seems missing `import * as path ...` ?

fisker (changes requested):
@ota-meshi Please take a look

rules/prefer-module.js
ota-meshi: We already check if it's an `ImportNamespaceSpecifier` when `checkKind === 'module'` 👍

ota-meshi:
Thank you for reviewing this PR!
I've made some changes to this PR, could you please check it again?

rules/prefer-module.js
fisker: Sorry, missed.

fisker:
Pushed some commits, most of them are making all kinds of check stricter and improve readability.

fisker:
Due to bug in `iterateProblemsFromFilename`, I found a broken case https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2607/commits/78169281fcea9f8025875a40fea90668e274e73d , can you fix?

ota-meshi:
I think it's the same test case as L406, and I think it's intentional. Any suggestions on how to fix it :thinking:

fisker:
> I think it's the same test case as L406, and I think it's intentional. Any suggestions on how to fix it 🤔

Sorry, I was wrong about it.

fisker:
Everything looks good to me, except one thing.

Do you think it's possible that we break this case?

```
const x = new URL(import.meta.url).pathname
const filename = isWindows() ? x.slice(1) :x
```

Since `new URL('file:///C:/file').pathname` is actually `/C:/file`, not `C:/file`

fisker:
Also, about [this](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2607#discussion_r2016025914)

> Should we give the community some time to drop Node.js 18, and add an option instead?

[Your answer](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2607#discussion_r2016480538) make sense to me 

> With Node v18 reaching EOL next month, I'm not sure if it should be an option 😅

But I don't want people to disable this rule only because their target environment doesn't have `import.meta.{dirname,filename}`.

Personally, I even prefer add this PR as a separate rule, this part also not technically "prefer-module".

ota-meshi:
> Since `new URL('file:///C:/file').pathname` is actually `/C:/file`, not `C:/file`

Thank you for your comment, I didn't know it would return different results for Windows paths 😅 
I will change it to not report `.pathname`.

ota-meshi:
> Personally, I even prefer add this PR as a separate rule, this part also not technically "prefer-module".

By the way, originally I wanted a rule to be added to convert `path.dirname(fileURLToPath(import.meta.url))` to `import.meta.dirname`. I searched for similar issues in this repository and found #2255, so I implemented it in `prefer-module` and opened a PR.
It's useful for me either in the `prefer-module` rule or in another rule 😄

sindresorhus:
It could be a separate rule, but isn't it easier just to keep it in here under an option? We still want this rule to enforce not using `__dirname`, and we want a fix-it for that, and that fix it should ideally be `import.meta.dirname` (when an option is enabled). So while a different rule could handle the `fileURLToPath(import.meta.url)` and `path.dirname(import.meta.filename)` case. It just seems easier to have it all here.

fisker:
> __dirname

We already provide both 
 `import.meta.dirname` and `path.dirname(url.fileURLToPath(import.meta.url))` as suggestion.

If the new rule enabled, it doesn't matter which one they choose, use will use `import.meta.dirname`.

Since you are not against seperate rule, let's do it. 

My suggestion for rule name is `prefer-import-meta-properties`.

ota-meshi:
> My suggestion for rule name is prefer-import-meta-properties.

👍 

I will change this PR to add the new rule.

fisker (approved):
Good job! Thank you!

PR #2611, opened 2025-03-27
`prefer-node-protocol`: Support `process.getBuiltinModule()`

fisker:


test/prefer-node-protocol.js
```diff
@@ -92,6 +92,29 @@ test.snapshot({
...
+		'const fs = notProcess.getBuiltinModule("fs")',
+		'const fs = process.getBuiltinModule("fs", extra)',
+		'const fs = process.getBuiltinModule(...["fs"])',
+		'const fs = process.getBuiltinModule()',
+		'const fs = process.getBuiltinModule("unicorn")',
+		// Not checking this to avoid false positive
+		outdent`
+			import {getBuiltinModule} from 'node:process';
+			const fs = getBuiltinModule("fs");
+		`,
```
fisker, line 111: This case need check where the `getBuiltinModule` from, let's ignore it now.

PR #2614, opened 2025-03-28
Add `no-unnecessary-array-splice-count` rule, Rename `no-length-as-slice-end` to `no-unnecessary-slice-end`, Support checking `Infinity` in `no-unnecessary-slice-end`

### LINKED ISSUE #2601: `no-length-as-slice-end`: support splice
### Description

I think .splice() has the same issue as .slice

### Examples

```js
// ❌
original.splice(0, original.length)

// ✅
original.splice(0)
```

### Additional Info

_No response_

fisker:
- Rename `no-length-as-slice-end` to `no-unnecessary-slice-end`
- Support checking `Infinity` in `no-unnecessary-slice-end`
- Add `no-unnecessary-array-splice-count`, use the same logic as `no-unnecessary-slice-end`

Fixes #2601


.gitignore
```diff
@@ -6,3 +6,4 @@ package-lock.json
 .cache-eslint-remote-tester
 eslint-remote-tester-results
 *.log
+.vscode
```
sindresorhus, line 9: https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

fisker: I'm not sure add this to global ignore is a good idea, since many repos actually stores `.vscode`.
Adding this because my VSCode keeps creating this empty directory.

PR #2619, opened 2025-04-01
Replace `read-package-up` with `find-up-simple`

### LINKED ISSUE #2151: Suggestion on slimming down dependencies
`eslint-plugin-unicorn` has 63 transitive dependencies: https://npm.anvaka.com/#/view/2d/eslint-plugin-unicorn

39 of them come from `read-pkg-up`: https://npm.anvaka.com/#/view/2d/read-pkg-up

This functionality isn't super complicated and could probably be done without any dependencies. E.g. here's Vite's implementation, which resides in a single file and is more complicated than than the current implementation in some ways since it has caching built-in: https://github.com/vitejs/vite/blob/ff3ce312a572ec126808afb97b5d3f0a6f9adcb1/packages/vite/src/node/packages.ts#L117

As an example of why we could do this with less code, `read-pkg-up` pulls in some dependencies to normalize the package data, but this plugin uses the `{normalize: false}` option, so that functionality would not be necessary.

beeequeue:
Fixes #2151

~~I decided to use [`lukeed/empathic`](https://github.com/lukeed/empathic) over `find-up` due to it being smaller, simpler, and dependency free.~~
Changed to `find-up-simple` as I missed the comment mentioning it.

I also extracted the similar logic into a shared utility (I wasn't sure whether to put it in `shared/` or `utils/`) and added some simple caching for it.


beeequeue:
Ah I only saw the `find-up` comment

rules/expiring-todo-comments.js
```diff
@@ -50,24 +50,12 @@ const messages = {
...
-	const hasPackage = Boolean(packageResult);
-	const packageJson = packageResult ? packageResult.packageJson : {};
+	const packageJsonResult = readPackageJson(dirname);
+	const hasPackage = Boolean(packageJsonResult);
 
 	const packageDependencies = {
-		...packageJson.dependencies,
-		...packageJson.devDependencies,
+		...packageJsonResult.packageJson.dependencies,
+		...packageJsonResult.packageJson.devDependencies,
```
fisker, line 58: Will this line throw if there is no `package.json`?

beeequeue: No, it falls back to an empty object.

beeequeue: You're right, I was confused by [this line](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2619/files#diff-983077e885cbad3d4bbd1379551c5912e39e4e0685457d3a0a9f8e320e53ac30R180) in my early morning haze

beeequeue: you can test the fix by running these lines in the node repl:

```js
const packageJsonResult = {}
const packageJson = packageJsonResult?.packageJson ?? {}
const test = { ...packageJson.dependencies, ...packageJson.devDependencies }
```

rules/shared/package-json.js
```diff
@@ -0,0 +1,52 @@
...
+			return {
+				path: packageJsonPath,
+				packageJson: dataCache.get(packageJsonPath),
+			};
+		}
+	}
+
+	const packageJsonPath = findUpSync('package.json', {cwd: dirname, type: 'file'});
+	if (!packageJsonPath) {
+		return;
```
fisker, line 29: I think we should also cache the result if it's not found.

Same on line 45

beeequeue: I added caching for missing file paths, but caching line 45 would mean that we would cache any errors in the file, which I believe would not be desired.

Especially when considering language servers

fisker: Simplified little bit, [`5188d85` (#2619)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2619/commits/5188d85427ca3e013416adbfb4949eb9c5d51004)

fisker (approved):

PR #2621, opened 2025-04-03
fix(`no-useless-undefined`): Support functions with allowed undefined return types and mixed returns

nsotiroglou96:
This update refines the `no-useless-undefined` rule to better handle cases where returning undefined is valid, specifically:

1. The rule now inspects the enclosing function’s return type annotation and skips reporting when that type includes undefined (e.g. Promise<number | undefined>). 
2. It also introduces an `allowMixedReturns` option so that functions with mixed return values (some branches returning a value, others undefined) aren’t flagged unnecessarily.

### DIFF
--- a/rules/no-useless-undefined.js
+++ b/rules/no-useless-undefined.js
@@ -106,6 +106,61 @@ const includesUndefined = (typeAnnotation) => {
 	}
 };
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Please make sure `npm test` passes locally.

You need to add tests.

PR #2647, opened 2025-05-08
Remove `defaultsDeep`

fisker:


test/import-style.js
```diff
@@ -626,6 +641,51 @@ test({
...
+			],
+			errors: [{
+				messageId: 'importStyle',
+				data: {
+					allowedStyles: 'named or default',
+					moduleName: 'node:util',
+				},
+			}],
+		},
+		// ...(see previous test), unless we disable `named` explicitly
```
fisker, line 666: Maybe this should not work this way, but this is how it works before, I'm not going to change it.

PR #2656, opened 2025-05-18
`prevent-abbreviations`: Add `ProcessEnv` to `defaultAllowList`

azat-io:
Hello!

I currently have a problem when using the `prevent-abbreviations` rule when typing `process.env`:

```
// environment.d.ts
declare global {
  namespace NodeJS {
    interface ProcessEnv {
      GITHUB_ACTIONS: string;
    }
  }
}

export {}
```

Error:

```
The variable `ProcessEnv` should be named `ProcessEnvironment`.
```

This is the only and natural way to describe types for `process.env`.

### DIFF
--- a/rules/shared/abbreviations.js
+++ b/rules/shared/abbreviations.js
@@ -254,6 +254,9 @@ export const defaultAllowList = {
 	getStaticProps: true,
 	// The name iOS is a standard name for an OS
 	iOS: true,
+	// Node.js global environment interface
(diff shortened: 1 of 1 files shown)


sindresorhus:
`NodeJS.ProcessEnv` is not the only use for such a name, and "process environment" can be valid in many cases, so it should not be in the default list. This is a good case for an inline ESLint ignore comment in your own code. We could potentially add a specific workaround for this case, but there would have to be a high quality pull request for such a thing.

Issue #2657, opened 2025-05-22
Rule proposal: `no-array-fill-with-reference-type`


legend80s:
## **Rule: `no-array-fill-with-reference-type`**  
**Description**:  
Disallows using `Array.fill()` with **reference types** (objects, arrays, functions, Maps, Sets, RegExp literals, etc.) to prevent unintended shared references across array elements. Encourages `Array.from()` or explicit iteration for creating independent instances.  

**Reason**:  
`Array(len).fill(value)` fills all elements with the **same reference** if `value` is non-primitive (e.g., `fill([])`), leading to bugs when one element’s mutations affect others.  

**Key Features**:  
- Catches **all reference types**: Objects, Arrays, Functions, `new` expressions, RegExp literals, and variables referencing them.  
- **Clear error messages**: Identifies the reference type (e.g., `Object`, `RegExp`, `variable (name)`).  
- **Recommends fixes**: Suggests `Array.from()` or similar patterns.  

**Note**: Primitive types (`number`, `string`, `boolean`, `null`, `undefined`, `symbol`, `bigint`) are always allowed.

Here's a polished version of your text with improved clarity and flow:

---

## Why Existing Rules Don't Address This Issue

The [`no-new-array`](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-new-array.md) rule prohibits using `new Array()` or `Array()` (maybe) constructor calls, but it doesn't specifically address the reference-sharing problem with `.fill()`. Since developers may disable this rule when they legitimately need array constructors, we need a dedicated rule to:

1. **Target the specific hazard** of reference sharing when using `.fill()`
2. **Provide focused guidance** on safe alternatives like `Array.from()`

This specialized rule complements rather than replaces `no-new-array`, offering more granular protection against this common pitfall.

---
## Need fix?

No fix needed because this rule is automatically fixable by `no-new-array`. or the `fix` logic can use the same as `no-new-array`.

### Examples

```js
// ❌
new Array(3).fill({});       // ✗ Object  
new Array(3).fill(new Map());       // ✗ Map
new Array(3).fill(new Set());       // ✗ Set
new Array(3).fill(/pattern/); // ✗ RegExp 
new Array(3).fill(new String('fff'));       // ✗ new String
new Array(3).fill(new Foo('fff'));       // ✗ new Class
new Array(3).fill(Class);       // ✗ Class or function

const map = new Map();  
[...]

PR #2661, opened 2025-05-29
feat: add new rule no-array-fill-with-reference-type without fix solution

legend80s:
https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2657


rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,141 @@
...
+		const {variables} = context.sourceCode.getScope(node);
+
+		log('variables 2:', variables);
+		if (!variables || variables.length === 0) {
+			// Variable declaration not found; it might be a global variable.
+			return false;
+		}
+	}
+
+	// Other cases: objects, arrays, functions, new expressions, regular expressions, etc.
```
fisker: I don't think functions should be reported.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,141 @@
...
+				} else if (fillArgument.type === 'Identifier') {
+					type = `variable (${fillArgument.name})`;
+				}
+			}
+		}
+
+		const actual = context.sourceCode.getText(node.callee.object.callee) === 'Array.from' ? 'Array.from().fill()' : 'Array.fill()';
+
+		return {
+			node,
```
fisker:
```suggestion
			node: node.arguments[0],
```
Let's report on `node.arguments[0]` instead.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,141 @@
...
+function isReferenceType(node, context) {
+	if (!node) {
+		return false;
+	}
+
+	// For null, number, string, boolean.
+	if (node.type === 'Literal') {
+		// Exclude regular expression literals (e.g., `/pattern/`, which are objects despite being literals).
+		return node.regex !== undefined;
+	}
```
fisker, line 267: I'm not sure about this, unless `lastIndex` is used, it normally doesn't matter I think.

rules/no-array-fill-with-reference-type.js
legend80s: function is also an object that can add property.

```js
funcs = Array(3).fill(function foo() {
  return 'foo'
})

funcs[0].property = 'function property'
console.log(funcs[1].property) // function property
```

rules/no-array-fill-with-reference-type.js
legend80s:
```js
regexpList = Array(3).fill(/foo/)

regexpList[0].property = 'add property to regexp'
console.log(regexpList[1].property) // => 'add property to regexp'
```

rules/no-array-fill-with-reference-type.js
fisker: Yes you can, but what's the consequence? It's rarely to see cases store and access data from a function.

legend80s: Yes it's rare so a `canFillWithFunction` default to `true` has added to the rule's option.

rules/no-array-fill-with-reference-type.js
legend80s: should I just ignore regexp and function? or add whitelist options?

fisker: functions and regular expressions are rarely going to cause problems, let's ignore them.

If you prefer an option to control, they should be ignored by default.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,184 @@
...
+		log('variables:', variables);
+		log('variable:', variable);
+		log('variable.defs[0].node:', definitionNode);
+
+		if (!variable || !definitionNode) {
+			return false;
+		}
+
+		// Check `const foo = []; Array(3).fill(foo);`
+		if (definitionNode.type === 'VariableDeclarator') {
```
fisker: Please make sure it's a `const`(check `parent.kind`), 

```js
let a = []
a = 2
new Array(3).fill(a)
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,184 @@
...
+			break;
+		}
+
+		case 'ArrayExpression': {
+			type = 'Array';
+			break;
+		}
+
+		case 'NewExpression': {
+			type = `new ${fillArgument.callee.name}()`;
```
fisker: `NewExpression.callee` not always have a name.

```js
new A.B()
```

```js
new class {}
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,184 @@
+import {isFunction} from './ast/index.js';
+
+// @ts-check
+const MESSAGE_ID_ERROR = 'no-array-fill-with-reference-type/error';
+const messages = {
+	[MESSAGE_ID_ERROR]: 'Avoid using `{{actual}}` with reference type{{type}}. Use `Array.from({ ... }, () => { ... })` instead to ensure independent instances.',
```
fisker: This message can be misleading

```
const object = {}

Array.from({length: 3}).fill(object)

Array.from({length: 3}, () => object) //  Still using reference
```

fisker: Speaking of this, maybe we should also report

```
const object = {}
Array.from({length: 3}, () => object)
```

rules/no-array-fill-with-reference-type.js
legend80s: options have been added and ignored by default.

rules/no-array-fill-with-reference-type.js
legend80s: Good point, it's fixed!
```
+> Error 1/1
+
+    `␊
+    > 1 | new Array(3).fill(new A.B());␊
+        |                   ^^^^^^^^^ Avoid using \`Array.fill()\` with reference type (new A.B). Use \`Array.from({ ... }, () => { return independent instance })\` instead to ensure no reference shared.␊
```

rules/no-array-fill-with-reference-type.js
legend80s: Only `Array.fill()` and `Array.from().fill()` are checked.

```
// Below are the cases though have the same reference problem but are not covered by the rule.
// Due to the rule name it will not check other than `Array.fill`, even if `Array.from` also fills in reference variable.
// It cannot be exhaustively checked, we only check `.fill`.
'const map = new Map(); Array.from({ length: 3 }, () => map);',
`
const map = new Map();
const list = [];
for (let i = 0; i < 3; i++) {
	list.push(map);
}
`,
```

rules/no-array-fill-with-reference-type.js
legend80s: Good point!

rules/no-array-fill-with-reference-type.js
legend80s: when we check 
```
const object = {}
Array.from({length: 3}, () => object)
```

the ci `npm run lint` failed. Should we add ignore comments for the line in test/prefer-negative-index.js

```
// It's safe due to no mutation to `errors`
// eslint-disable-next-line unicorn/no-array-fill-with-reference-type
errors: Array.from({length: 4}, () => error),
```

or change it to?

```
errors: Array.from({length: 4}, () => ({ ...error })),
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,324 @@
...
+
+const debugging = false;
+const log = (...arguments_) => debugging && console.log(...arguments_);
+
+/** @param {import('eslint').Rule.RuleContext} context */
+const create = context => ({
+	CallExpression(node) {
+		// Array.fill or Array.from().fill
+		const isArrayFill = node.callee.type === 'MemberExpression'
+			&& ((node.callee.object.callee?.name === 'Array') || (context.sourceCode.getText(node.callee.object.callee) === 'Array.from'))
```
fisker: Let's use `isMemberExpression` instead of checking text.

`Array.fill` do you mean `new Array().fill()`? Please use `isNewExpression`.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,324 @@
...
+		return {returnNode: node.body, declaredInCurrentFunction: true};
+	}
+
+	const returnStatement = node.body.body.find(node => node.type === 'ReturnStatement');
+	const name = returnStatement?.argument?.name;
+	if (!name) {
+		return {returnNode: returnStatement?.argument, declaredInCurrentFunction: true};
+	}
+
+	const declaredInCurrentFunction = node.body.body.some(node => node.type === 'VariableDeclaration' && node.declarations.some(declaration => declaration.id.name === name));
```
fisker: Use `findVariable` instead.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,324 @@
...
+	{
+		type: 'object',
+		additionalProperties: false,
+		properties: {
+			canFillWithFunction: {
+				type: 'boolean',
+			},
+			canFillWithRegexp: {
+				type: 'boolean',
+			},
```
fisker: How about `allowFunctions` and `allowRegularExpressions`?

legend80s: OK 👍

rules/no-array-fill-with-reference-type.js
legend80s: The function `findVariable` in @eslint-community/eslint-utils is designed to locate variables that are defined in or outside the current function. However, my requirement is slightly different. I want to restrict the search to variables declared within the current function. Even if the variable is a reference object, as long as it is declared in the current function (`declaredInCurrentFunction`), it should be considered valid.

```ts
// valid case 
const map = new Map();
const list = Array.from({ length: 3 }, () => {
	const map = new Map(); # valid, map is defined in it's function body
	return map
});
```

rules/no-array-fill-with-reference-type.js
legend80s: Okay it has been changed to `const isArrayFill = isMemberExpression(node.callee)`

fisker: Please check how `isMemberExpression`/`isMethodCall` used in other rules.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,354 @@
...
+			// Not check if the identifier is not a function
+			return {returnNode: node, declaredInCurrentFunction: true};
+		}
+
+		return getReturnIdentifier(init, context);
+	}
+
+	// Console.log('node:', node);
+
+	// No check member expression as callback `Array.from(element.querySelectorAll('ng2 li')).map(angular.element);`
```
fisker, line 134: Did you mean `.fill`?

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,354 @@
...
+ @returns {string}
+ */
+function getType(fillArgument, context) {
+	let type = '';
+
+	switch (fillArgument.type) {
+		case 'ObjectExpression': {
+			type = 'Object';
+			break;
+		}
```
fisker, line 183:
```suggestion
		case 'ObjectExpression': {
			return 'Object';
		}
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,354 @@
...
+	// Symbol (such as `Symbol('name')`)
+	if (node.type === 'CallExpression' && node.callee.name === 'Symbol') {
+		const {variables} = context.sourceCode.getScope(node);
+
+		log('variables 2:', variables);
+		if (!variables || variables.length === 0) {
+			// Variable declaration not found; it might be a global variable.
+			return false;
+		}
+	}
```
fisker, line 281: Should we check `Symbol.for('description')` and `Symbol.iterator`(well-known symbols)?

legend80s: No it should not be checked. Code will be added to deal with this wrong checking.

rules/no-array-fill-with-reference-type.js
legend80s: Member expression node has added to checklist.

rules/no-array-fill-with-reference-type.js
legend80s: Code is simpler after using those utility functions. Thanks!

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
+// @ts-check
+import {findVariable} from '@eslint-community/eslint-utils';
+import {
+	isCallExpression,
+	isFunction, isMemberExpression, isMethodCall, isNewExpression, isRegexLiteral,
+} from './ast/index.js';
```
fisker, line 10:
```suggestion
import {
	isCallExpression,
	isFunction,
	isMemberExpression,
	isMethodCall,
	isNewExpression,
	isRegexLiteral,
} from './ast/index.js';
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+
+const debugging = false;
```
fisker:
```suggestion
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+};
+
+const DEFAULTS = {
+	// Not check for function expressions by default because it is rare to fill an array with a function and add properties to it.
+	allowFunctions: true,
+	// The same reason as above.
+	allowRegularExpressions: true,
+};
+
+const log = (...arguments_) => debugging && console.log(...arguments_);
```
fisker:
```suggestion
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+
+	return 'new ()';
+}
+
+/**
+ @param {Node} node
+ @param {import('eslint').Rule.RuleContext} context
+ @returns {[is: false] | [is: true, node: Node]}
+ */
+function isReferenceType(node, context) {
```
fisker: Can't we use the node to determine if it's a reference?

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+		const currentPropertyInChain = chain[index].property;
+		// .log(`#${index}`, 'currentPropertyInChain:', currentPropertyInChain?.type, currentPropertyInChain);
+		// .log(`#${index}`, 'currentObject:', currentObject);
+		if (!currentObject || currentObject.type !== 'ObjectExpression') {
+			return;
+		}
+
+		const property = currentObject.properties.find(
+			// @ts-expect-error
+			p => p.key.type === 'Identifier'
```
fisker: The parameter name should be `property`

test/string-content.js
```diff
@@ -223,6 +223,7 @@ test({
 			code: 'const foo = `no${foo}no${foo}no`',
 			output: 'const foo = `yes${foo}yes${foo}yes`',
 			options: [{patterns: noToYesPattern}],
+
```
fisker:
```suggestion
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+ */
+function getReturnIdentifier(node, context) {
+	if (node.type === 'Identifier') {
+		const scope = context.sourceCode.getScope(node);
+
+		const variable = findVariable(scope, node);
+
+		if (!variable) {
+			// Not check if the identifier is not declared
+			return {returnNode: node, declaredInCurrentFunction: true};
```
fisker: Can you explain what this `declaredInCurrentFunction` mean?

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+	// Array.from({ length: 3 }, () => (new Map))
+	// Array.from({ length: 3 }, () => ({}))
+	// Array.from({ length: 3 }, () => {})
+	if (!nodeBody.body) {
+		return {returnNode: nodeBody, declaredInCurrentFunction: true};
+	}
+
+	const returnStatement = nodeBody.body.find(node => node.type === 'ReturnStatement');
+	const name = returnStatement?.argument?.name;
+	if (!name) {
```
fisker: Are we checking `Identifier`? If so please check node.type  explicitly

rules/no-array-fill-with-reference-type.js
legend80s: If the node is an `Literal` we can just use node, but for variable identifiers we should recursively check its declaration until the `Literal` has been found. Therefore the `context` is required.

rules/no-array-fill-with-reference-type.js
legend80s: Fixed.

rules/no-array-fill-with-reference-type.js
legend80s: If the return node is declared in current callback of Array.from then it is valid even if it is a reference type, because it is created at every callback called in another word it is not a *shared* reference.
Consider this one in test/no-array-fill-with-reference-type.js:

```js
// Valid because it returns a new map every time
		`
		const map = new Map();
		const list = Array.from({ length: 3 }, () => {
			const map = new Map();
			return map
		});
		`,

```
But if this line `const map = new Map();` is omitted the case should be invalid.

rules/no-array-fill-with-reference-type.js
fisker: I mean the return type. Shouldn't enough to return just the node?

rules/no-array-fill-with-reference-type.js
legend80s: `if (!name)` is for case below:
```js
const list = Array.from({ length: 3 }, () => {
  return {} // the returnStatement  has no name
});
```

rules/no-array-fill-with-reference-type.js
legend80s:
```js
const [is, resolvedNode] = isReferenceType(fillArgument, context);

		if (!is) {
			return;
		}

		const actual = isMethodCall(
			// @ts-expect-error
			node.callee.object,
			{object: 'Array', method: 'from'},
		)
			? 'Array.from().fill()'
			: 'Array.fill()';

		const type = getType(resolvedNode, context);
```
Yes `isReferenceType` can return just the node: return `node` if it is a reference and return `nil` otherwise. but the function name is **is**ReferenceType it should return a boolean as its name suggests. but the node is needed to `getType` from. Therefore I decided to return a tuple. But if you want it to return only the node, I can change that.

rules/no-array-fill-with-reference-type.js
fisker: Use `returnStatement?.argument?.type !== 'Identifier'` then.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+	/**
+	  @param {import('estree').CallExpression} node
+	 */
+	CallExpression(node) {
+		// Check all `fill` method call even if the object is not Array because we don't know if its runtime type is array
+		// `arr.fill()` or `new Array().fill()` or `Array.from().fill()`
+		const isArrayFill = isMethodCall(node, {method: 'fill', argumentsLength: 1});
+
+		// `Array.from().map` or `Array.from(arrayLike, mapper)`
+		const isArrayFrom = isMethodCall(node, {object: 'Array', method: 'from'});
```
fisker:
```suggestion
		const isArrayFrom = isMethodCall(node, {object: 'Array', method: 'from', argumentsLength: 2});
```

fisker: If we are going to check `Array.from().map()`, better just check `anything.map()`

rules/no-array-fill-with-reference-type.js
fisker: If `!variable`, I think it's because it's global reference? Shouldn't this case be `declaredInCurrentFunction: false`?

docs/rules/no-array-fill-with-reference-type.md
```diff
@@ -0,0 +1,161 @@
...
+```js
+// ❌ Class
+class BarClass {};
+new Array(3).fill(new BarClass());
+Array(3).fill(new BarClass());
+Array.from({ length: 3 }).fill(new BarClass());
+
+// ✅
+Array.from({ length: 3 }, () => new BarClass());
+```
```
fisker: These two are the same cases.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,517 @@
...
+	},
+});
+
+/**
+
+ @param {import('estree').CallExpression} node
+ @param {RuleContext} context
+ @returns
+ */
+function getArrayFromReturnNode(node, context) {
```
fisker: The logic in this function is ambiguous, please pass the function explicitly. And make sure `node` is a `function` before pass to it.

rules/no-array-fill-with-reference-type.js
legend80s: Yes, I did consider expanding the check scope, but that would likely trigger false positives. Since the `map` method can be declared on any custom class, we only check `Array.from().map`.

Take the case in `test/no-array-fill-with-reference-type.js` as an example.

The following code will be flagged incorrectly:

```js
const object = {};
Array.from({ length: 3 }, () => object);
```

```js
const object = {};
Array.from({ length: 31 }).map(() => object);
```

However, this won't be flagged:

```js
const myInstance = { map: () => {} };
const sharedObject = {};
myInstance.map(() => sharedObject);
```

docs/rules/no-array-fill-with-reference-type.md
legend80s: Yes in essence, they are the same,  but I want developers to know that this rule can check both native and custom classes.
OK the second case has been deleted.

rules/no-array-fill-with-reference-type.js
legend80s: To clear the ambiguousness I renamed it to `function getReturnNodeOfArrayDotFrom(functionNode, context) {` both the function name and the first param name.
But the check of "node is a function" is redundant because the node must be a `CallExpression`.
image

rules/no-array-fill-with-reference-type.js
legend80s: Should we check `Array.from().map` ? If not I will remove  the case
```js
const myInstance = { map: () => {} };
const sharedObject = {};
myInstance.map(() => sharedObject);
```

and modify to

```diff
-		const isArrayFrom = isMethodCall(node, {object: 'Array', method: 'from'});
+		const isArrayFrom = isMethodCall(node, {object: 'Array', method: 'from', argumentsLength: 2});
```

rules/no-array-fill-with-reference-type.js
legend80s: Yes, thanks—it’s much more readable now, and the code’s intent is more obvious.

rules/no-array-fill-with-reference-type.js
legend80s: You're right—`declaredInCurrentFunction` currently carries two meanings, which makes it confusing:  
1. A variable that is literally declared in the current function.  
2. A global variable that we also want to skip.  

Both cases tell us “no need to check,” so I crammed them into one name. I’ll split them into two clearer flags.

rules/no-array-fill-with-reference-type.js
fisker: Since we don't care about non-reference type, you can change this to `getReference`

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,535 @@
...
+
+	return {returnNode: returnStatement?.argument, declaredInCurrentFunction};
+}
+
+/**
+ @param {*} fillArgument
+ @param {import('eslint').Rule.RuleContext} context
+ @returns {string}
+ */
+function getType(fillArgument, context) {
```
fisker:
```suggestion
function getDescription(node, context) {
```

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,535 @@
...
+	// Array.from({ length: 10 }, () => { return sharedObject; });
+	let result;
+	if (secondArgument && isFunction(secondArgument)) {
+		result = getReturnIdentifier(secondArgument, context);
+		// @ts-expect-error node always has a parent
+	} else if (isMemberExpression(functionNode.parent, 'map')) {
+		// Array.from({ length: 10 }).map(() => { return sharedObject; });
+		// @ts-expect-error node always has a parent
+		result = getReturnIdentifier(functionNode.parent.parent.arguments[0], context);
+	}
```
fisker: This part should be done in the `CallExpression`, and pass the callback function to this function.

rules/no-array-fill-with-reference-type.js
```diff
@@ -0,0 +1,535 @@
...
+	return fillArgument;
+}
+
+/**
+
+ @param {import('estree').FunctionExpression | Node} node - callback for map
+ @param {RuleContext} context
+ @returns {{ returnNode: Node, declaredInCurrentFunction: boolean }}
+ */
+function getReturnIdentifier(node, context) {
```
fisker, line 108: This function should also only care about the reference type, I don't think `declaredInCurrentFunction` is necessary to return.

legend80s: Yes but if we return more information(`declaredInCurrentFunction` and `isGlobal`) other than node itself the code is more readable
```js
if (result?.declaredInCurrentFunction || result?.isGlobal) {
		return;
	}
```

> Should not check reference type if the identifier is declared in the current function or is global

rules/no-array-fill-with-reference-type.js
legend80s: OK

legend80s: OK

rules/no-array-fill-with-reference-type.js
legend80s: Yes!

sindresorhus:
Some tests are failing.

PR #2667, opened 2025-06-03
`prefer-string-replace-all`: Preserve escape

### LINKED ISSUE #2666: `rule-name`: better-regex / prefer-string-replace-all
### Description

Add an option for the quick fix, that doesn't convert unicode from codePoint into its string.

### Examples

```js
// ❌
str.replace(/\u200B/g, '')
str.replace(/\x20/g, '')

// Now they both will be fixed as:
str.replaceAll(' ​', '')
```

Sometimes, the unicode string isn't readable, so the code point is better to read.
Like the example above, the space character cannot be recognized if it deals with multiple space types.

```js
// Want an option to fix it as:
str.replaceAll(/\u200B/g, '')
```

### Additional Info

_No response_

fisker:
Fix #2666

I'm not confident with the current code, hope it won't break anything.

### DIFF
--- a/rules/prefer-string-replace-all.js
+++ b/rules/prefer-string-replace-all.js
@@ -10,7 +10,8 @@ const messages = {
 	[MESSAGE_ID_USE_STRING]: 'This pattern can be replaced with {{replacement}}.',
 };
 
-const QUOTE = '\''
+const QUOTE = '\'';
+
 function getPatternReplacement(node) {
 	if (!isRegexLiteral(node)) {
 		return;


rules/prefer-string-replace-all.js
```diff
@@ -39,10 +38,36 @@ function getPatternReplacement(node) {
...
+			if (codePoint === 10) {
+				return String.raw`\n`;
+			}
+
+			if (codePoint === 9) {
+				return String.raw`\t`;
+			}
+
+			return `\\u{${codePoint.toString(16)}}`;
+		}
```
fisker: This rewrites `\cX` to `\u{XX}`, but except `\r` `\n` `\t`.

PR #2676, opened 2025-06-12
Add `no-error-capture-stack-trace` rule

### LINKED ISSUE #2627: Rule proposal: `no-error-capture-stack-trace`
### Description

I'm not sure about other engines, but at least on Node.js if an error is subclassing `Error`, [Error.captureStackTrace()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/captureStackTrace) is unnecessary to call. It should already done by `Error` constructor.

### Examples

```js
class MyError extends Error {
	constructor() {
		// ❌
		Error.captureStackTrace(this, MyError);
		// ❌
		Error.captureStackTrace?.(this, MyError);
		// ❌
		Error.captureStackTrace(this, this.constructor);
		// ❌
		Error.captureStackTrace?.(this, this.constructor);
		// ❌
		Error.captureStackTrace(this, new.target);
		// ❌
		Error.captureStackTrace?.(this, new.target);
	}
}
```

### Proposed rule name

no-error-capture-stack-trace

### Additional Info

_No response_

fisker:
Fixes #2627


sindresorhus:
Should it be `no-useless-error-capture-stack-trace`? Since it doesn't prevent all uses of `Error.captureStackTrace`, just ones inside error subclass.

PR #2682, opened 2025-06-13
Update ESLint and use the new `isGlobalReference` to make rule stricter

fisker:


rules/shared/builtin-errors.js
```diff
@@ -7,7 +7,6 @@ const builtinErrors = [
 	'SyntaxError',
 	'TypeError',
 	'URIError',
-	'InternalError',
```
fisker, line 10: This is non-standard https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/InternalError

kachkaev:
👋 @fisker, It’d be great to see this change released! `eslint-plugin-unicorn@59.0.1` (latest) depends on `@eslint/plugin-kit 0.2.8` which triggers https://github.com/advisories/GHSA-xffm-g5w8-qvg7.

PR #2690, opened 2025-06-22
Add more test cases for `prefer-string-raw` rule

som-sm:
This PR adds a couple of new test cases that cover the following:

- ```js
  String.raw`a = '\\'`
  ```
  This covers the following check, which is currently _not_ covered by any of the tests.
  https://github.com/sindresorhus/eslint-plugin-unicorn/blob/944cb50f138a92a4a237d6615463db29108efe45/rules/prefer-string-raw.js#L59

- ```js
   String.raw`a = 'a\\b\''`
   String.raw`a = "a\\b\""`
   ```
   These cover the `nextCharacter === quote` check in the following line, which again is _not_ currently covered.
   https://github.com/sindresorhus/eslint-plugin-unicorn/blob/944cb50f138a92a4a237d6615463db29108efe45/rules/prefer-string-raw.js#L21
   
   
- ```js
   String.raw`a = 'a\\b\"'`
   ```
   Finally, this just highlights that if the string contains unnecessarily escaped quotes (or unnecessary escapes in general), this rule would stop complaining.


som-sm:
Also, would you be open for the following refactor?

<br>

Simplifying the following:
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/944cb50f138a92a4a237d6615463db29108efe45/rules/prefer-string-raw.js#L11-L32

To this:
```js
function unescapeBackslash(raw) {
	const quote = raw.charAt(0);

	return raw
		.slice(1, -1)
		.replaceAll(new RegExp(String.raw`\\([\\${quote}])`, 'g'), '$1');
}
```

fisker (approved):

PR #2691, opened 2025-06-23
`prefer-string-raw`: Add support for template literals

### LINKED ISSUE #2653: `prefer-string-raw`: Support template literals
### Description

Make sense to add it.

### Examples

```js
// ❌
`
   \\a
`

// ✅
String.raw`
   \a
`
```

### Additional Info

_No response_

som-sm:
Closes #2653


rules/prefer-string-raw.js
```diff
@@ -64,6 +61,52 @@ const create = context => {
...
+
+			if (cooked.includes(BACKSLASH)) {
+				hasBackslash = true;
+			}
+
+			if (index > 0) {
+				unescaped += '${' + context.sourceCode.getText(node.expressions[index - 1]) + '}';
+			}
+
+			unescaped += unescapedQuasi;
```
fisker: We don't need concat all the string, just save the replacements and replace `TemplateElement`s one by one later.

For example, current fix are removing spaces and comments in `expression`s.

rules/prefer-string-raw.js
```diff
@@ -64,6 +61,45 @@ const create = context => {
...
+			return;
+		}
+
+		return {
+			node,
+			messageId: MESSAGE_ID,
+			* fix(fixer) {
+				yield * fixSpaceAroundKeyword(fixer, node, context.sourceCode);
+				yield * node.quasis.map(quasi => replaceTemplateElement(fixer, quasi, quasi.value.cooked));
+				yield fixer.insertTextBefore(node, 'String.raw');
```
som-sm, line 78: The order matters over here, so the `fixSpaceAroundKeyword` fix should be before the `insertTextBefore` fix.

<br>

@fisker Is it better to do it like this instead?
```ts
* fix(fixer) {
	const quoteStart = context.sourceCode.getRange(node)[0];
	
	yield * node.quasis.map(quasi => replaceTemplateElement(fixer, quasi, quasi.value.cooked));
	yield fixer.replaceTextRange([quoteStart, quoteStart + 1], 'String.raw`');
	yield * fixSpaceAroundKeyword(fixer, node, context.sourceCode);
}
```
In this case, the order wouldn't matter.

rules/prefer-string-raw.js
som-sm: Fixed in [30ed709](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2691/commits/30ed7093dc62c010dca1dc7284f047c20db3a38b).

rules/prefer-string-raw.js
fisker: Order always matters, we don't need change to this.

test/prefer-string-raw.js
```diff
@@ -47,6 +47,53 @@ test.snapshot({
...
+		'a = `\\\\a${foo}b\\\\${foo}`',
+
+		// Tagged template expression
+		'a = String.raw`a\\\\b`',
+
+		// Slash before newline (spread into multiple lines)
+		outdent`
+			a = \`\\\\a \\
+			b\`
+		`,
```
fisker, line 77: Can you add test for 

```js
`\${`
```

and 

```js
`$\{`
```

?

rules/prefer-string-raw.js
som-sm: > Order always matters

I meant that in this specific case, the order of operations shouldn't matter. Whether we call `insertTextBefore` first and then `replaceTextRange`, or the other way around, the end result should be the same.

Consider this example:
Source:
```js
let foo = 1
```
Rule:
```js
Identifier(node) {
  if (node.name !== "foo") {
    return;
  }

  context.report({
    node,
    message: "Demo",
    fix(fixer) {
      return [
        fixer.insertTextBefore(node, "a"),
        fixer.insertTextBefore(node, "b")
      ];
    }
  });
}
```
Output:
```js
let abfoo = 1
```
So, after inserting `"a"` before `foo`, we get `afoo`. Then, inserting `"b"` again before `foo` inserts it before the original position of `foo`, pushing `"a"` forward—resulting in `abfoo`.

<br>

Now, if we reverse the order of the two inserts, the result flips to `bafoo`.

```js
return [
  fixer.insertTextBefore(node, "b"),
  fixer.insertTextBefore(node, "a")
]
```
```js
let bafoo = 1
```

<br>

However, if we use `replaceTextRange` instead of `insertTextBefore` for the `"b"` fix, then the order wouldn't matter:
```js
return [
  fixer.replaceTextRange([node.range[0], node.range[0] + 1], "bf"),
  fixer.insertTextBefore(node, "a")
]
```
```js
return [
  fixer.insertTextBefore(node, "a"),
  fixer.replaceTextRange([node.range[0], node.range[0] + 1], "bf"),
];
```
Both the above would produce the same output of `let abfoo = 1`.

<br>

Anyways, I'd keep it as-is, but did that make sense?

test/prefer-string-raw.js
som-sm: There was already a test for `\${`. Added one for `$\{` in [629191f](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2691/commits/629191f3f0a93d7d1d25ec361aa04be119ee582d).

rules/prefer-string-raw.js
```diff
@@ -64,6 +61,45 @@ const create = context => {
...
+			}
+
+			if (cooked.includes(BACKSLASH)) {
+				hasBackslash = true;
+			}
+		}
+
+		if (!hasBackslash) {
+			return;
+		}
```
fisker, line 71: The logic looks good, but I think the above can replace with `Array#some()`?

rules/prefer-string-raw.js
```diff
@@ -64,6 +61,45 @@ const create = context => {
...
+
+		return {
+			node,
+			messageId: MESSAGE_ID,
+			* fix(fixer) {
+				yield * fixSpaceAroundKeyword(fixer, node, context.sourceCode);
+				yield * node.quasis.map(quasi => replaceTemplateElement(fixer, quasi, quasi.value.cooked));
```
fisker: Let's use a for loop to avoid creating a new array, we should also skip unnecessary replacements.

som-sm: Yeah, makes sense, done in [6d81306](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2691/commits/6d81306540c492e7e6194b0e70ffd2182f742e3b).

rules/prefer-string-raw.js
som-sm: Yeah, we can replace the entire `for of` loop with this:
```js
const isFixable = node.quasis.every(({value: {raw, cooked}}) =>
	cooked.at(-1) !== BACKSLASH && unescapeBackslash(raw) === cooked,
);

const hasBackslash = node.quasis.some(quasi =>
	quasi.value.cooked.includes(BACKSLASH),
);

if (!isFixable || !hasBackslash) {
	return;
}
```
Just that it requires two loops instead of one, let me know if this makes more sense?

fisker: Yes, you are right, we need loop twice, but seems more clear/readable to me, what do you think?

[`5ef2291` (#2691)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2691/commits/5ef2291bc66da80285541728f10e74e4295b615c)

som-sm: Yeah, LGTM!

fisker (approved):

fisker:
This one.

som-sm:
> This one.

Ok, I'm not sure if I follow, will `\r` behave any differently from other escaped characters?

If the string contains `\r`, then `raw` shouldn't be equal to `cooked` and it shouldn't complain.

PR #2694, opened 2025-06-24
`no-array-callback-reference`: fix optional chaining expressions ignored

### LINKED ISSUE #2294: `unicorn/no-array-callback-reference` does not report when optional chaining

k-yle:
This PR fixes an inconsistency where unsafe syntax is allowed if the optional-chaining operator is used:

```js
x.map(String.fromCharCode); // flagged
x?.map(String.fromCharCode); // not flagged 🤔
```

Now both cases are flagged by the rule.

Closes #2294

### DIFF
--- a/rules/no-array-callback-reference.js
+++ b/rules/no-array-callback-reference.js
@@ -234,7 +234,6 @@ const create = context => ({
 				minimumArguments: 1,
 				maximumArguments: 2,
 				optionalCall: false,
-				optionalMember: false,
 				computed: false,
 			})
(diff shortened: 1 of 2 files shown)


currentyfx-debug:
I'm sorry for disturbing you

PR #2695, opened 2025-06-25
`prefer-string-raw`: Forbid unnecessary `String.raw`

### LINKED ISSUE #2652: Rule proposal: opposite of `prefer-string-raw`
### Description

During refactoring, some strings may not need `String.raw` anymore, should use a string literal instead.

### Examples

```js
// ❌
String.raw`a`

// ✅
'a'
```

### Proposed rule name

no-string-raw

### Additional Info

Not sure if we want to add a separate rule, or just add to `prefer-string-raw`.

som-sm:
Closes #2652

This PR simply loops over all the `quasis` and checks if any of them contain a backslash. If none of them do, then the rule starts complaining.

Also, the suggestion is a template string(`` ` ``) only if there are `expressions` or the string contains line breaks, otherwise, the suggestion is always a single quotes string (`'`).


test/prefer-string-raw.js
```diff
@@ -47,6 +47,44 @@ test.snapshot({
 	],
 });
 
+test.snapshot({
+	valid: [
+		'a = foo`ab`',
+		'a = foo().bar`ab`',
+		'a = foo.bar()`ab`',
+		'a = String["raw"]`ab`',
```
som-sm, line 87: I think it's fine to ignore this case, wdyt?

fisker: Yes, fine to ignore.

rules/prefer-string-raw.js
```diff
@@ -64,6 +66,40 @@ const create = context => {
...
+		const {quasi, tag} = node;
+
+		if (tag.type !== 'MemberExpression'
+			|| tag.object.type !== 'Identifier'
+			|| tag.property.type !== 'Identifier'
+			|| tag.object.name !== 'String'
+			|| tag.property.name !== 'raw'
+		) {
+			return;
+		}
```
fisker: There is a `isMemberExpression` you can use.

rules/prefer-string-raw.js
```diff
@@ -64,6 +66,40 @@ const create = context => {
...
+			: `'${rawQuasi.slice(1, -1).replaceAll('\'', String.raw`\'`)}'`;
+
+		return {
+			node,
+			messageId: MESSAGE_ID_UNNECESSARY_STRING_RAW,
+			* fix(fixer) {
+				yield fixer.replaceText(node, suggestion);
+			},
+		};
+	});
```
fisker, line 206: We need check this.

```
`\u0040`
'@'
String.raw`\u0040`
'\\u0040'
```

fisker: Maybe check `.cooked === .raw`?

rules/prefer-string-raw.js
```diff
@@ -64,6 +66,40 @@ const create = context => {
...
+		const rawQuasi = context.sourceCode.getText(quasi);
+		const suggestion = quasi.expressions.length > 0 || /\r?\n/.test(rawQuasi)
+			? rawQuasi
+			: `'${rawQuasi.slice(1, -1).replaceAll('\'', String.raw`\'`)}'`;
+
+		return {
+			node,
+			messageId: MESSAGE_ID_UNNECESSARY_STRING_RAW,
+			* fix(fixer) {
+				yield fixer.replaceText(node, suggestion);
```
fisker: There is an ASI problem need to be fix.

```
foo

String.raw``
```

We can't simply remove `String.raw`, we need to insert a semicolon.

rules/prefer-string-raw.js
```diff
@@ -64,6 +66,40 @@ const create = context => {
...
+		);
+
+		if (hasBackslash) {
+			return;
+		}
+
+		const rawQuasi = context.sourceCode.getText(quasi);
+		const suggestion = quasi.expressions.length > 0 || /\r?\n/.test(rawQuasi)
+			? rawQuasi
+			: `'${rawQuasi.slice(1, -1).replaceAll('\'', String.raw`\'`)}'`;
```
fisker, line 177: Will this cause an infinite loop? We should already complainting ``'\''``.

rules/prefer-string-raw.js
som-sm: Sorry, I didn't get this.

<br>

The following tests already work:
```js
valid: [
  'a = String.raw`\\u0040`'
]
```

```js
invalid: [
  'a = String.raw`\u0040`'
]
```

rules/prefer-string-raw.js
som-sm: > We should already complainting `'\''`.

Umm...I don't get this.

<br>

> Will this cause an infinite loop?

It shouldn't, right?


Let's say the code is:

```js
let a = String.raw`a"b`;
```

This fixer will update the code to:

```js
let a = 'a"b';
```

Now, let's say the quotes preference is double quotes (`"`), so that fixer will update the code to:

```js
let a = "a\"b";
```

And the process should stop here, am I missing something?

rules/prefer-string-raw.js
som-sm: Done in [ae37793](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/ae37793dd7d5f274ba9c12f17a67f4b272c6d285).

rules/prefer-string-raw.js
som-sm: Done in [be1ece7](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/be1ece719cee06d8d7e32502dfbe5d9653f591ea).

rules/prefer-string-raw.js
fisker: I forgot about details, but we supposed to suggest

```js
// ❌
'\''

// ✅
String.raw`'`
```

before this change.

rules/prefer-string-raw.js
fisker: Okay, I thought we suggest `` `\u0040` `` over `` String.raw`\u0040` ``, I must made a mistake.

Anyway, do you think we can  check `.cooked === .raw` instead of checking contains `\`?

rules/prefer-string-raw.js
```diff
@@ -64,6 +67,41 @@ const create = context => {
 			},
 		};
 	});
+
+	context.on('TaggedTemplateExpression', node => {
+		const {quasi, tag} = node;
+		const {sourceCode} = context;
+
+		if (!isMemberExpression(tag, {object: 'String', property: 'raw'})) {
```
fisker: Currently

```js
String?.raw``
```

is a syntax error, but just to be safe, let's add `optional: false`.

fisker: Actually, the typescript-eslint parser seems to allow it.

rules/prefer-string-raw.js
fisker: There is also a potential risk (very low, we can ignore, but just want to point out) of converting it to a string literal, that the string can become a directive.

```
String.raw`use strict` // <-- an expression
```

```
'use strict' // <-- a directive
```

rules/prefer-string-raw.js
```diff
@@ -64,6 +67,41 @@ const create = context => {
...
+			return;
+		}
+
+		const rawQuasi = sourceCode.getText(quasi);
+		const suggestion = quasi.expressions.length > 0 || /\r?\n/.test(rawQuasi)
+			? rawQuasi
+			: `'${rawQuasi.slice(1, -1).replaceAll('\'', String.raw`\'`)}'`;
+
+		return {
+			node,
```
fisker: Let's change the report range to point to `String.raw`.

```suggestion
			node: tag,
```

rules/prefer-string-raw.js
```diff
@@ -64,6 +67,41 @@ const create = context => {
...
+			messageId: MESSAGE_ID_UNNECESSARY_STRING_RAW,
+			* fix(fixer) {
+				const tokenBefore = sourceCode.getTokenBefore(node);
+				if (needsSemicolon(tokenBefore, sourceCode, suggestion)) {
+					yield fixer.insertTextAfter(tokenBefore, ';');
```
fisker: This should insert before 
the `TaggedTemplateExpression`,

```js
a
String.raw``;
```

Should be

```js
a
;``;
```

Since this is how people write code in "no semi" style.

test/prefer-string-raw.js
```diff
@@ -47,6 +47,53 @@ test.snapshot({
...
+		`,
+		outdent`
+			a
+			String.raw\`abc\`
+		`,
+		// ASI
+		outdent`
+			a
+			String.raw\`a\${b}\`
+		`,
```
fisker, line 130: Please add a test for


```
function foo() {
	return String.raw
			// Comment
			``
}
```

rules/prefer-string-raw.js
som-sm: > I forgot about details, but we supposed to suggest
> 
> ```js
> // ❌
> '\''
> 
> // ✅
> String.raw`'`
> ```
> 
> before this change.

No, we don't suggest anything for `'\''`, because it doesn't contain any backslashes. 

We do suggest  ``String.raw`\'` `` for `'\\\''`, and this should be fine because we wouldn't suggest anything further for ``String.raw`\'` ``.

fisker: Alright, I checked the original PR, looks like I ignored that, https://github.com/sindresorhus/eslint-plugin-unicorn/blob/2dc27d90c79a3ad97345e547c08482fd06ca455b/rules/prefer-string-raw.js#L45

Do you think we shoud prefer ``String.raw`'` `` over `'\''`?

fisker: I guess I ignored it because Prettier can convert it to `"'"` to avoid escaping.

som-sm: > Do you think we shoud prefer `` String.raw`'` `` over `'\''`?

IMO no, because `String.raw` has nothing to do with quotes.

If we change to `` String.raw`'` ``, it will then get changed to `` `'` ``, which means we are essentially changing `'\''` to `` `'` ``, which seems unrelated to this rule.

fisker: Make sense.

som-sm: > it will then get changed to `` `'` ``

Quick correction: It will actually get changed to `'\''`, which will then cause a circular issue.

But anyways, it's better not to suggest quote related changes in this rule.

rules/prefer-string-raw.js
som-sm: > Actually, the typescript-eslint parser seems to allow it.

Yeah, but then `tag` would become a `ChainExpression`, so the `isMemberExpression` conditional would anyways fail.

image<br>

Anyways, I've added `optional: false` in ~[0c30a3c](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/0c30a3c02c3eb113121ab087a71380fb26fda12b)~ [15781f5](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/15781f55d24917a205f25aff9ec5d3187aa0621c).

rules/prefer-string-raw.js
som-sm: > Anyway, do you think we can check `.cooked === .raw` instead of checking contains `\`?

Yeah, we can, updated in ~[2eb20bf](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/2eb20bf18f19e53cdf5748839526a6fa1f6fbb98)~.

som-sm: The above update, surfaced a _very_ interesting issue. 

So, after this update, I'm getting a parsing error in the `puppeteer` codebase at the following line:
https://github.com/puppeteer/puppeteer/blob/832831d21c42ff542300c194c2b4f07274d8f82f/test/src/queryhandler.spec.ts#L373

<br>

Here's a simple reproduction of the same:

Source code:
```js
let a = "\\38";
```
This _correctly_ gets fixed to:
```js
let a = String.raw`\38`;
```
However, the above `` String.raw`\38` `` expression fails with the `typescript` parser, throwing the following error:
```
Parsing error: Octal escape sequences are not allowed. Use the syntax '\x03'.
```

The `includes(BACKSLASH)` check wasn’t failing because it allowed the function to exit early when a backslash was detected.

But, with the `.cooked === .raw` check, it fails because the `cooked` value seems to be incorrect in this case. So, both `cooked` and `raw` values are `'\\38'`—causing the check to pass, but the `cooked` value should instead have been `'38'`. And, it looks it was `'38'` in `v5`, refer to this AST explorer example—https://astexplorer.net/#/gist/554e60873250681db6c92107269b3df6/ec6f75f2d3e991cd293013973e54825f585d4811
```js
value: { cooked: '\\38', raw: '\\38' }
```

This happens to work with the `babel` parser because in that case the `cooked` value is `null`, which is _also_ incorrect but happens to get caught by our check.
```js
value: { raw: '\\38', cooked: null }
```

<br>

So, I think it's better to keep the `includes(BACKSLASH)` check.

rules/prefer-string-raw.js
som-sm: Updated in [ee17e72](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/ee17e724108ae44aa36f4a76a03d188350aacd14).

rules/prefer-string-raw.js
som-sm: Updated in [15781f5](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/15781f55d24917a205f25aff9ec5d3187aa0621c).

test/prefer-string-raw.js
som-sm: Comments like these weren't being preserved, had to update the implementation a bit, refer [cc0a128](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2695/commits/cc0a128ecce34fbd9dda948d4dd172f2926747fc).

rules/prefer-string-raw.js
som-sm: The above update, surfaced an _interesting_ issue. 

So, after this update, I'm getting a parsing error in the `puppeteer` codebase at the following line:
https://github.com/puppeteer/puppeteer/blob/832831d21c42ff542300c194c2b4f07274d8f82f/test/src/queryhandler.spec.ts#L373

<br>

Here's a simple reproduction of the same:

Source code:
```js
let a = "\\38";
```
This _correctly_ gets fixed to:
```js
let a = String.raw`\38`;
```
However, the above `` String.raw`\38` `` expression fails with the `typescript` parser, throwing the following error:
```
Parsing error: Octal escape sequences are not allowed. Use the syntax '\x03'.
```

The `includes(BACKSLASH)` check wasn’t failing because it allowed the function to exit early when a backslash was detected in the `raw` value.

But, with the `.cooked === .raw` check, it fails because both `cooked` and `raw` values here are the same (`'\\38'`), causing the check to pass. 
`cooked` value can't be `'\38'` because that `'\38'` is syntactically incorrect, maybe because of this the parser keeps the `cooked` value same as `raw`.
```js
value: { cooked: '\\38', raw: '\\38' }
```

This works fine with the `babel` parser because looks like it sets the `cooked` value to `null` in cases like these.
```js
value: { raw: '\\38', cooked: null }
```

<br>

So, I think it's better to keep the `includes(BACKSLASH)` check because it operates on the `raw` value, which seems to be a better choice in this case.

LMK if this makes sense, I'll add the ``let a = String.raw`\38`;`` test with the `typescript` parser.

rules/prefer-string-raw.js
```diff
@@ -64,6 +67,42 @@ const create = context => {
...
+			node: tag,
+			messageId: MESSAGE_ID_UNNECESSARY_STRING_RAW,
+			* fix(fixer) {
+				const tokenBefore = sourceCode.getTokenBefore(node);
+				if (needsSemicolon(tokenBefore, sourceCode, suggestion)) {
+					yield fixer.insertTextBefore(node, ';');
+				}
+
+				yield fixer.replaceText(node.quasi, suggestion);
+				yield fixer.remove(node.tag);
```
fisker, line 203: This is incorrect, since tag can be parenthesis.

```
((String.raw))``
```

rules/prefer-string-raw.js
fisker: This is bug in typescript-eslint, can you report to them? I can do it if you don't want to.

test/prefer-string-raw.js
fisker: The comment is not the point, in this case, we can't simply remove `String.raw` need add parenthesis after `return`.

fisker: Ideally, it should fix to 

```js
function foo() {
	return (
			// Comment
			``)
}
```

There are similar cases exists, please search other rules.

fisker: Search `addParenthesizesToReturnOrThrowExpression`

rules/prefer-string-raw.js
fisker: You can use `removeParentheses`

test/prefer-string-raw.js
som-sm: Gotcha, apologies for the confusion.

rules/prefer-string-raw.js
fisker: https://github.com/typescript-eslint/typescript-eslint/issues/11352

rules/prefer-string-raw.js
som-sm: I'd appreciate if you could, I'm fairly new to parsers, so I might not be able to explain it clearly.

fisker: No problem.

fisker: https://github.com/typescript-eslint/typescript-eslint/issues/11353

rules/prefer-string-raw.js
```diff
@@ -64,6 +67,42 @@ const create = context => {
...
+		if (hasBackslash) {
+			return;
+		}
+
+		const rawQuasi = sourceCode.getText(quasi);
+		const suggestion = quasi.expressions.length > 0 || /\r?\n/.test(rawQuasi)
+			? rawQuasi
+			: `'${rawQuasi.slice(1, -1).replaceAll('\'', String.raw`\'`)}'`;
+
+		return {
```
fisker, line 179: There is also an edge case that can't autofix (possible, but hard).

```js
String.raw`use strict`;
```

This will become a "Directive", I don't think we need handle it, but let's leave a comment.

Or simply remove the fix if it's an `ExpressionStatement` if you want.

som-sm: Added a test with note for this.

rules/prefer-string-raw.js
som-sm: Added `removeParentheses`.

test/prefer-string-raw.js
som-sm: Added `addParenthesizesToReturnOrThrowExpression`.

Issue #2214, opened 2023-11-01
Rule proposal: no-restricted-function-dependencies


mmkal:
### Description

It'd be nice to have a rule that makes sure certain functions are free of external dependencies. For example, this won't work:

```ts
import makeSynchronous from 'make-synchronous'

export const fetchSync = () => {
  const url = 'https://example.com'
  const getText = makeSynchronous(async () => {
    const res = await fetch(url)
    return res.text()
  })
  console.log(getText())
}
```

(from make-synchronous docs: "The given function is executed in a subprocess, so you cannot use any variables/imports from outside the scope of the function. You can pass in arguments to the function. To import dependencies, use await import(…) in the function body.")

There are other scenarios where functions should/must avoid using variables from outside their scopes, like [server actions](https://nextjs.org/docs/app/api-reference/functions/server-actions) (maybe?), or if using `myFn.toString()`.

I don't know exactly what the conditions for banning scope-external variables should be, but maybe `makeSynchronous(...)`-style calling would be a good start, given it's part of the Sindresorhus Cinematic Universe. Config could look similar to [no-restricted-syntax](https://eslint.org/docs/latest/rules/no-restricted-syntax):

```
'restrict-function-dependencies': ['error', [{
  selector: 'CallExpression[name=makeSynchronous]',
]],
```

### Fail

```ts
import makeSynchronous from 'make-synchronous'

export const fetchSync = () => {
  const url = 'https://example.com'
  const getText = makeSynchronous(async () => {
    const res = await fetch(url)
    return res.text()
  })
  console.log(getText())
}
```

### Pass

```ts
import makeSynchronous from 'make-synchronous'

export const fetchSync = () => {
  const getText = makeSynchronous(async () => {
    const url = 'https://example.com'
    const res = await fetch(url)
    return res.text()
  })
  console.log(getText())
}
```

```ts
import makeSynchronous from 'make-synchronous'

export const fetchSync = () => {
  const getText = makeSynchronous(async (url) => {
    const res = await fetch(url)
    return res.text()
  })
  console.log(getText('https://example.com'))
}
```

### Additional Info

The devil would be in the details here, and it'd probably be impossible to completely get rid of false positives or false negatives but it could still be useful to (try to) remind people when they can't use normal function scoping.

I had a look for similar ideas in the docs/issues but maybe I missed something? Does this exist?

sindresorhus:
I like it.


---

[...]

PR #2701, opened 2025-07-02
Add `isolated-functions` rule

mmkal:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2214

From docs:

---

# Prevent usage of variables from outside the scope of isolated functions

💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).

Some functions need to be isolated from their surrounding scope due to execution context constraints. For example, functions passed to [`makeSynchronous()`](https://github.com/sindresorhus/make-synchronous) are executed in a worker or subprocess and cannot access variables from outside their scope. This rule helps identify when functions are using external variables that may cause runtime errors.

Common scenarios where functions must be isolated:

- Functions passed to `makeSynchronous()` (executed in worker)
- Functions that will be serialized via `Function.prototype.toString()`
- Server actions or other remote execution contexts
- Functions with specific JSDoc annotations

By default, this rule allows global variables (like `console`, `fetch`, etc.) in isolated functions, but prevents usage of variables from the surrounding scope.

## Examples

```js
import makeSynchronous from 'make-synchronous';

export const fetchSync = () => {
	const url = 'https://example.com';

	const getText = makeSynchronous(async () => {
		const res = await fetch(url); // ❌ 'url' is not defined in isolated function scope
		return res.text();
	});

	console.log(getText());
};

// ✅ Define all variables within isolated function's scope
export const fetchSync = () => {
	const getText = makeSynchronous(async () => {
		const url = 'https://example.com'; // Variable defined within function scope
		const res = await fetch(url);
		return res.text();
	});

	console.log(getText());
};

// ✅ Alternative: Pass as parameter
export const fetchSync = () => {
	const getText = makeSynchronous(async (url) => { // Variable passed as parameter
		const res = await fetch(url);
		return res.text();
	});

	console.log(getText('https://example.com'));
};
```
```

```js
const foo = 'hi';

/** @isolated */
function abc() {
	return foo.slice(); // ❌ 'foo' is not defined in isolated function scope
}

// ✅
/** @isolated */
function abc() {
	const foo = 'hi'; // Variable defined within function scope
	return foo.slice();
}
```

## Options

Type: `object`

### functions

Type: `string[]`\
Default: `['makeSynchronous']`

Array of function names that create isolated execution contexts. Functions passed as arguments to these functions will be considered isolated.

### selectors

Type: `string[]`\
Default: `[]`

Array of [ESLint selectors](https://eslint.org/docs/developer-guide/selectors) to identify isolated functions. Useful for custom naming conventions or framework-specific patterns.

```js
{
	'unicorn/isolated-functions': [
		'error',
		{
			selectors: [
				'FunctionDeclaration[id.name=/lambdaHandler.*/]'
			]
		}
	]
}
```

### comments

Type: `string[]`\
Default: `['@isolated']`

Array of comment strings that mark functions as isolated. Functions with JSDoc comments containing these strings will be considered isolated.

```js
{
	'unicorn/isolated-functions': [
		'error',
		{
			comments: [
				'@isolated',
				'@remote'
			]
		}
	]
}
```

### globals

Type: `boolean | string[]`\
Default: `true`

Controls how global variables are handled:

- `false`: Global variables are not allowed in isolated functions
- `true` (default): All globals from ESLint's language options are allowed
- `string[]`: Only the specified global variable names are allowed

```js
{
	'unicorn/isolated-functions': [
		'error',
		{
			globals: ['console', 'fetch'] // Only allow these globals
		}
	]
}
```

### Custom function names

```js
{
	'unicorn/isolated-functions': [
		'error',
		{
			functions: [
				'makeSynchronous',
				'createWorker',
				'serializeFunction'
			]
		}
	]
}
```

### Lambda function naming convention

```js
{
	'unicorn/isolated-functions': [
		'error',
		{
			selectors: [
				'FunctionDeclaration[id.name=/lambdaHandler.*/]'
			]
		}
	]
}
```

```js
const foo = 'hi';

function lambdaHandlerFoo() { // ❌ Will be flagged as isolated
	return foo.slice();
}

function someOtherFunction() { // ✅ Not flagged
	return foo.slice();
}

createLambda({
	name: 'fooLambda',
	code: lambdaHandlerFoo.toString(), // Function will be serialized
});
```

### Allowing specific globals

```js
{
	'unicorn/isolated-functions': [
		'error',
		{
			globals: [
				'console',
				'fetch',
				'URL'
			]
		}
	]
}
```

```js
// ✅ All globals used are explicitly allowed
makeSynchronous(async () => {
	console.log('Starting...'); // ✅ Allowed global
	const response = await fetch('https://api.example.com'); // ✅ Allowed global
	const url = new URL(response.url); // ✅ Allowed global
	return response.text();
});

makeSynchronous(async () => {
	const response = await fetch('https://api.example.com', {
		headers: {
			'Authorization': `Bearer ${process.env.API_TOKEN}` // ❌ 'process' is not in allowed globals
		}
	});
	const url = new URL(response.url);
	return response.text();
});
``` 

---

Note: I didn't go as far as suggested in that issue https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2214#issuecomment-1801269748 - that is, to track imports from specific modules like `import makeSync from 'make-synchronous'`, to allow for arbitrary aliasing from that package. Instead I went for fairly specific naming conventions. But I thought the rule still had value while relying on naming convention, so opening a pull request now since I don't know when I'll get some more time to spend on it. I think the module-import-tracking thing could go in later as a new feature, if there's enough interest in it.

### DIFF
--- a/docs/rules/isolated-functions.md
+++ b/docs/rules/isolated-functions.md
@@ -112,7 +112,7 @@ Array of comment strings that mark functions as isolated. Functions with inline,
 }
 ```
 
-### globals
+### overrideGlobals
 
 Type: `object`\
 Default: `undefined` (uses ESLint's language options globals)
@@ -128,7 +128,7 @@ Controls how global variables are handled. When not specified, uses ESLint's lan
 	'unicorn/isolated-functions': [
 		'error',
 		{
-			globals: {
+			overrideGlobals: {
 				console: 'writable',     // Allowed and writable
 				fetch: 'readonly',       // Allowed but readonly
 				process: 'off'           // Not allowed
@@ -200,28 +200,6 @@ makeSynchronous(async () => {
 });
 ```
 
-### Disallowing all globals
-
-```js
-{
-	'unicorn/isolated-functions': [
-		'error',
-		{
-			globals: {} // Empty object disallows all globals except language globals
-		}
-	]
-}
-```
-
-```js
-// ❌ All globals are disallowed
-makeSynchronous(async () => {
-	console.log('Starting...'); // ❌ 'console' is not allowed
-	const response = await fetch('https://api.example.com'); // ❌ 'fetch' is not allowed
-	return response.text();
-});
-```
-
 ### Allowing specific globals
 
 ```js
@@ -229,7 +207,7 @@ makeSynchronous(async () => {
 	'unicorn/isolated-functions': [
 		'error',
 		{
-			globals: {
+			overrideGlobals: {
 				console: 'writable',     // Allowed and writable
 				fetch: 'readonly',       // Allowed but readonly
 				URL: 'readonly'          // Allowed but readonly
--- a/rules/isolated-functions.js
+++ b/rules/isolated-functions.js
@@ -16,11 +16,12 @@ const parseEsquerySelector = selector => {
 	return parsedEsquerySelectors.get(selector);
 };
 
-/** @type {{functions: string[], selectors: string[], comments: string[], globals?: import('eslint').Linter.Globals}} */
+/** @type {{functions: string[], selectors: string[], comments: string[], overrideGlobals?: import('eslint').Linter.Globals}} */
 const defaultOptions = {
 	functions: ['makeSynchronous'],
 	selectors: [],
 	comments: ['@isolated'],
+	overrideGlobals: {},
 };
 
 /** @param {import('eslint').Rule.RuleContext} context */
@@ -36,7 +37,8 @@ const create = context => {
 
 	const allowedGlobals = {
 		...(globals[`es${context.languageOptions.ecmaVersion}`] ?? globals.builtins),
-		...(options.globals ?? context.languageOptions.globals),
+		...context.languageOptions.globals,
+		...options.overrideGlobals,
 	};
 
 	/** @param {import('estree').Node} node */
@@ -151,7 +153,7 @@ const schema = [
 		type: 'object',
 		additionalProperties: false,
 		properties: {
-			globals: {
+			overrideGlobals: {
 				additionalProperties: {
 					anyOf: [{type: 'boolean'}, {type: 'string', enum: ['readonly', 'writable', 'writeable', 'off']}],
 				},
--- a/test/isolated-functions.js
+++ b/test/isolated-functions.js
@@ -54,9 +54,9 @@ test({
 			code: 'makeSynchronous(() => new Array())',
 		},
 		{
-			name: 'global Array w globals: {} still works',
+			name: 'global Array w overrideGlobals: {} still works',
 			code: 'makeSynchronous(() => new Array())',
-			options: [{globals: {}}],
+			options: [{overrideGlobals: {}}],
 		},
 		{
 			name: 'can implicitly allow global variables from language options',
@@ -70,7 +70,7 @@ test({
 		{
 			name: 'allow global variables separate from language options',
 			languageOptions: {globals: {abc: true}},
-			options: [{globals: {foo: true}}],
+			options: [{overrideGlobals: {foo: true}}],
 			code: outdent`
 				makeSynchronous(function () {
 					return foo.slice();
@@ -218,19 +218,8 @@ test({
 			],
 		},
 		{
-			name: 'all global variables can be explicitly disallowed',
-			languageOptions: {globals: {foo: true}},
-			options: [{globals: {}}],
-			code: outdent`
-				makeSynchronous(function () {
-					return foo.slice();
-				});
-			`,
-			errors: [fooInMakeSynchronousError],
-		},
-		{
 			name: 'individual global variables can be explicitly disallowed',
-			options: [{globals: {URLSearchParams: 'readonly', URL: 'off'}}],
+			options: [{overrideGlobals: {URLSearchParams: 'readonly', URL: 'off'}}],
 			code: outdent`
 				makeSynchronous(function () {
 					return new URL('https://example.com?') + new URLSearchParams({a: 'b'}).toString();
@@ -282,7 +271,7 @@ test({
 		{
 			name: 'can explicitly turn off ecmascript globals',
 			code: 'makeSynchronous(() => new Array())',
-			options: [{globals: {Array: 'off'}}],
+			options: [{overrideGlobals: {Array: 'off'}}],
 			errors: [error({name: 'Array', reason: 'callee of function named "makeSynchronous"'})],
 		},
 	],


rules/isolated-functions.js
```diff
@@ -0,0 +1,165 @@
...
+		if (!reason) {
+			return;
+		}
+
+		const nodeScope = sourceCode.getScope(node);
+
+		for (const reference of nodeScope.through) {
+			const {identifier} = reference;
+
+			if (allowedGlobals.includes(identifier.name)) {
```
fisker: Use a `Set` please.

docs/rules/isolated-functions.md
```diff
@@ -0,0 +1,252 @@
...
+Common scenarios where functions must be isolated:
+
+- Functions passed to `makeSynchronous()` (executed in worker)
+- Functions that will be serialized via `Function.prototype.toString()`
+- Server actions or other remote execution contexts
+- Functions with specific JSDoc annotations
+
+By default, this rule allows global variables (like `console`, `fetch`, etc.) in isolated functions, but prevents usage of variables from the surrounding scope.
+
+## Fail
```
sindresorhus: The examples should follow our new style where we try to group fail and pass closer to each other: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/escape-case.md#examples https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-global-this.md#examples

fisker:
> so end-users don't have to manually install globals

Users should always have `globals` installed, I don't think we should worry about it, we can accept the `languageOptions.globals` style, anyway, I think language builtins should always allowed.

test/isolated-functions.js
```diff
@@ -0,0 +1,218 @@
...
+
+test({
+	/** @type {import('eslint').RuleTester.InvalidTestCase[]} */
+	invalid: [
+		{
+			name: 'out of scope variable under makeSynchronous (arrow function)',
+			code: stripIndent(`
+				const foo = 'hi';
+				makeSynchronous(() => foo.slice());
+			`),
```
fisker: Use `outdent` please.

rules/isolated-functions.js
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	return parsedEsquerySelectors.get(selector);
+};
+
+/** @type {{functions: string[], selectors: string[], comments: string[], globals?: import('eslint').Linter.Globals}} */
+const defaultOptions = {
+	functions: ['makeSynchronous'],
+	selectors: [],
+	comments: ['@isolated'],
```
fisker, line 23: Should it be just `isolated` instead of `@isolated` ?

docs/rules/isolated-functions.md
```diff
@@ -0,0 +1,303 @@
...
+
+### globals
+
+Type: `object`\
+Default: `undefined` (uses ESLint's language options globals)
+
+Controls how global variables are handled. When not specified, uses ESLint's language options globals. When specified as an object, each key is a global variable name and the value controls its behavior:
+
+- `'readonly'`: Global variable is allowed but cannot be written to (deprecated form `false` also accepted)
+- `'writable'`: Global variable is allowed and can be read/written (deprecated forms `true` and `'writeable'` also accepted)
```
sindresorhus: While it is accepted, I don't see the point of documenting `writeable` (the typo) as accepted.

rules/isolated-functions.js
mmkal: sure

rules/isolated-functions.js
```diff
@@ -0,0 +1,171 @@
...
+				&& (commentableNode.parent.type === 'VariableDeclarator' || commentableNode.parent.type === 'VariableDeclaration')
+			) {
+				// Search up to find jsdoc comments on the parent declaration `/** @isolated */ const foo = () => abc`
+				commentableNode = commentableNode.parent;
+				previousToken = sourceCode.getTokenBefore(commentableNode, {includeComments: true});
+			}
+
+			if (previousToken?.type === 'Block' || previousToken?.type === 'Line') {
+				const previousComment = previousToken.value.trim().toLowerCase();
+				const match = options.comments.find(comment => previousComment.includes(comment));
```
fisker: I don't think we should check with `.includes()`, the comment should exactly the same as configured.

mmkal: Seems unrealistic/not useful to have to write out the exact comment. That way you'd lose protection if someone innocently changed a comment from `/** isolated */` to `/** isolated - do not reference any external variables because this is serialized via .toString() */`. So you'd be disincentivising good practices.

Reducing the Scunthorpe problem is why I originally had it as `'@isolated'`, but you suggested I change it to `'isolated'`: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2701#discussion_r2206128794

Compromise: either `'isolated'` or `'@isolated'` at the start of any line in the comment?

so this fn would be isolated:
```ts
// isolated: this function is serialized via .toString()
const fn1 = () => 'hi'
```

but this wouldn't

```ts
// unlike the one above, this function is not isolated, go ahead and use external variables
const fn2 = () => process.env.HI
```

fisker: What if user wrote something totally unrelated, but contains `isolated`, I prefer explicit comment.

sindresorhus: I think I prefer `@isolated` to make it clear it's some kind of tag. I prefer it to be strict. We should not allow any inconsistent format, but we should allow an explanatory string like ESLint, so maybe either `@isolated` or `@isolated - Foo bar`.

mmkal: done

rules/isolated-functions.js
```diff
@@ -0,0 +1,174 @@
...
+	const reasonForBeingIsolatedFunction = node => {
+		if (options.comments.length > 0) {
+			let previousToken = sourceCode.getTokenBefore(node, {includeComments: true});
+			let commentableNode = node;
+			while (
+				(previousToken?.type !== 'Block' && previousToken?.type !== 'Line')
+				&& (commentableNode.parent.type === 'VariableDeclarator' || commentableNode.parent.type === 'VariableDeclaration')
+			) {
+				// Search up to find jsdoc comments on the parent declaration `/** @isolated */ const foo = () => abc`
+				commentableNode = commentableNode.parent;
```
fisker: This seems  incorrect to me, shouldn't we only search  comments before node? Why searching parents?

mmkal: Because here we're looking at function types: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/c44ea2641b55a6f244f648010b372db359ff4707/rules/ast/function-types.js - and in practice comments are usually on their "parent" declarators rather than the function node itself. e.g.

```ts
// comment on VariableDeclaration
const a = () => 1

// comment on ExportNamedDeclaration
export const b = () => 1

// comment on ExportNamedDeclaration
export function c() {}

// comment ExportDefaultDeclaration
export default () => 1
```

https://astexplorer.net/#/gist/1158a35a4e586559bbdba7e826e7caf2/4476ed5f4bb9ffd8480d4bf19be1bdd0dfbe5ba4

But I refactored this a bit to clarify via some comments, and added support for exports which I had missed.

fisker: Yeah, make sense, and the new code is clear, amazing.

rules/isolated-functions.js
```diff
@@ -71,22 +71,39 @@ const create = context => {
...
+			&& (commentableNode.parent.type === 'VariableDeclarator'
+				|| commentableNode.parent.type === 'VariableDeclaration'
+				|| commentableNode.parent.type === 'ExportNamedDeclaration'
+				|| commentableNode.parent.type === 'ExportDefaultDeclaration')
+		) {
+			commentableNode = commentableNode.parent;
+			previousToken = sourceCode.getTokenBefore(commentableNode, {includeComments: true});
+		}
+
+		if (previousToken?.type === 'Block' || previousToken?.type === 'Line') {
```
fisker: Let's extract an `isComment` helper?

rules/isolated-functions.js
```diff
@@ -0,0 +1,191 @@
...
+	const reasonForBeingIsolatedFunction = node => {
+		if (options.comments.length > 0) {
+			let previousComment = findComment(node);
+
+			if (previousComment) {
+				previousComment = previousComment
+					.replace(/(\*\s*)*/, '') // JSDoc comments like `/** @isolated */` are parsed into `* @isolated`. And `/**\n * @isolated */` is parsed into `*\n * @isolated`
+					.trim()
+					.toLowerCase();
+				const match = options.comments.find(comment => previousComment === comment || previousComment.startsWith(`${comment} - `));
```
fisker:
```suggestion
				const match = options.comments.find(comment => previousComment === comment || previousComment.startsWith(`${comment} - `) ||  previousComment.startsWith(`${comment} -- `));
```

rules/isolated-functions.js
```diff
@@ -0,0 +1,193 @@
+import esquery from 'esquery';
+import functionTypes from './ast/function-types.js';
+
+const MESSAGE_ID_EXTERNALLY_SCOPED_VARIABLE = 'externally-scoped-variable';
+const messages = {
+	[MESSAGE_ID_EXTERNALLY_SCOPED_VARIABLE]: 'Variable {{name}} not defined in scope of isolated function. Function is isolated because: {{reason}}.',
```
fisker, line 7: Is this reason necessary to show in the message? When user see "in scope of isolated function", they should realize that it's isolated.

fisker:
I didn't see handling of ES builtin globals.

```
/** @isolated */
function foo() {
	const map = new Map() // Will this `Map` be reported?
}
```

I think it should be ignored unless `globals: {Map: 'off'}`

rules/isolated-functions.js
fisker: I prefer to leave the "isolated reason" alone, but we can keep the "global variable is not writable", it can be useful.

mmkal: More information is more useful? If someone is new to a project they may not know that this rule exists, that the function is isolated, or what the team's conventions are. It's just a bit more of a hint about why the rule applies. LLMs will also likely be better at addressing the problem based on an error message which includes the reason. What's the downside to including it?

mmkal:
@fisker i see - good idea, done.

docs/rules/isolated-functions.md
```diff
@@ -0,0 +1,303 @@
...
+```js
+import makeSynchronous from 'make-synchronous';
+
+export const fetchSync = () => {
+	const url = 'https://example.com';
+
+	const getText = makeSynchronous(async () => {
+		const response = await fetch(url); // ❌ 'url' is not defined in isolated function scope
+		return response.text();
+	});
```
fisker: Let's remove the outer `fetchSync`? Just make the code shorter to read.

rules/isolated-functions.js
```diff
@@ -0,0 +1,197 @@
...
+	 Find the string "reason" that a function (node) should be considered isolated. For passing in to `context.report(...)` when out-of-scope variables are found. Returns undefined if the function should not be considered isolated.
+	 @param {import('estree').Node & {parent?: import('estree').Node}} node
+	 */
+	const reasonForBeingIsolatedFunction = node => {
+		if (options.comments.length > 0) {
+			let previousComment = findComment(node);
+
+			if (previousComment) {
+				previousComment = previousComment
+					.replace(/(\*\s*)*/, '') // JSDoc comments like `/** @isolated */` are parsed into `* @isolated`. And `/**\n * @isolated */` is parsed into `*\n * @isolated`
```
fisker:
```suggestion
					.replace(/(?:\*\s*)*/, '') // JSDoc comments like `/** @isolated */` are parsed into `* @isolated`. And `/**\n * @isolated */` is parsed into `*\n * @isolated`
```

rules/isolated-functions.js
```diff
@@ -0,0 +1,197 @@
+import globals from 'globals';
+import esquery from 'esquery';
+import functionTypes from './ast/function-types.js';
```
fisker:
```suggestion
import {functionTypes} from './ast/index.js';
```

docs/rules/isolated-functions.md
```diff
@@ -0,0 +1,303 @@
...
+```
+
+### Disallowing all globals
+
+```js
+{
+	'unicorn/isolated-functions': [
+		'error',
+		{
+			globals: {} // Empty object disallows all globals
```
fisker:
```suggestion
			globals: {} // Empty object disallows all globals except language globals
```

fisker: Maybe we should mention that language globals are always enabled; they need to be disabled explicitly.

fisker: Do you think worth changing the option to

```
{
	esGlobals: boolean,
	eslintGlobals: boolean,
	overrideGlobals: {
		...
	}
}
```

?

fisker (approved):
Everything looks good, just some nit comments.

docs/rules/isolated-functions.md
```diff
@@ -21,37 +21,31 @@ By default, this rule uses ESLint's language options globals and allows global v
...
-		return response.text();
-	});
+const getText = makeSynchronous(async () => {
+	const response = await fetch(url); // ❌ 'url' is not defined in isolated function scope
+	return response.text();
+});
 
-	console.log(getText());
-};
+console.log(getText());
```
fisker: Let's remove this `console.log` too

docs/rules/isolated-functions.md
mmkal: How about I change `globals` to `overrideGlobals` and we can add the other ones if/when somebody asks for it?

mmkal: Done. I think the behaviour is less confusing now - before `globals: {}` could turn off _everything_ in language options, and I don't think it was good for `globals: undefined` to be so dramatically different from `globals: {}`. 

Now, `overrideGlobals` does just that - it overrides globals that otherwise are defined, so if you want to turn off globals you turn them off explicitly.

mmkal: commit: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2701/commits/d9836a08ddfcadf56348b3efbccbd5c0cda347d6

fisker (approved):

PR #2704, opened 2025-07-07
`prefer-includes`: Fix compatibility with Vue SFC

### LINKED ISSUE #2702: .vue file `indexOf` combined with the `not equal to` operator Exception
This is the initial code
```vue
<script setup lang="ts">
console.log([].indexOf(1) != -1);   
</script>

<template>
  <div></div>
</template>
```

The vscode ide will freeze for a while, and then the final vue file will be linted as the following code
```vue
<script setup lang="ts">
console.log([].indexOf(1) != -1);   
</script>

<template>
  <div></div>
</template>.includes(1));
</script>

<template>
  <div/>
</template>.includes(1));
</script>

...
```

`unicorn/prefer-includes`

 [repo](https://github.com/jevin98/eslint-plugin-vue-test-demo) The above is the repository address. After you clone the repository, run `pnpm install` and then save the file in `1.vue`

#2643 I saw an issue of type `issue` but I don't know if it's the same issue. If it's a duplicate, please close the issue.

fisker:
Fixes #2702


rules/prefer-includes.js
```diff
@@ -18,7 +18,11 @@ const getProblem = (context, node, target, argumentsNodes) => {
 	const tokenStore = sourceCode.parserServices.getTemplateBodyTokenStore?.() ?? sourceCode;
 
 	const memberExpressionNode = target.parent;
-	const dotToken = tokenStore.getTokenBefore(memberExpressionNode.property);
+	let dotToken = tokenStore.getTokenBefore(memberExpressionNode.property);
+	if (dotToken.value !== '.') {
+		dotToken = sourceCode.getTokenBefore(memberExpressionNode.property);
+	}
```
fisker: Hello @ota-meshi, sorry for bothering you.

In this case I need find the dot in

```
foo.indexOf
```

When it's in element attribute eg `v-if`, `sourceCode.parserServices.getTemplateBodyTokenStore().getTokenBefore(memberExpressionNode.property)`, I get the correct token, but when it's in `<script>`, I get the wrong token, what's the correct approach here? Do I need check if the JS code in attribute/template?

ota-meshi: Sorry for the late reply.

When parsing with vue-eslint-parser, we need to use `getTemplateBodyTokenStore()` to get a token inside a `<template>`, and `sourceCode` to get a token inside a `<script>`.

```js
	const tokenStore = (
			sourceCode.ast.templateBody
			&& sourceCode.ast.templateBody.range[0] <= target.range[0]
			&& target.range[1] <= sourceCode.ast.templateBody.range[1]
			&& sourceCode.parserServices.getTemplateBodyTokenStore?.()
		)
		|| sourceCode
```

fisker: Thanks, so we are using it all wrong, I'll update.

Issue #2634, opened 2025-04-29
Remove `empty-brace-spaces` from recommended


golopot:
### Description

Formatting problems are better handled by code formatters like prettier. This code formatting rule should be removed from recommended rules, because it produce noisy warnings if you already have a code formatter configured, which I believe is true for most projects.

### Examples

.

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/issues/896

sindresorhus:
The recommend preset is our recommendation. We are considering an unopinionated preset which would not include the rule.

golopot:
Thanks for the heads-up

PR #2715, opened 2025-08-07
feat: add 'unopinionated' config

### LINKED ISSUE #896: Add `unopinionated` preset

JoshuaKGoldberg:
Fixes #896.

Adds all rules from `recommended` to a new `unopinionated` config, with the exception of rules that add in subjective opinions >1% of users might not want (per my vague guesstimation). This list was generated by looking through #896 & linked issues, then running on my personal repos:

| Rule                        | Reason                                                 | Mentioned |
| --------------------------- | ------------------------------------------------------ | --------- |
| catch-error-name            | Subjective naming preference                           | #896      |
| consistent-function-scoping | Subjective preference; doesn't impact control flow     |           |
| empty-brace-spaces          | Formatters also handle this                            | #2634     |
| explicit-length-check       | Subjective preference; increases character count       |           |
| filename-case               | Subjective naming preference                           | #896      |
| no-array-callback-reference | Subjective preference; not always cleaner code         |           |
| no-array-reduce             | Subjective code structure preference                   | #896      |
| no-await-expression-member  | Subjective preference; not always cleaner code         |           |
| no-for-loop                 | Subjective preference; not always optimal              |           |
| no-nested-ternary           | Very opinionated; some projects need this              | #896      |
| no-null                     | Subjective preference some libraries don't work with   | #896      |
| prefer-export-from          | Subjective preference; not always less code            |           |
| prefer-query-selector       | Subjective preference not relevant to all projects     | #896      |
| prefer-spread               | Subjective preference; doesn't impact control flow     |           |
| prevent-abbreviations       | Subjective naming preference                           | #896      |
| switch-case-braces          | Subjective preference; not always cleaner code         |           |
| template-indent             | Conflicts with Vitest-style `.toMatchInlineSnapshot()` |           |

This `unopinionated` config does include the following rules that were mentioned as being candidates for removal:

| Rule                 | Reason                                                                  | Mentioned |
| -------------------- | ----------------------------------------------------------------------- | --------- |
| no-useless-undefined | Legitimate improvement; `consistent-return` doesn't enforce correctness | #2715     |

I recognize I'm not an expert on this plugin and look forward to learning why some rules should(n't) be on those lists. ☺️


fisker:
I don't think "Type safety makes this much less useful" is a good reason to make rule 'opinionated'.

If I understand it correctly. It should only include rules for correctness, there should be only a few rules.

sindresorhus (approved):

docs/rules/catch-error-name.md
```diff
@@ -1,6 +1,6 @@
 # Enforce a specific parameter name in catch clauses
 
-💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
+💼🚫 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config). This rule is _disabled_ in the ☑️ `unopinionated` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
```
fisker, line 3: I think this rule is opinionated.

docs/rules/consistent-assert.md
```diff
@@ -1,6 +1,6 @@
 # Enforce consistent assertion style with `node:assert`
 
-💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
+💼🚫 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config). This rule is _disabled_ in the ☑️ `unopinionated` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
```
fisker, line 3: I think this rule is opinionated.

docs/rules/consistent-empty-array-spread.md
```diff
@@ -1,6 +1,6 @@
 # Prefer consistent types when spreading a ternary in an array literal
 
-💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
+💼🚫 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config). This rule is _disabled_ in the ☑️ `unopinionated` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
```
fisker, line 3: I think this rule is opinionated.

docs/rules/consistent-existence-index-check.md
```diff
@@ -1,6 +1,6 @@
 # Enforce consistent style for element existence checks with `indexOf()`, `lastIndexOf()`, `findIndex()`, and `findLastIndex()`
 
-💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config).
+💼 This rule is enabled in the following [configs](https://github.com/sindresorhus/eslint-plugin-unicorn#recommended-config): ✅ `recommended`, ☑️ `unopinionated`.
```
fisker, line 3: I think this rule is opinionated.

docs/rules/catch-error-name.md
JoshuaKGoldberg: I'm not following, it's disabled in the unopinionated config. I didn't mark this rule as opinionated. Is this a point of agreement, or...?

fisker: Sorry, I thought it's enabled...

fisker: I'll take another look.

fisker (approved):
LGTM, thanks!

JoshuaKGoldberg:
Update: also the new `require-module-attributes` rule.

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v61.0.0

Issue #2720, opened 2025-08-29
`no-array-reduce` - doesn't report optional chaining


ipanasenko:
Title is pretty self-explanatory. `no-array-reduce` doesn't report optional chaining calls of `reduce`, e.g. `array?.reduce()`

Rule: `no-array-reduce`

```js
array.reduce((str, item) => str += item, "") // reports correctly as error
array?.reduce((str, item) => str += item, "") // doesn't report this one
```

I see in rule's code `optionalMember: false` is explicitly passed. So, is this really a bug or an intended behaviour?
I've created a PR with new test cases - they fail: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2721

Thanks!

ipanasenko:
All mutating methods rules also don't check optional chaining, did I get that right? E.g. `no-array-sort`

ipanasenko:
Enabled reporting of optionally chained and called `.reduce()` in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2721

PR #2721, opened 2025-08-29
`no-array-reduce` - enable reporting of optional chaining and calling of `.reduce()`

ipanasenko:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2720

This PR enables checking for optionally chainged `.reduce()` calls for `no-array-reduce` rule.

### DIFF
--- a/rules/no-array-reduce.js
+++ b/rules/no-array-reduce.js
@@ -16,7 +16,6 @@ const cases = [
 				methods: ['reduce', 'reduceRight'],
 				minimumArguments: 1,
 				maximumArguments: 2,
-				optionalCall: false,
 			})
 			&& !isNodeValueNotFunction(callExpression.arguments[0]),
 		getMethodNode: callExpression => callExpression.callee.property,
@@ -46,7 +45,6 @@ const cases = [
 		test: callExpression =>
 			isMethodCall(callExpression, {
 				method: 'call',
-				optionalCall: false,
 			})
 			&& isArrayPrototypeProperty(callExpression.callee.object, {
 				properties: ['reduce', 'reduceRight'],
@@ -62,7 +60,6 @@ const cases = [
 		test: callExpression =>
 			isMethodCall(callExpression, {
 				method: 'apply',
-				optionalCall: false,
 			})
 			&& isArrayPrototypeProperty(callExpression.callee.object, {
 				properties: ['reduce', 'reduceRight'],
--- a/rules/no-array-reduce.js
+++ b/rules/no-array-reduce.js
@@ -16,6 +16,7 @@ const cases = [
 				methods: ['reduce', 'reduceRight'],
 				minimumArguments: 1,
 				maximumArguments: 2,
+				optionalCall: false,
 			})
 			&& !isNodeValueNotFunction(callExpression.arguments[0]),
 		getMethodNode: callExpression => callExpression.callee.property,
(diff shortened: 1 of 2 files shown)


fregante (approved):

test/no-array-reduce.js
```diff
@@ -12,6 +12,8 @@ test({
 		'a[b.reduce]()',
 		'a(b.reduce)',
 		'a.reduce()',
+		'a?.reduce()',
+		'a.reduce?.()',
```
fisker, line 16: It doesn't make sense  to allow optional call.

rules/no-array-reduce.js
```diff
@@ -64,8 +60,6 @@ const cases = [
 		test: callExpression =>
 			isMethodCall(callExpression, {
 				method: 'apply',
-				optionalCall: false,
-				optionalMember: false,
```
fisker: This one should not allow optional member or optional call.

rules/no-array-reduce.js
```diff
@@ -47,8 +45,6 @@ const cases = [
 		test: callExpression =>
 			isMethodCall(callExpression, {
 				method: 'call',
-				optionalCall: false,
-				optionalMember: false,
```
fisker: This one should not allow optional member or optional call.

test/no-array-reduce.js
ipanasenko: Added it because I saw it in `no-array-sort` and `no-array-reverse` 🤷‍♂️
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/no-array-sort.js#L14
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/test/no-array-reverse.js#L14

fisker: They are **valid** cases in these two rules, and they both use `optionalCall: false,` https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d96dfb710e2dd00b82f8bb85b8b16023f6505c68/rules/shared/no-array-mutate-rule.js#L17

fisker: I mean `a?.reduce()` should be **invalid**, but `a.reduce?.()` should be **valid**.

fisker: Sorry, I comment this in a misleading place, I mean L112

ipanasenko: I don't understand, sorry :(
Why one should be valid and another invalid? I thought this rule finds usages of `reduce`, and reports it (except trivial callback cases, if corresponding option is provided). Why reporting of `a?.reduce()` should be different to `a.reduce?.()` and `a.reduce()`?
Or `array.reduce((str, item) => str += item, "")` different to `array?.reduce((str, item) => str += item, "")` and `array.reduce?.((str, item) => str += item, "")` if we talk about L112?

fisker: Because it's doesn't make sense to check `Array#reduce`, if user using `foo.reduce?.()` then `foo` can be a `non-array` object.

Explained in this commit [`8240f4f` (#2721)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2721/commits/8240f4ff9433625de2184f57726bb0a083772294)

fisker (approved):

fisker:
@sindresorhus I plan to go through all rules to check if `optionalMemeber` check is needed this week, do not release until I finish.

PR #2725, opened 2025-09-01
Add `require-module-attributes` rule

### LINKED ISSUE #2681: Rule proposal: require attributes in import/export declaration
### Description

Similar to #2680

### Examples

```js
// ❌
import foo from "foo" with {};

// ✅
import foo from "foo";
```

```js

// ❌
export {foo} from "foo" with {};

// ✅
export {foo} from "foo";
```

### Proposed rule name

require-module-attributes

### Additional Info

_No response_

fisker:
Fixes #2681


docs/rules/require-module-attributes.md
```diff
@@ -0,0 +1,36 @@
+# Require non-empty attribute list in import and export statements
```
fisker: Do we need to mention "import expression" here?

docs/rules/require-module-attributes.md
```diff
@@ -0,0 +1,44 @@
...
```
sindresorhus:
```suggestion
# Require non-empty module attributes for imports and exports
```

PR #2728, opened 2025-09-02
Improve `removeArgument`

fisker:
Preserve comments inside the argument.


test/snapshots/no-unnecessary-array-splice-count.js.md
```diff
@@ -15,7 +15,7 @@ Generated by [AVA](https://avajs.dev).
 > Output
 
     `␊
-      1 | foo.splice(1)␊
+      1 | foo.splice(1 )␊
```
sindresorhus, line 18: I don't think this is correct.

PR #2729, opened 2025-09-02
`prefer-type-error`: ignore error type check

### LINKED ISSUE #2608: `prefer-type-error`: ignore error type check
### Description

This rule make sense when checking other value types, but doesn't make sense when checking error.

### Examples

```js
try {
} catch (error) {
  if (error instanceof SomeError) {
    throw new Error(...) // <- doesn't make sense to throw `TypeError` here
  }

  throw error
}
```

### Additional Info

Not sure if we should just skip the error from catch binding or skip the whole catch block, but I guess we can just ignore catch binding for now.

fisker:
Fixes #2608

### DIFF
--- a/rules/prefer-type-error.js
+++ b/rules/prefer-type-error.js
@@ -73,6 +73,19 @@ const isTypecheckingMemberExpression = (node, callExpression) => {
 	return false;
 };
 
+const errorNameRegexp = /^(?:[A-Z][\da-z]*)*Error$/;
+const isErrorConstructor = node => {
+	if (node.type === 'Identifier') {
+		return errorNameRegexp.test(node.name);
+	}
+
+	if (node.type === 'MemberExpression' && !node.optional && !node.computed && node.property.type === 'Identifier') {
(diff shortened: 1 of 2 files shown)


rules/prefer-type-error.js
```diff
@@ -95,11 +108,14 @@ const isTypecheckingExpression = (node, callExpression) => {
...
-				|| isTypecheckingExpression(node.right, callExpression)
-			);
+			const {operator, left, right} = node;
+
+			if (operator === 'instanceof') {
+				return !isErrorConstructor(right);
+			}
+
+			return isTypecheckingExpression(left, callExpression)
+				|| isTypecheckingExpression(right, callExpression);
```
fisker, line 118: This is actually wrong, there many operators https://github.com/estree/estree/blob/master/es5.md#binaryoperator , should not blindly check left/right.

fisker: Not going to do anything now.

PR #2730, opened 2025-09-03
`no-array-reverse`: better examples in doc

Sysix:
This is a bad example of incorrect code, because it will create a new array and does not modify the original one.


sindresorhus:
The point of the example is to show the equivalant before the new method was introduced. `.reverse()` modifies in-place and is not the same.

Sysix:
> before the new method was introduced. 

This was not clear by the docs. :/ 

Offtopic: I would prefer a fix for `[...array].reversed()` to `array.toReversed()`.

PR #2732, opened 2025-09-06
`prefer-classlist-toggle`: omit `Element#classLIst.contains()` call when fixing

### LINKED ISSUE #2709: Rule proposal: `prefer-classlist-toggle`
### Description

This rule would simplify adding or removing a class based on a condition.

### Examples

```js
// ❌
if (condition) {
  element.classList.add('success');
} else {
  element.classList.remove('success');
}

// ✅
element.classList.toggle('success', condition);
```

```js
// ❌
if (condition) {
  element.classList.remove('success');
} else {
  element.classList.add('success');
}

// ✅
// remove() is first, so negate the condition
element.classList.toggle('success', !(condition));
```

### Proposed rule name

prefer-classlist-toggle

### Additional Info

Very much in the same vein as the [prefer-ternary](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-ternary.md) rule.

fisker:
Ref: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2709#issuecomment-3257873501

Closes #2709


fisker:
Main change is this commit [`9b92a53` (#2732)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2732/commits/9b92a53fe0d30cbaf9c85c68a47570c16cb98284)

PR #2740, opened 2025-09-08
`throw-new-error`: Fix false positive for Effect library

### LINKED ISSUE #2654: `throw-new-error` false positive on class extension
The rule is erroring in cases where it's not relevant, eg extending another class into a new error type :) 

unicorn/throw-new-error

```ts
//   3:33  error  Use `new` when creating an error  unicorn/throw-new-error
//------------------------------v
export class QueryError extends Data.TaggedError('QueryError')<{
  cause: unknown
}> {}
```

kachkaev:
Fixes #2654

### DIFF
--- a/rules/throw-new-error.js
+++ b/rules/throw-new-error.js
@@ -1,4 +1,4 @@
-import isMemberExpression from './ast/is-member-expression.js';
+import {isMemberExpression} from './ast/index.js';
 import {switchCallExpressionToNewExpression} from './fix/index.js';
 
 const messageId = 'throw-new-error';


rules/throw-new-error.js
```diff
@@ -23,6 +23,18 @@ const create = context => ({
...
 
+		// https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2654 (Effect library)
+		if (
+			callee.type === 'MemberExpression'
+			&& !callee.computed
+			&& callee.object.type === 'Identifier'
+			&& callee.object.name === 'Data'
+			&& callee.property.type === 'Identifier'
+			&& callee.property.name === 'TaggedError'
+		) {
```
fisker, line 34: Let's use `isMemberExpression` or  `isNodeMatches`.

fisker (approved):

rules/throw-new-error.js
kachkaev: Done in [`70c9ddc` (#2740)](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2740/commits/70c9ddc3af4a0b6ee325285dfda299c26b4892e5)

fisker (approved):

PR #2742, opened 2025-09-09
doc: fix copy/paste typo in `no-array-sort`.md rule: reversed -> sorted

corneliusroemer:
This rule was likely copy/pasted from reversed and the editing was incomplete.


fisker (approved):
Thanks!

PR #2766, opened 2025-09-25
`prefer-code-point`: Report cases that `String.fromCharCode` not called

### LINKED ISSUE #2613: [prefer-code-point] doesn't report on `const fromCharCode = String.fromCharCode.bind(String)`
```js
const fromCharCode = String.fromCharCode.bind(String)
```

baevm:
Fixed #2613 bug with `String.fromCharCode.bind(String)`, i hope I understood this [comment](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2613#issuecomment-2796543695) correctly


rules/prefer-code-point.js
```diff
@@ -12,16 +12,26 @@ const getReplacement = node => {
...
+	}
+
+	// String.fromCharCode.bind(String)
+	if (isMemberExpression(node.callee.object, {
+		object: 'String',
+		property: 'fromCharCode',
+		optional: false,
+	})) {
+		return ['fromCodePoint', node.callee.object.property];
 	}
```
fisker: This is not the correct solution, we should change the listerner to `MemberExpresion`.

Meaning, no matter how `String.fromCharCode` is used, it should be reported.

```
const x = String.fromCharCode

String.fromCharCode + 1
```

fisker: And the `String.fromCharCode(code)` condition above should be removed.

fisker:
```js
const create = () => ({
	CallExpression(node) {
		// Report `charCodeAt` call
	},
	MemberExpression() {
		// Report `String.fromCharCode`
	}
});
```

fisker: You can even try to use `GlobalReferenceTracker` which handles more cases, but it's hard to understand. I prefer `MemberExpression` check for now which is good enough.

baevm: my bad 🤦 , fixed in e075d0721630fcd9c5806e99af76483df81385cc, also added extra test case for `const x = String.fromCharCode`

fisker: The logic is correct now, thank you!

We can improve the code quantity a little bit,

1. We can reuse the problem reporting part. something like `function getProblem(node, replacement){...}` should be good.
2. `getReplacement...()` can be inlined since they dont need return "replacement", we already know which case we are dealing.

baevm: done

PR #2771, opened 2025-10-07
fix(`prefer-export-from`): type-import removed when using namespace import

k-yle:
Currently, this code:
```ts
import type * as X from 'foo';
export type { X };
```
is autofixed to this:
```ts
export * as X from 'foo';
```
which is missing the [`type` modifier](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export).

This PR fixes the issue. First commit adds a test case, second commit fixes the https://github.com/sindresorhus/eslint-plugin-unicorn/labels/bug


test/snapshots/prefer-export-from.js.md
```diff
@@ -2227,6 +2227,106 @@ Generated by [AVA](https://avajs.dev).
...
+      1 | import * as X from 'foo';␊
+      2 | export type { X };␊
+    `
+
+> Output
+
+    `␊
+      1 |␊
+      2 |␊
+      3 | export type * as X from 'foo';␊
```
fisker: This one seems unsafe to autofix, the imported module may have side effects.

k-yle: good point, fixed in [`deda324...4345915`](https://github.com/sindresorhus/eslint-plugin-unicorn/compare/deda324fde1039ef45c512db9e499572bc27e869..43459159840868c3dbdb66c944c473e812cf8629). 

the other cases get `import type`, this case does not.

test/snapshots/prefer-export-from.js.md
```diff
@@ -2227,6 +2227,106 @@ Generated by [AVA](https://avajs.dev).
...
+      1 | import * as X from 'foo';␊
+      2 | export type { X };␊
+    `
+
+> Output
+
+    `␊
+      1 |␊
+      2 |␊
+      3 | export * as X from 'foo';␊
```
fisker: I think this also change the behavior, it only expose type before, now it expose values. Maybe we can export type but not remove the import.

k-yle: done, this also affects one existing test case

test/snapshots/prefer-export-from.js.md
```diff
@@ -1871,7 +1871,7 @@ Generated by [AVA](https://avajs.dev).
         |               ^^^ Use \`export…from\` to re-export \`foo\`.␊
     ␊
     Output:␊
-      1 |␊
+      1 | import { foo } from 'foo';␊
```
fisker, line 1874: We can't remove this (unless it's type import).

fisker: `removeSpecifier` can control whether the declaration should be removed if the specifier is the only one. https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/fix/remove-specifier.js#L12

k-yle: sorry, what do you mean by we _can't_ remove this? the diff here is adding the import, not removing it

fisker (approved):

Issue #2773, opened 2025-10-09
[`prefer-node-protocol`]: does not flag `typeof import("...")`


sadan4:
prefer-node-protocol does not check `typeof import("...")`

```ts
const assert = require("node:assert") as typeof import("assert");
//                                                      ^^^^^^
// This should be flagged as requiring the node protocol
```

I would be willing to file a PR to fix this.

PR #2774, opened 2025-10-10
`prefer-node-protocol`: Handle typescript import types

sadan4:
closes: #2773

### DIFF
--- a/rules/prefer-node-protocol.js
+++ b/rules/prefer-node-protocol.js
@@ -11,8 +11,8 @@ const messages = {
 const NODE_PROTOCOL = 'node:';
 
 /**
- @param {import('eslint').Rule.RuleContext} context
- */
+@param {import('eslint').Rule.RuleContext} context
+*/
 const create = context => ({
 	Literal(node) {
 		if (!(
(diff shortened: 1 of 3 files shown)


rules/prefer-node-protocol.js
```diff
@@ -34,6 +37,7 @@ const create = context => ({
 				)
 				&& node.parent.arguments[0] === node
 			)
+			|| (node.type === 'Literal' && node.parent.type === 'TSLiteralType' && node.parent.parent.type === 'TSImportType')
```
fisker:
```suggestion
			|| (node.parent.type === 'TSLiteralType' && node.parent.literal === node && node.parent.parent.type === 'TSImportType' && node.parent.parent.argument === node.parent)
```

It's unnecessary to check `node.type`, but better check the child is the correct child.

test/prefer-node-protocol.js
```diff
@@ -132,3 +132,35 @@ test.babel({
...
+	valid: [
+		'const fs = require("node:fs") as typeof import("node:fs");',
+		'const fs = require("node:fs") as typeof SomeType<"fs">;',
+		'const fs = require("node:fs") as typeof fs;',
+		'const fs = require("node:fs") as "fs";',
+		'const fs = require("node:fs") as any;',
+	],
+	invalid: [
+		{
+			code: 'const fs = require("fs") as typeof import("fs");',
```
fisker, line 145: Can you add a test that `import("fs")` is not used in `typeof`?

```ts
type Fs = import("fs")
```

fisker (approved):

PR #2778, opened 2025-10-13
Add `prefer-response-static-json` rule

### LINKED ISSUE #1960: Rule proposal: `prefer-static-response-json` instead of `new Response(JSON.stringify())`
### Description

Prefer the static [`Response.json()`](https://fetch.spec.whatwg.org/#dom-response-json) method instead of `new Response(JSON.stringify())`.

More info: https://github.com/whatwg/fetch/issues/1389 / https://github.com/whatwg/fetch/pull/1392

### Fail

```js
const response = new Response(JSON.stringify({ foo: "bar" }));
```

### Pass

```js
const response = Response.json({ foo: "bar" });
```

### Additional Info

_No response_

fisker:
Fixes #1960

Idealy, we should also remove the useless `'Content-Type': 'application/json'` header, but it's deep in the options which is hard to fix, and as I tested it seems still behave the same.


test/prefer-response-static-json.js
```diff
@@ -0,0 +1,34 @@
...
+		'new Response(JSON.stringify(data), extraArgument)',
+		'new Response( (( JSON.stringify( (( 0, data )), ) )), )',
+		outdent`
+			function foo() {
+				return new // comment
+					Response(JSON.stringify(data))
+			}
+		`,
+	],
+});
```
sindresorhus, line 51: Add:

```
new Response(JSON.stringify(data), {status: 200})
```

test/snapshots/prefer-static-response-json.js.md
```diff
@@ -0,0 +1,98 @@
...
+> Input
+
+    `␊
+      1 | new Response(JSON.stringify(data))␊
+    `
+
+> Output
+
+    `␊
+      1 | Response(data)␊
```
sindresorhus: The output is wrong. Should be:

```suggestion
      1 | Response.json(data)␊
```

rules/prefer-response-static-json.js
```diff
@@ -0,0 +1,72 @@
...
+		})) {
+			return;
+		}
+
+		return {
+			node: jsonStringifyNode.callee,
+			messageId: MESSAGE_ID,
+			/** @param {import('eslint').Rule.RuleFixer} fixer */
+			* fix(fixer) {
+				const {sourceCode} = context;
```
sindresorhus, line 42:
```suggestion
				const {sourceCode} = context;
				yield fixer.insertTextAfter(newExpression.callee, '.json');
```

docs/rules/prefer-static-response-json.md
```diff
@@ -0,0 +1,20 @@
+# Prefer the static `Response.json()` method instead of `new Response(JSON.stringify())`
```
sindresorhus:
```suggestion
# Prefer `Response.json()` over `new Response(JSON.stringify())`
```

rules/prefer-static-response-json.js
```diff
@@ -0,0 +1,72 @@
...
+	},
+});
+
+/** @type {import('eslint').Rule.RuleModule} */
+const config = {
+	create,
+	meta: {
+		type: 'suggestion',
+		docs: {
+			description: 'Prefer the static `Response.json()` method instead of `new Response(JSON.stringify())`.',
```
sindresorhus:
```suggestion
			description: 'Prefer `Response.json()` over `new Response(JSON.stringify())`.',
```

rules/prefer-response-static-json.js
fisker: I'm dum

fisker:
Moved to the issue https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1960#issuecomment-3410519056 for more people to involve.

Should we rename it as `prefer-response-static-json`?
If one day we want to add a rule for the instance `json` method, we definitely don't want use `prefer-{instance,prototype}-response-json`.

PR #2780, opened 2025-10-13
`text-encoding-identifier-case`: Add `withDash` option

carlosjeurissen:
Closes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1926
Closes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2547

Feel free to tweak the docs.

### DIFF
--- a/docs/rules/text-encoding-identifier-case.md
+++ b/docs/rules/text-encoding-identifier-case.md
@@ -47,7 +47,7 @@ Type: `object`
 Type: `boolean`\
 Default: `false`
 
-Prefer the WHATWG standard encoding notation. I.e. `utf-8` will be preferred over `utf8`. Pass `withDash: true` to prefer `utf-8`.
+Use WHATWG standard encoding notation with dashes (e.g., ⁠`utf-8` instead of `⁠utf8`).
 
 ```js
 // ❌


docs/rules/text-encoding-identifier-case.md
```diff
@@ -12,28 +12,42 @@
...
+await fs.readFile(file, 'ascii');
+```
+
+```js
 const string = buffer.toString('utf8');
 ```
+
+## Options
+
+### whatwgEncoding
```
sindresorhus:
```suggestion
### preferDash
```

Note that it's inverted.

carlosjeurissen: Happy with preferDash. whatwgEncoding was originally proposed by @fisker in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1926#issuecomment-2742537455. What is inverted?

fisker: I was thinking that we may add more encodings in future. And the preferred one in web is not the same case as `UTF-8`

fisker: Personally, I still think `whatwgEncoding` is a better name.

docs/rules/text-encoding-identifier-case.md
```diff
@@ -12,28 +12,42 @@
...
+await fs.readFile(file, 'ascii');
+```
+
+```js
 const string = buffer.toString('utf8');
 ```
+
+## Options
+
+### preferDash
```
carlosjeurissen:
```suggestion
### whatwgEncoding
```

docs/rules/text-encoding-identifier-case.md
fisker: @sindresorhus

test/text-encoding-identifier-case.js
```diff
@@ -36,6 +36,58 @@ test.snapshot({
 	],
 });
 
+// 'whatwgEncoding'
+test.snapshot(avoidTestTitleConflict({
```
fisker: Please do not duplicate all the tests, just pick a few and test with the new options. Maybe just

```
"utf8"
"UTF8"
"utf-8"
"UTF-8"
```

and the jsx case.

rules/text-encoding-identifier-case.js
```diff
@@ -35,62 +36,80 @@ const isFsReadFileEncoding = node =>
...
-			&& node.parent.name.name.toLowerCase() === 'charset'
-			&& node.parent.parent.type === 'JSXOpeningElement'
-			&& node.parent.parent.attributes.includes(node.parent)
-			&& node.parent.parent.name.type === 'JSXIdentifier'
-			&& node.parent.parent.name.name.toLowerCase() === 'meta'
-		) {
-			return;
-		}
+	return {
+		Literal(node) {
```
fisker: If we change to `context.on('Literal', () => {})`, will it reduce indentation? If so we can make the diff smaller.

docs/rules/text-encoding-identifier-case.md
sindresorhus: I don't see us adding more encodings. Anything other than UTF-8 is a super edge-case these days.

Most users would not know what `whatwg` is though. And there may be other places that uses encoding with a dash. However, I don't feel strongly about it. Maybe `useWhatwgEncoding` would be better name then, since it's a boolean.

fisker: Speaking of this, I saw ESLint adding descriptions for the schema, not sure how it useful, I'll take a look right now. Opened #2781

carlosjeurissen: A potential addition could be people preferring the uppercase version in some situations.

test/text-encoding-identifier-case.js
carlosjeurissen: Addressed!

rules/text-encoding-identifier-case.js
carlosjeurissen: @fisker great suggestion. applied this change!

docs/rules/text-encoding-identifier-case.md
fisker: > Most users would not know what whatwg is though. 

That's a good point, let's use `preferDash` then. @carlosjeurissen

test/text-encoding-identifier-case.js
fisker: Not the JSX case.

fisker: Since there are only a couple of tests, can we avoid conflicts in code, something like

```
foo = 'utf-8'
bar = 'utf-8'
'utf-8' // random comment
```

And `avoidTestTitleConflict` requires a comment, you can see it changed code to `code /* undefined */`. https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2780/files#diff-b7f64d82189a67c726cf1dc986e01c6917e2f3ffdcefda5385c49c62bd764b19R400

docs/rules/text-encoding-identifier-case.md
carlosjeurissen: `preferDash` implies the default is without dash. Even tho right now it seems you do want to keep "utf8" the default for now, this may change in the future. Agreed `preferDash` is more understandable than `useWhatwgEncoding`. A more neutral version could be `{withDash: true}`. @fisker @sindresorhus

test/text-encoding-identifier-case.js
carlosjeurissen: @fisker good point. Added additional JSX cases, and removed the use of avoidTestTitleConflict. Thanks for the note on avoidTestTitleConflict requiring a comment!

docs/rules/text-encoding-identifier-case.md
fisker: Fine with both.

sindresorhus: Let's go with `withDash`

carlosjeurissen: updated to withDash.

docs/rules/text-encoding-identifier-case.md
```diff
@@ -37,3 +37,15 @@ const string = buffer.toString('utf-8');
...
+
+### withDash
+
+Type: `boolean`\
+Default: `false`
+
+- `false` (default)
+  - Prefer `utf8` without a dash (Node.js style)
+- `true`
+  - Prefer `utf-8` with a dash (WHATWG standard, required for HTML)
```
fisker: Can you take [this](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-undefined.md#checkarguments) as example and update the docs?

carlosjeurissen: @fisker Done.

docs/rules/text-encoding-identifier-case.md
```diff
@@ -37,3 +37,26 @@ const string = buffer.toString('utf-8');
...
+## Options
+
+Type: `object`
+
+### withDash
+
+Type: `boolean`\
+Default: `false`
+
+Prefer the WHATWG standard encoding notation. I.e. `utf-8` will be preferred over `utf8`. Pass `checkArguments: true` to prefer `utf-8`.
```
fisker: Incorrect.

carlosjeurissen: @fisker good catch! fixed.

fisker (approved):

PR #2782, opened 2025-10-14
Meta tweak, fix crash on `no-useless-undefined`

fisker:
- Update dependencies
- Fix integration test (next.js)
- Add webpack repo to integration test

---

`no-useless-undefined` crashes when `undefined` used in callexpressions that arguments are parenthesized.


fisker:
Yes

PR #2784, opened 2025-10-21
`prefer-bigint-literals`: Support signed numbers and strings

### LINKED ISSUE #2733: `prefer-bigint-literal`: Support number with `+`/`-` sign
### Description

I wasn't aware that `BigInt('-1')` and `BigInt('+1')` is also valid.

### Examples

```js
// ❌
BigInt('-1')

// ✅
-1n
```

```js
// ❌
BigInt(-1)

// ✅
-1n
```

```js
// ❌
BigInt('+1')

// ✅
1n
```

### Additional Info

The string cases should already supported, need tests.
The negative integer case should be fixed.

baevm:
Fixes #2733, added support for `BigInt('+1')`, `BigInt('-1')`, `BigInt(+1)`, `BigInt(-1)`. Not sure about `BigInt(!1)` and `BigInt(~1)`, looks like those also valid?

image


fisker:
I opened #2733, because I think there are maybe people writes `BigInt("-1")`, but I don't think anyone would write `BigInt(!1)`, it should be a mistake.

rules/prefer-bigint-literals.js
```diff
@@ -36,19 +34,32 @@ const canUseNumericLiteralRaw = numericLiteral => {
...
+		let raw = valueNode.raw.slice(1, -1);
 		try {
 			BigInt(raw);
 		} catch {
 			return;
 		}
 
+		// BigInt("+1") -> 1n
+		if (raw[0] === '+') {
+			raw = raw.slice(1);
```
fisker: Can you add test for `BigInt("++1")`?

fisker:
Need test `-BigInt(-1)` and `-BigInt("-1")`, I think they will fail.

fisker:
There can also ASI issue since the fixed code inserts `-`.

```
foo // <-- no semicolon here
BigInt("-1")
```

rules/prefer-bigint-literals.js
fisker: And `BigInt("+ 1")`

fisker:
```
BigInt(" -1 ")
```
```
BigInt(" +1 ")
```

rules/prefer-bigint-literals.js
```diff
@@ -36,19 +34,32 @@ const canUseNumericLiteralRaw = numericLiteral => {
...
-		const raw = valueNode.raw.slice(1, -1);
+		let raw = valueNode.raw.slice(1, -1);
 		try {
 			BigInt(raw);
 		} catch {
 			return;
 		}
 
+		// BigInt("+1") -> 1n
+		if (raw[0] === '+') {
```
fisker: Unlike `"-1"`, `"+1"` probably a mistake, let's use suggestion instead of autofix. Same for `UnaryExpression`

rules/prefer-bigint-literals.js
baevm: added tests:
`BigInt("++1")` - valid 
`BigInt("+ 1")` - valid

`BigInt(" -1 ")` - invalid
`BigInt(" +1 ")` - invalid with suggestion

rules/prefer-bigint-literals.js
baevm: replaced with suggestion, also fixed logic for cases with spaces `BigInt(" +1 ")`

test/snapshots/prefer-bigint-literals.js.md
```diff
@@ -424,3 +424,296 @@ Generated by [AVA](https://avajs.dev).
...
+      3 | 		BigInt("-1")␊
+      4 | 		␊
+    `
+
+> Output
+
+    `␊
+      1 |␊
+      2 | 		foo␊
+      3 | 		-1n␊
```
fisker: This should be

```
foo
;-1n
```

fisker: image

fisker: There is a `needsSemicolon` utility for it.

test/snapshots/prefer-bigint-literals.js.md
```diff
@@ -505,6 +522,109 @@ Generated by [AVA](https://avajs.dev).
...
+> Input
+
+    `␊
+      1 | -BigInt("-1")␊
+    `
+
+> Output
+
+    `␊
+      1 | 1n␊
```
fisker, line 574: Instead of `1n`, better fix to `-(-1n)`

test/snapshots/prefer-bigint-literals.js.md
baevm: fixed

test/snapshots/prefer-bigint-literals.js.md
baevm: fixed both for `-BigInt("-1")` -> `-(-1n)` and `-BigInt(-1)` ->  `-(-1n)`

rules/prefer-bigint-literals.js
```diff
@@ -83,17 +101,36 @@ const create = context => {
...
+		// -BigInt(-1) -> -(-1n)
+		// -BigInt("-1") -> -(-1n)
+		if (
+			parent.type === 'UnaryExpression'
+			&& parent.operator === '-'
+			&& text.startsWith('-')
+		) {
+			nodeToReplace = parent;
+			text = `-(-${text.slice(1)})`;
+		}
```
fisker: This is not enough, `BigInt(-1).toString()`. We should always add `()` for negative numbers.

baevm: fixed: `BigInt(-1).toString()` -> `(-1n).toString()`

rules/prefer-bigint-literals.js
```diff
@@ -118,7 +123,20 @@ const create = context => {
...
 
 		const tokenBefore = context.sourceCode.getTokenBefore(nodeToReplace);
-		if (needsSemicolon(tokenBefore, context, text)) {
+
+		if (text.startsWith('-') && !text.startsWith('-(')) {
+			const needsParentheses
+				// BigInt(-1).toString() -> (-1n).toString()
+				= (parent.type === 'MemberExpression' && parent.object === callExpression)
+				// 2n -BigInt("-1") -> 2n -(-1n)
+				|| (tokenBefore?.type === 'Punctuator' && (tokenBefore.value === '-' || tokenBefore.value === '+'));
```
fisker: Why are we doing this? Just always wrap with `()` when it's negative.

baevm: you mean like this? `BigInt("-1")` -> `(-1n)`, but why? isnt -1n also valid?

should be fixed in [5e3158a](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2784/commits/5e3158ae44f2dbd55a49f31b4623f794ed963eaa)

test/prefer-bigint-literals.js
```diff
@@ -52,5 +64,42 @@ test.snapshot({
...
+		'BigInt(-1).toString()',
+		'functionCall(-BigInt(1))',
+		'obj[BigInt(1)]',
+		'arr[BigInt(0)]',
+		'void BigInt(1)',
+		'typeof BigInt(1)',
+		'BigInt(1) + BigInt(2)',
+		'BigInt(1) ** 2n',
+		'condition ? BigInt(1) : BigInt(2)',
+		'+BigInt(1)',
```
fisker, line 95: Same runtime error, this is not what we should care.

```
+BigInt(1)
VM782:1 Uncaught TypeError: Cannot convert a BigInt value to a number
    at <anonymous>:1:1
(anonymous) @ VM782:1

+1n
VM792:1 Uncaught TypeError: Cannot convert a BigInt value to a number
    at <anonymous>:1:1
```

PR #2794, opened 2025-10-25
Improve performance of parentheses related check

fisker:
Since @eslint-community/eslint-utils only provides a [`isParenthesized`](https://eslint-community.github.io/eslint-utils/api/ast-utils.html#isparenthesized) to check if a node is parenthesized, it is super inefficient in the  original `getParenthesizedTimes`

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/8a132aceea7927cb7209887d7284e40f9adf55ed/rules/utils/parentheses.js#L15-L23

And `getParentheses`, `getParenthesizedRange`, and `getParenthesizedText` all rely on it.

Since these methods are also heavily used in almost every rule, I added a `WeakMap` to cache the result too.


sindresorhus:
Conflict

PR #2808, opened 2025-11-01
Various fix

fisker:
I was playing with the AI, it found these.

- Missed a `.` in error message
- An unexpected sequence expression

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -47,6 +47,7 @@ const config = [
 	},
 	{
 		rules: {
+			'no-sequences': ['error', {allowInParentheses: false}],
 			'unicorn/escape-case': 'off',
 			'unicorn/expiring-todo-comments': 'off',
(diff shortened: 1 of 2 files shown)


eslint.config.js
```diff
@@ -47,6 +47,7 @@ const config = [
 	},
 	{
 		rules: {
+			'no-sequences': ['error', {allowInParentheses: false}],
```
fisker, line 50: Maybe you want add this to xo

https://github.com/xojs/eslint-config-xo/blob/0662bab31739b34fc28dfbd0c441588b129e9351/index.js#L172

sindresorhus: Will do in the next XO version.

https://github.com/xojs/eslint-config-xo/commit/391d4c8fc2d583631c2f13c302eb624663c165ce

PR #2814, opened 2025-11-03
`explicit-timer-delay`: Add rule

### LINKED ISSUE #2809: Rule proposal: `explicit-timer-delay`
### Description

Enforce or disallow explicit delay argument for `setTimeout()` and `setInterval()`

Allow configuring whether `setTimeout()` and `setInterval()` must have an explicit delay argument. This provides flexibility for different team preferences.

### Examples

```typescript
{
  "unicorn/explicit-timer-delay": ["error", "always" | "never"]
}
```

- `"always"` (default): Require explicit delay argument
- `"never"`: Disallow explicit delay when it's 0 (prefer implicit default)

**With `"always"`:**

```js
// ❌ Fail
setTimeout(() => console.log('Hello'));
setInterval(callback);

// ✅ Pass
setTimeout(() => console.log('Hello'), 0);
setInterval(callback, 0);
```

**With `"never"`:**

```js
// ❌ Fail
setTimeout(() => console.log('Hello'), 0);

// ✅ Pass
setTimeout(() => console.log('Hello'));
setInterval(callback, 1000); // Non-zero is still OK
```

### Proposed rule name

explicit-timer-delay

### Additional Info

_No response_

azat-io:
Close #2809


docs/rules/explicit-timer-delay.md
```diff
@@ -0,0 +1,77 @@
...
+- `"always"` (default): Require explicit `delay` argument for clarity
+- `"never"`: Disallow explicit `0` delay (prefer implicit default)
+
+## Examples
+
+### With `"always"` (default)
+
+```js
+// ❌
+setTimeout(() => console.log('Hello'));
```
sindresorhus, line 18: Please see other rules for formatting here. Whenever possible, show fail/pass cases together.

sindresorhus:
Match the common options style used in other docs.
- Use `Type` + `Default` together (for example: `Type: 'string'\ Default: 'always'`).
- Use single-quoted option values (`'always'`, `'never'`) to match repo convention.
- Prefer JS config examples over JSON for consistency with most rule docs.

azat-io:
@sindresorhus Thanks for the review! I've fixed your comments!

sindresorhus:
I rechecked this manually and there are still a few blockers:

- The option schema is missing a `description`. This currently trips `eslint-plugin/require-meta-schema-description`, so CI fails until the enum schema object gets one.
- The generated rule doc header is stale. The generator expects the rule name as the heading plus the generated description line, but this has a hand-written heading instead. Running the docs generator after the schema fix should sort that out.
- The fixer call formatting also fails `npx xo` for the new rule because of the newline around the function call parens. Not a behavior issue, but definitley needs cleanup before merge.

azat-io:
Fixed

sindresorhus:
Thanks for contributing :)

PR #2822, opened 2025-11-13
fix: `isolated-functions`: allow typescript types

mmkal:
On trying it out I noticed I got errors for typescript types, which there's no value in.

### DIFF
--- a/rules/isolated-functions.js
+++ b/rules/isolated-functions.js
@@ -54,7 +54,7 @@ const create = context => {
 		for (const reference of nodeScope.through) {
 			const {identifier} = reference;
 
-			if (identifier?.parent?.type === 'TSTypeReference' || identifier?.parent?.type === 'TSTypeQuery') {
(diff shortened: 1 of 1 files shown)


rules/isolated-functions.js
```diff
@@ -53,6 +53,11 @@ const create = context => {
 		// `through`: "The array of references which could not be resolved in this scope" https://eslint.org/docs/latest/extend/scope-manager-interface#scope-interface
 		for (const reference of nodeScope.through) {
 			const {identifier} = reference;
+
+			if (identifier?.parent?.type === 'TSTypeReference' || identifier?.parent?.type === 'TSTypeQuery') {
+				continue;
```
fisker, line 58: Are you sure these `?.` needed? `identifier` can't be nullish, their parents should also exists.

mmkal: Nope, not based on anything, I can remove

Issue #2745, opened 2025-09-11
`expiring-todo-comments`: should not flag "eslint disable" comments


cdavie-artium:
I like the `expiring-todo-comments` rule, but it is annoying to try to disable it for a single line when `allowWarningComments: false` is set.

I expected this to work:

```js
// eslint-disable-next-line unicorn/expiring-todo-comments
// TODO without a date
```

But the lint rule then flags the `eslint-disable-next-line` line itself as problematic!

The cumbersome workaround I've found is to use block disable comments instead (plus an extra next-line comment is required):

```js
/* eslint-disable unicorn/expiring-todo-comments */
// TODO without a date
// eslint-disable-next-line unicorn/expiring-todo-comments
/* eslint-enable unicorn/expiring-todo-comments */
```

I think there are two bugs here:

1. It should ignore eslint-disable lines entirely.
2. It should not match substrings that happen to have "todo" within them (eg, "Mastodon" or "-todo-").

PR #2828, opened 2025-11-22
`expiring-todo-comments`: should not flag eslint disable comments

Vladyslav-Soldatenko:
fix for #2745


rules/expiring-todo-comments.js
```diff
@@ -284,6 +284,13 @@ const create = context => {
 
 	const {sourceCode} = context;
 	const comments = sourceCode.getAllComments();
+
+	const isEslintDirectiveComment = comment => {
+		// Strip leading whitespace and optional `*` from block comments
+		const normalizedComment = comment.value.trimStart().replace(/^\*\s*/, '');
+		return /^eslint(?:-(?:en|dis)able)?(?:-(?:next-)?line)?\b/.test(normalizedComment);
+	};
```
fisker: Use `ConfigCommentParser` from `@eslint/plugin-kit` please.

You can take https://github.com/sindresorhus/eslint-plugin-unicorn/commit/0522c912869207b678cb7e9baf8ecdbfa1d1b451 as example

Vladyslav-Soldatenko: Good point, thanks. Done

rules/expiring-todo-comments.js
```diff
@@ -319,6 +322,13 @@ const create = context => {
 
 	// eslint-disable-next-line complexity
 	function processComment(comment) {
+		configCommentParser ??= new ConfigCommentParser();
+
+		const directive = configCommentParser.parseDirective(comment.value);
+		if (directive?.label?.startsWith('eslint')) {
```
fisker: Please only match

```
	'eslint-disable',
	'eslint-disable-line',
	'eslint-disable-next-line',
```

fisker: You can extract the code in https://github.com/sindresorhus/eslint-plugin-unicorn/commit/0522c912869207b678cb7e9baf8ecdbfa1d1b451

To a function `parseDirective` and add a `isEslintDisableDirective` property to reuse.

Vladyslav-Soldatenko: 1) I'm not sure I understand "extract code to a function "parseDirective". "parseDirective" is a method of plugin-kit and I can't modify it
2) I can't match ONLY disable directives, because otherwise the rule starts to complain even about "/* eslint-enable unicorn/expiring-todo-comments */" lines (see original issue)

Vladyslav-Soldatenko: so "enable" directives also must be checked and ignored

fisker: > I'm not sure I understand "extract code to a function "parseDirective". "parseDirective" is a method of plugin-kit and I can't modify it

Something like

```
function parseDirective() {
	const result = configCommentParser.parseDirective(comment.value);

	if (!result) {
		return
	}

	return {
		...result,
		isEslintDisableDirective: ...,
		isEslintEnableDirective: ...,
	}
}
```


>  so "enable" directives also must be checked and ignored

Add `eslint-enable` then.

Vladyslav-Soldatenko: done

rules/utils/parse-directive-comment.js
```diff
@@ -0,0 +1,47 @@
+import {ConfigCommentParser} from '@eslint/plugin-kit';
+
+const ESLINT_DISABLE_DIRECTIVES = new Set([
```
fisker: Please keep the link

```
// https://github.com/eslint/eslint/blob/ecd0ede7fd2ccbb4c0daf0e4732e97ea0f49db1b/lib/linter/linter.js#L509-L512
```

rules/utils/parse-directive.js
```diff
@@ -0,0 +1,47 @@
...
+const ESLINT_DISABLE_DIRECTIVES = new Set([
+	'eslint-disable',
+	'eslint-disable-line',
+	'eslint-disable-next-line',
+]);
+
+const ESLINT_ENABLE_DIRECTIVES = new Set([
+	'eslint-enable',
+	'eslint-enable-line',
+	'eslint-enable-next-line',
```
fisker: There is no such directive, see the link above.

rules/utils/parse-directive-comment.js
Vladyslav-Soldatenko: done

rules/utils/parse-directive.js
Vladyslav-Soldatenko: yeah, my bad. Done

fisker (approved):
Thanks!

fisker:
@Vladyslav-Soldatenko Turns out there is an `sourceCode.getDisableDirectives()` we can use directly!

rules/no-abusive-eslint-disable.js
```diff
@@ -1,45 +1,31 @@
...
-				// It's a eslint-disable comment
-				eslintDisableDirectives.has(result?.label)
-				// But it did not specify any rules
-				&& !result?.value
-			)) {
-				return;
+	context.on('Program', function * () {
+		for (const directive of getEslintDisableDirectives(context)) {
+			if (directive.value) {
+				continue;
```
fisker, line 15: This is a bug, using `return` instead of `continue`, introduced by me in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2571

PR #2829, opened 2025-11-23
`prefer-set-has`: Check `Iterator#toArray()` and `String#split()`

fisker:


sindresorhus:
CI seems to be failing.

PR #2845, opened 2025-12-31
Add `eslint-scope` dependency

kovsu:
Missing `eslint-scope` dependency.

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/b80e75bc3b5bc8a048fdf08476d65af98a813f21/scripts/parsers.js#L6

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
 		"ci-info": "^4.3.1",
 		"clean-regexp": "^1.0.0",
 		"core-js-compat": "^3.46.0",
-		"eslint-scope": "^9.0.0",
 		"find-up-simple": "^1.0.1",
 		"globals": "^16.4.0",
 		"indent-string": "^5.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -63,6 +63,7 @@
 		"ci-info": "^4.3.1",
 		"clean-regexp": "^1.0.0",
 		"core-js-compat": "^3.46.0",
+		"eslint-scope": "^9.0.0",
```
fisker: Only required as devDependencies, right?

package.json
```diff
@@ -122,5 +123,6 @@
 			"text",
 			"lcov"
 		]
-	}
+	},
+	"packageManager": "pnpm@10.26.2+sha512.0e308ff2005fc7410366f154f625f6631ab2b16b1d2e70238444dd6ae9d630a8482d92a451144debc492416896ed16f7b114a86ec68b8404b2443869e68ffda6"
```
fisker: Please revert this.

kovsu: oh, forget this.

package.json
kovsu: Yes, you are right.

fisker (approved):

Issue #2854, opened 2026-01-31
ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "unicorn"


aladdin-add:
Node version: v25.2.1
npm version: v11.7.0
Local ESLint version: v9.39.2 (Currently used)
Global ESLint version: Not found
Operating System: linux 6.6.87.2-microsoft-standard-WSL2
eslint-plugin-unicorn: v62.0.0

The official ESLint documentation now recommends using `defineConfig` and `extends`, but currently, this plugin will report an error when used this way.

n/a.

```js
import unicorn from 'eslint-plugin-unicorn';
import { defineConfig } from 'eslint/config';

export default defineConfig([
    {
        files: ['**/*.{js,ts}'],
        plugins: { unicorn },
        extends: ['unicorn/recommended'],
    }
])
```

```bash
npx eslint .

Oops! Something went wrong! :(

ESLint: 9.39.2

ConfigError: Config (unnamed): Key "plugins": Cannot redefine plugin "unicorn".
    at rethrowConfigError (/home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/@eslint/config-array/dist/cjs/index.cjs:343:8)
    at /home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/@eslint/config-array/dist/cjs/index.cjs:1405:5
    at Array.reduce (<anonymous>)
    at FlatConfigArray.getConfigWithStatus (/home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/@eslint/config-array/dist/cjs/index.cjs:1398:43)
    at FlatConfigArray.getConfig (/home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/@eslint/config-array/dist/cjs/index.cjs:1427:15)
    at /home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/eslint/lib/eslint/eslint.js:630:27
    at Array.map (<anonymous>)
    at lintFilesWithoutMultithreading (/home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/eslint/lib/eslint/eslint.js:628:13)
    at ESLint.lintFiles (/home/weiran/repos/eslint-community/eslint-plugin-eslint-plugin/node_modules/eslint/lib/eslint/eslint.js:1060:20)
[...]

PR #2855, opened 2026-01-31
fix: export the same object in configs

aladdin-add:
fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2854

This pull request simplifies the export structure in `index.js` by attaching the `configs` object directly to the `unicorn` object and exporting `unicorn` as the default export, instead of creating a separate `allConfigs` object.

Export structure simplification:

* Modified `index.js` to attach the `configs` object directly to the `unicorn` object and export `unicorn` as the default export, removing the intermediate `allConfigs` object.


aladdin-add:
@fisker

PR #2859, opened 2026-02-07
`new-for-builtins`: Fix `Date()` suggestion dropping arguments

### LINKED ISSUE #2802: `new-for-builtin`: Invalid suggestion
```js
## invalid(6): const foo = Date(bar);

> Input

    `␊
      1 | const foo = Date(bar);␊
    `

> Error 1/1

    `␊
    Message:␊
    > 1 | const foo = Date(bar);␊
        |             ^^^^^^^^^ Use \`String(new Date())\` instead of \`Date()\`.␊
    ␊
    Suggestion 1/1: Switch to \`String(new Date())\`.:␊
      1 | const foo = String(new Date());␊
                                      ^ The argument is removed.
    `
```

sindresorhus:
Fixes #2802


test/snapshots/new-for-builtins.js.md
```diff
@@ -1438,8 +1438,8 @@ Generated by [AVA](https://avajs.dev).
     `␊
     Message:␊
     > 1 | const foo = Date(bar);␊
-        |             ^^^^^^^^^ Use \`String(new Date())\` instead of \`Date()\`.␊
+        |             ^^^^^^^^^ Use \`new Date()\` instead of \`Date()\`.␊
     ␊
-    Suggestion 1/1: Switch to \`String(new Date())\`.:␊
-      1 | const foo = String(new Date());␊
+    Output:␊
+      1 | const foo = new Date(bar);␊
```
fisker, line 1444: This is incorrect.


> Any arguments given in a Date() function call (without the new keyword) are ignored; regardless of whether it's called with an invalid date string — or even called with any arbitrary object or other primitive as an argument — it always returns a string representation of the current date and time.


I forgot why I opened that issue. I think current behavior is correct. `Date(bar)` is equivalent to `String(new Date())`, and we already use suggestion.

I think we should close that  issue directly .

PR #2862, opened 2026-02-07
`no-array-sort`: Ignore `.sort()` calls with non-function arguments

### LINKED ISSUE #2744: `no-array-sort` should trigger only with `Array.prototype.sort()`
```js
import { type Collection } from 'mongodb'

const collection = await this.getColection()
const docs = await collection.find({ id }).sort({ expireAt: -1 }).limit(1).toArray()
```

sindresorhus:
Fixes #2744

### DIFF
--- a/rules/shared/no-array-mutate-rule.js
+++ b/rules/shared/no-array-mutate-rule.js
@@ -1,25 +1,10 @@
 import {isMethodCall} from '../ast/index.js';
-import {getParenthesizedText} from '../utils/index.js';
+import {getParenthesizedText, isNodeValueNotFunction} from '../utils/index.js';
 
 const MESSAGE_ID_ERROR = 'error';
 const MESSAGE_ID_SUGGESTION_APPLY_REPLACEMENT = 'suggestion-apply-replacement';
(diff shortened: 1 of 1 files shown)


rules/shared/no-array-mutate-rule.js
```diff
@@ -6,6 +6,21 @@ const MESSAGE_ID_SUGGESTION_APPLY_REPLACEMENT = 'suggestion-apply-replacement';
...
+	'ObjectExpression',
+	'ArrayExpression',
+	'Literal',
+	'TemplateLiteral',
+]);
+
+const isUnaryLiteralExpression = node => node?.type === 'UnaryExpression'
+	&& node.argument.type === 'Literal';
+
+const isNonFunctionNode = node => nonFunctionTypes.has(node?.type)
```
fisker: Same as https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2865#discussion_r2779365849.

Maybe we can reuse `isNodeValueNotFunction`

sindresorhus: Done

PR #2865, opened 2026-02-07
`prefer-add-event-listener`: Don't autofix non-function assignments

### LINKED ISSUE #2440: `prefer-add-event-listener`: Bad autofix when assignment a non-function object
This is actually in my project, although the code is unreasonable.

```diff
var target

- target.ondrag = true
+ target.addEventListener('drag', true)
```

In addition, if the parameters do not match the types, an exception will be thrown

```js
addEventListener('click', true) 

// Uncaught TypeError: EventTarget.addEventListener: Argument 2 is not an object.
```

sindresorhus:
Fixes #2440


rules/prefer-add-event-listener.js
```diff
@@ -49,6 +49,19 @@ const shouldFixBeforeUnload = (assignedExpression, nodeReturnsSomething) => {
 
 const isClearing = node => isUndefined(node) || isNullLiteral(node);
 
+const notFunctionNodeTypes = new Set([
```
fisker: Reuse   `isNodeValueNotFunction`?

sindresorhus: Done

PR #2867, opened 2026-02-08
`prefer-regexp-test`: Add test for non-regexp variable argument

### LINKED ISSUE #1380: prefer-regexp-test: incorrect autofix for union type
```js
const pattern: string | RegExp = 'xxx';

// source
'yyy'.match(pattern)

// fixed
pattern.test('yyy')
```

sindresorhus:
Fixes #1380

I think it was fixed at some point.


fisker (approved):
Fine to close without using type information.

PR #2870, opened 2026-02-08

### LINKED ISSUE #1661: Rule proposal: No top-level side effects
### Description

Related rules: 

### Fail

```js
document.title = 'gone'
```

### Pass

```js
const details = new Map();
```

```js 
function init () {
  document.title = 'gone'
}
```

```js
function init () {
  document.title = 'gone'
}
```

```js
function init () {
  document.title = 'gone'
}

init(); 
```

```js
const details = new Map();
```

aifunmobi:
## Summary
- New rule that disallows side-effect statements at the top level of module files
- Importing a module should not produce observable side effects
- Side effects should be wrapped in exported functions that consumers call explicitly

## Exemptions (per maintainer specs in #1661)
- Files with hashbang (`#!/usr/bin/env node`) — CLI scripts
- Files with no exports — entry points
- Variable declarations (`const x = fetch()` is fine)
- CJS `module.exports` / `exports.foo` assignments
- `'use strict'` directives

## Test plan
- [x] 23 test cases (16 valid, 7 invalid)
- [x] Snapshot tests for all invalid cases
- [x] Documentation with examples

Fixes #1661

🤖 Generated with [Claude Code](https://claude.com/claude-code)


test/no-top-level-side-effects.js
```diff
@@ -0,0 +1,256 @@
...
+		`,
+		// Object.assign on imported binding is a side effect
+		outdent`
+			import shared from 'x';
+			Object.assign(shared, { flag: true });
+			export default shared;
+		`,
+		// Export default with side-effecting call expression
+		outdent`
+			export default setup();
```
fisker, line 294: I think we should support pure notations

`export default /* @__PURE__ */ setup();`
`export default /* #__PURE__ */ setup();`

https://github.com/javascript-compiler-hints/compiler-notations-spec/tree/main

aifunmobi: Done — added `hasPureAnnotation()` that checks for `/* @__PURE__ */` and `/* #__PURE__ */` block comments via `sourceCode.getCommentsBefore()`. Works for both `export default /* @__PURE__ */ setup()` and bare `/* @__PURE__ */ setup()` expression statements.

Also added `isPureExpression()` to handle `export default { ... }` with no calls (which fixes the `run-rules-on-codebase` CI failure on `isolated-functions.js`).

fisker: `#__NO_SIDE_EFFECTS__` notation?

aifunmobi: Done — added support for `@__NO_SIDE_EFFECTS__` and `#__NO_SIDE_EFFECTS__` annotations. The regex now matches all four notations: `@__PURE__`, `#__PURE__`, `@__NO_SIDE_EFFECTS__`, and `#__NO_SIDE_EFFECTS__`.

aifunmobi: Done — added function-declaration-level `#__NO_SIDE_EFFECTS__` / `@__NO_SIDE_EFFECTS__` support. When a function is annotated, all top-level calls to it are treated as pure (matching Rollup/esbuild/Webpack semantics).

Previously we only supported these annotations at the call site. Now both patterns work:

```js
/* #__NO_SIDE_EFFECTS__ */
function setup() { return {}; }
const app = setup(); // allowed
export default createConfig(); // also allowed if annotated
```

3 new test cases added. All 59 tests pass.

PR #2871, opened 2026-02-08
fix(`prefer-spread`): skip TypedArray and ArrayBuffer constructor calls

### LINKED ISSUE #2818: `unicorn/prefer-spread`: ArrayBuffer.prototype.slice incorrectly linted
`unicorn/prefer-spread` is turning `arrayBuffer.slice(0)` into `[...arrayBuffer]`, which is not valid and errors in TS:

```
Type 'ArrayBuffer' must have a '[Symbol.iterator]()' method that returns an iterator. (ts 2488)
```

kaigritun:
Fixes #2818

## Problem

The `prefer-spread` rule was incorrectly suggesting to spread `ArrayBuffer` and `TypedArray` instances when using `.slice()`. This is problematic because:

1. **ArrayBuffer has no `[Symbol.iterator]`**, so spreading fails:
   ```
   Type 'ArrayBuffer' must have a '[Symbol.iterator]()' method that returns an iterator. (ts 2488)
   ```

2. **TypedArray.slice() returns the same typed array** (e.g., `Uint8Array`), but spreading converts it to `number[]`, silently breaking code expecting the original type.

   As mentioned in the issue:
   ```js
   var sliced = u8.slice(0); // <--- Uint8Array
   var spread = [...u8]; // <--- number[]
   ```

## Solution

Added detection for direct constructor calls to skip reporting:
- `new ArrayBuffer(10).slice()`
- `new SharedArrayBuffer(10).slice()`
- `new Uint8Array([1,2,3]).slice(0)`
- `new Blob([]).slice()`
- `new File([], 'test').slice()`
- All TypedArray variants (Int8Array, Uint8ClampedArray, Int16Array, Uint16Array, Int32Array, Uint32Array, Float32Array, Float64Array, BigInt64Array, BigUint64Array)

## Limitations

This fix detects direct constructor calls (`new Uint8Array(...).slice()`) but does **not** detect when a variable holds a TypedArray instance, as that would require type information not available to ESLint rules. The existing workaround of naming variables `arrayBuffer`, `blob`, etc. still applies for those cases.

## Tests

- Moved `new Uint8Array([10, 20, 30, 40, 50]).slice()` from invalid to valid tests
- Added comprehensive test cases for all TypedArray constructors with `.slice()` and `.slice(0)`


rules/prefer-spread.js
```diff
@@ -41,6 +41,41 @@ const ignoredSliceCallee = [
...
+	'Int16Array',
+	'Uint16Array',
+	'Int32Array',
+	'Uint32Array',
+	'Float32Array',
+	'Float64Array',
+	'BigInt64Array',
+	'BigUint64Array',
+	'Blob',
+	'File',
```
sindresorhus: These are not typed array.

__

and you can use `rules/shared/typed-array.js`

PR #2872, opened 2026-02-09
Add `no-useless-iterator-to-array` rule

### LINKED ISSUE #2830: Rule proposal: `no-useless-iterator-to-array`
### Description

Very similar to [no-useless-spread](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-spread.md), but I don't think it make sense to add to `no-useless-spread`

### Examples

```js
// ❌
const set = new Set(iterator.toArray());

// ✅
const set = new Set(iterator);
```

```js
// ❌
for (const foo of iterator.toArray());

// ✅
for (const foo of iterator);
```

More examples see [no-useless-spread](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-useless-spread.md), replace `...` to `.toArray()`

### Proposed rule name

no-useless-iterator-to-array

### Additional Info

_No response_

costajohnt:
Fixes #2830

Adds a new rule that detects unnecessary `.toArray()` calls on iterators.

## What it detects

1. **Iterable-accepting constructors/methods**: `new Set(iterator.toArray())` → `new Set(iterator)`
2. **`for…of` loops**: `for (const x of iterator.toArray())` → `for (const x of iterator)`
3. **`yield*` delegation**: `yield* iterator.toArray()` → `yield* iterator`
4. **Iterator methods available on both Iterator and Array**: `iterator.toArray().every(fn)` → `iterator.every(fn)`

Shared methods flagged: `every`, `find`, `forEach`, `reduce`, `some`.

## What it does NOT flag

- `.filter()`, `.map()`, `.flatMap()` — these have different semantics between Iterator (returns iterator) and Array (returns array)
- Array-only methods like `.slice()`, `.sort()`, `.includes()`, `.indexOf()`, `.join()`, `.at()`, `.flat()`
- Optional chaining (`iterator?.toArray()`, `iterator.toArray?.()`)
- `.toArray()` calls with arguments (not `Iterator.prototype.toArray`)
- Standalone `.toArray()` usage (`const arr = iterator.toArray()`)

## Detection strategy

Per fisker's guidance in #2830: objects with a `.toArray()` method are assumed to be Iterator instances. No complex type checking needed.


rules/no-useless-iterator-to-array.js
```diff
@@ -0,0 +1,217 @@
...
+		return {
+			node: callerObject.callee.property,
+			messageId: MESSAGE_ID_ITERATOR_METHOD,
+			data: {method: node.callee.property.name},
+			suggest: [
+				{
+					messageId: MESSAGE_ID_SUGGESTION_ITERATOR_METHOD,
+					data: {method: node.callee.property.name},
+					fix: fixer => removeMethodCall(fixer, callerObject, context),
+				},
```
fisker: I think we should provide autofix if the callback is known not using the  3rd argument and `this`.

test/no-useless-iterator-to-array.js
```diff
@@ -0,0 +1,165 @@
...
+
+		// Spread — not `.toArray()`
+		'[...iterator]',
+		'call(...iterator)',
+
+		// Spread — optional chaining
+		'[...iterator?.toArray()]',
+		'[...iterator.toArray?.()]',
+		'call(...iterator?.toArray())',
+		'call(...iterator.toArray?.())',
```
sindresorhus, line 58:
```suggestion
		'call(...iterator.toArray?.())',
		
		// Object spread should not be flagged
		'({...iterator.toArray()})',
```

test/no-useless-iterator-to-array.js
```diff
@@ -0,0 +1,165 @@
...
+		'[...iterator.toArray(), b]',
+
+		// Case 7: Spread in function call / new expression
+		'call(...iterator.toArray())',
+		'call(a, ...iterator.toArray())',
+		'new Foo(...iterator.toArray())',
+
+		// Parenthesized
+		'new Set((iterator.toArray()))',
+		'for (const x of (iterator.toArray()));',
```
sindresorhus, line 177:
```suggestion
		// Parenthesized
		'new Set((iterator.toArray()))',
		'Promise.all((iterator.toArray()))',
		'Array.from((iterator.toArray()))',
		'Object.fromEntries((iterator.toArray()))',
		'for (const x of (iterator.toArray()));',
		'[...(iterator.toArray())]',
		'call(...(iterator.toArray()))',
		'new Foo(...(iterator.toArray()))',
```

test/no-useless-iterator-to-array.js
```diff
@@ -0,0 +1,179 @@
...
+
+		// Iterator method with optional chaining
+		'iterator.toArray()?.every(fn)',
+		'iterator.toArray().every?.(fn)',
+
+		// Iterator methods with `thisArg` — Array accepts it, Iterator does not
+		'iterator.toArray().every(fn, thisArg)',
+		'iterator.toArray().find(fn, thisArg)',
+		'iterator.toArray().forEach(fn, thisArg)',
+		'iterator.toArray().some(fn, thisArg)',
```
sindresorhus, line 95:
```suggestion
		'iterator.toArray().some(fn, thisArg)',
		// Array callbacks receive the 3rd `array` argument, Iterator callbacks do not
		'iterator.toArray().every((value, index, array) => array.length === 1)',
```

PR #2873, opened 2026-02-10
Add 
o-top-level-side-effects rule

### LINKED ISSUE #1661: Rule proposal: No top-level side effects

swjcpy:
Fixes #1661.

### DIFF
--- a/rules/no-top-level-side-effects.js
+++ b/rules/no-top-level-side-effects.js
@@ -4,52 +4,81 @@ const messages = {
 	[MESSAGE_ID]: 'Do not use top-level side effects.',
 };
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Closing as this user only does low-quality AI PRs without even checking them.

PR #2874, opened 2026-02-11
`no-unnecessary-polyfills`: Improve performance

sindresorhus:
The rule was doing too many regex checks per import. I changed it to preselect likely matches before running regex tests. It now builds token indexes from core-js feature names and only checks candidate patterns. I also normalized polyfill-prefixed imports (polyfill-* and mdn-polyfills/*) so they use the same narrowed path instead of scanning everything. 

Result: overhead on synthetic 4000-import files dropped from roughly 2x+ in some cases to about 1.0-1.06x.


aditya-ai-architect:
Acknowledged, I will review this.

PR #2876, opened 2026-02-11
`no-useless-undefined`: Flag return undefined for explicit undefined return types

### LINKED ISSUE #2747: `no-useless-undefined`: should work when a function has a return type specified
When enabling the `no-useless-undefined` rule in Typescript, I expect both of the following to get flagged:

```ts
function getsFlaggedCorrectly() {
  return undefined; // 👍 flagged
}

function shouldBeFlaggedButIsNot(): undefined {
  return undefined; // ☹️ this SHOULD be flagged, but is NOT
}
```

sindresorhus:
Fixes #2747

### DIFF
--- a/rules/no-useless-undefined.js
+++ b/rules/no-useless-undefined.js
@@ -77,6 +77,9 @@ const getFunction = scope => {
 	}
 };
 
+const isUndefinedReturnType = functionNode =>
+	functionNode.returnType?.typeAnnotation?.type === 'TSUndefinedKeyword';
+
 const isFunctionBindCall = node =>
 	!node.optional
 	&& node.callee.type === 'MemberExpression'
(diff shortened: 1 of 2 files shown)


rules/no-useless-undefined.js
```diff
@@ -118,10 +121,14 @@ const create = context => {
 			&& node.parent.type === 'ReturnStatement'
 			&& node.parent.argument === node
 		) {
+			const functionNode = getFunction(sourceCode.getScope(node));
+			if (functionNode?.returnType && !isUndefinedReturnType(functionNode)) {
```
fisker, line 125: We should check union types too.

sindresorhus: Not sure I understand. Unions are already handled. `isUndefinedReturnType()` only matches exact `: undefined`, so `: string | undefined` skips autofix.

fisker: I'm not sure I understand either.

> Unions are already handled. 

You mean this is handled?

```
const foo = (): string | undefined => {
	if (foo) {
	  return undefined;
	}
	return 'string'
}
```

I didn't see related logic.

fisker: Ah. sorry, misunderstood.

PR #2877, opened 2026-02-11
`prefer-array-flat`: Don’t report on obvious non-array receivers

### LINKED ISSUE #2660: `prefer-array-flat` fix is *not* safe
Same issue in Biome linter: https://github.com/biomejs/biome/issues/6142

[Rule prefer-array-flat](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-array-flat.md) fix is not safe.

```js
const randomObject = {
  flatMap(f) {
    f()
  }
}

randomObject.flatMap((x) => x)
```
## testcase
https://biomejs.dev/playground/?code=YwBvAG4AcwB0ACAAcgBhAG4AZABvAG0ATwBiAGoAZQBjAHQAIAA9ACAAewAKACAAIABmAGwAYQB0AE0AYQBwACgAZgApACAAewAKACAAIAAgACAAZgAoACkACgAgACAAIAAgAGMAbwBuAHMAbwBsAGUALgBsAG8AZwAoACIAaABpACIAKQAKACAAIAB9AAoAfQAKAAoAcgBhAG4AZABvAG0ATwBiAGoAZQBjAHQALgBmAGwAYQB0AE0AYQBwACgAKAB4ACkAIAA9AD4AIAB4ACkA

sindresorhus:
Fixes #2660

### DIFF
--- a/test/prefer-array-flat.js
+++ b/test/prefer-array-flat.js
@@ -80,6 +80,19 @@ test({
 			output: 'const Items = []; Items.flat();',
 			errors: 1,
 		},
+		{
+			code: outdent`
+				for (const value of values) {
+					value.flatMap(x => x);
+				}
+			`,
+			output: outdent`
+				for (const value of values) {
+					value.flat();
+				}
+			`,
+			errors: 1,
+		},
 	],
 });
 


rules/prefer-array-flat.js
```diff
@@ -209,11 +279,15 @@ function create(context) {
 	];
 
 	context.on('CallExpression', function * (node) {
-		for (const {testFunction, description, getArrayNode, shouldSwitchToArray, isOptionalArray} of cases) {
+		for (const {testFunction, description, getArrayNode, shouldSwitchToArray, isOptionalArray, shouldIgnoreCase} of cases) {
```
fisker: The `shouldIgnoreCase` logic can merge into  `testFunction`.

rules/prefer-array-flat.js
```diff
@@ -157,6 +159,72 @@ const lodashFlattenFunctions = [
...
+	) {
+		return;
+	}
+
+	return definition.node.init;
+};
+
+const isConstNonArrayVariable = (node, context) => {
+	const initializer = getConstVariableInitializer(node, context);
+	return Boolean(initializer && isDefinitelyNonArrayExpression(initializer));
```
fisker, line 216: Let's add a test for this line?

`VariableDeclarator` with `const` always has an initializer unless it's in a `for..of` loop.

PR #2878, opened 2026-02-11
`prefer-object-from-entries`: Avoid auto-fixing generic reduce

### LINKED ISSUE #2519: `prefer-object-from-entries`: autofix is unsafe if `reduce` type parameter is explicitly specified
If an explicit type parameter is present on `reduce`, it's unsafe for `prefer-object-from-entries` to fix the problem because this type parameter will carry over to `map` which will likely be incompatible:

```ts
array.reduce<Record<string, Data & { b?: string }>>(
  (res, entry) => ({ ...res, [entry.id]: entry.data }),
  {}
);

// is transformed by the rule to ...

//             but this type parameter v is wrong for `map`
Object.fromEntries(array.map<Record<string, Data & { b?: string }>>(
  ( entry) => [entry.id, entry.data]
));
```

Full reproduction: https://stackblitz.com/edit/vitejs-vite-axdfht2u

sindresorhus:
Fixes #2519

### DIFF
--- a/rules/prefer-object-from-entries.js
+++ b/rules/prefer-object-from-entries.js
@@ -191,7 +191,7 @@ function create(context) {
 				messageId: MESSAGE_ID_REDUCE,
 			};
 
-			if (!(callExpression.typeArguments ?? callExpression.typeParameters)) {
+			if (!callExpression.typeArguments) {
 				problem.fix = fixReduceAssignOrSpread({
 					context,
 					callExpression,


rules/prefer-object-from-entries.js
```diff
@@ -186,15 +186,20 @@ function create(context) {
...
 			}
 
-			yield {
+			const problem = {
 				node: callExpression.callee.property,
 				messageId: MESSAGE_ID_REDUCE,
-				fix: fixReduceAssignOrSpread({
+			};
+
+			if (!(callExpression.typeArguments ?? callExpression.typeParameters)) {
```
fisker: Why `typeParameters`? I don't think current version of  typescript-eslint generate this property.

fisker: AI?

sindresorhus: Yeah, from Babel, it seems: https://github.com/babel/babel/issues/16679 Will remove.

PR #2879, opened 2026-02-11
`prefer-global-this`: Fix for window-specific `in` checks

### LINKED ISSUE #2484: `prefer-global-this` should also exclude `'open' in window`
The rule has exceptions: 

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/2b469bee475a8f3f2767f4669864acdd89654017/rules/prefer-global-this.js#L8-L121

But these are not checked in existence checks, so the rule just suggested I replace `'open' in window` while leaving `window.open()` behind

sindresorhus:
Fixes #2484


rules/prefer-global-this.js
```diff
@@ -125,6 +127,11 @@ const webWorkerSpecificAPIs = new Set([
 	'onconnect',
 ]);
 
+const environmentSpecificAPIByGlobalIdentifier = {
```
fisker: `environmentSpecificAPIByGlobalIdentifier` -> `environmentSpecificApiByGlobalIdentifier`

rules/prefer-global-this.js
```diff
@@ -160,6 +167,30 @@ function isComputedMemberExpressionObject(identifier) {
...
+Check if the identifier is used in an existence check for a known environment-specific API.
+
+@param {import('estree').Identifier} identifier
+@returns {boolean}
+*/
+function isKnownSpecificAPIExistenceCheck(identifier) {
+	const specificAPIs = environmentSpecificApiByGlobalIdentifier[identifier.name];
+	if (!specificAPIs) {
+		return false;
+	}
```
fisker: This is unsafe to me, should either turn `environmentSpecificApiByGlobalIdentifier` into a `Map`, or check with `Object.hasOwn()` first.

`specificAPIs` should also be `specificApis`

PR #2880, opened 2026-02-11
Refactor boolean utilities

### LINKED ISSUE #2449: Let's change the utils `isBooleanNode(node)` to `isInControlFlow(node)`
https://github.com/sindresorhus/eslint-plugin-unicorn/blob/d3e4b805da31c6ed7275e2e2e770b6b0fbcf11c2/rules/utils/boolean.js#L26-L37

Currently, this function is too complicated, judging both the node itself and the node's parent.

It is easy to [false positives](https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1840). Let's split this into two functions

- `isBooleanExpression(node)` - [Implemented](https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2446/files#diff-845a2a47912eb9ac3ef5b583bfd706f5732931dd2a7b85dae22fc5515ae9ad7a)
- `isInControlFlow(node)` - create a new function

sindresorhus:
Fixes #2449

### DIFF
--- a/rules/utils/boolean.js
+++ b/rules/utils/boolean.js
@@ -22,11 +22,11 @@ const isVueBooleanAttributeValue = node =>
 	);
 const isDirectControlFlowTest = node =>
 	(
-		node.parent?.type === 'IfStatement'
-		|| node.parent?.type === 'ConditionalExpression'
-		|| node.parent?.type === 'WhileStatement'
-		|| node.parent?.type === 'DoWhileStatement'
-		|| node.parent?.type === 'ForStatement'
+		node.parent.type === 'IfStatement'
+		|| node.parent.type === 'ConditionalExpression'
+		|| node.parent.type === 'WhileStatement'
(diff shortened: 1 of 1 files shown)


rules/utils/boolean.js
```diff
@@ -20,43 +20,52 @@ const isVueBooleanAttributeValue = node =>
 		|| node.parent.key.name.rawName === 'else-if'
 		|| node.parent.key.name.rawName === 'show'
 	);
+const isDirectControlFlowTest = node =>
+	(
+		node.parent?.type === 'IfStatement'
+		|| node.parent?.type === 'ConditionalExpression'
+		|| node.parent?.type === 'WhileStatement'
+		|| node.parent?.type === 'DoWhileStatement'
+		|| node.parent?.type === 'ForStatement'
```
fisker: Should remove `?`, we are not going to test node without `parent` (`Program`)

PR #2881, opened 2026-02-11
`better-regex`: Fix empty-pattern autofix

### LINKED ISSUE #2314: `better-regex` autofix breaks regular expression syntax: `/|/g` -> `//g`
An empty regular expression such as `//` is invalid and parsed as a line comment instead, which breaks other syntax within the code. I would expect the `better-regex` rule to not propose code with empty regular expressions.

The following is a minimal example:

```js
const expression = /|/;
```

I would expect this expression to be corrected to something like:

```js
const expression = /(?:)/;
```

sindresorhus:
Fixes #2314

A better fix is to fix the library, but I doubt that will happen, so we can work around it for now.

### DIFF
--- a/test/better-regex.js
+++ b/test/better-regex.js
@@ -100,6 +100,16 @@ test({
 			output: String.raw`const expression = /(?:)/g`,
 		},
 		{
+			code: String.raw`const expression = /(?:)|/`,
+			errors: createError(String.raw`/(?:)|/`, String.raw`/(?:)/`),
+			output: String.raw`const expression = /(?:)/`,
+		},
+		{
+			code: String.raw`const expression = /(?:)|/g`,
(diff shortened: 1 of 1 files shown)


rules/better-regex.js
```diff
@@ -36,6 +44,7 @@ const create = context => {
 
 		try {
 			optimized = regexpTree.optimize(original, undefined, {blacklist: ignoreList}).toString();
+			optimized = normalizeOptimizedRegexLiteral(optimized);
```
fisker, line 47: This only fixes `/|/`? 


If so why not 

```
if (original === '/|/') {
  // ...
}
```

?

sindresorhus: It is not only `/|/`. I have added a few more test cases.

PR #2882, opened 2026-02-11
`prefer-set-has`: Fix string false positives

### LINKED ISSUE #2216: Bug: prefer-set-has false positive for strings in JS

sindresorhus:
Fixes #2216

### DIFF
--- a/rules/prefer-set-has.js
+++ b/rules/prefer-set-has.js
@@ -125,9 +125,7 @@ const create = context => {
 				parent.parent.parent.type === 'ExportNamedDeclaration'
 				&& parent.parent.parent.declaration === parent.parent
 			)
-			&& (
-				isArrayMethodCall(parent.init)
-			)
+			&& isArrayMethodCall(parent.init)
 		)) {
 			return;
 		}


rules/prefer-set-has.js
```diff
@@ -116,6 +133,7 @@ const create = context => {
 					optionalMember: false,
 				})
 			)
+			&& !isArrayOrStringMethodCallOnStringLikeNode(parent.init)
```
fisker: I don't  like this, we should exclude `slice` and `concate` from  `methodsReturnsArray` and  check them stricter instead.

rules/prefer-set-has.js
```diff
@@ -95,26 +122,7 @@ const create = context => {
...
-					optional: false,
-				})
-				// `Array.from()` and `Array.of()`
-				|| isMethodCall(parent.init, {
-					object: 'Array',
-					methods: ['from', 'of'],
-					optionalCall: false,
-					optionalMember: false,
-				})
-				// Methods that return an array
```
fisker, line 112: Can we add these comments back?

test/prefer-set-has.js
```diff
@@ -23,6 +23,13 @@ const methodsReturnsArray = [
 	'split',
 ];
 
+const methodsReturnsArrayOrString = [
+	'concat',
+	'slice',
+];
+
+const methodsReturnsOnlyArray = methodsReturnsArray.filter(method => !methodsReturnsArrayOrString.includes(method));
```
fisker: This variable should be defined directly. Since the current `methodsReturnsArray` is useless now, it should just remove `concat` and `slice` from `methodsReturnsArray` and use it directly.

I feel like babysitting the AI.. Getting tired

sindresorhus: > I feel like babysitting the AI.. Getting tired

I feel this is a bit unfair. I manually review every PR before submitting. I missed this too. Also, it's just minor nitpick. I think you have to do 5x more babysitting for other PRs than the ones I have been doing with AI assistance (not fully AI generated by the way). So in my opinion, this is a net win.

rules/prefer-set-has.js
```diff
@@ -95,26 +126,7 @@ const create = context => {
...
-				|| isMethodCall(parent.init, {
-					methods: methodsReturnsArray,
-					optionalCall: false,
-					optionalMember: false,
-				})
+				isArrayMethodCall(parent.init)
 			)
```
fisker:
```suggestion
			&& isArrayMethodCall(parent.init)
```

fisker: The diff is strange, I mean, should remove `()` around.

test/prefer-set-has.js
fisker: I don't know, but I feel not interested review some code at all.

For example, this test https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2880/changes#diff-6b5d516c201323584b5cbaa4ec4118b5603a7b401f769c58143d5697312180cbR1.

If was added by a human, I'd review carefully and suggest using an ESLint plugin to run tests, since you don't need to do the `setParent` and `findNode` part.

But as it's generated by AI, I can't say it's wrong, and feel meaningless to tell it run tests in another way.

sindresorhus: > If was added by a human, I'd review carefully and suggest using an ESLint plugin to run tests, since you don't need to do the setParent and findNode part.

I don't see how this is any different. I went with the parse stuff as it seemed the simplest to me. AI suggested and **I** agreed. I didn't even think to use ESLint plugin for this. Is this what you had in mind? https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2880/commits/bdfe8401508c9f913685dd79c1f2bc0972841661

rules/prefer-set-has.js
sindresorhus: Sorry, I don't get what should be removed.

fisker: L128

fisker:
```diff
-			&& (
-				isArrayMethodCall(parent.init)
-			)
+			&& isArrayMethodCall(parent.init)
```

PR #2883, opened 2026-02-11
`explicit-length-check`: Avoid unsafe autofix in negated comparisons

### LINKED ISSUE #2190: Wrong autofix in `explicit-length-check`
There is a wrong autofix for the `explicit-length-check` rule.

My code:
```ts
if (!foo.length > 0) {
 ...
```

Code after autofix:

```ts
if (foo.length === 0 > 0) {
 ...
```

Expected replacement

```ts
if (foo.length === 0) {
 ...
```

sindresorhus:
Fixes #2190

### DIFF
--- a/docs/rules/explicit-length-check.md
+++ b/docs/rules/explicit-length-check.md
@@ -165,7 +165,7 @@ The `non-zero` option can be configured with one of the following:
 
 ## Unsafe to fix case
 
-`.length` check inside `LogicalExpression`s are not safe to fix.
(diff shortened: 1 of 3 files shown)


fisker:
Maybe it doesn't make much sense, but 

```js
if (!foo.length > 0);
```

should fix to 

```js
if ((foo.length === 0) > 0);
```

the check logic is correct, but the problem is the fix logic missing parens in such case.

PR #2884, opened 2026-02-12
`prefer-top-level-await`: Fix false positive with `Promise.all`

### LINKED ISSUE #2177: False positive in `prefer-top-level-await`
Should not report on following cases.

```js
async function getStat() {}

const [core, pure, bundle] = await Promise.all([
  getStat('core-js'),
  ALL && getStat('core-js-pure'),
  ALL && getStat('core-js-bundle'),
]);
```

https://github.com/zloirock/core-js/blob/ab5f0252fcbfb58aebd636d7c3b6585b2f4cc320/scripts/downloads-by-versions.mjs#L18-L24

sindresorhus:
Fixes #2177

### DIFF
--- a/rules/prefer-top-level-await.js
+++ b/rules/prefer-top-level-await.js
@@ -51,6 +51,13 @@ const isAwaitExpressionArgument = node => {
 const isArrayElementWrapper = node => (
 	(node.parent.type === 'ChainExpression' && node.parent.expression === node)
 	|| (
+		node.parent.type === 'ConditionalExpression'
+		&& (
+			node.parent.consequent === node
+			|| node.parent.alternate === node
+		)
+	)
+	|| (
 		isLogicalExpression(node.parent)
 		&& (
 			node.parent.right === node
--- a/test/prefer-top-level-await.js
+++ b/test/prefer-top-level-await.js
@@ -226,6 +226,12 @@ test.snapshot({
 			]);
 		`,
 		outdent`
+			async function getStat() {}
+			const [core] = await Promise.all([
+				ALL ? getStat('core-js') : getStat('core-js-pure'),
+			]);
+		`,
+		outdent`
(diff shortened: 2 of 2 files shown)


rules/prefer-top-level-await.js
```diff
@@ -47,16 +48,45 @@ const isAwaitExpressionArgument = node => {
 	return node.parent.type === 'AwaitExpression' && node.parent.argument === node;
 };
 
+const isArrayElementWrapper = node => (
+	(node.parent.type === 'ChainExpression' && node.parent.expression === node)
```
fisker, line 52: This not tested.

rules/prefer-top-level-await.js
```diff
@@ -47,16 +48,45 @@ const isAwaitExpressionArgument = node => {
...
+
 // `Promise.{all,allSettled,any,race}([foo()])`
-const isInPromiseMethods = node =>
-	node.parent.type === 'ArrayExpression'
-	&& node.parent.elements.includes(node)
-	&& isMethodCall(node.parent.parent, {
+const isInPromiseMethods = node => {
+	let expression = node;
+	while (
+		expression.parent
```
fisker: `expression.parent` check should be removed. I don't think it's possible to be nullish.

rules/prefer-top-level-await.js
```diff
@@ -47,16 +48,42 @@ const isAwaitExpressionArgument = node => {
...
+	|| (
+		isLogicalExpression(node.parent)
+		&& (
+			node.parent.right === node
+			|| (
+				node.parent.left === node
+				&& node.parent.operator !== '&&'
+			)
+		)
+	)
```
fisker, line 69: Ternary is also common, maybe we can also check it.

fisker (approved):

PR #2890, opened 2026-02-13
`no-unused-properties`: Fix for JSX member access

### LINKED ISSUE #1799: no-unused-properties rule does not support styled when using as react component
When using react styled components (and similar libraries like emotion, stitches etc.) as styles rule don't work.

no-unused-properties

```tsx
const styles = {
  wrapper: styled('div', {}),
  unused: styled('div', {}),
}

function Component() {
  return <styles.wrapper />
}
```

sindresorhus:
Fixes #1799

### DIFF
--- a/rules/no-unused-properties.js
+++ b/rules/no-unused-properties.js
@@ -37,13 +37,7 @@ const getPropertyKeyName = key => {
 
 const propertyKeysEqual = (keyA, keyB) => {
 	const keyNameA = getPropertyKeyName(keyA);
-	const keyNameB = getPropertyKeyName(keyB);
-
-	if (keyNameA !== undefined && keyNameB !== undefined) {
-		return keyNameA === keyNameB;
-	}
-
-	return false;
+	return keyNameA !== undefined && keyNameA === getPropertyKeyName(keyB);
 };
 
 const objectPatternMatchesObjectExprPropertyKey = (pattern, key) =>


rules/no-unused-properties.js
```diff
@@ -25,7 +25,17 @@ const specialProtoPropertyKey = {
 	name: '__proto__',
 };
 
+const normalizePropertyKey = key => key.type === 'JSXIdentifier'
+	? {
+		type: 'Identifier',
+		name: key.name,
+	}
```
fisker, line 31: I don't think it's a good idea to fake it as `Identifier`, should use `JSXIdentifier` directly.

rules/no-unused-properties.js
```diff
@@ -25,25 +25,22 @@ const specialProtoPropertyKey = {
...
+	const keyNameB = getPropertyKeyName(keyB);
 
-		if (keyB.type === 'Literal') {
-			return keyA.value === keyB.value;
-		}
+	if (keyNameA !== undefined && keyNameB !== undefined) {
+		return keyNameA === keyNameB;
 	}
 
 	return false;
```
fisker:
```suggestion
const propertyKeysEqual = (keyA, keyB) => {
	const keyNameA = getPropertyKeyName(keyA);
	return keyNameA !== undefined && keyNameA === getPropertyKeyName(keyB);
```

PR #2893, opened 2026-02-13
`consistent-destructuring`: Fix false positive after reassignment

### LINKED ISSUE #1407: consistent-destructuring: incorrect report if the value is reassigned
```js
const { children } = parent
parent.children = combine(children)
for(const child of parent.children) { // incorrect report here
}
```

sindresorhus:
Fixes #1407

### DIFF
--- a/rules/consistent-destructuring.js
+++ b/rules/consistent-destructuring.js
@@ -55,12 +55,17 @@ const isRootVariableReassigned = (declaration, memberExpressionNode, memberScope
 			return false;
 		}
 
-		if (reference.from.variableScope !== memberScope.variableScope) {
+		const [referenceStart] = sourceCode.getRange(reference.identifier);
+		if (referenceStart < declarationEnd) {
 			return false;
 		}
 
-		const [referenceStart, referenceEnd] = sourceCode.getRange(reference.identifier);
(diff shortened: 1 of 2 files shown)


rules/consistent-destructuring.js
```diff
@@ -33,6 +34,36 @@ const isChildInParentScope = (child, parent) => {
...
 
+const getRootIdentifier = expression => {
+	while (expression?.type === 'MemberExpression') {
+		expression = expression.object;
+	}
+
+	return expression?.type === 'Identifier' ? expression : undefined;
+};
+
+const isRootVariableReassigned = (declaration, memberExpressionNode, memberScope, sourceCode) => {
```
fisker, line 45: I don't think this is correct, the reassign code can be anywhere that have access to the variable.

For example,

```
let foo = bar;

const {a} = foo;

reassign();

console.log(foo.a);

function reassign(){
  foo = baz;
}
```

Anyway, I don't have a good solution either.

sindresorhus: I have adjusted reassignment checks to avoid unsafe stuff while still reporting reads before same-scope reassignments. I think that's the best we can do.

PR #2902, opened 2026-02-24
Add `prefer-simple-condition-first` rule

### LINKED ISSUE #2853: Rule proposal: `simple-condition-first`
### Description

When writing multiple conditions, the simple one should come first.

I believe this can improve performance. The direct boolean check and variable `===`/`!==` comparison are obviously cheaper.

### Examples

```js
// ❌
if (check(foo) && bar);

// ✅
if (bar && check(foo));
```

```js
// ❌
if (foo.bar.baz === 1 && bar === 2);

// ✅
if (bar === 2 && foo.bar.baz === 1);
```

### Proposed rule name

simple-condition-first

### Additional Info

There is an internal rule in Prettier codebase checking a particular case https://github.com/prettier/prettier/blob/46061a160f7a6726d11d047647d59036435a3279/scripts/tools/eslint-plugin-prettier-internal-rules/test.js#L39-L44 , but it will be great to have a generic rule.

costajohnt:
Fixes #2853.

Adds a new rule that flags logical expressions (`&&`, `||`) where a simple condition appears on the right but a complex one appears on the left, and offers to swap them.

## What's "simple"

Per fisker's scoping in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2853#issuecomment-2652755982:

1. Bare identifier (`foo`)
2. `identifier === literal` or `identifier !== literal` (strict equality/inequality between identifiers and/or literals)

Everything else is considered "complex."

## Fix safety

- **Auto-fix** when the complex side has no function calls — safe to reorder
- **Suggestion** when the complex side contains `CallExpression` or `NewExpression` — reordering changes short-circuit evaluation timing, as noted by @axetroy

## Examples

```js
// ❌
if (check(foo) && bar);
if (foo.bar.baz === 1 && bar === 2);

// ✅
if (bar && check(foo));
if (bar === 2 && foo.bar.baz === 1);
```


test/prefer-simple-condition-first.js
```diff
@@ -0,0 +1,53 @@
...
+		// Neither simple — not flagged in v1
+		'if (a.b && c.d);',
+		'if (foo() && bar());',
+		'if (a.b.c && d.e.f);',
+
+		// Single condition — not a LogicalExpression
+		'if (foo);',
+
+		// Simple on left, complex on right — correct order
+		'if (bar || foo());',
```
sindresorhus, line 31:
```suggestion
		'if (bar || foo());',
		
		// Potentially unsafe to reorder (side effects / throws)
		'if ((state.ready = true) && ok);',
		'if (++counter && ok);',
		'if (object.deep.value && ok);',
		'const x = object.deep.value || ok;',
		'if (tag`x` && ok);',
```

rules/no-static-only-class.js
```diff
@@ -33,7 +33,7 @@ function isStaticMember(node) {
 		return false;
 	}
 
-	if (!isStatic || isPrivate || key.type === 'PrivateIdentifier') {
+	if (isPrivate || !isStatic || key.type === 'PrivateIdentifier') {
```
fisker: Let's consider `!foo` as "simple"?

costajohnt: Done, `!identifier` is now treated as simple. Reverted the swaps in no-static-only-class and text-encoding-identifier-case since both sides are simple with this change.

PR #2905, opened 2026-02-26
`text-encoding-identifier-case`: Support template literals

FloEdelmann:


test/text-encoding-identifier-case.js
```diff
@@ -4,12 +4,15 @@ const {test} = getTester(import.meta);
 
 test.snapshot({
 	valid: [
-		'`UTF-8`',
+		'`utf8`',
+		// eslint-disable-next-line no-template-curly-in-string
+		'`UTF-${8}`',
```
sindresorhus, line 9:
```suggestion
		'`UTF-${8}`',
		'`\\u0055tf8`',
		'tag`\\unicode`',
```

PR #2906, opened 2026-02-26
chore: replace `is-builtin-module` with native `node:module` `isBuiltin`

roli-lpci:
## Summary

- Replaces the `is-builtin-module` package with Node.js built-in `module.isBuiltin()` (available since Node 16.x)
- The package already requires `node >= 20.10`, so the native API is always available
- Removes 1 production dependency (and its transitive dep `builtin-modules`)

## Notes

The native `isBuiltin()` returns `true` for deprecated modules (`punycode`, `sys`) that `is-builtin-module` explicitly excludes. A small exclusion set preserves the existing behavior — the `prefer-node-protocol` rule will continue to skip deprecated builtins, matching current test expectations.

## Changes

- `rules/prefer-node-protocol.js` — `import isBuiltinModule` → `import { isBuiltin } from 'node:module'`, added `deprecatedBuiltins` set
- `package.json` — removed `is-builtin-module` from dependencies

## Testing

All existing tests pass, including the `punycode` edge case (`import "punycode"` remains valid without triggering the rule).

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,6 @@
 		"find-up-simple": "^1.0.1",
 		"globals": "^16.4.0",
 		"indent-string": "^5.0.0",
-		"is-builtin-module": "^5.0.0",
 		"jsesc": "^3.1.0",
 		"pluralize": "^8.0.0",
 		"regexp-tree": "^0.1.27",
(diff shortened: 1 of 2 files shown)


sindresorhus:
Duplicate of https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2475

PR #2911, opened 2026-03-17
fix(`no-for-loop`): handle cached-length pattern for (let i = 0, j = arr.length; i < j; i++) (fixes #295)

### LINKED ISSUE #295: A case not handled by `no-for-loop`

argusagent:
The rule previously bailed out whenever the loop init had more than one declarator, so this common optimization pattern was silently ignored:

`js
for (let i = 0, j = arr.length; i < j; i += 1) { ... }
`

**Changes:**
- getIndexIdentifierName: allow exactly two declarators when the second is j = someArray.length. Validate eagerly so or (let i = 0, j = 0; ...) still passes.
- getArrayIdentifier: fall back to the cached-length path when the standard i < arr.length test does not match.
- Two new invalid test cases.

All 117 existing tests pass.

Fixes #295
Related to #250


sindresorhus:
I double-checked this locally, and I think there's stil one unsafe autofix case here.

If the cached-length variable is shadowed inside the loop body, the new guard resolves the inner variable instead of the one declared in the `for` header. That means the real header variable can still be referenced after the loop, but the fixer no longer sees it as escaping.

For example:

```js
for (var i = 0, j = arr.length; i < j; i++) {
	let j = 1;
	console.log(arr[i]);
}

console.log(j);
```

In this case the body-scope lookup finds the inner `let j`, while the header `var j` is the one used by `console.log(j)` after the loop. So the fix can still remove the header declarator and leave that trailing `j` behind.

sindresorhus:
First, the new two-declarator path still falls back to the old `i < arr.length` test. That means an unrelated second initializer gets removed even though it is not the cached length used by the condition:

```js
for (let i = 0, j = other.length; i < arr.length; i++) {
	console.log(arr[i]);
}
```

This autofixes to:

```js
for (const element of arr) {
	console.log(element);
}
```

So `other.length` is dropped. That can be observable through a getter/proxy, and it is outside the cached-length pattern this PR is trying to support. I think the cached-length branch should only apply when the condition is actually `i < j` / `j > i`, or the extra initializer should make the fix bail out.

Second, `var` cached-length variables are still unsafe when used after the loop:

```js
for (var i = 0, j = arr.length; i < j; i++) {
	console.log(arr[i]);
}
console.log(j);
```

This autofixes to:

```js
for (const element of arr) {
	console.log(element);
}
console.log(j);
```

The current through-reference check catches the unresolved `let` version, but `var j` after the loop is resolved, so it still slips through while the fixer removes the declaration.

Also, the focused XO run still fails with 9 style errors and 1 complexity warning.

PR #2938, opened 2026-04-10
Add `prefer-iterator-to-array-at-end` rule

### LINKED ISSUE #2767: Rule proposal: Enforce `.toArray()` at the end of `Iterator` methods chain
### Description

> Note that it's a good idea to call `toArray()` as a last step of your processing. For example, `fibonacci().take(10).toArray().filter(...)` is less efficient, because iterator helpers are lazy and avoids creating a temporary array.

From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/toArray

### Examples

```js
// ❌
foo.toArray().filter(...)
//            ^^^^^^ Array#filter()

// ✅
foo.filter(...).toArray()
//  ^^^^^^ Iterator#filter()
```

### Proposed rule name

not sure

### Additional Info

_No response_

nami8824:
Closes #2767 

## What                                                                                                                                                                                
                                                                                                                                                                                       
Adds a new rule that enforces calling `.toArray()` at the end of an iterator method chain, rather than in the middle.                                                                                        
                                                                                                                                                                                         
## Notes                                                                                                                                                                                  
                                                              
- `flatMap` is intentionally excluded: `Iterator#flatMap` requires the callback to return an iterable, while `Iterator#flatMap` requires the callback to return an iterable, whereas `Array#flatMap` does not.
- Callbacks with more than 2 parameters are skipped: `Array#filter/map` passes a third argument (the array itself), but `Iterator#filter/map` does not.


test/snapshots/prefer-iterator-to-array-at-end.js.md
```diff
@@ -0,0 +1,157 @@
...
+
+> Error 1/1
+
+    `␊
+    Message:␊
+    > 1 | iterator.toArray().filter(fn)␊
+        |          ^^^^^^^ Move \`.toArray()\` to the end of the iterator chain; use \`Iterator#filter()\` instead of \`Array#filter()\`.␊
+    ␊
+    Output:␊
+      1 | iterator.filter(fn).toArray()␊
```
fisker: Since the `fn` call runs differently

- `Iterator#filter()` calls with two arguments, while `Array#filter()` calls with 3 arguments
- `this` is different

We can't autofix this case unless we know how `fn` works(eg: arrow function with less than 3  parameters).

rules/prefer-iterator-to-array-at-end.js
```diff
@@ -0,0 +1,79 @@
...
+				optionalMember: false,
+			})
+		) {
+			return;
+		}
+
+		// Iterator#filter/map call the callback with two arguments (element, index).
+		// Array versions also pass a third argument (the array reference). If the callback
+		// declares more than two parameters, the third (array) would be `undefined`
+		// after the fix, so skip.
```
fisker: Ha, I didn't see this before I wrote https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2938/changes#r3309392988, but seems the logic should be changed to only fix the safe cases.

test/snapshots/prefer-iterator-to-array-at-end.js.md
nami8824: Thanks for pointing this out.
I’ve changed the autofix to only run for arrow callbacks that cannot observe the extra `Array#filter()` arguments

https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2938/changes/1db91908fc4fdf1d74b0714211d4afeb97feb470

nami8824:
Closing this PR as it has been addressed by #3041.

PR #2940, opened 2026-04-13
Add `no-unused-array-method-return` rule

### LINKED ISSUE #741: Rule proposal: `no-unused-return-value`
Details here: https://github.com/eslint/eslint/issues/12980

The idea is that many builtin functions return values and if those return values are not used, the code is essentialy a noop and can be removed. Notably if getters are involved, the rule may have false positives but I generally see getters as a obscure feature that should not be used.

The discussion got stale because it seems hard for a plugin to correctly judge the type of a variable which is essential for a rule like this so it may require Typescript to really be useful.

### Fail
```` js
const arr = [{name: 'eslint'}];
arr.map(item => item.name); // error: Unused Array.prototype.map return value
console.log(arr);
````
### Pass
```` js
const arr = [{name: 'eslint'}];
console.log(arr);
````

### PR DESCRIPTION
**Pull request by sindresorhus**
Fixes #741

### CODE REVIEW

**Comment by silverwind** [comment]
### Critical: rule crashes on common code

`rules/no-unused-array-method-return.js:251` — `getVariableValue` returns `definition.node.init` unconditionally. When a declarator has no initializer (uninitialized `let`, `for…of`, `for…in`, `for (let x; …)`), `init` is `null`. `resolveReceiver` then recurses with `null` and throws `TypeError: Cannot read properties of null (reading 'type')` at line 171.

Reproduced via `npx ava test/no-unused-array-method-return.js` after dropping these inputs into the existing test file:

```js
'let array; array.map(fn);'                    // CRASH
'for (const x of arr) x.map(fn);'              // CRASH
'for (let x of arr) x.map(fn);'                // CRASH
'for (var x of arr) x.map(fn);'                // CRASH
'for (const x in obj) x.map(fn);'              // CRASH
'for (let x; x.length;) x.map(fn);'            // CRASH
```

Suggested fix: add `&& definition.node.init` to the condition at lines 244–250, returning `uncertainValue` for these cases (matches the documented "for…of bindings unsupported" intent). Worth adding at least one `for…of` and one bare `let x;` to the test file to prevent regression.

### Code quality

- L9–40 — `methodNames` is only used to build `methods`; inline as `const methods = new Set([...]);`.
- L224–238 and L289–304 — ~30 lines of inline boundary commentary largely repeat what the docs already explain. Trimming would match the terser style of nearby rules.
- Mix of `const x = (…) =>` and `function x(…)` declarations in the same file is inconsistent.

### Heuristic gaps

- `.slice` / `.at` / `.includes` / `.indexOf` / `.concat` / `.lastIndexOf` exist on strings; unresolved receivers will false-positive. Acknowledged in the docs, but worth a final risk check before promoting beyond `unopinionated`.
- Statement-level `SequenceExpression` (e.g. `sideEffect(), arr.map(fn);`) is silently skipped — listed as intentional but a real false negative.

---
Review by Claude Opus 4.7

PR #2950, opened 2026-04-29
`prefer-string-slice`: Improve autofix

### LINKED ISSUE #2505: Improve `prefer-string-slice` autofix
unicorn/prefer-string-slice is good when all the parameters are constants, or when there is only one parameter. But when there are two parameters and they're variables, the autofixes will do things like this:

`editsummary = editsummary.substr(0, editsummary.length - 2); // remove trailing comma`
->
`editsummary = editsummary.slice(0, Math.max(0, editsummary.length - 2)); // remove trailing comma`

I find the Math.max pretty hard to read. I'd prefer that it not autofix that particular sub-pattern. Any interest in adding an option to this rule so that it can be configured to skip that particular sub-pattern when autofixing?

Promise2679:
closes #2505 

I noticed this issue and attempted an implementation.

The rule now produces `str.slice(0, -x)` instead of `str.slice(0, Math.max(0, str.length - x))` when x is a positive numeric literal.

Please let me know if there’s anything I should adjust.

### DIFF
--- a/rules/prefer-string-slice.js
+++ b/rules/prefer-string-slice.js
@@ -1,5 +1,5 @@
 import {getStaticValue} from '@eslint-community/eslint-utils';
-import {getParenthesizedText, getParenthesizedRange} from './utils/index.js';
+import {getParenthesizedText, getParenthesizedRange, isDecimalInteger} from './utils/index.js';
 import {replaceArgument} from './fix/index.js';
 import {isNumericLiteral, isMethodCall} from './ast/index.js';
 import {getNegativeIndexLengthNode} from './shared/negative-index.js';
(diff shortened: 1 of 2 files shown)


sindresorhus:
This still looks unsafe for non-interger numeric literals. The new shortcut accepts `foo.length - 1.5` and fixes it to `foo.slice(0, -1.5)`, but that changes behavior.

For example, with `const value = 'abcdef'`, both `value.substr(0, value.length - 1.5)` and `value.substring(0, value.length - 1.5)` return `abcd`, while `value.slice(0, -1.5)` returns `abcde`. `0.5` is even worse because `slice(0, -0.5)` becomes an empty string.

This should only apply when the subtracted literal is a positive integer, not just any positive number.

Promise2679:
Thanks for the review! I've fixed it by adding an `isDecimalInteger` check and included the corresponding test cases.

sindresorhus:
This shortcut still drops comments that are inside the argument.

For example:
```js
foo.substr(0, foo.length - /* keep */ 1);
foo.substring(0, foo.length - /* keep */ 1);
```
Both autofix to:
```js
foo.slice(0, -1);
```
which deletes the comment. The fallback keeps it as `Math.max(0, foo.length - /* keep */ 1)`, so I think the shortcut should either bail out when the matched argument contains comments or preserve them somehow. Would be good to add a regression test for this so it does not regress agian.

PR #2956, opened 2026-05-11
Add no-top-level-side-effects rule

### LINKED ISSUE #1661: Rule proposal: No top-level side effects

raydeveloperF:
Fixes #1661

This adds a new `no-top-level-side-effects` rule that reports top-level side-effect statements in modules with exports, while ignoring files without exports and hashbang files.

Covered cases include:
- top-level calls, `new`, top-level `await`, and side-effect-only imports in exported modules
- top-level control-flow statements in exported modules
- ignored scripts without exports
- ignored hashbang files
- allowed declarations and assignments

Tests:
- `npm run test:js -- test/no-top-level-side-effects.js`
- `npm run lint:js -- rules/no-top-level-side-effects.js test/no-top-level-side-effects.js`
- `npm run lint:eslint-docs`
- `npx markdownlint docs/rules/no-top-level-side-effects.md readme.md`
- `git diff --check`


sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2957#issuecomment-4419187554

PR #2957, opened 2026-05-11
Add no-top-level-side-effects rule

### LINKED ISSUE #1661: Rule proposal: No top-level side effects

MingGe0:
Fixes #1661.

This adds a new `no-top-level-side-effects` rule that reports side-effectful top-level expression statements in exported modules.

The implementation follows the constraints discussed in the issue:

- Ignores hashbang files.
- Ignores files without exports.
- Allows assignments.
- Does not report declarations or code inside functions/classes.
- Does not provide an autofix, because removing top-level side effects can change runtime behavior.

Verification:

- `npx ava test/no-top-level-side-effects.js`
- `npx eslint rules/no-top-level-side-effects.js test/no-top-level-side-effects.js`
- `npm run lint:eslint-docs`

I also ran `npm test`; it currently fails on the existing `CLAUDE.md` markdownlint issue (`MD041`/`MD047`), unrelated to this change.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1661: Rule proposal: `no-top-level-side-effects`](https://oss.issuehunt.io/repos/55832243/issues/1661)
---
</details>


sindresorhus:
We don't accept fully AI-generated PRs.

PR #2958, opened 2026-05-12

### LINKED ISSUE #1661: Rule proposal: No top-level side effects

dot-Justin:
Closes #1661

## What this does

Adds the `no-top-level-side-effects` rule that disallows statements that execute code at module load time.

### Files with no exports are ignored

Files without any `export` statement are scripts or entry points (like `ReactDOM.render(<App/>)`) — intentional execution contexts. These are silently skipped.

### Hashbang files are ignored

Files starting with `#!/usr/bin/env node` are CLIs. These are also skipped.

### Assignments are allowed

`const server = createServer()` and `x = foo()` at top level are allowed, as requested.

### Addresses the issues raised in the two existing PRs

- `export default foo()` — now flagged (the declaration is a `CallExpression`)
- `export default await foo()` — now flagged
- `void foo()` bypass — flagged (entire `ExpressionStatement` is reported, not just the call)
- `condition && foo()` / `condition || foo()` — flagged
- `foo(), bar()` sequence — flagged
- `if (condition) { foo(); }` at top level — flagged

### Pure annotations suppress errors

`/* @__PURE__ */ foo();` and `/* #__PURE__ */`, `/* @__NO_SIDE_EFFECTS__ */`, `/* #__NO_SIDE_EFFECTS__ */` all work as escape hatches, per the compiler-notations-spec.

## Tests

40 tests, all passing. Snapshots included.

---

*This PR is linked to a funded bounty on [IssueHunt](https://oss.issuehunt.io/r/sindresorhus/eslint-plugin-unicorn/issues/1661).*


sindresorhus:
We do not accept fully AI-generated PRs.

Issue #1852, opened 2022-07-04
`unicorn/better-regex` conflicts with `regexp/strict` and `regexp/match-any`


JounQin:
### Description

**Information:**
- ESLint version: v8.19.0
- `eslint-plugin-regexp` version: 1.7.0

**Description**

```js
/\[([^[\]]+)]/g
```

error from this plugin

```log
Unescaped source character ']'
```

After fixed:

```js
/\[([^[\]]+)\]/g
```

error from `unicorn/better-regex`

```log
/\[([^[\]]+)\]/g can be optimized to /\[([^[\]]+)]/g
```

```js
/{{([\S\s]+?)}}/g
```

`regexp/match-any` is reported

```js
/{{([\s\S]+?)}}/g
```

`unicorn/better-regex` is reported

### Fail

As above

### Pass

As above

### Additional Info

I hope these two plugins can work better together by default, some rule can be extracted from `unicorn` to `regexp` instead?

JounQin:
A related previous issue #469, we need to clarify why current rule `unicon/better-regex` is `better`.

sindresorhus:
It was removed: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/89dc5dc04389929ae786e171bac2af522fe2e7d8

Issue #1754, opened 2022-03-23
better-regex rule breaks regex


SimonSiefke:
```js
/\p{L}/u.test('Θ') // true
```

get transformed into

```js
/p{L}/u.test('Θ') // false
```

which produces wrong output.

## Possible solutions

Short term solution: Disable `better-regex` rule by default
Long term solution: Fixing the bug 😃

fisker:
Can you report this to [`safe-regex`](https://github.com/davisjam/safe-regex)? This rules uses it under the hood.

sindresorhus:
It was removed: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/89dc5dc04389929ae786e171bac2af522fe2e7d8

Issue #1626, opened 2021-12-01
better-regex proposes to replace [~^] with [^~]


runspired:
These are not equivalent expressions.

```js
-      aString.replace(/[~^]/g, "");
+      aString.replace(/[^~]/g, "");
```

sindresorhus:
They are equivalent.

runspired:
@sindresorhus they are not equivalent. For example:

![image](https://user-images.githubusercontent.com/650309/144219209-c1c22057-eda7-40a9-b680-447240e9058f.png)

B2o5T:
Yeah autofix should add backslash before `^`

sindresorhus:
Looks like a bug in the package we use to simplify/normalize regexes: https://github.com/DmitrySoshnikov/regexp-tree

fisker:
https://github.com/DmitrySoshnikov/regexp-tree/issues/237

metal0:
Hi any update on this? sorry for asking so soon

fregante:
The rule has been removed from the recommended config (however it still exists here)

sindresorhus:
It was removed: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/89dc5dc04389929ae786e171bac2af522fe2e7d8

Issue #1500, opened 2021-08-25
Rule proposal: `regex-strict-mode`


Nantris:
This rule would be beneficial to allow users to get the benefits of `unicorn/better-regex` without interfering with the [`regexp/strict` rule](https://ota-meshi.github.io/eslint-plugin-regexp/rules/strict.html) from the [`eslint-plugin-regexp`](https://github.com/ota-meshi/eslint-plugin-regexp) project.

## Fail

```js
const demo = /\}\{/ // Flagged by unicorn/better-regex if it has the unnecessary escape characters - and by `regexp/strict` if it doesn't.
```

## Pass

Pass would be leaving the code untouched and without warning, as strict mode encourages including the escape characters. This also has the benefit of improving syntax highlighting, as the `{` and `}` are clearly highlighted as characters to match, as opposed to being highlighted as quantifiers.

```js
const demo = /\}\{/ // Same code, just don't warn
```

papb:
Your request is actually for an option to `better-regex`, not for a new rule, right?

Slapbox:
I suppose that's accurate @papb.

sindresorhus:
It was removed: https://github.com/sindresorhus/eslint-plugin-unicorn/commit/89dc5dc04389929ae786e171bac2af522fe2e7d8

PR #2968, opened 2026-05-28
Drop `better-regex` rule

fregante:
- https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2443

For the same reason why it was removed from the "recommended" config two years ago, I think it should just be removed now. The underlying package has severe bugs and hasn't seen a release in 3+ years: https://github.com/DmitrySoshnikov/regexp-tree/tags

Dedicated, supported, more granular plugins are better suited for regex improvements:

- https://github.com/ota-meshi/eslint-plugin-regexp

Then we can close:
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/895
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1754
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1626
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1852
- https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1500


docs/deleted-and-deprecated-rules.md
```diff
@@ -86,4 +86,6 @@ This rule was renamed to [`prefer-string-trim-start-end`](rules/prefer-string-tr
 
 ### ~regex-shorthand~
 
-This rule was renamed to [`better-regex`](rules/better-regex.md) as it does more than just preferring the shorthand.
+### ~better-regex~
+
+This rule was renamed to [`better-regex`](rules/better-regex.md) and alter removed.
```
fisker:
```suggestion
This rule was renamed to [`better-regex`](rules/better-regex.md) as it does more than just preferring the shorthand.

### ~better-regex~

[Reason for removing `better-regex`]
```

fisker:
Normally, we add an empty rule to https://github.com/sindresorhus/eslint-plugin-unicorn/blob/2d51e3058e7c3280c37c37a69827d37c39007f31/index.js#L9 before actually removing.

So user don't need to update their config.

fregante:
Like this?

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/bc439d871c3faa6a899bd19a7d53b8fbb3535636/index.js#L10-L16

fisker (approved):
There should be some deps can be removed, otherwise looks good.

fregante:
> Normally, we add an empty rule to

Tests fail because the rule appears in the "deprecated" list while also completely being "removed". Should the test expectations be changed to allow this?

fisker:
> > Normally, we add an empty rule to
> 
> Tests fail because the rule appears in the "deprecated" list while also completely being "removed". Should the test expectations be changed to allow this?

Let's put in `Deprecated rules` for now?

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/bye-better-regex/docs/deleted-and-deprecated-rules.md#deprecated-rules

We removed implementations when deprecate rules before.

PR #2969, opened 2026-05-28
Avoid running CI twice for local PRs

fregante:
This is wasting GHA minutes:

Screenshot 11

I had this config in Refined GitHub for a while: https://github.com/refined-github/refined-github/blob/f93c8b2305e3eed82e12447a5780ba90e47c2553/.github/workflows/test.yml#L3-L10

It means:

- only run CI on local pushes to `main`
- other branches will run CI only when PRs are open

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,7 +1,13 @@
 name: CI
+
 on:
-  - push
-  - pull_request
+  pull_request:
+    branches:
+      - '*'
+  push:
+    branches:
+      - main
+
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -1,7 +1,13 @@
 name: CI
+
 on:
-  - push
-  - pull_request
+  pull_request:
+    branches:
+      - '*'
```
fisker:
```suggestion
  pull_request:
```

This should still work

fisker (approved):

PR #3073, opened 2026-06-10
Add `prefer-number-is-safe-integer` rule

### LINKED ISSUE #2623: `rule-name`: `prefer-number-is-safe-integer`
### Description

I think in all cases use `Number.isInteger`, [`Number.isSafeInteger`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger) should be preferred.

### Examples

```js
// ❌
if (!Number.isInteger(x)) {}

// ✅
if (!Number.isSafeInteger(x)) {}
```

### Additional Info

_No response_

sindresorhus:
Fixes #2623


sin-strokes (commented):
Leave it

PR #3080, opened 2026-06-10
Fix coverage report

fisker:


sindresorhus:
Same issue still: https://github.com/sindresorhus/eslint-plugin-unicorn/actions/runs/27303235166/job/80654250354?pr=3081#step:6:249

I'm going to disable it again. I don't really use it and it wastes a lot of time when it's always flaky.

PR #3090, opened 2026-06-11
Require Node.js 22 and ESLint 10.4

sindresorhus:


yoyo837:
Thanks for pushing this forward! One concern: `@babel/eslint-parser` (the most popular custom ESLint parser used by many projects that also depend on `eslint-plugin-unicorn`) still does not support ESLint 10 — its latest version (7.29.7) has peer deps limited to `^7.5.0 || ^8.0.0 || ^9.0.0`. The related babel issue (https://github.com/babel/babel/issues/17791) was closed without a fix.

For projects that use both `eslint-plugin-unicorn` and `@babel/eslint-parser`, this bump effectively creates an unresolvable peer dependency conflict. Would it make sense to keep the ESLint 9 range alongside 10 for a transitional period, or at least document this known incompatibility in the release notes so downstream users are aware?

sindresorhus:
ESLint 9 support was kept for some time. No plans to bring it back. You can just hold off on updating this plugin until the Babel thing is compatible with ESLint 10.

It's not up to us to document how another project is incompatible with another project. That will eventually become outdated and be incorrect too.

PR #3118, opened 2026-06-11
Add `prefer-short-arrow-method` rule 

### LINKED ISSUE #1737: Rule proposal: `prefer-short-arrow-method`
### Description

ESLint has various rules to enforce methods or function declarations vs arrow functions. One that’s missing is a way to enforce the use of a short arrow function vs the use of a method with a single return value.

I.e., given `this` is not used the following are equivalent:

```js
// 1
const person = {
  greet(name) {
    return `Hello, ${name}`
  }
}
```

```js
// 2
const person = {
  greet: (name) => {
    return `Hello, ${name}`
  }
}
```

```js
// 3
const person = {
  greet: (name) => `Hello, ${name}`
}
```

The [`object-shorthand`](https://eslint.org/docs/rules/object-shorthand) rule can be used to prefer option 1 over option 2. The [`arrow-body-style`](https://eslint.org/docs/rules/arrow-body-style) rule can be used to enforce option 3 over option 2. However, there’s no way to enforce option 1 or 3 over the other.

I suggest to add a rule to prefer a short arrow function over a method with a single return value. I’m also fine with a rule that’s the other way around. For me it’s about consistency.

### Fail

```js
const person = {
  greet(name) {
    return `Hello, ${name}`
  }
}
```

### Pass

```js
const person = {
  greet: (name) => `Hello, ${name}`
}
```

Use the [`arrow-body-style`](https://eslint.org/docs/rules/arrow-body-style) rule to deal with this situation.

```js
const person = {
  greet: (name) => {
    return `Hello, ${name}`
  }
}
```

sindresorhus:
Fixes #1737


fregante:
Not sure about this one, in this specific case it just makes the methods inconsistent:

https://github.com/refined-github/refined-github/blob/fdfd796edc97ff278531edf2caf22b96fd03fd46/source/background.ts#L36-L62

Screenshot 3

fregante:
`@stylistic/object-curly-newline` has `consistent` boolean option that xo uses: https://github.com/xojs/eslint-config-xo/blob/9f9b5ac57ed4a0d9d38e589f7f4cda2d268e262a/source/javascript-rules.js#L461-L484

Maybe `prefer-short-arrow-method` deserves it too?

fregante:
Neat, I just saw the `consistent-as-needed` option

PR #3131, opened 2026-06-12
Add `prefer-early-return` rule

### LINKED ISSUE #16: `return-early` rule
As discussed in https://github.com/sindresorhus/execa/pull/15#discussion-diff-60815913. How I think when coding is to handle errors and exceptions as soon as possible and return early. If we take a look at the two examples below, the latter one makes more sense to me. `foobar()` in this case is the "real" body while the code in the `if` statements deal with exceptions. When using the style in the former one, this isn't as clear to me.

``` js
if (conditon) {
    someFn();
} else if (someOtherCondition) {
    someOtherFn();
} else {
    foobar();
}
```

``` js
if (condition) {
    someFn();
    return;
}

if (someOtherCondition) {
    someOtherFn();
    return;
}

foobar();
```

This might be worth reading too http://blog.timoxley.com/post/47041269194/avoid-else-return-early.

sindresorhus:
Fixes #16


fregante:
I'd love to see an autofix on this one. It's a pain to flip `if` blocks manually and it should be rather easy to stick a `!` in it via autofix

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3212

fregante:
Strangely it didn't autofix this:

```ts
export default mem((selector: string | ((clickedItem: HTMLElement) => string)): EventHandler => event => {
	if (event.altKey && event.isTrusted) {
		const clickedItem = event.delegateTarget;

		// `parentElement` is the anchor because `clickedItem` might be hidden/replaced after the click
		const resetScroll = preserveScroll(clickedItem.parentElement!);
		clickAllExcept(typeof selector === 'string' ? selector : selector(clickedItem), clickedItem);
		resetScroll();
	}
});
```

fregante:
Actually a number of them haven't been fixed. You can check out https://github.com/refined-github/refined-github/commit/b115c055adca9733a84fcf951af3c57a18d76a5a and run `npx eslint .`

PR #3139, opened 2026-06-12
Add `no-declarations-before-early-exit` rule

### LINKED ISSUE #2772: Rule proposal: forbid declarations that are not used before an early return
### Description

Flag variables that are declared before a possible early exit (return, throw, break, continue) but are only used after that exit point. Such variables may be unnecessary in the early-exit path and can be safely moved down to reduce scope and avoid unneeded initialisation.

### Examples

```js
// ❌
function foo(bar) {
  const res = getRes();
  if (!bar) return;
  console.log(res);
}

// ✅
function foo(bar) {
  if (!bar) return;
  const res = getRes();
  console.log(res);
}
```

### Proposed rule name

no-unused-vars-before-exit

### Additional Info

Introducing this rule promotes clearer, more efficient control flow by encouraging developers to declare variables only when they are truly needed. In functions or loops that contain early exits (like **return**, **throw**, **break**, or **continue**), variables declared before these exit points but used only afterwards are unnecessary in those paths. By delaying such declarations, developers can avoid performing unneeded work—especially when it involves expensive operations or function calls.

This rule also helps minimise variable scope, making code easier to read and reason about. Keeping variables close to where they are used aligns with modern JavaScript practices and complements existing rules like no-unused-vars and prefer-const. Overall, it encourages a more intentional structure, better performance in early-exit scenarios, and improved code maintainability.

sindresorhus:
Fixes #2772


fregante:
It reported this line:

https://github.com/refined-github/refined-github/blob/fdfd796edc97ff278531edf2caf22b96fd03fd46/source/github-helpers/parse-compare-url.ts#L29

but the line has a side effect via Array#pop

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3238

akx:
This also wreaks some havoc with React hooks; code like

```
export default function RuleView() {
  const { code } = useParams<{ code: string }>();
  const facets = React.useMemo(...);
  if (!code) {
    return <NoDataAlert />;
  }
  return ...
}
```
tags `const facets = React.useMemo(...)` to be moved after the early return, which would then violate the rules of hooks.

I think `no-declarations-before-early-exit` could maybe use some tribal knowledge about `React.use*` and `use*` in general?

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/commit/de47dcece1118b1b41ea7dc2c5ada989fd65372d

PR #3161, opened 2026-06-12
Add `no-unnecessary-global-this` rule

### LINKED ISSUE #2419: Rule proposal: Avoid or prefer using globals as `globalThis.something`
### Description

_Originally posted in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2410#issuecomment-2255792543_
            
> you should still prefer nothing at all over `globalThis` in some cases (unless you're using `?.`), like you wouldn't use `new globalThis.Image()` or `globalThis.Array.from()` or `globalThis.globalThis`

I'd lean towards never using `globalThis` (and `window`, `self`, `global`, etc) **for known globals**. `globalThis.jQuery = jQuery` should still be allowed though.

### Fail

```js
globalThis.alert();
globalThis.fetch();
globalThis.console.log();
globalThis.addEventListener('navigate', fn);
globalThis.JSON.stringify();
globalThis.window.self.close()
new globalThis.Image();
globalThis.RegExp.$1;
```

### Pass

```js
alert();
fetch();
console.log();
addEventListener('navigate', fn);
JSON.stringify();
self.close(); // Partial exception
new Image();
RegExp.$1;
```

The _exception_ is the same as the one described in https://github.com/sindresorhus/eslint-plugin-unicorn/pull/2410#discussion_r1695312427

### Proposed rule name

`global-properties` or just `avoid-global-propertoes`

### Additional Info

_No response_

sindresorhus:
Fixes #2419


fregante:
Note that this will conflict with the current setup of no-confusing-globals (or whatever that rule is) so that rule needs to be updated/disabled in XO. 

It was good when it prevented usage of globals like `name` and `event` but at some point in the past year it stated forcing me to use globalThis.whatever for almost any global, including those mentioned in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/2419

sindresorhus:
Yeah, I will handle it when I update to this change.

fregante:
I ran the latest commit on Refined GitHub and it autofixed this:

Screenshot

fregante:
I'm not sure what the final decision here is, I upgraded to the latest eslint-config-xo (unicorn v67) and it's marking lines previously fixed by this rule as `Unexpected use of 'location' no-restricted-globals`

```js
const actionUrl = new URL('search', location.origin);
```

PR #3184, opened 2026-06-13

### LINKED ISSUE #1661: Rule proposal: No top-level side effects

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1661

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#1661: Rule proposal: `no-top-level-side-effects`](https://oss.issuehunt.io/repos/55832243/issues/1661)
---
</details>


rules/no-top-level-side-effects.js
```diff
@@ -0,0 +1,74 @@
...
+	[MESSAGE_ID]: 'Do not use top-level side effects in exported modules.',
+};
+
+const exportDeclarationTypes = new Set([
+	'ExportAllDeclaration',
+	'ExportDefaultDeclaration',
+	'ExportNamedDeclaration',
+]);
+
+const isExportDeclaration = node => exportDeclarationTypes.has(node.type);
```
andersk, line 15:
```suggestion
const isExportDeclaration = node => exportDeclarationTypes.has(node.type) && node.exportKind !== 'type';
```

- #3402

PR #3191, opened 2026-06-13
Add `no-asterisk-prefix-in-documentation-comments` rule

### LINKED ISSUE #1607: Rule proposal: Doc comment Asterix prefix style
### Description

I prefer not using Asterix prefix on doc comments as I find them noisy and they also make it much harder to edit comments.

I'm aware you can almost get this with the JSDoc ESLint plugin, but it's a bit heavy to add that just for this and doc comments do not necessarily need to be JSDoc, so this rule would be more general.

### Fail

```js
/**
 * Add two numbers.
 * @param {number} num1 The first number.
 * @param {number} num2 The second number.
 * @returns {number} The sum of the two numbers.
 */
```

### Pass

```js
/**
Add two numbers.
@param {number} num1 The first number.
@param {number} num2 The second number.
@returns {number} The sum of the two numbers.
*/
```

(Notice how `*/` moved too)

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1607


fregante:
How come this is disabled? Also how is it different form `jsdoc/require-asterisk-prefix` which is currently used in XO?

Related: #3724 

Also I noticed that jsdoc's rule doesn't report/drop the extra space, so maybe _that_ could be the differentiator?

```js
Input:
/**
 * Hello
 */

Output:
/**
 Hello
 */
```

👆 see spaces before Hello and `*/`

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3725

PR #3208, opened 2026-06-13
`prefer-array-some`: Ignore non-array find checks

### LINKED ISSUE #3198: False positive in the `prefer-array-some` rule
For Mozilla's PDF.js project we found that upgrading from `eslint-plugin-unicorn` version 64.0.0 to 65.0.1 triggered a new `prefer-array-some` error for the line at https://github.com/mozilla/pdf.js/blob/master/src/core/xfa/factory.js#L104:

```
src/core/xfa/factory.js
  104:54  error  Prefer `.some(…)` over `.find(…)`  unicorn/prefer-array-some
```

However, this is a false positive because on this line we're not dealing with an array but with a `FontFinder` instance instead, which is instantiated at https://github.com/mozilla/pdf.js/blob/01948aff236d76578a170ae5043ed3bac2871518/src/core/xfa/factory.js#L100 and doesn't have a `.some()` method. Interestingly it does _not_ trigger on https://github.com/mozilla/pdf.js/blob/01948aff236d76578a170ae5043ed3bac2871518/src/core/xfa/text.js#L42 which is similar, so I suspect that the way it's called (using inline square brackets for computed property names) confuses the rule.

This possibly introduced by patch https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3028 because that is the only change related to this rule that I could find in the changelog.

For now we're working around this in https://github.com/mozilla/pdf.js/pull/21446/changes/68a5ec1403be1a68bac51cb391944c939f5ce191 by ignoring the rule for this particular line, but we figured it'd be good to report it here too in the hopes that it can be fixed. Thank you!

sindresorhus:
Fixes #3198


timvandermeij:
Thanks a lot for fixing this so quickly!

PR #3222, opened 2026-06-14
Add `prefer-minimal-ternary` rule

### LINKED ISSUE #1625: Rule proposal: `minimum-ternary`
### Description

The ternary part should be as small as possible.

### Fail

```js
const foo = test ? call(a) : call(b);
```

```js
const foo = test ? a(test) : b(test);
```

```js
const foo = test ? a + 1 : b + 1;
```

```js
const foo = test ? Math.min(a, 100) : Math.max(b, 100);
```

### Pass

```js
const foo = call(test ? a : b);
```

```js
const foo = (test ? a : b)(test);
```

```js
const foo = (test ? a : b) + 1;
```

```js
const foo = Math[test ? "min" : "max"](a, 100);
```

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1625


fregante:
I think I hate and disable every ternary rule. Here it's suggesting I "fix" `isMac ? event.metaKey : event.ctrlKey`

sindresorhus:
Agreed. I'm making it an option that's off by default: https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3241

PR #3226, opened 2026-06-14
Add `consistent-class-member-order` rule

### LINKED ISSUE #1110: Rule proposal: consistent class member order
Enforce statements in this order

- Private field
- Public field
- Static field
- constructor
- Private method
- Public method
- Static method
- Static block(https://github.com/tc39/proposal-class-static-block)

_Originally posted by @fisker in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/990#issuecomment-753474656_

Can you open a separate issue for the class member order rule idea? Similar rule for TS: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md

_Originally posted by @sindresorhus in https://github.com/sindresorhus/eslint-plugin-unicorn/issues/990#issuecomment-755267134_

sindresorhus:
Fixes #1110


fregante:
This appears to be impossible to satisfy:

https://github.com/refined-github/refined-github/blob/main/source/github-helpers/search-query.ts

> Expected private field to come before static method.

then I move `url` and `queryParts` to the top and I get:

> Member escapeValue should be declared before all private instance field definitions.

sindresorhus:
You have a conflict with the TS rule, but also https://github.com/sindresorhus/eslint-plugin-unicorn/pull/3239

PR #3245, opened 2026-06-14
`no-unreadable-object-destructuring`, `no-unreadable-array-destructuring`: Disallow property targets

### LINKED ISSUE #1660: Rule proposal: No property references in destructuring patterns
### Description

While using property references in destructuring patterns can be expressive, it can suffer from two problems:

1. It is confuseable with destructuring default values, e.g., `({ a: this.b } = obj);` and `({ a = this.b } = obj);`

2. Tracing the however intended side effects may be confusing as one might be inclined to gloss over full property references on the left hand side where they don't normally occur and where normally only some straightforward local declarations occur.

3. Property references on the left-hand side may allow extended complex and difficult to read expressions:
```js
var a = true, b = {d: 4}, c = {d: 88};
({x: (a ? b : c).d} = {x: 5}); // b.d === 5
b = false;
({x: (a ? b : c).d} = {x: 5}); // c.d === 5
```
While local variables are also defined within the left-hand-side, these are only local in scope and easier to visually track).

4. Left-hand property reference expressions as above may have been more likely intended for the right side.

### Fail

```js
var obj = {};
({ a: this.b } = obj);
```
```js
var obj = {};
({ a: (true).b } = obj);
```
```js
var arr = [];
([ this.a ] = arr);
```

### Pass

```js
var c;
var obj = {a: 5};
({ a: c } = obj);
```
```js
var obj = {a: {b: 15}};
({ a: {b: {c} }} = obj);
```
```js
var a;
var arr = [3];
[ a ] = arr;
```

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/1660


brettz9:
This is great--along with all the others--thanks very much!

PR #3377, opened 2026-06-19
`custom-error-definition`: Don't require `options` parameter when forwarded to `super()` inline

### LINKED ISSUE #3374: `custom-error-definition` is complaining about options as second parameter when it shouldn't
As the title says `custom-error-definition` is being a bit too eager complaining about the options parameter location.

We have loads of custom errors in our codebase because we rely on the class instances rather than error messages or codes to check for equality, so there's errors like

```js
export class TimeoutError extends FetchError {
  constructor(cause) {
    super('The request timed out', { cause });
    this.name = 'TimeoutError';
  }
}
```

and other similar patterns where the message is always hard-coded because it's by design, we don't want to do things like `throw new MissingPaymentIdentifier('...', { cause });`

The message there can only be one, there's no need for a message input.

sindresorhus:
Fixes #3374

### DIFF
--- a/test/custom-error-definition.js
+++ b/test/custom-error-definition.js
@@ -89,6 +89,15 @@ const tests = {
 				}
 			}
 		`,
+		// `options` is forwarded to `super()` inline, so no dedicated `options` parameter is needed
+		outdent`
+			class FooError extends Error {
+				constructor(cause) {
+					super('The request timed out', {cause});
+					this.name = 'FooError';
+				}
+			}
+		`,
 		outdent`
 			class FooError extends Error {
 				constructor() {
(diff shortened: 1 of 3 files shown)


test/custom-error-definition.js
```diff
@@ -685,6 +694,45 @@ const tests = {
 				}
 			`,
 		},
+		{
+			code: outdent`
+				class FooError extends Error {
+					constructor(cause) {
+						super('Fixed message', cause);
```
alfaproject, line 701: I suppose that this example would also be fine, because it would be better written as:

```suggestion
					constructor(options) {
						super('Fixed message', options);
```

rapopan89-dotcom (commented):
wtf

PR #3462, opened 2026-06-29
`no-collection-bracket-access`: Handle conditional member keys

### LINKED ISSUE #3461: `no-collection-bracket-access`: false positive on dynamic method call
`no-collection-bracket-access` flags computed bracket access even when the key is a collection method (here used dynamically):

```js
const set = new Set();
set[set.has('x') ? 'delete' : 'add']('x'); // should be ok
```

Suggesting to never flag bracket notation accessing collection methods like `add`, `delete` etc, which means `set["add"]` likely should be valid.

sindresorhus:
Fixes #3461


silverwind (commented):
lgtm

PR #3487, opened 2026-07-02
`consistent-boolean-name`: Skip React hooks by default

### LINKED ISSUE #3486: `consistent-boolean-name`: add `skipPrefixes` for framework required prefixes
### Description

The rule misfires on react hooks that return a boolean because they have to begin with `use` (e.g. `**use**IsMyBooleanHook`). The rule is still useful for React hooks so the `ignore` option is overkill; instead, we just want to ignore the prefix itself, and allow the rule to consider the rest of the name for validity.

```ts
  "unicorn/consistent-boolean-name": ["error", { skipPrefixes: ["use"] }],
```

### Examples

```ts
// ❌
declare function useMyFlag(): boolean;
declare function useIsCool(): number;

// ✅
declare function useIsMyFlag(): boolean;
declare function useCool(): number;
```

### Additional Info

There is currently no workaround except *redefining* the list of acceptable prefixes manually and add `use` to each one in a loop inside the config, this isn't portable or resilient to future changes to the current list of prefixes.

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/3486

### DIFF
--- a/docs/rules/consistent-boolean-name.md
+++ b/docs/rules/consistent-boolean-name.md
@@ -35,6 +35,8 @@ The plural prefixes (`are`, `have`, `were`) allow names for boolean collections,
 
 The prefix must be a distinct word part. `isReady`, `is_ready`, and `IS_READY` are allowed, but `island` is not considered to have the `is` prefix.
 
+React hooks are checked after the required `use` prefix. For example, `useIsReady` is treated as a boolean hook name, while `useReady` is not.
+
 This rule intentionally does not check destructuring bindings, imports, class names, or catch parameters.
 
 TypeScript type annotation checks resolve local type aliases and callable interfaces, but not qualified or namespaced type references.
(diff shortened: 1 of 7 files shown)


eranhirsch:
Thanks!

lohart13:
This PR doesn't resolve issue #3497. I've still got the same eslint errors after updating to v71.0.0.
The below code still gives an error: `hasConsentReference` starts with `has`, so it should be boolean
```typescript
const hasConsentReference = useRef(false);
```

sindresorhus:
https://github.com/sindresorhus/eslint-plugin-unicorn/releases/tag/v71.1.0

lohart13:
Thanks, can confirm rule is working with useRef correctly now.

Joozty:
Hi @sindresorhus, this doesn't work when I use this setting. I'm still forced to use the `Reference` suffix instead of `Ref`. Thanks!

```			
'unicorn/name-replacements': [
				'error',
				{
					replacements: {
						param: false,
						params: false,
						prev: false,
						prop: false,
						props: false,
						ref: false,
						refs: false,
					},
				},
			],
```

PR #3524, opened 2026-07-11
Update `defineConfig` config docs

sindresorhus:


sindresorhus:
@fregante Not sure if it's better or not. Maybe better in the sense that it's more consistent. Thoughts?

fregante:
Very conflicted. 

I liked the original way of just adding `...plugin.config.recommended,` to the array, but having done `...plugin.config.map(changeSomething)` before, I understand why this change is useful.

For unicorn I think the challenge comes from the many existing and upcoming supported file extensions. With `...plugin.config.recommended,` they would work out of the box, while with `extend` each user has to specify all the extensions manually.

I think the most pressing change would be to update all examples to use `defineConfig` instead of a plain array.


> Maybe better in the sense that it's more consistent

ESLint has done a poor job of informing the plugin developers of this change so from what I can tell most of the existing plugins still do not use `extends`

- https://npmx.dev/package/eslint-plugin-react-hooks#user-content-flat-config-eslintconfigjsts
- https://npmx.dev/package/eslint-plugin-import-x#user-content-configuration-new-eslintconfig
- https://npmx.dev/package/eslint-plugin-react#user-content-flat-configs
- https://npmx.dev/package/eslint-plugin-prettier#user-content-configuration-new-eslintconfigjs
- https://npmx.dev/package/eslint-plugin-jsx-a11y#user-content-shareable-configs

With the sole exceptions I found being TS-ESLint:

- https://typescript-eslint.io/getting-started/#step-2-configuration

which matches what Unicorn currently has on `main`

sindresorhus:
> For unicorn I think the challenge comes from the many existing and upcoming supported file extensions. With ...plugin.config.recommended, they would work out of the box, while with extend each user has to specify all the extensions manually.

Unicorn does not define any file patterns in the config though.

fisker:
> I liked the original way of just adding `...plugin.config.recommended`

👍

fisker (approved):

docs/rules/filename-case.md
```diff
@@ -121,6 +128,14 @@ To enforce PascalCase only for JSX/TSX files, use ESLint's `files` config:
 	},
 	{
 		files: ['**/*.{jsx,tsx}'],
```
fregante: Nit: this should probably be ts/tsx. JSX does not need any parsers I think https://eslint.org/blog/2014/11/es6-jsx-support/

PR #3543, opened 2026-07-13
`no-useless-template-literals`: Ignore const assertions

### LINKED ISSUE #3537: `no-useless-template-literals`: using `String` instead of template literal widens the type of the expression to just `string`
Template literal -> `String(...)` conversion is not always equivalent and desirable. This code is reported:

```ts
const code = `${Math.random() > 0.5 ? 1 : 0}` as const; // Type: `"1" | "0"`
```

This is what the rule suggests as an alternative, but it breaks the expression type:

```ts
const code = String(Math.random() > 0.5 ? 1 : 0); // Type: `string`
```

TS playground for all the examples above: https://www.typescriptlang.org/play/?#code/MYewdgzgLgBKAmBTGBeGADAJAbwLIEMoALAOgCd8x4QBbACgEoYA+GABhIFYYB+GARhgAudgF90MfBDjhoAbgBQMZTAD0qmAD0eChaEiwEiAKpgoATwAOieKhgBlKGQCWYAOZ0CxcpWr0mrBzcfIIibAyKKmoa2kA

(If also actually keeps the `as const` part which causes a hard TS error `A 'const' assertion can only be applied to references to enum members, or string, number, boolean, array, or object literals`)

sindresorhus:
Fixes #3537


andreww2012:
Not only the const assertions should be ignored. I simplified my issue example; the more real use case for me is:

```ts
someFunctionCall({
  // eslint-disable-next-line unicorn/no-useless-template-literals
  someParam: `${someCondition ? 1 : 0}`,
});
```

This will still be reported even with this fix.

PR #3563, opened 2026-07-17
Add rationales to rule documentation

sindresorhus:


docs/rules/empty-brace-spaces.md
```diff
@@ -9,6 +9,8 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
+Keeping empty braces together makes empty blocks and object literals easier to scan and avoids whitespace that has no meaning.
```
fregante, line 12: The style ones are a bit of a stretch. It's more like "I think it looks better" or "It's good to keep a consistent style"

docs/rules/no-nested-ternary.md
```diff
@@ -11,6 +11,8 @@
 
 Improved version of the [`no-nested-ternary`](https://eslint.org/docs/latest/rules/no-nested-ternary) ESLint rule. This rule allows cases where the nested ternary is only one level and wrapped in parens.
 
+Nested ternaries require readers to track multiple conditions and branches at once, making them harder to read and maintain.
```
fregante, line 14: By the way, I find nested ternaries acceptable at any depth, but only when they're `if else if else`, never `if if else else`. If properly formatted they're no different than actual `if else if else` blocks (wrap, no parens)

will open an issue

fregante (commented):
Sorry I didn't get a chance to read these before. They're good for the most part.

PR #3572, opened 2026-07-19
`consistent-boolean-name`: Add `wrappers` option

### LINKED ISSUE #3556: `consistent-boolean-name`: exceptions for other boolean wrappers?
### Description

Like `Promise<boolean>`, the value is actually an object, not a boolean, but it's a common object and native type, so it's easy to have an exception for it. 

However I think that there are other "wrapper" types that ultimately _resolve_ to a boolean value. `() => boolean` is another native "object that resolves to a boolean".

Do you think there's a way to exclude/allow types automatically or via allowlist?

### Examples

Just one example: https://github.com/fregante/webext-storage/blob/main/source/storage-item.md

```js
import {StorageItem} from "webext-storage";

const isUnicorn: StorageItem<unknown, boolean> = new StorageItem('isUnicorn')
```

I'm not sure if it's possible to add an option to parse/understand the _resolved type_ of `StorageItem` 

### Additional Info

At the moment I'm not using the rule, I'm only checking if this is possible/practical.

sindresorhus:
Fixes #3556


fregante:
Checking the getter method makes sense 👌

PR #3577, opened 2026-07-21
`no-unnecessary-string-trim`: Handle Zod string schemas

### LINKED ISSUE #3576: `no-unnecessary-string-trim` incorrectly flags Zod `.trim()` method as unnecessary before `.startsWith()`
The `no-unnecessary-string-trim` rule incorrectly flags `.trim()` calls on Zod schemas (`ZodString`) as unnecessary when chained before `.startsWith()`. Zod's `.trim()` is not `String.prototype.trim()` being called on an already-known string value. The rule appears to match purely on method name without verifying that the target is actually a string.

Rule: `no-unnecessary-string-trim`

This incorrectly triggers the rule:

```ts
import {z} from "zod"

const schema = z.string().trim().startsWith("asdf")
```

sindresorhus:
Fixes #3576


nicholasmt-code (commented):
Assist.Protection.Security

[RA1393804]
(BRP Residence Permit Urban CountryLink Services Australia)

BioMetrics SYD Kingsford Hugo Boss

@nick.does.insta

PR #3583, opened 2026-07-23
`consistent-export-decorator-position`: Fix decorator position reporting

### LINKED ISSUE #3571: `consistent-export-decorator-position`: reported decorator is always `decorators[0]` — intentional?
### Description

The reported node is always `decorators[0]`. Whether flagging it happens to be misleading or not isn't determined by any check in the code — it depends on incidental factors: which decorator sits first in source order, and which `expectedStyle` is configured.

Same source, three different `expectedStyle` options, same `actualStyle: mixed` result:

```js
@foo export @bar class Foo {}
```

**`above`** (test `invalid(11)`):
```
1 | @foo export @bar class Foo {}
  | ^^^^ Expected export decorators to be positioned `above`, but found `mixed`.
```
Neither `@foo` nor `@bar` conforms to `above`, so flagging `@foo` isn't misleading here — but only because there's no correctly-placed decorator to contrast it against, not because the rule identified `@foo` as the offender.

**`before`** (test `invalid(17)`):
```
1 | @foo export @bar class Foo {}
  | ^^^^ Expected export decorators to be positioned `before`, but found `mixed`.
```
`@foo` already conforms. `@bar` is the actual deviation. The warning is anchored on `@foo`, the decorator that's already correctly placed, while `@bar` is never flagged. This is the misleading case.

**`after`** (test `invalid(26)`):
```
1 | @foo export @bar class Foo {}
  | ^^^^ Expected export decorators to be positioned `after`, but found `mixed`.
```
`@bar` already conforms. `@foo` is the actual deviation, and the warning happens to land on it — but only because `@foo` is `decorators[0]` and coincidentally also the sole offender, not because the rule identified it as such. The same three-way pattern repeats identically with `export default` in `invalid(12)` / `invalid(18)` / `invalid(27)`.

Source: `getActualStyle` computes a single style for the whole decorator group and never inspects individual decorators beyond that; `getProblem` always anchors on `decorators[0]`:

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/rules/consistent-export-decorator-position.js#L120-L128

### Question

Is anchoring the report on `decorators[0]`, regardless of which decorator actually deviates, intentional?

sindresorhus:
Fixes #3571


koyulin24747-arch (commented):
Fix 3853

PR #3629, opened 2026-08-14
`prefer-default-parameters`: Support logical assignment operators

### LINKED ISSUE #3628: `prefer-default-parameters` rule is broken in some cases
# `unicorn/prefer-default-parameters` rule doesn't provide editor suggestions in some cases.

## Description
The rule doesn't suggest fixes when equivalent syntax, `||=` or `??=` are used.

With the `logical-assignment-operators` rule, cases that would yield editor suggestions are auto-fixed into cases that do not:
1.  `foo = foo || 'bar'` ✔️ ➡️ `foo ||= 'bar'` ✖️
1.  `foo = foo ?? 'bar'` ✔️ ➡️ `foo ??= 'bar'` ✖️

## Examples
1. This usage is not detected as incorrect
   ```js
   function abc(foo) {
     foo ||= 'bar';
   }
   ```

   It should have a suggested fix to be:
   ```js
   function abc(foo = 'bar') {
   }
   ```

2. The same is true for optional chaining
   ```js
   function abc(foo) {
     foo ??= 'bar';
   }
   ```

   It should have a suggested fix to be:
   ```js
   function abc(foo = 'bar') {
   }
   ```

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/3628

### DIFF
--- a/docs/rules/prefer-default-parameters.md
+++ b/docs/rules/prefer-default-parameters.md
@@ -9,7 +9,7 @@
 <!-- end auto-generated rule header -->
 <!-- Do not manually modify this header. Run: `npm run fix:eslint-docs` -->
 
-Instead of reassigning a function parameter, default parameters should be used. The `foo = foo || 123` statement evaluates to `123` when `foo` is falsy, possibly leading to confusing behavior, whereas default parameters only apply when passed an `undefined` value. This rule only reports reassignments to literal values.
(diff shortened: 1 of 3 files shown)


jpaul11-code (commented):
PANORAFUS IS THE MOST FAMOUS FOR READING THE SCRIPTURES AND UNDERSTANDING SCHATOLOGY.

docs/rules/prefer-default-parameters.md
Michaelalcala14, line 7: - - [ ] - **[@[__****__]()]()**

PR #3631, opened 2026-08-25
`no-nonstandard-builtin-properties`: Support stage 4 proposals

### LINKED ISSUE #3630: `no-nonstandard-builtin-properties`: `Symbol.asyncDispose` is incorrectly reported as non-standard
`unicorn/no-nonstandard-builtin-properties` reports `Symbol.asyncDispose` as a non-standard property of `Symbol`.

`Symbol.asyncDispose` is part of the ECMAScript standard as part of Explicit Resource Management, so I would expect it to be included in the rule's static list of standard built-in properties.

The rule documentation states that:

> The static list tracks the latest published ECMAScript edition and selected web-standard built-ins, not runtime-specific extensions or separate proposal drafts.

Example:

```js
const asyncDispose = Symbol.asyncDispose;
```

Currently reports:

```text
`asyncDispose` is not a standard property of `Symbol`. unicorn/no-nonstandard-builtin-properties
```

The same issue occurs when implementing an async disposable resource in TypeScript:

```ts
type TemporaryDirectory = {
    path: string;
    [Symbol.asyncDispose](): Promise<void>;
};
```

Expected behavior: No error should be reported for `Symbol.asyncDispose`.

sindresorhus:
Fixes #3630


luisalucena (commented):
👍

PR #3716, opened 2026-09-17
`prefer-minimal-ternary`: Add autofix

### LINKED ISSUE #3713: `prefer-minimal-ternary`: Autofix when no side effects
### Description

The docs for p-m-t say

> The rule is not autofixable, since moving the ternary can change evaluation order.

This is true, but uncommon. We should use `hasSideEffects`, which is already used elsewhere in this plugin, to provide autofixes when safe.

### Examples

// All the same reports, just autofixes available when safe

### Additional Info

_No response_

sindresorhus:
Fixes https://github.com/sindresorhus/eslint-plugin-unicorn/issues/3713


Kristeanhoward45-creator (commented):
5000000
Hempire

